PR #84, opened 2015-03-15
Fix Lilypond detection

maxalbert:
I just stumbled upon `music21`, tried to run some examples from the documentation and got a weird error with no obvious error message. After some debugging it turned out that this was due to Lilypond not being installed on my system, but the missing Lilypond executable was not detected correctly.

Currently the logic in `music21/lily/translate.py` (line 194) assumes that a `KeyboardInterrupt` is thrown if Lilypond is not installed, but I don't see how this would ever be triggered by trying to split the version string (which is what happens in the `try...` block in line 190ff).

This PR implements a cleaner way. By avoiding to use `shell=True` in the call to `subprocess.Popen` an `OSError` is triggered if Lilypond is not found, which can be caught directly.

### DIFF
--- a/music21/lily/translate.py
+++ b/music21/lily/translate.py
@@ -182,19 +182,20 @@ class LilypondConverter(object):
 
     def setupTools(self):
         LILYEXEC = self.findLilyExec()
-        command = LILYEXEC + ' --version'
-        proc = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE)
-        versionString = proc.stdout.readline()
-        if six.PY3:
-            versionString = versionString.decode(encoding='utf-8')     
+        command = [LILYEXEC, '--version']
         try:
-            versionString = versionString.split()[-1]
-            versionString = versionString.strip()
-            versionPieces = versionString.split('.')
-        except KeyboardInterrupt:
+            proc = subprocess.Popen(command, stdout=subprocess.PIPE)
+        except OSError:
             raise LilyTranslateException("Cannot find a copy of Lilypond installed on your system. " +
                                          "Please be sure it is installed. And that your " +
                                          "environment.UserSettings()['lilypondPath'] is set to find it.")
+        stdout, _ = proc.communicate()
+        if six.PY3:
+            stdout = stdout.decode(encoding='utf-8')
+        versionString = stdout.split('\n')[0]
+        versionString = versionString.split()[-1]
+        versionString = versionString.strip()
+        versionPieces = versionString.split('.')
         
         self.majorVersion = versionPieces[0]
         self.minorVersion = versionPieces[1]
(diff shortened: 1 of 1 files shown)


mscuthbert:
Thanks.  I didn't write these lines directly, so I don't know why we were catching KeyboardInterrupt.  It looks like it's good to me.  Any chance you were able to test on Windows? I don't have an automatic Windows tester and no longer use it myself.

mscuthbert:
I'll merge now, but with the possibility of a revert if Windows creates a problem.

mscuthbert:
welcome to music21, btw.

maxalbert:
Thanks for the quick reply and the merge! I don't have any option to test on Windows, I'm afraid, but I did take a look at the [`subprocess` documentation](https://docs.python.org/2/library/subprocess.html#popen-constructor) before implementing the change and it says: "the only time you need to specify `shell=True` on Windows is when the command you wish to execute is built into the shell (e.g. dir or copy). You do not need `shell=True` to run a batch file or console-based executable." So I hope it's fine. But by all means do revert the change if it turns out that it introduces any problems.

mscuthbert:
Did you see this email on the list... wondering if you could help.  It might just mean checking if anything gets split or something.  Not sure why this error should rise...

----
apologies for the late night grumpiness. I found the latest windows exe and have music21 installed but I am having trouble with show() and get the error when running sBach.show()  or anything requiring show()  I have installed lilypond and set the PATH

  File "<ipython-input-3-a2b371c913f4>", line 1, in <module>
    sBach.show()

  File "C:\Users\Greg\Anaconda\lib\site-packages\music21\base.py", line 2293, in show
    return formatWriter.show(self, regularizedConverterFormat, app=app, subformats=subformats, **keywords)

  File "C:\Users\Greg\Anaconda\lib\site-packages\music21\converter\subConverters.py", line 230, in show
    fp = helperSubConverter.write(obj, helperFormat, subformats=helperSubformats)

  File "C:\Users\Greg\Anaconda\lib\site-packages\music21\converter\subConverters.py", line 246, in write
    conv = lily.translate.LilypondConverter()

  File "C:\Users\Greg\Anaconda\lib\site-packages\music21\lily\translate.py", line 147, in __init__
    self.setupTools()

  File "C:\Users\Greg\Anaconda\lib\site-packages\music21\lily\translate.py", line 189, in setupTools
    versionString = versionString.split()[-1]

IndexError: list index out of range

maxalbert:
No, I hadn't seen the email but just sent a reply. This piece of code assumes that `versionString` contains the first output line of `lilypond --version`, which *should* read something like `'GNU LilyPond 2.16.2'`. It then tries to split it so that it can grab the last part, which should be the version string `2.16.2`. Not sure what's going wrong in this case, but hopefully we'll get some more debugging information.

PR #86, opened 2015-03-19
Minor docstring addition

maxalbert:
Mention in docstring that numberEnd=None is interpreted as 'last measure'.


mscuthbert:
thanks!

PR #87, opened 2015-03-22
Add Interval.__str__() for more convenient pretty-printing of intervals

maxalbert:
Currently when printing intervals in a human-readable, concise form you have to write something like `print(interval.directedName)`, which is not very intuitive and a bit hard to discover.

This PR implements the `__str__()` method so that intervals can be easily printed via `print(interval)`. The default representation `<music21.interval.Interval ...>` is not changed. Example output:

```python
>>> from music21 import *
>>> note1 = note.Note('C3')
>>> note2 = note.Note('G3')
>>> interval = interval.Interval(note1, note2)
>>> interval
<music21.interval.Interval P5>
>>>print(interval)
P5

### DIFF
--- a/music21/interval.py
+++ b/music21/interval.py
@@ -1738,13 +1738,16 @@ class Interval(base.Music21Object):
             self.isStep = self.isChromaticStep or self.isDiatonicStep
 
     def __repr__(self):
+        return "<music21.interval.Interval {}>".format(self.__str__())
+
+    def __str__(self):
         from music21 import pitch
         shift = self._diatonicIntervalCentShift()
         if shift != 0:
             micro = pitch.Microtone(shift)
-            return "<music21.interval.Interval %s %s>" % (self.directedName, micro)
+            return "%s %s" % (self.directedName, micro)
         else:
-            return "<music21.interval.Interval %s>" % self.directedName
+            return "%s" % self.directedName
 
     def isConsonant(self):
         '''


mscuthbert:
I'm fine with the change, but make sure that all tests pass and that the difference between repr() and str() is made clear and that similar changes take place for the lower-level interval classes, including for the case of chromatic intervals without a diatonic representation.

maxalbert:
Oops, I should indeed have run the test suite before submitting the PR. Sorry about that! I'm currently having trouble with some failing tests even on current master (They seem to be issues with the doctests, but I don't know what's causing them because they pass on Travis.) I probably won't have time to dig into this until later this week, but I'll try to tidy things up and add the methods to the other classes as well.

Btw, what do you mean by: "make sure that the difference between repr() and str() is made clear"? The distinction between them is explained in the [Python documentation](https://docs.python.org/2/reference/datamodel.html#object.__repr__) Are you suggesting to add a summary of this info to docstrings in the `__repr__()` and `__str__()` methods of the interval classes?

mscuthbert:
What I mean is, up to now the difference between Interval.__repr__() and Interval.__str__() is none, so if there's going to be a distinction now, it'd be good to demonstrate it in the docs.  Odds are that few people were trying to print(Interval) and expecting the <music21.interval.Interval P5> representation, but you never know.  Thanks!

PR #90, opened 2015-03-31
Add toplevel __version__ attribute

maxalbert:
This is in compliance with PEP 396 (see [specification](https://www.python.org/dev/peps/pep-0396/#specification))

### DIFF
--- a/music21/__init__.py
+++ b/music21/__init__.py
@@ -152,6 +152,7 @@ __all__ = [
 from music21 import base
 from music21.base import VERSION
 from music21.base import VERSION_STR
+from music21.base import VERSION_STR as __version__
 from music21.base import Music21Exception
 from music21.base import SitesException
 from music21.base import Music21ObjectException


mscuthbert:
Rejected.  Sorry. PEP 396 is not an accepted PEP and music21 has had its own version specification since before the PEP existed.

maxalbert:
That's fair enough. I completely agree that music21 should not change its versioning specification. But is there any downside in making `__version__` available as an additional way of *accessing* the version number? It won't break any existing code, and there is no potential for conflict since it just imports the existing versioning string under a different name.

I'm just asking because `__version__` is the de-facto standard amongst all the major Python packages and it took me a bit to find out the version of my music21 installation when I was looking for it, so it might be nice to allow this way of accessing the version number, too. But it's your decision of course.

mscuthbert:
You're right. No reason not to have this...

maxalbert:
Thanks!

PR #93, opened 2015-04-28
Fixed: installer.command updated to work on OS X

ahankinson:
Previously the installer command file was malformed and did not work on OSX (it complained of file not found). This was because the shell interpreted DIRFILE as a command, not a variable.

This commit fixes this by removing spaces around the DIRFILE variable assignment.


mscuthbert:
Apologies of missing this.  yes, definitely.

Issue #28, opened 2013-11-18
Mid-measure clef changes from MusicXML (to MusicXML also?)


mscuthbert:
Midmeasure clef changes are not being supported (should be a clef change after 1st and 3rd eighth notes)

Support file:
https://raw.github.com/cuthbertLab/cuthbertLab-supportFiles/master/mid_measure_clef_change.xml

mscuthbert:
Thanks Frank!

PR #95, opened 2015-05-02
musicxml mid-measure clefs

fzalkow:
Hi,
I solved the issue #28 for the case where I needed it. I don't know if it solves the problem in all respects (I don't know enough about the music21 internals for that), but for my cases it works.
Hope this helps.
Best,
Frank


mscuthbert:
Can you add a test case to the tests that shows how this works? It looks good to me but I don't want to include code without a test.

fzalkow:
Hi. So far I did only m21 to mx mid-measure clefs. Now I also added mx 2 m21 mid-measure clefs and a test file.

fzalkow:
Added the tests. Seems to work with multiple staves! :)

mscuthbert:
Looks good...going with the merge.  Thank you very much for contributing!

PR #97, opened 2015-06-05
Adding initial commit for Hindustani tala structure

swapnilgt:
Added a module inside the music21 package which defines the basic elements of a tala in Hindustani music:

1) Matra (smallest time unit in a tala cycle)
2) Vibhaga (partitions whose onset represent the emphasis points in a tala cycle - made up of Matra(s))
3) Avartan (a complete tala cycle - made up of Vibhaga(s))
4) TeenTal (object to represent one avartan of TeenTal)

### DIFF
--- a/music21/tala.py
+++ b/music21/tala.py
@@ -1,0 +1,152 @@
+'''
+This module is for adding support for talas support in music21 in Indian Art Music.
+'''
+from music21 import base
+from __builtin__ import int
+
+class Matra(base.Music21Object):
+    ''' 
+    This class represents the matra in a tala cycle 
+    '''
+    
+    khali = 0
+    tali = 1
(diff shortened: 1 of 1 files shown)


mscuthbert:
It's looking really good -- some comments:

[1] Can you add a usage scenario/demonstration to the top of the file? My goal for music21 is eventually to have everything have a demo of how a musicologist/researcher could use these tools.

[2] Can we move it from a top-level module to a sub module?  We can refactor instrument so that it becomes a folder with additional attributes (best if future commits will be for future instruments) or create a top-level "southAsia" folder to hold modules.  I'm trying to reduce the number of top-level modules (contour.py is on the chopping/moving list.  Thanks!

PR #101, opened 2015-07-22
Fixed LilyPond colors

ringw:
Hi, I noticed that the LilyPond output doesn't actually display colors in the `note.editorial.color` property. In fact, there were some errors on the current version, e.g.:

    >>> import music21.note
    >>> n = music21.note.Note('G')
    >>> n.editorial.color = 'red'
    >>> n.show('lily')
    GNU LilyPond 2.18.2
    Changing working directory to: `.../music21'
    Processing `.../music21/tmpyxX2wAly'
    Parsing...
    .../music21/tmpyxX2wAly:5:45: error: GUILE signaled an error for the expression beginning here
            \once \override NoteHead #'color = #
                                                (x11-color $color)
    ... (similar errors)

It still displays the note, but it is black. I'm not familiar with Scheme myself, but I changed the lines such as:

    \once \override NoteHead #'color = #(x11-color $color)

to use `color` instead of `$color` in the LilyPond output, and that seems to fix the errors and generate correct output.

### DIFF
--- a/music21/lily/translate.py
+++ b/music21/lily/translate.py
@@ -89,10 +89,10 @@ class LilypondConverter(object):
     colorDef = \
     r'''
     color = #(define-music-function (parser location color) (string?) #{
-        \once \override NoteHead #'color = #(x11-color $color)
-        \once \override Stem #'color = #(x11-color $color)
-        \once \override Rest #'color = #(x11-color $color)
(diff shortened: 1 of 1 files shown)


mscuthbert:
thank you! this used to work, but something changed at some point and I wasn't ever able to fix it.

PR #106, opened 2015-09-15
Changed normalization in MelodicIntervalHistogramFeature histogram should sum to 1

fzalkow:
In `features.jSymbolic.MelodicIntervalHistogramFeature` normalization was done the wrong way. Before it was done in such a way, that the maximum histogram-bin always was equal to one. Now it is done in such a way, that the sum of all histogram-bins is equal to one.

Relevant passage from the jMIR thesis:

> The histogram is then normalized, so that the magnitude of each bin indicates the fraction of all melodic intervals that correspond to the melodic interval of the given bin.

Cory McKay: *Automatic Music Classification with jMIR*, PhD 2010, p. 227, [PDF](http://www.music.mcgill.ca/~cmckay/papers/musictech/mckay10dissertation.pdf).


mscuthbert:
Thanks! We missed this definition -- I think we ended up putting M5's definition into M1.  Thanks!

PR #108, opened 2015-10-06
open rstFilePath explicitly in utf8 in IPythonNotebookReSTWriter.conv…

bagratte:
…ertOneNotebook when writing converted lines.


mscuthbert:
Thank you!  Wonderful.

PR #112, opened 2015-10-26
guess best enharmonic name for notes in chords

fzalkow:
I fixed an issue that bothered me. I am not sure if it is solved in the best way, but maybe this could find its way into music21 somehow?

The issue is the following: If I give a chord only pitch classes or midi names, it could take it in a very a-musical way. Here is an example where a B/C- major triad is totally mistaken:

    >>> from music21 import *
    >>> c = chord.Chord([11,3,6])
    >>> c.pitches
    (<music21.pitch.Pitch B>, <music21.pitch.Pitch E->, <music21.pitch.Pitch F#>)
    >>> c.pitchedCommonName
    'E--major triad'

I fixed it, by giving the Pitch a 'context'-keyword. This keyword gets a list of pitches and it is used for finding the best name for a pitch by maximizing the consonant intervals of the pitch with its context:

    >>> c = chord.Chord([11,3,6])
    >>> c.pitches
    (<music21.pitch.Pitch B>, <music21.pitch.Pitch D#>, <music21.pitch.Pitch F#>)
    >>> c.pitchedCommonName
    'B-major triad'

Feedback appreciated!

### DIFF
--- a/music21/chord/__init__.py
+++ b/music21/chord/__init__.py
@@ -243,19 +243,17 @@ class Chord(note.NotRest):
                     self.duration = n.duration
                     del keywords['duration']
                     quickDuration = False
-            elif isinstance(n, six.string_types):
+            elif isinstance(n, six.string_types) or isinstance(n, int):
                 if 'duration' in keywords:
                     self._notes.append(note.Note(n, duration=keywords['duration']))
                 else:
                     self._notes.append(note.Note(n))
                 #self._notes.append({'pitch':music21.pitch.Pitch(n)})
-            elif isinstance(n, int):
-                if 'duration' in keywords:
-                    self._notes.append(note.Note(n, duration=keywords['duration'], context=self._notes))
-                else:
-                    self._notes.append(note.Note(n), context=self._notes)
             else:
                 raise ChordException("Could not process input argument %s" % n)
+
+        if all(isinstance(n, int) for n in notes):
+            self.simplifyEnharmonics(inPlace=True)
 
         if quickDuration is True:
             del keywords['duration']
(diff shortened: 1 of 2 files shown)


mscuthbert:
THANK YOU! This has bugged me for a while.  I love the efficiency and the solution (and having tests); I don't like, however, adding a "context" attribute to Note -- I'm trying to make Note simpler and I believe that this will make it more complex for only a single use case as it's currently written.

What I suggest is that we make it a little less efficient, but more powerful and conceptually simpler -- have the Note in the chord be created as before, let's make in pitch.py (but not in pitch.Pitch object) a simplifyMultipleEnharmonics routine that can be quite powerful.  It takes in a list of pitch objects (not Notes) or numbers (pitch class or MIDI) and it can have different selection criteria, of which what could be the default for now (and only one that would need to be implemented for this pull request unless you wanted to do more) could be 'optimizeConsonance'.  It should take an optional Key or KeySignature object as well.  It should return a list of pitches so that there's a one-to-one correspondence between the original pitch index and the new one.  Probably all pitches should be new objects even if no change needs to be made to the pitch.

Chord could have its own method "simplifyEnharmonics" (default inPlace=False, but called with True on __init__) that takes care to replace the pitches on the Notes. Then this routine could be called from Chord if all(pitches) are numbers to change the pitches of the notes.  Mixtures of numbers and pitch names/Note objects would not get optimized (who would do that?).  It would allow for all notes to be considered before making a determination rather than determining the best spelling at any given moment.

The reason I want this approach, besides leaving Notes out of it, is that it could be used, for instance, to spell MIDI input as best as possible (something I desperately need).  A quick parse could allow for determining the key of the piece (.analyze('key')) and this could be used to minimize written accidentals (using the KeySignature to determine which accidentals would need to be written). Another approach could be to first minimize augmented intervals, then to optimize for consonance, then to minimize unisons, then seconds, then minimize diminished intervals.  For writing out a chord, the minimization may be most important.  C# + Eb may be preferable to C#-D# even though the first is a diminished interval and the latter is a M2.  Both should be preferred to Db-D# of course.

This is something that I've wanted/needed/desired for some time but haven't gotten around to it.  I hope that it's okay that I'm asking for it to do a bit more even though the current approach definitely solves your problem. Thanks!

mscuthbert:
Oh, and as the failed test will make you aware -- pitch classes (0-11) should come without an octave number, while MIDI numbers (12-127) should have octaves afterwards.  Be careful (and test) that a B#4 gets properly flipped to C5 (and vice versa for C-5 to B4); the current flipEnharmonic routines make sure that this works.

fzalkow:
Thanks, that are great suggestions!

I have tried to implement that (without the minimization of accidentals yet) and am open for further suggestions.

mscuthbert:
AHA! I missed that it was the thing removed.  Okay!  If you could add a test with a key to show that it works, that'd be great.  

When we're talking nanoseconds, it doesn't matter too much.  But here's the apples to apples comparison of time since you'll need to reimport music21.key just to do the isinstance check:

```
In [1]: k = key.KeySignature()

In [2]: %timeit 'KeySignature' in k.classes
The slowest run took 68.17 times longer than the fastest. This could mean that an intermediate result is being cached 
1000000 loops, best of 3: 396 ns per loop

In [3]: %timeit from music21 import key; isinstance(k, key.KeySignature)
The slowest run took 6.42 times longer than the fastest. This could mean that an intermediate result is being cached 
1000000 loops, best of 3: 1.57 µs per loop
```

In music21 we use 'X' in classes in order to avoid needing to import extra modules just to do isinstance checking -- it's just the paradigm we've employed throughout.  Thanks!  If we can get the last test passing, I'll accept and merge.  It's great! And thanks for the added comments in the code.

fzalkow:
Ok, thanks. Also thanks for your clarification with the timing of the class check!
I'm happy to contribute!

mscuthbert:
thanks Frank! btw -- would you mind submitting a dummy pull request (change a space somewhere or something... :-)  )  -- I think I've integrated pylint into the pull request system to catch bugs that the tests don't, and I'd like to test whether it works, but it's hard for me to do that without creating a new account, forking, etc. Thanks!

fzalkow:
Done, see #114. :-)

mscuthbert:
Thanks! I have pylint grumbling on multiple newlines turned off, but just that it checked it was the most important thing!

PR #114, opened 2015-10-31
dummy pull request, added newlines

fzalkow:
A dummy pull request (for why see #112)


mscuthbert:
Thanks Frank -- this is going to be awesome.  Apparently Scrutinizer only checks for new issues, so I can turn on some of the code checks that the current code breaks and fix them as I have time but still prevent issues of the same type from appearing in changed code.  Brilliant! Thanks so much.

PR #116, opened 2015-11-09
custom consonance function for simplifyMultipleEnharmonics

fzalkow:
I spot an undesirable behaviour in `simplifyMultipleEnharmonics`. Since it relied on `interval.isConsonant()`, e.g. it considered a perfect fourth and an augmented third equal -- both are non-consonant. So I wrote a simple custom consonance function.

Before:

    In [1]: pitch.simplifyMultipleEnharmonics([3, 8, 0])
    Out[1]: [<music21.pitch.Pitch E->, <music21.pitch.Pitch G#>, <music21.pitch.Pitch C>]

Now:

    In [1]: pitch.simplifyMultipleEnharmonics([3, 8, 0])
    [<music21.pitch.Pitch E->, <music21.pitch.Pitch A->, <music21.pitch.Pitch C>]

### DIFF
--- a/music21/chord/__init__.py
+++ b/music21/chord/__init__.py
@@ -1394,10 +1394,10 @@ class Chord(note.NotRest):
         '''Return a Z relation if it exists, otherwise return None.
 
         >>> chord.fromIntervalVector((1,1,1,1,1,1))
-        <music21.chord.Chord C C# E F#>
+         <music21.chord.Chord C D- F- G->
 
         >>> chord.fromIntervalVector((1,1,1,1,1,1)).getZRelation()
-        <music21.chord.Chord C C# E- G>
+        <music21.chord.Chord C D- E- G>
 
 
         >>> chord.Chord('C E G').getZRelation() is None
@@ -4399,7 +4399,7 @@ def fromForteClass(notation):
     <music21.chord.Chord C E- G>
 
     >>> chord.fromForteClass((11,1))
-    <music21.chord.Chord C C# D E- E F F# G G# A B->
+    <music21.chord.Chord C D- D E- E F G- G A- A B->
 
     '''
     card = None
@@ -4446,13 +4446,13 @@ def fromIntervalVector(notation, getZRelation=False):
     True
 
     >>> chord.fromIntervalVector((1,1,1,1,1,1))
-    <music21.chord.Chord C C# E F#>
+    <music21.chord.Chord C D- F- G->
 
     >>> chord.fromIntervalVector((1,1,1,1,1,1), getZRelation=True)
-    <music21.chord.Chord C C# E- G>
+    <music21.chord.Chord C D- E- G>
 
     >>> chord.fromIntervalVector((1,1,1,1,1,1)).getZRelation()
-    <music21.chord.Chord C C# E- G>
+    <music21.chord.Chord C D- E- G>
 
     '''
     addressList = None
--- a/music21/chord/__init__.py
+++ b/music21/chord/__init__.py
@@ -1394,7 +1394,7 @@ class Chord(note.NotRest):
         '''Return a Z relation if it exists, otherwise return None.
 
         >>> chord.fromIntervalVector((1,1,1,1,1,1))
-         <music21.chord.Chord C D- F- G->
+        <music21.chord.Chord C D- F- G->
 
         >>> chord.fromIntervalVector((1,1,1,1,1,1)).getZRelation()
         <music21.chord.Chord C D- E- G>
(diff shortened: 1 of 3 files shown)


mscuthbert:
The output in chord.py seems sane to me now, so worth fixing the test results to fit the output. I wonder though if in the case of a tie if the "more common" enharmonic could be used (the list of most common are defined somewhere in pitch.py or interval.py). In the absence of other information, I prefer F# to G- and C# to D-.  But perhaps it's that lone B- that is causing the favoring of flats in the intervalClass(11,1) calculation.

fzalkow:
Ok, had another idea: I implemented a function to get the ratio of an interval in pythagorean tuning. (Very naive way. Don't know if there is a better one?)

This can than be used as a consonance function, just the inverse of the product of both ratio numbers.

According to the more common enharmonics... Hm, I think that would just be possible, if `simplifyMultipleEnharmonics` would not be called by default on chords. Or do you have some ideas?

mscuthbert:
chord.postTonalChordsA has stalled -- infinite loop?  Have you done timing checks on the Pythagorean routine? I suspect that creating 6+ Interval objects per call is likely to be very expensive.  It might be better in a separate module because there's a lot of disagreement about "correct" Pythagorean tuning of intervals once we get beyond the basic few (e.g., is M3 5:4 or 81:80? I think the latter).

mscuthbert:
There's a (now deleted) module called ratios.py that you can find in the codebase somewhere (probably still included up to 1.7 or 1.9.3) that has some basic ideas on what a Pythagorean tuning support system could have.  It wasn't well integrated into M21 so I removed it.

fzalkow:
I didn't found ratios.py, but I fixed my bugs anyway. Hm, conceptually Pythagorean tuning in music21 should be straightforward, I think. Just the shortest way of stacked fifth upwards or downwards, followed by the octave adjustment. M3 should definetly be 81:64, I think.

The infinite loop was a bug. But I introduced a cache for performance nevertheless.

Do you think that should go into a seperate module or should it stay in interval.py?

mscuthbert:
It sounds good to leave as it is now so almost a merge. But can you put into while not_found_flag a bailout if counter gets unreasonably big (say to triple or quadruple flats). I do this for all while statements and I believe that it'd still be possible to trigger an infinite loop here: try testing on a quarter-tone accidental. :-)

fzalkow:
Ui, you are completly right. Didn't thought of something other than semitones. :-)

The counter is now restricted to max. 36 -- than stacking fifth wrap back to its origin because of music21's limiting of accidentals.

Because this function is called anytime when a chord is constructed from numers, I don't want to throw an Error... Now I just print a warning and return `nan`. Is that good practice?

mscuthbert:
Tell me when you're sick of tweaking and I'll accept and finish myself. :-) But it's awesome working with you this way and I know we have plans to work more closely, so figure it's a good chance for each other to influence the other's style (for instance, you've already convinced me that it doesn't matter if internal variables are underscore_connected as opposed to camelCase).

My preference is to have the function raise an exception (with the message of the print statement) if it can't handle the input and then have the chord-spelling/pitch-spelling routine catch the exception and handle it in some elegant way such as simply returning the original pitch spelling without simplifying. (Is that even possible to raise the exception from Chord? do we check if the pitch specifiers are numbers or Ints?) If we're printing and returning NaN then it'll be hard to figure out what's going wrong later when we suddenly have a chord missing the microtones.  Thanks!

fzalkow:
Not at all! This is fun and I am learning much about music21's internals! So let's keep continue this way! (Of course I cannot guarantee to do things promptly if it doesn't relates to my work, but all my pull requests a very relevant for my work.)

Yeah, I am used to underscore_names for variables and even functions. But now I am aware that this isn't music21-style. If you think, I should change the variable names, just say so!

I think your idea with the exception is great. One does not need to return anything if the exception is catched. The consonance-score in this case is just 0 and if there are alternative pitch spellings with a better score, it will probably use one of them.

To your questions: 1. It would be possible to raise the exception from Chord, but I see no reason for that? 2. In `simplifyMultipleEnharmonics` we convert non-Pitch-instances to Pitch instances for getting them spelled the best way. In `chord.__init__` we only use this procedure if all notes are ints.

mscuthbert:
Looks great to me.  I used camelCase in m21 only because I came from Perl where it was the norm. Since then I've read that underscores are more accessible to foreign language readers and to screen readers (we have at least one blind contributor to music21) so I wish I had used them from the start, but it's too late to change the standard for m21 and I don't want to mix them in the API. That said, there's no problem with them for function-internal variable names.

Merging. Thanks!

PR #121, opened 2015-12-01
Error with subito when getting scalar of dynamics

fzalkow:
In the case of subito dynamics I came across this little bug:

    d = dynamics.Dynamic('sfp')
    n = note.Note('C4')
    s = stream.Stream()
    s.append(d)
    s.append(n)
    n.volume # TypeError: unsupported operand type(s) for *: 'NoneType' and 'float'

I fixed this by ignoring subito in any case. Not sure if this is the very best solution? It fixes the error anyway.

(The small fix in the documentation is unrelated.)

### DIFF
--- a/music21/documentation/source/usersGuide/usersGuide_04_stream1.ipynb
+++ b/music21/documentation/source/usersGuide/usersGuide_04_stream1.ipynb
@@ -173,7 +173,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "But suppose you had thirty notes?  Then it'd be a pain to type \"`print noteX.step`\"\n",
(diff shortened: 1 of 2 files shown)


mscuthbert:
Thanks! 

I think though the underlying problem is that we can somehow generate dynamics that create float * None comparisons.  If you do Dynamic('asdfjka') it should be meaningless, but not cause an Exception.  I'll merge this and then make another fix for that.

Btw -- I don't think the "s" here generally stands for "subito" as in "subito forte-piano" which would be redundant-- It's "sf" + "p" or "sforzando" + "piano".  

Thanks also for catching the print statement/function!

Issue #125, opened 2015-12-14
MEI transposition information of instruments not parsed


sonovice:
It seems like the transposition information of instruments in an MEI file are not parsed.
Trying to get the actual sounding pitch of the notes always results in `music21.exceptions21.StreamException: atSoundingPitch is unknown: cannot transpose`

The MEI file does include the transpositions in a correct manner, for example `trans.semi="-7"`.

Minimal example:
```python
stream = converter.parseFile('Hummel_Concerto_for_trumpet.mei', None, "mei", True)
stream.toSoundingPitch(inPlace=True)
```

The MEI file used: [Hummel Concerto](http://www.verovio.org/examples/downloads/Hummel_Concerto_for_trumpet.mei)

crantila:
Thanks for this report, @sonovice. It looks like music21 isn't pulling in any of the instrument information in that file. I'm investigating as we speak!

crantila:
After some investigation, I can reproduce the issue on the master branch, and on the branch in pull request #126.

As for finding the problem, ``music21.mei.base.staffGrpFromElement()`` appears to return ``Instrument`` objects as expected, so the problem must be at a higher level in the score hierarchy, or a later point in processing. I can also reproduce the issue with a modified version of the built-in "test_file.mei" that adds a ``@label`` attribute to the ``<staffDef>`` and should therefore produce a named ``Instrument``.

I know this used to work, so the next step is a git bisection to find the commit that broke things.

mscuthbert:
I know our transposition routines but I don't know the MEI code well at all.  Ping me if I can help.

sonovice:
Any progress on this? Sorry for bumping randomly... If i can help in any way, please let me know.

crantila:
No progress, but you can help by fixing it. :smile: 

I'll take a look today. This isn't a high priority for me at the moment, so I will forget about it. Therefore please do ask periodically if there's no activity on this issue!

crantila:
Okay, in today's episode, I determined that instrument data was never properly imported. I really thought it was too.

sonovice:
Periodical bump ;-)

crantila:
I added a test in that there commit 0664941. I also used the small MEI document in that test to find the problem.

Currently, the MEI importer puts Instrument objects in Measure objects, like this:
```
>>> zed.parts[0]
<music21.stream.Part 0x7f10731b2710>
>>> zed.parts[0].show('text')
[...]

PR #126, opened 2015-12-17
MEI Improvements

crantila:
Before working on cuthbertLab/music21#125, I thought I should submit a pull request for all of this stuff. I don't really remember doing this work, but it seems like I added support for lyrics, ``<space>`` elements, and ``<barLine>`` elements, and rewrote and added tests for score-wide metadata handling.

Of course, suggestions for revision are welcome.


mscuthbert:
Looks excellent. Thanks Chris! Merging.

PR #128, opened 2015-12-21
simplifyMultipleEnharmonics with better scoring

fzalkow:
I tried to improve the scoring in simplifyMultipleEnharmonics. Before it just took the Pythagorean ratio into account. Now it also takes into account that double-accidentals should be avoided and there thirds should be rewarded.

I also refactored a bit.

Myke, you also suggested some Monte Carlo method for doing better than greedy-left-to-right. As far as I know Monte Carlo would be stochastic and so there could be a different result every time. So a deterministic global optimization method would be needed... I am not having an idea yet, but maybe something would be possible.

### DIFF
--- a/music21/chord/__init__.py
+++ b/music21/chord/__init__.py
@@ -1394,7 +1394,7 @@ class Chord(note.NotRest):
         '''Return a Z relation if it exists, otherwise return None.
 
         >>> chord.fromIntervalVector((1,1,1,1,1,1))
-        <music21.chord.Chord C D- F- G->
+        <music21.chord.Chord C C# E F#>
 
         >>> chord.fromIntervalVector((1,1,1,1,1,1)).getZRelation()
         <music21.chord.Chord C D- E- G>
@@ -4444,7 +4444,7 @@ def fromIntervalVector(notation, getZRelation=False):
     True
 
     >>> chord.fromIntervalVector((1,1,1,1,1,1))
-    <music21.chord.Chord C D- F- G->
+    <music21.chord.Chord C C# E F#>
 
     >>> chord.fromIntervalVector((1,1,1,1,1,1), getZRelation=True)
     <music21.chord.Chord C D- E- G>
--- a/music21/chord/__init__.py
+++ b/music21/chord/__init__.py
@@ -3391,9 +3391,7 @@ class Chord(note.NotRest):
         else:
             returnObj = copy.deepcopy(self)
 
-        pitches = pitch.simplifyMultipleEnharmonics(self.pitches, 
-                                                    criterion='maximizeConsonance', 
-                                                    keyContext=None)
+        pitches = pitch.simplifyMultipleEnharmonics(self.pitches)
         returnObj.pitches = pitches
 
         if inPlace is False:
--- a/music21/interval.py
+++ b/music21/interval.py
@@ -495,7 +495,7 @@ def intervalToPythagoreanRatio(intervalObj):
 
         _pythagorean_cache[end_pitch_wanted.name] = end_pitch, ratio
 
-    octaves = int((end_pitch_wanted.midi - end_pitch.midi)/12)
+    octaves = int((end_pitch_wanted.ps - end_pitch.ps)/12)
     return ratio * Fraction(2, 1) ** octaves
 
 #-------------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


mscuthbert:
Thanks, looks awesome.  Can you import itertools and be sure that that's working.  Thanks!

fzalkow:
Oops. Forgot to import itertools.

Wait a moment before accepting... I'll have a look if [`scipy.optimize.linprog`](http://docs.scipy.org/doc/scipy-0.16.0/reference/generated/scipy.optimize.linprog.html) could be used for optimization...

mscuthbert:
Scipy isn't a required part of music21 so make sure that it's optional and that there's a decent work-around if it doesn't exist. Thanks!

fzalkow:
Before I go further into this: Is it a valid option, that in the case scipy is installed, global optimization is used and in other cases, it stays with the greedy method? That means pitch spellings results will be different depending if scipy is installed.

mscuthbert:
Thinking more about this, it seems like the quick and dirty and no-external libraries, should be used no matter whether SciPy is installed or not, so that the results will always be the same. Then if you'd like, we could have a much better algorithm in a module in analysis that could be called explicitly for users who want the best matches. That seems to be more the music21 way.

I'd like to put the cap on the run time of the non-scipy quick and dirty routine at 100-microseconds if possible.  Do you have a sense of where it's at now?  Thanks!

(oops division by zero -- should probably short circuit early if len(pitches) is 0 or 1)

fzalkow:
Yes, ok. As far as I see, there wouldn't be a a simple out-of-the-box method in scipy, anyway. At least I have no idea if this problem could be formulated as an integer programming problem.

To your runtime question: I fear we are way slower... Of course it depends on the number of pitches, because each interval is checked, so it is kind of quadratic behaviour. 

On my machine:

    In [2]: %timeit pitch.simplifyMultipleEnharmonics([0])
    The slowest run took 15.43 times longer than the fastest. This could mean that an intermediate result is being cached 
    100000 loops, best of 3: 12.5 µs per loop

    In [3]: %timeit pitch.simplifyMultipleEnharmonics([0, 4])
    The slowest run took 4.99 times longer than the fastest. This could mean that an intermediate result is being cached 
    100 loops, best of 3: 1.87 ms per loop

    In [4]: %timeit pitch.simplifyMultipleEnharmonics([0, 4, 7])
    100 loops, best of 3: 10.8 ms per loop

    In [5]: %timeit pitch.simplifyMultipleEnharmonics([0, 4, 7, 10])
    10 loops, best of 3: 56.1 ms per loop

    In [6]: %timeit pitch.simplifyMultipleEnharmonics([0, 4, 7, 10, 2])
    10 loops, best of 3: 22.9 ms per loop

    In [7]: %timeit pitch.simplifyMultipleEnharmonics([0, 4, 7, 10, 2, 5])
    10 loops, best of 3: 37.8 ms per loop

The decrease in runtime form 4 to 5 pitches corresponds to a switch from brute-force to greedy.

One possibility for runtime would be always using the greedy version. It seems to work quite fine in general. (That would boost to 100-microseconds)

mscuthbert:
Do you know about the base40 module in the musedata folder? This was something that Walter Hewlett figured out back in the days when computing power and memory was short -- it's a system for getting exact intervals and enharmonics using an extremely simple system where notes are encoded as a number between 1-40 for each octave (encompassing up to double flats and double sharps) and intervals are encoded as a number between 0-39 and the mod40 difference between two note's base40 representation gives the interval (up to augmented and diminished, though the system could be expanded in most cases to doubly augmented and doubly diminished, I believe).

I think that this could by used to extremely quickly calculate potential names.  Enharmonic names (up to double sharp and flat) are always 4 off on either side. String lengths give the number of accidentals. Etc. the only hard thing is to calculate the octave for the note in cases where the octave is not implicit, but that wouldn't be too hard.  I might merge and then try it.

mscuthbert:
Here is some work towards a fast solution.  Look at base40.py now:

```
In [1]: from music21 import *

In [2]: p = pitch.Pitch('C#')

In [3]: %timeit musedata.base40.quickEnharmonicString(p.name)
The slowest run took 5.26 times longer than the fastest. This could mean that an intermediate result is being cached 
100000 loops, best of 3: 7.81 µs per loop

In [5]: %timeit p.getAllCommonEnharmonics()
1000 loops, best of 3: 757 µs per loop
```
This routine doesn't give a pitch object in the end, just a string, and doesn't work for very exotic notes (C###), but it could give a huge speedup in intermediary calculations.

PR #132, opened 2016-01-07
new branch for having simplifyMultipleEnharmonics based on base40 module

fzalkow:
Hi,

as discussed with Michael, I have rewritten the `simplifyMultipleEnharmonics` procedure, to internally use `base40` pitches for some speed up. I have not replaced the old one and haven't written docstrings, tests, etc. because I first would like to hear your opinion.

Here a speed comparison:

```
In [1]: %timeit pitch.simplifyMultipleEnharmonics(['C4', 'D#4', 'G4', 'A#4'])
10 loops, best of 3: 42.1 ms per loop

In [2]: %timeit pitch.simplifyMultipleEnharmonics_b40(['C4', 'D#4', 'G4', 'A#4'])
100 loops, best of 3: 16.7 ms per loop
```

This example is about 2.5x faster in the new version. But it pays the price of the code being less easy to understand and to maintain by other users as well as being less extensible. So I am not quite sure what weights more?

The other thing is, that base40 has some limitations. Obviously, more than two accidents are not possible. Furthermore currently not all kinds of intervals as well as notes in octave below 0 are not possible. (That is something that could be solved, somehow...)

Should I replace the old version and put in docstrings, etc.?
Because I am less sure about this than with my pull requests before, I have moved it to a separate branch.

Best,

Frank

### DIFF
--- a/music21/chord/__init__.py
+++ b/music21/chord/__init__.py
@@ -4395,7 +4395,7 @@ def fromForteClass(notation):
     <music21.chord.Chord C E- G>
 
     >>> chord.fromForteClass((11,1))
-    <music21.chord.Chord C D- D E- E F G- G A- A B->
+    <music21.chord.Chord C D- D E- F- F G- G A- A B->
 
     '''
     card = None
--- a/music21/musedata/base40.py
+++ b/music21/musedata/base40.py
@@ -106,6 +106,18 @@ base40Representation = {'C--': 1,
                         'B##': 40,
                         }
 
+base40FromMidi =   {0: 3,
+                    1: 4,
+                    2: 9,
+                    3: 14,
+                    4: 15,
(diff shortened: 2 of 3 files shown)


mscuthbert:
hmmm... you're right.  I'm really shocked it's not tons and tons faster.  let me try a bit to play with some speed cases.

mscuthbert:
This is still on my agenda.  Apologies for the time slowing on it.

mscuthbert:
Hi @fzalkow ! Look at #253 -- we used some of this code to create a new simplifyEnharmonics version also using base40  -- you were totally on the right track and @MarkGotham and I couldn't've done it without you.  Thank you!

fzalkow:
Nice to see, that it found some use! Great work, guys! :+1:

PR #134, opened 2016-01-20
Give default volumeScalar to unknown dynamics

fzalkow:
I made a simple fix to give the default volumeScalar to any dynamic string that is unknown. Related issue: #121. Example:

    from music21 import *
    d = dynamics.Dynamic('sfz')
    n = note.Note('C4')
    s = stream.Stream()
    s.append(d)
    s.append(n)
    n.volume

Before this returned an error, now it returns `<music21.volume.Volume realized=0.99>`.


mscuthbert:
aweesome.....!!!!

Issue #131, opened 2016-01-03
Py3In2OutputChecker is a bit too naive about removing b' from strings


shimpe:
This is the reason the python 2.7 doctest for pull request https://github.com/cuthbertLab/music21/pull/130 fails.

The expected fragment contains b' (a lilypond note) and this b' is stripped away by Py3In2OutputChecker, presumably because it is considered the start of a byte string.

I will change the doctest to work around this problem, but it's an interesting bug nevertheless.

shimpe:
(probably not very important to solve, but it 's important to be aware of at least)

mscuthbert:
Ah thanks. I think the solution is probably to have the checker run the tests with the fixups and then if it fails to try again without the fixups. I can put in a patch for that.

Sometime this year I'm going to turn off doctests on Python 2 or at least always have the expected output match Python 3 and turn off any fixups for Py3.

The big pain with Py3 and doctests is that dictionary keys are usually in the same order on one computer but can be different on other computers.

Thanks for solving this. Now that you have I remember this occurring with Lilypond tests in the past.

Best,
Michael

---
Sent from a phone (413-575-6024)
Apologies for typos or general brevity

On Jan 3, 2016, at 13:31, shiihs <notifications@github.com<mailto:notifications@github.com>> wrote:


This is the reason the python 27 doctest for pull request https://githubcom/cuthbertLab/music21/pull/130 fails

The expected fragment contains b' (a lilypond note) and this b' is stripped away by Py3In2OutputChecker, presumably because it is considered the start of a byte string

I will change the doctest to work around this problem, but it's an interesting bug nevertheless

shimpe:
To be clear, I haven't solved this problem. I just changed my doctest to work around it.
If you plan to disable py2 doctests in the foreseeable future, perhaps it's not worth solving.

mscuthbert:
Fixed in #136

PR #136, opened 2016-01-20
let binary fixup tests pass...

mscuthbert:
Tests on Python2 now pass if the doctest output passes either after compensating for Py3->Py2 differences, OR if the unaltered code passes.

The main place where this was a bug was in Lilypond output of " b' " meaning b above middle c, but which was being read as the start of a Py3 binary string.

Now there could still be places where some strings should be read as binary and some as not, but this seems like a good start...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ sudo: false
 python:
     - "2.7"
     - "3.3"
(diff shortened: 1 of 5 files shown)


mscuthbert:
Fixes #131 .

PR #138, opened 2016-01-24
For issue #125

crantila:
As usual, I would be happy to make any style or functional requirements as requested. Thanks in advance!


mscuthbert:
This is great.  Thanks Chris.  I'll marge and then see about changing .getInstruments()

PR #143, opened 2016-02-22
new branch from cuthbertLab/master with omr midi changes

emzhang:
started new branch with omr/midi fixings


emzhang:
just to see where i am

PR #155, opened 2016-04-21
bugfix: metronomeMarkBoundaries for multiple MetronomeMarks in nested stream

fzalkow:
Consider the following:

    s = stream.Score()
    m1 = stream.Measure()
    m2 = stream.Measure()
    m1.append(tempo.MetronomeMark('Adagio'))
    m1.append(note.Note('D4', quarterLength=4))
    m2.append(tempo.MetronomeMark('Allegro'))
    m2.append(note.Note('C4', quarterLength=4))
    s.append(m1)
    s.append(m2)

Calling `s.metronomeMarkBoundaries()` raised this error because the MetronomeMarks have been searched in a non-flatted stream:

    an entry for this object <music21.tempo.MetronomeMark Adagio Quarter=56> is not stored in stream <music21.stream.Score 0x7f0fcde08cc0>

Now it correctly returns:

    [(0.0, 4.0, <music21.tempo.MetronomeMark Adagio Quarter=56>),
     (4.0, 8.0, <music21.tempo.MetronomeMark Allegro Quarter=132>)]

### DIFF
--- a/music21/stream/__init__.py
+++ b/music21/stream/__init__.py
@@ -7371,7 +7371,7 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
         else:
             offsetPairs = []
             for ti in tiStream:
-                o = ti.getOffsetBySite(srcObj)
+                o = ti.getOffsetBySite(srcFlat)
                 offsetPairs.append([o, ti.getSoundingMetronomeMark()])
             # fill boundaries
             # if lowest region not defined, supply as default


mscuthbert:
This is obviously the correct solution for the problem here, but "metronomeMarkBoundaries" is really too specific to belong in Stream; "srcObj" is strange strange to have in the call signature, and working on .flat. is generally a no-no in Stream parlance.  I think that this was one of those methods that me or Chris wrote really quickly just as a paper due date was coming up that isn't thought through nearly as  well as it could have.  

Curious about what use you're having for it?  Thanks!

fzalkow:
I write an extractor of chromagram features. It is already done for "musical" time units (beats or measures), but we also want it for "physical" time units. So with the help of `.secondsMap` (which in turn uses `metronomeMarkBoundaries`) I estimate the physical time points for begin and end of all notes.

Why is working on `.flat` a no-no? I use `.secondsMap` on flatted parts. Is there anything wrong with this?

mscuthbert:
Cool.  that's really sweet work.

As far as `.flat` goes, there's totally nothing wrong with the user using `.flat` in his or her own work -- I use it all the time (though `.recurse()` has passed it in terms of frequency).  It is that individual methods in `.stream()` should not, generally, be working with `.flat`.  

Here's why: say there's a method called "gCounter(self)" that counts the number of (pitch) G's in a Stream.  There's two possible interpretations for what this might mean: count all the Gs in notes/chords in the stream, or count all the Gs in the notes/chords in the stream and any substreams (i.e., `.flat`).  The second one might sound more useful, so it might make sense to have `gCounter()` begin with something like `sf = self.flat`.  But what would happen if you wanted the former (no-substream) interpretation?  Then there's no way to do it.  If, however, `gCounter()` worked only on the stream and no substreams, then it would be easy to do either version: call `self.gCounter()` to get the no-substream version, and `self.flat.gCounter()` to get the substream-aware version.

If a method needs to do something more sophisticated (such as reaching into substreams but still be passed the hierarchical version of the stream), like I think `.secondsMap()` does, then it should have a `{recurse=True}` configurable option, or something like that.  I might get to patching this soon, but I'll make sure that the interface to `.secondsMap` doesn't change without letting you know. :-)

Thanks!

fzalkow:
Makes sense. Thanks a lot for the clarification! :-)

PR #161, opened 2016-05-12
changed how quarterLengthDivisor is handled

emzhang:
quarterLengthDivisor is now set in **keywords when parsing a midi file,
so you can pass it in as an argument. if it’s not set, the default of
[4, 3] is used.


mscuthbert:
Great! :-)

Issue #165, opened 2016-05-19
Instrument parsing unstable


sonovice:
One of the latest commits changed the parsing of instrument names leading to some new errors.
This is a minimal MEI file with the instrument label "Clarinetti in A", which is obviously not a key in the `instrumentNames` dictionary.

``` xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<?xml-model href="http://www.music-encoding.org/schema/2.1.1/mei-all.rng" type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
<?xml-model href="http://www.music-encoding.org/schema/2.1.1/mei-all.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" xmlns:xlink="http://www.w3.org/1999/xlink" meiversion="2013">
    <meiHead>
        <fileDesc>
            <titleStmt>
                <title/>
            </titleStmt>
            <pubStmt/>
        </fileDesc>
    </meiHead>
    <music>
        <body>
            <mdiv>
                <score>
                    <scoreDef meter.count="4" meter.unit="4" meter.sym="common" key.sig="3s" key.mode="major">
                        <staffGrp n="1" barthru="false">
                            <staffDef n="1" lines="5" label="Clarinetti in A" key.sig="0" trans.semi="-3" trans.diat="-2" clef.line="2" clef.shape="G"/>
                            <staffDef n="2" lines="5" label="Piano" clef.line="2" clef.shape="G"/>
                        </staffGrp>
                    </scoreDef>
                    <section>
                        <measure n="1" xml:id="d23378e68">
                            <staff n="1">
                                <layer n="1">
                                    <note xml:id="d23378e100" pname="d" accid="s" oct="5" dur="4"/>
                                </layer>
                            </staff>
                            <staff n="2">
                                <layer n="1">
                                    <note xml:id="d23378e325" pname="c" oct="4" dur="4"/>
                                </layer>
                            </staff>
                        </measure>
                    </section>
                </score>
            </mdiv>
        </body>
    </music>
</mei>
```

[(Download)](https://www.dropbox.com/s/a8melv3l1ct0xwn/minimal.mei?dl=1)

Trying to read this file and export it to MIDI gives 2 warnings, that completely screw up the file:

```
[...]

PR #166, opened 2016-05-20
fixed import of mei files with unknown instruments

sonovice:
fix for #165


sonovice:
I accidently added my test file to the repository, sorry. First steps with git...

music21/mei/base.py
```diff
@@ -1751,20 +1751,14 @@ def staffDefFromElement(elem, slurBundle=None):  # pylint: disable=unused-argume
     if post is not None:
         post = {'instrument': instrDefFromElement(post)}
     else:
-        try:
-            post = {'instrument': instrument.fromString(elem.get('label', ''))}
```
ahankinson: I'm not entirely sure this is correct. If `@label` contains a valid instrument name it should still return the valid instrument, and only fall back to "UnspecifiedInstrument" if it can't find it.

sonovice: Thanks, that should indeed be the usual behaviour.

ahankinson:
You should probably also include a unit test for this particular functionality (an MEI file or fragment with a bogus instrument and verify that music21 parses it out to an unknown instrument...)

mscuthbert:
Is this PR still active?  Agreeing with @ahankinson that a test is needed.

sonovice:
I will look into this. Time is limited right now though...

mscuthbert:
Just a ping on whether you'd like to try to get this into v.4, wait for v.5 or close?  Thanks!

mscuthbert:
Closing for now.  Reopen please if work becomes active.

Issue #85, opened 2015-03-15
No information how to run the tests


maxalbert:
I couldn't find any info on how to run the tests in the repository. It would be helpful to mention this in the toplevel README file. Alternatively (or even better in addition) this could be mentioned in the [Developer Reference section](http://web.mit.edu/music21/doc/index.html) of the documentation. It may also be useful to have some quick guidelines in that section on how to write new tests and where they should be added.

mscuthbert:
It's at http://web.mit.edu/music21/doc/developerReference/usingGit.html?highlight=multiprocesstest at the bottom.  I will separate this out to another document since people not using Eclipse will miss it.  It doesn't belong in the top level README file since we have about 50 people using the system for everyone developing it. (regrettably).

maxalbert:
Ah thanks, I missed that. Yes, I agree it would be good to separate it out and place it somewhere more prominent (when I searched for "test" in the Developer Reference section there was no match). I can see that the toplevel README file is not the right place for this kind of information, but it could be a good idea to add another README file in `music21/test` because just from looking in that directory it's not obvious which file is the main test file to run (for example, I personally would never have thought of running `multiprocessTest.py`). I wasn't even sure whether it's the correct directory to look at because when I grepped for "test" in the code base it also returned a lot of matches in the regular code files, so I thought that perhaps running "py.test" from the toplevel directory would be enough (and indeed it returns 172 passed tests, but `multiprocessTest.py` appears to run some additional ones).

mscuthbert:
Max -- it's probably not so high priority any more now that TravisCI will flag any pull requests with problems and point to the test program.

maxalbert:
Fair enough. It's still nice to be able to run the tests locally, though, and I guess a brief entry in a README file would be sufficient so that someone looking for this information can discover it. It's just that there is a variety of testing frameworks out there, and almost every open source project seems to apply its own little twist to them so it's helpful to know what the correct incantation is to run the tests without having to guess.

mscuthbert:
Okay, I'll keep it as a to-do to improve the developer docs for that or happy to take a pull request.

shimpe:
does running a test locally always imply one needs to install first? or is there some clever PYTHONPATH manipulation or other trick that will allow testing the local changes without installing?

mscuthbert:
[...]

PR #170, opened 2016-07-07
Fix VexFlow output

infojunkie:
This PR, along with https://github.com/cuthbertLab/music21j/pull/16, enables music21j / VexFlow output for iPython. I've opted to display a "scrollable player", rather than plain VexFlow output, because it makes things more interactive (which is the point of iPython).

Here's how to test it:
- clone the `music21j` PR branch inside `music21` - such that the folder is `music21/music21j`
- run your notebook from `music21` root
- in your notebook, write the following cells:

```
from music21 import *
```

```
%%javascript
require.undef('music21')
require.config({
  paths: {
      music21: 'http://localhost:8888/tree/music21j/src/music21'
  }
});
require(['music21'], function() {
    console.log(music21);
});
```
- then try out some commands:

```
c = chord.Chord("C4 E4- G4")
c.show('ipython')
```

```
sc = scale.PhrygianScale('g')
s = stream.Stream()
x = [s.append(note.Note(sc.pitchFromDegree(i % 11), quarterLength=.25)) for i in range(60)]
s.show('ipython')
```

Note that music21j cannot handle all music21 objects yet, for example the following will fail:

```
b = corpus.parse('bach/bwv66.6')
b.show('ipython')
```

I'll be working on this as time permits.


mscuthbert:
Hi the coverage % decreased with the PR, which generally means that something isn't being fully tested here -- the `vfshow()` routine was commented out because it was too unstable to encourage using.  Uncommenting it without a test showing that it works properly can't be merged yet, I'm afraid.

infojunkie:
To be fair, I didn't just "uncomment" the code, but fixed it to make it work :-)

That said, happy to contribute tests. Is there a doc explaining the automated tests infrastructure?

infojunkie:
Noting that this issue addresses running tests: https://github.com/cuthbertlab/music21/issues/85

infojunkie:
Okay! I added just one test for the time being, please let me know how else you'd like to test the VexFlow output.

Note: To run all tests, I imitated Travis:
`python -c 'from music21.test.testSingleCoreAll import travisMain as tm; tm()'`

To run my new test on its own, I did:
`python -c 'from music21.converter.subConverters import Test; t = Test(); t.testVexFlowShow()'`

mscuthbert:
I'll be looking into this today -- one of the things that's missing from the system is a test to make sure that the vexflow output from music21j looks right -- that's a visual regression test of the type that VexFlow has that will be a bit bigger task.  The Test() class shouldn't pop anything up on the screen; TestExternal() can have consequences such as playing music, opening web browsers, etc. 

I'll give more info about the testing mechanism soon.  Thanks for being patient -- this summer I'm the only developer on staff at MIT and I have a number of other projects that I need to do also, so I'll need to keep asking for patience while being so so happy about the enthusiasm here.  Thanks!

mscuthbert:
Could the test that shows vexflow be moved to testExternal?  When I run the test suite there shouldn't be any windows popping up.  Thanks!

music21/converter/subConverters.py
```diff
@@ -260,35 +260,34 @@ class ConverterIPython(SubConverter):
...
+        </script>
+        <div id='""" + divId + """'></div>
+        """)
     
     def show(self, obj, fmt, app=None, subformats=None, **keywords):
         '''
         show using the appropriate subformat.
         '''
-        if subformats is None:
+        if not subformats:
```
mscuthbert: this is quite different.  If I wanted subformats = [] how would I do this now?

infojunkie: I changed it because `subformats` always arrived here as empty array, instead of None. Since no tests broke, it seemed okay to do it. I cannot anticipate your future needs.

PR #176, opened 2016-08-02
Fix duplicate id problem.

twcook:
Though there are still issues with the ids. This fixes the core problem of object ids being reused randomly and causing failure in later appending stream objects, i.e. measures to parts and parts to scores after a set of notes were created.

### DIFF
--- a/music21/base.py
+++ b/music21/base.py
@@ -42,6 +42,7 @@ import copy
 import sys
 import types
 import unittest
+import time
 
 from collections import namedtuple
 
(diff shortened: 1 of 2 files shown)


ahankinson:
There seem to be a lot of formatting changes in this PR, which makes reviewing your change difficult. You may want to revert and resubmit without the changes to indentation and extra spaces at the ends of lines.

mscuthbert:
Your change introduced an inconsistency with the exception message. Please fix and check that there are no problems with scores that use the same id for two different parts (two flutes For instance or a part and a transposed version of the same part). You may have fixed something for your own work but there are other use cases that will be broken by this.

mscuthbert:
Thank you Andrew also for that. Yes substantive and formatting PRs are both helpful but shouldn't be mixed. Thanks!

twcook:
Just delete the PR.  This is just a hobby for me and I will be busy for the next few weeks anyway.

PR #177, opened 2016-08-02
Fix duplicate id problem.

twcook:
Apologies for the spaces issue at the end of lines.  I have WingIDE set to strip those but apparently it isn't working. 

The changes are base.py 
lines; 45, 400 

stream/core.py
lines; 280, 289 & 290.

### DIFF
--- a/music21/base.py
+++ b/music21/base.py
@@ -42,6 +42,7 @@ import copy
 import sys
 import types
 import unittest
+import time
 
 from collections import namedtuple
 
@@ -56,19 +57,19 @@ from music21._version import __version__, __version_info__
 VERSION = __version_info__
 VERSION_STR = __version__
 #------------------------------------------------------------------------------
-__all__ = ['Music21Exception', 
-           'SitesException', 
+__all__ = ['Music21Exception',
+           'SitesException',
            'Music21ObjectException',
-           'ElementException', 
-           
-           'Groups',            
+           'ElementException',
+
+           'Groups',
            'Music21Object',
            'ElementWrapper',
-           
-           'VERSION', 
-           'VERSION_STR',            
+
+           'VERSION',
+           'VERSION_STR',
            ]
-## N.B. for eclipse "all" import working, we need to list this 
+## N.B. for eclipse "all" import working, we need to list this
 #       separately in "music21/__init__.py"
 ##      so make sure to update in both places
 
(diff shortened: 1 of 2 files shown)


ahankinson:
I rather meant that you *were* stripping the spaces at the ends of lines, which makes it difficult to actually discover the changes. In general, while stripping whitespace and formatting can be good things to do, they typically go under separate commits since they don't affect functionality.

twcook:
Looks like Travis CI is failing on that last update to 
music21/alpha/theoryAnalysis/theoryAnalyzer.py", line 1539

This isn't one of my changes.

mscuthbert:
Sorry, but I can't use time.time() as a way of making a GUID -- many systems only report millisecond or even second accuracy, which means that in a multiprocessor environment, it is very likely that two objects could be created:

```
*time.time()*
Return the time in seconds since the epoch as a floating point number. 
Note that even though the time is always returned as a floating point number, 
not all systems provide time with a better precision than 1 second. 
While this function normally returns non-decreasing values, 
it can return a lower value than a previous call if the system clock has been 
set back between the two calls.
```

You sent me access to files that were going to show why this part of music21 wasn't working for you but nothing in those files uses music21.

Thanks.
M

mscuthbert:
There are a few tests that are broken in the latest Travis-CI build because of something I introduced (was massively interrupted this afternoon, figured no one was using four-part-multi-staff braille, so breaking a test that wasn't working in the first place would not be a big deal...), but the vast majority of the broken tests here are all from this patch.  Including everything in theoryAnalysis.

twcook:
You sent me access to files that were going to show why this part of music21 wasn't working for you but nothing in those files uses music21.

Did you try the music21_based branch as I suggested?

Thanks for the open source code.  Some of it will be useful.  But as a whole the fundamental approach of using python object ids as a unique ID is a bad idea.  I can't think of a modern POSIX compliant/compatible system that the time.time() warning applies to anymore. I am certain you won't find a multiprocessor environment running an OS that doesn't report microsecond accuracy.  Think about it.

mscuthbert:
ah, I didn't see that there was a separate branch -- will check.

mscuthbert:
I had a chance to look at the Stream.core() change and I don't think it'll solve whatever problem is arising for you.  The id(el) check is just a fast way of saying, "does the stream THINK it might have the element in it already?" and only if that returns yes, do we actually do a slow, iterative "IS" search checking whether the element is in the stream.  I suspect that the problem, if any, should be elsewhere. Thanks!

PR #179, opened 2016-08-17
closing file descriptors for temp files

fzalkow:
File descriptors for temp files have not beeen properly closed. This is resolved now.

My use case: I have an restrictive limit of opened files:

    $ ulimit -n
    1024

If I write quite some files with music21, it returned an error:

    >>> import os
    >>> from music21 import *
    >>> s = stream.Stream()
    >>> s.append(note.Note('C4'))
    >>> for i in range(1025):
    ...     fp = s.write('musicxml')
    ...     os.remove(fp)
    ... 
      File "<stdin>", line 2, in <module>
      File "/usr/local/lib/python3.5/dist-packages/music21/stream/__init__.py", line 249, in write
      File "/usr/local/lib/python3.5/dist-packages/music21/base.py", line 2514, in write
      File "/usr/local/lib/python3.5/dist-packages/music21/converter/subConverters.py", line 802, in write
      File "/usr/local/lib/python3.5/dist-packages/music21/converter/subConverters.py", line 785, in writeDataStream
    OSError: [Errno 24] Too many open files: '/tmp/music21/tmprio57gox.xml

### DIFF
--- a/music21/environment.py
+++ b/music21/environment.py
@@ -56,7 +56,7 @@ def etIndent(elem, level=0, spaces=2):
     else:
         if level and (not elem.tail or not elem.tail.strip()):
             elem.tail = i
- 
+
 #------------------------------------------------------------------------------
 
 
@@ -130,7 +130,7 @@ class _EnvironmentCore(object):
         value = self._ref[key]
         if six.PY3 and isinstance(value, bytes):
             value = value.decode(encoding='utf-8', errors='replace')
-        
+
         valueStr = str(value).lower()
 
         if key == 'debug':  # debug expects a number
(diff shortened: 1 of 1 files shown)


mscuthbert:
At the time that music21 first came out there were OSes (I believe including Windows) where closing the file descriptor on a temp file meant that it would immediately be deleted (it's a temp file with no open connections, so CLEANUP TIME!).  Back then we were supporting from Python 2.2 onwards, so there may be conditions that have changed (there's a delete=False parameter that we can now pass).  I'll need to see evidence that it works for Windows users consistently in the normal before changing the behavior for a condition that is rare to have arise.  Can you test on windows and report back? Thanks!

fzalkow:
No problems on Windows 10 as far as I see:

    >>> import os
    >>> import win32file
    >>> from music21 import *
    >>> win32file._getmaxstdio()
    512
    >>> s = stream.Stream()
    >>> s.append(note.Note('C4'))
    >>> for i in range(1025):
    ...     fp = s.write('musicxml')
    ...     os.remove(fp)
    >>>

For Windows the opened files are closed, regardless of my pull request. (because it will get in the else-clause of the if-statement that starts on line 488.

mscuthbert:
Ah! Of course! Merging.

mscuthbert:
THANKS!

PR #182, opened 2016-08-21
Update m21ToXml.py and xmlToM21.py for MuseScore 2

doctor-schmidt:
before

[output before.txt](https://github.com/cuthbertLab/music21/files/428727/output.before.txt)

after

[output after.txt](https://github.com/cuthbertLab/music21/files/428728/output.after.txt)

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -3380,13 +3380,13 @@ class MeasureExporter(XMLExporterBase):
         <harmony>
           <root>
             <root-step>E</root-step>
-            <root-alter>-1.0</root-alter>
+            <root-alter>-1</root-alter>
           </root>
           <kind text="M">major</kind>
           <inversion>2</inversion>
           <bass>
             <bass-step>B</bass-step>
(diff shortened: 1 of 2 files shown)


doctor-schmidt:
Errors are with accidentals in 'root-alter' and 'root-base' musicxml tags
e.g
>>> accidental = int('-1.0')
ValueError: invalid literal for int() with base 10: '-1.0'
>>> accidental = int(float('-1.0'))
>>> accidental
-1

mscuthbert:
Great patch.  One issue though: http://usermanuals.musicxml.com/MusicXML/Content/ST-MusicXML-semitones.htm -- reports that 0.5 is acceptable, so we should only be converting to int if (int(alter) == float(alter)), but leave it alone in case it's not.  Could you extend this patch to do so and add a test case for that?  Thanks!

mscuthbert:
Actually, I know what would be best -- I'm going to merge this and then I'll immediately put that in, because there are a lot of places where it would be useful.  Thanks!

mscuthbert:
Ha!  I already had the right routine:  common.numToIntOrFloat()

PR #189, opened 2016-08-26
Chord symbol Bass notes missing in MusicXML output

doctor-schmidt:
Chord symbol bass does not display in MuseScore with music21 V2 and V3.
(It worked fine in version 1.6) This definitely should be *Undun*!

Check out the bridge in this MusicXML file for testing:
[test file.zip](https://github.com/cuthbertLab/music21/files/439719/test.file.zip)
>>> s = converter.parse('Undun.xml')
>>> s.show()

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -3536,7 +3536,7 @@ class MeasureExporter(XMLExporterBase):
         mxHarmony = Element('harmony')
 
         csRoot = cs.root()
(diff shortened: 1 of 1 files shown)


mscuthbert:
Pull requests without tests or an examination of why this was changed in the first place will be rejected.  See https://github.com/cuthbertLab/music21/blob/c7f44493eb0e232b1c463e4131f97291c21e9604/music21/musicxml/toMxObjects.py and move forward in time.

Issue #191, opened 2016-09-08
IndexError when running pitch detection on .wav file.


jjrob13:
When trying to transcribe .wav files using the monophonicStreamFromFile function, I received an IndexError coming from the autocorrelation function.

Here is the stack trace:

`/music21/audioSearch/transcriber.pyc in monophonicStreamFromFile(fileName, useScale)
    124     from music21 import audioSearch as audioSearchBase
    125
--> 126     freqFromAQList = audioSearchBase.getFrequenciesFromAudioFile(waveFilename=fileName)
    127
    128     detectedPitchesFreq = audioSearchBase.detectPitchFrequencies(freqFromAQList, useScale)

/music21/audioSearch/__init__.pyc in getFrequenciesFromAudioFile(waveFilename)
    384     for data in storedWaveSampleList:
    385         samps = numpy.fromstring(data, dtype=numpy.int16)
--> 386         freqFromAQList.append(autocorrelationFunction(samps, recordSampleRate))
    387     wv.close()
    388

/music21/audioSearch/__init__.pyc in autocorrelationFunction(recordedSignal, recordSampleRateIn)
    143         beginning = positiveDifferences[0]
    144         peak = numpy.argmax(correlation[beginning:]) + beginning
--> 145         vertex = interpolation(correlation, peak)
    146         finalResult = recordSampleRateIn / vertex
    147     return finalResult

/music21/audioSearch/__init__.pyc in interpolation(correlation, peak)
    212     3.21428571...
    213     '''
--> 214     vertex = (correlation[peak - 1] - correlation[peak + 1]) / (
    215                 correlation[peak - 1] - 2.0 * correlation[peak] + correlation[peak + 1])
    216     vertex = vertex * 0.5 + peak

IndexError: index 1024 is out of bounds for axis 0 with size 1024
`

Is this something on my end?

This is happening on about half of the 50 .wav files that I am trying to transcribe.

I am running Ubuntu 16.04 and installed the latest music21 from the master branch.

Please let me know if you need any additional information.

Thanks,

JJ

mscuthbert:
Is there any way to run the same tests on a Mac or PC to narrow down if the problem is an Ubuntu specific problem or a general music21 problem?  thanks!

jjrob13:
I am waiting on a Mac to attempt to reproduce it.  I will have the results later this week.

riyansetiadji:
I got the same issue on Mac and CentOS(Docker).

[...]

PR #194, opened 2016-09-26
Fix interpolation function in audioSearch

jjrob13:
The interpolation returns the index of the peak, but for the edge cases when peak == 0 or peak == len(correlation) - 1, the function was returning the value instead of the index.


jjrob13:
Fix for #191

mscuthbert:
Thanks for the catch John!  Merged.

PR #199, opened 2016-11-12
created normalOrderString method

emzhang:
added in method and doctest so that deprecation message for
normalFormString matches references something that exists


mscuthbert:
I'm guessing that the coverage decrease is a rounding error because it looks like coverage would be improved or the same. Merging.

PR #200, opened 2016-12-14
BWV 386: remove duplicate E from bar 4 beat 4

alexcoplan:
Hello,

I noticed while writing some code to process the Bach chorales that there is a chord in bar 4 beat 4 of BWV 386 which consists of two (duplicate) E5 crotchets. This breaks any code that expects notes instead of chords.

This PR replaces the chord with a single note.

Thanks,
Alex


mscuthbert:
Thanks! obvious error, so accepting.

However, there are many chorales that have chords in them, including a couple where the melody lines split into chords.  As Bach wrote them this way, we cannot edit the chorales to remove them for ease of processing -- the code should treat each note as a potential chord and if only one pitch can be processed, .pitches[0] will need to be selected.  We can accept changes into the corpus that correct errors in the editions (for instance, there was one chorale where the tenor line was the bass line for four measures; we fixed that from the Bach Gesellschaft edition).

alexcoplan:
Great, thanks! That sounds sensible regarding chords in the Chorales in general, thanks for the heads up.

PR #204, opened 2017-01-28
fixing the problem, for putting only sound tag where no metronome is …

AlmogCohen:
…needed (Metronomemark without text)


AlmogCohen:
wasn't synced with master....so I remerged everything and submitted all the commits. It is working again

AlmogCohen:
Sorry guys but I forgot adding any documentation and having problems checking out this branch again. this is what you should do to see the problem:

To test this code just add this object to the first measure and test it
```
mm = tempo.MetronomeMark()
mm.numberSounding = 60
measure.insert(0, mm)
```

AlmogCohen:
added relevant doc tests to the function

mscuthbert:
Thanks for the work (and the doc tests).  Some questions:

My sense is that <words> requires some sort of text, not an empty tag: http://usermanuals.musicxml.com/MusicXML/Content/CT-MusicXML-formatted-text.htm
http://usermanuals.musicxml.com/MusicXML/Content/EL-MusicXML-words.htm

Can the <direction-type> be omitted in this case?  How does one give a simple "[qtr] = 60" notation?

mscuthbert:
My sense is that our MetronomeMark ontology/schema is in a pretty bad place, and much under-tested, but it currently works in most cases.  It'd be good if we could determine current behavior (via doctests) for each of the four common cases:   Words only (+ sound),  Metronome mark only (+sound), Words + Metronome Mark (+sound), and no words or metronome mark (+sound).  And these should also be tested for where the beat is not a quarter note as well.

You probably didn't sign up for all this!  But want to make sure we don't have an reversions before moving forward.  I'll help as I can.  Thanks!

AlmogCohen:
I've been working with an musicXML expert who explained the need for an empty <words> tag. I experimented with removing the <direction-type> but the xml was interpreted as invalid in all cases by MuseScore.

Regarding your second comment, I have not completely understood what do you mean. Does my addition break anything else that was done there before?

mscuthbert:
I think that the ideal fix would not to create a <direction> or <direction-type> tag at all, but if there are no words, to put the <sound> tag inside the <measure> hierarchy itself, no?

mscuthbert:
The very good is better than the perfect...merging. :-)

PR #207, opened 2017-02-13
Issue 14 hotfix

johnMamish:
Fixes the most urgent needs for issue 14.  Namely, getting the composer and title from humdrum metadata.  All the parsing machinery was there, it looks like a little bit of plumbing was just missing.

The titles still require HTML unescaping.  For instance, `Herr Jesu Christ, ich weiß gar wohl` will be entered into metadata as `Herr Jesu Christ, ich wei&szlig; gar wohl`.

Naive attempts to resolve this remaining issue resulted in the Lilypond parser crashing when I run `.show()` on the title containing the ß.

### DIFF
--- a/music21/humdrum/spineParser.py
+++ b/music21/humdrum/spineParser.py
@@ -717,7 +717,6 @@ class HumdrumDataCollection(object):
         # metadata.  We need to make a new metadata object and put the
         # information from insertList into the metadata structure.
         self.stream.insert(0, metadata.Metadata())
-        print('foo')
         for offset, el in insertList:
             self.addGlobalEventToMetadata(el)
         if insertList:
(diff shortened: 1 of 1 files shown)


johnMamish:
Hmm... it appears that the CI tests do not expect the humdrum files to have any metadata.  That is where my tests are failing.

```
Expected:
    {0.0} <music21.stream.Part spine_0>
    ......

Got:
    {0.0} <music21.metadata.Metadata object at 0x7f8c34e19b10>
    {0.0} <music21.stream.Part spine_0>
    .....
```

Suggestions?

mscuthbert:
I imagine that with your change, the authoritative answer to some tests will have changed too. It seems to make sense that there will now be a metadata object there so you can probably update the tests to expect a metadata object before the first Part

johnMamish:
I need to put this on the backburner for a little bit.

mscuthbert:
Fixed using some of your code but a slightly different approach.  Thank you!

PR #211, opened 2017-02-26
demo that shows how to build a MidiFile directly

PeterMitrano:
Simple example of working directly with MidiFile/MidiTrack. Not sure if there should be more explanation accompanying this, or if there are links to these demos that need to be updated?

I have a few more useful demo scripts I'd like to contribute once I understand the process.

### DIFF
--- a/music21/demos/build_melody.py
+++ b/music21/demos/build_melody.py
@@ -34,7 +34,7 @@ def populate_midi_track_from_data(mt, data):
         mt.events.append(dt)
 
         me = midi.MidiEvent(mt)
-        me.type = "NOTE_ON"
+        me.type = "NOTE_OFF"
         me.channel = 1
         me.time = None  # d
         me.pitch = p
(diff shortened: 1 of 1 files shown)


mscuthbert:
Hi -- it looks very promising -- can you make sure that the whole demo is wrapped in a def main() clause and is protected from running unless it's run from the command line ("if __name__ == '__main__': main()") and I think we'll need an explanation at the top about what this does.

For the writing Music21 should only write out to files in the music21temp directory specified in the environment.Environment() variable:

```
env = environment.Environment()
fn = env.getTempFile('.mid')
```

Looking very promising!  Thanks!

mscuthbert:
Sorry I missed this! It looks great! Should the second NOTE_ON be NOTE_OFF?

PeterMitrano:
Well, that probably makes more logical sense. Currently it is a NOTE_ON
with velocity zero. I'll change it. Not sure why I didn't use NOTE_OFF

mscuthbert:
All set! COngrats!

PR #227, opened 2017-07-07
Fixed: key.mode was required in MEI import

ahankinson:
Previously, the MEI importer could crash if @key.mode was not specified since it would pass in 'None' for the mode argument on KeySignature.asKey.

This commit fixes this by providing a default value so that if key.mode is None it assumes a 'major' mode.

### DIFF
--- a/music21/mei/base.py
+++ b/music21/mei/base.py
@@ -1016,8 +1016,9 @@ def _keySigFromAttrs(elem):
                        mode=elem.get('key.mode', ''))
     else:
         # @key.sig, @key.mode
+        # If @key.mode is null, assume it is a 'major' key (default for ks.asKey)
         ks = key.KeySignature(sharps=_sharpsFromAttr(elem.get('key.sig')))
-        return ks.asKey(mode=elem.get('key.mode'))
+        return ks.asKey(mode=elem.get('key.mode', 'major'))
 
 
 def _transpositionFromAttrs(elem):


mscuthbert:
Would it be better to leave the element as a KeySignature object to indicate that mode was not specified? Happy to merge either way but it seems to me that this might be closer to the intended input.

ahankinson:
Would it be best to return `Key` or `KeySignature` in this instance? `Key` seems like it's making an assumption that major/minor will be available, which in many cases it will not be, but `key.sig` will almost always be specified (and if it isn't, it's assumed to be 0 s/f).

I would think that the safest assumption is to always return a KeySignature, since there is no implicit assumption there about the Key of the work. Perhaps @crantila has some input here?

In either case, the comments should be updated at the very least since this will always return a `Key` object, so don't merge it just yet.

mscuthbert:
Ah!  I had time to look at the code better and now I understand.  Yes, if a pitch name is being specified then it's probably better to return a key object and to assume they mean major rather than anything else.  (Or to get major or minor from the case of the pname if that's something you do.)

Btw -- music21 does support identifying a key but having a different number of sharps in the signature.  So you can do (as Bach often does), d-minor with a no-sharp/flat key signature.  So you may want to allow this also.  But I think that what you have is a good fix for here.  So merging and we can re-merge if you want to add this feature.  MusicXML does not support this.

Issue #230, opened 2017-07-17
 SubConverterFileIOException: png file of xml not found. Is your file >999 pages?


japrogramer:
I have this notebook that was working a few versions ago, i decided to go back and run it and i get an error.
architechture:
4.11.9-1-ARCH
linux 4.11.9-1

Name            : musescore
Version         : 2.1.0-1
Description     : Create, play and print beautiful sheet music
Architecture    : x86_64
URL             : http://musescore.org/
Licenses        : GPL
Groups          : None
Provides        : None
Depends On      : desktop-file-utils  gtk-update-icon-cache  libpulse  portaudio  qt5-quickcontrols  qt5-svg  qt5-tools  qt5-webkit  shared-mime-info
Optional Deps   : lame: MP3 export [installed]
Required By     : None
Optional For    : None
Conflicts With  : None
Replaces        : None
Installed Size  : 58.50 MiB
Packager        : Maxime Gauduin <alucryd@archlinux.org>
Build Date      : Sun 07 May 2017 05:44:32 AM CDT
Install Date    : Wed 17 May 2017 08:21:34 AM CDT
Install Reason  : Explicitly installed
Install Script  : No
Validated By    : Signature

music21==3.1.0

```

from music21 import *
us = environment.UserSettings()
import IPython.display
#print(y)
print(us.getSettingsPath())
##us['musicxmlPath'] = '/usr/bin/musescore'
print(us['musicxmlPath'])
print(us['musescoreDirectPNGPath'])

    /home/archangel/.music21rc
    /usr/bin/musescore
    /usr/bin/musescore

converter.parse('tinynotation: 3/4 c4 d8 f g16 a g f#').show()
```

error message 
```
converter.parse('tinynotation: 3/4 c4 d8 f g16 a g f#').show()

#IPython.display

---------------------------------------------------------------------------
SubConverterFileIOException               Traceback (most recent call last)
<ipython-input-3-07d160cac093> in <module>()
----> 1 converter.parse('tinynotation: 3/4 c4 d8 f g16 a g f#').show()
      2 #IPython.display

~/VENV/music_gpu3.6/lib/python3.6/site-packages/music21/stream/__init__.py in show(self, *args, **kwargs)
    255         if self.isSorted is False and self.autoSort:
    256             self.sort()
--> 257         return super(Stream, self).show(*args, **kwargs)
    258 
    259     #---------------------------------------------------------------------------

[...]

PR #231, opened 2017-07-18
fixed issue for systems running headless, and better execution of musescore, better test for testXMLtoPNGtooLong

japrogramer:
I found a solution, to my problem in issue #230 it was a little difficult because all errors where being suppressed and by the time the execution got to the part where it checks for png output music21 had no idea something had gone wrong. I based this branch of master but the patch should also go into 3.*

### DIFF
--- a/music21/converter/subConverters.py
+++ b/music21/converter/subConverters.py
@@ -1323,11 +1323,11 @@ class Test(unittest.TestCase):
         os.rename(tempfp, tempfp + "-0001.png")
         tempfp += "-0001.png"
         if six.PY3:
-            from unittest import mock  # @UnusedImport # pylint: disable=no-name-in-module
+            from unittest import mock, MagicMock  # @UnusedImport # pylint: disable=no-name-in-module
         else:
-            from music21.ext import mock # @Reimport
+            from music21.ext import mock, MagicMock # @Reimport
         with mock.patch('music21.converter.subConverters.findPNGfpFromXMLfp.found') as mockConv:
-            mockConv.__len__ = 1000
+            mockConv.__len__ = MagicMock(return_value=1000)
             xmlconverter = ConverterMusicXML()
             self.assertRaises(SubConverterFileIOException, xmlconverter.findPNGfpFromXMLfp, xmlfp)
 
--- a/music21/converter/subConverters.py
+++ b/music21/converter/subConverters.py
@@ -819,7 +819,7 @@ class ConverterMusicXML(SubConverter):
                              , stderr=subprocess.STDOUT, env=env)
         except subprocess.CalledProcessError as cpe:
             print("program error:", cpe, file=sys.stderr)
-            raise Exception(err)
+            raise Exception(cpe)
 
         if subformatExtension == 'png':
             return self.findPNGfpFromXMLfp(fpOut)
--- a/music21/converter/subConverters.py
+++ b/music21/converter/subConverters.py
@@ -1326,8 +1326,11 @@ class Test(unittest.TestCase):
             from unittest import mock  # @UnusedImport # pylint: disable=no-name-in-module
         else:
             from music21.ext import mock # @Reimport
-        with mock.patch('music21.converter.subConverters.ConverterMusicXML.findPNGfpFromXMLfp.found') as mockConv:
-            mockConv.__len__ = 1000
+        with mock.patch('music21.converter.subConverters.len') as mockConv:
+            mockConv.return_value = 1000
+            xmlconverter = ConverterMusicXML()
+            self.assertRaises(SubConverterFileIOException, xmlconverter.findPNGfpFromXMLfp, xmlfp)
+            mockConv.return_value = 0
             xmlconverter = ConverterMusicXML()
             self.assertRaises(SubConverterFileIOException, xmlconverter.findPNGfpFromXMLfp, xmlfp)
 
--- a/music21/converter/subConverters.py
+++ b/music21/converter/subConverters.py
@@ -1324,10 +1324,10 @@ class Test(unittest.TestCase):
         else:
             from music21.ext import mock # @Reimport
         with mock.patch('music21.converter.subConverters.glob.glob') as mockConv:
-            mockConv.return_value = [x for x in range(1)]
+            mockConv.return_value = [0]
             xmlconverter = ConverterMusicXML()
             xmlconverter.findPNGfpFromXMLfp(tempfp)
-            mockConv.return_value = [x for x in range(0)]
+            mockConv.return_value = []
             xmlconverter = ConverterMusicXML()
             self.assertRaises(SubConverterFileIOException, xmlconverter.findPNGfpFromXMLfp, tempfp)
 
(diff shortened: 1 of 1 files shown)


japrogramer:
@mscuthbert my tests are passing but for python 2.7 the test don't even begin and the build exits with
 
`The command "python -m compileall music21" failed and exited with 1 during .`

music21/converter/subConverters.py
```diff
@@ -732,12 +732,10 @@ def findPNGfpFromXMLfp(self, xmlFilePath):
...
         '''
-        if os.path.exists(xmlFilePath[0:len(xmlFilePath) - 4] + "-1.png"):
-            pngfp = xmlFilePath[0:len(xmlFilePath) - 4] + "-1.png"
-        elif os.path.exists(xmlFilePath[0:len(xmlFilePath) - 4] + "-01.png"):
-            pngfp = xmlFilePath[0:len(xmlFilePath) - 4] + "-01.png"
-        elif os.path.exists(xmlFilePath[0:len(xmlFilePath) - 4] + "-001.png"):
-            pngfp = xmlFilePath[0:len(xmlFilePath) - 4] + "-001.png"
+        import glob
+        found = sorted(glob.glob(xmlFilePath[0:len(xmlFilePath) - 4] + "-*.png"))
+        if found and len(found) < 999:
```
mscuthbert: we proibably don't need to check for < 999 now, since your glob will catch even longer generations.

music21/converter/subConverters.py
```diff
@@ -806,19 +804,22 @@ def runThroughMusescore(self, fp, subformats=None, **keywords):
         fpOut = fp[0:len(fp) - 3]
         fpOut += subformatExtension
 
-        musescoreRun = '"' + musescorePath + '" ' + fp + " -o " + fpOut + " -T 0 "
+        musescoreRun = '' + musescorePath + ' ' + fp + " -o " + fpOut + " -T 0 "
```
mscuthbert: this seems quite dangerous together with the split below.  Many users have spaces in their musescorePath and their fp, fpOut paths -- if using subprocess, instead of concationating and splitting, we should build the array directly.  Make sure that this commit works for paths involving spaces.

music21/converter/subConverters.py
```diff
@@ -806,19 +804,22 @@ def runThroughMusescore(self, fp, subformats=None, **keywords):
...
-        fileLikeOpen.close()
-        sys.stderr = storedStrErr
+        import subprocess
+        command = musescoreRun.split()
+        try:
+            env = dict(os.environ, QT_QPA_PLATFORM='offscreen')
+            output = subprocess.check_output(command
+                             , stderr=subprocess.STDOUT, env=env)
+        except subprocess.CalledProcessError as cpe:
+            print("program error:", cpe, file=sys.stderr)
```
mscuthbert: use environLocal.warn() instead of print -- this lets people capture all warnings directly along with where they come from (they may have redirected it to somewhere instead of sys.stderr, for instance when running music21 as a service).  Or don't print anything simply let the message bubble up.

music21/converter/subConverters.py
```diff
@@ -738,6 +738,7 @@ def findPNGfpFromXMLfp(self, xmlFilePath):
             pngfp = found[0]
         else:
             raise SubConverterFileIOException("png file of xml not found. Is your file >999 pages?")
+        print(found)
```
mscuthbert: Make sure that this disappears before the final commit.  Thanks!

music21/converter/subConverters.py
```diff
@@ -722,6 +722,7 @@ class ConverterMusicXML(SubConverter):
     registerOutputSubformatExtensions = {'png': 'png',
                                          'pdf': 'pdf',
                                          }
+    glob = __import__('glob')
```
mscuthbert: Can you explain why you need this usage rather than simply "import glob"?

japrogramer: No particular reason, im going to change it to an import in my next push.

music21/converter/subConverters.py
```diff
@@ -1326,11 +1326,7 @@ def testXMLtoPNGtooLong(self):
         with mock.patch('music21.converter.subConverters.glob.glob') as mockConv:
             mockConv.return_value = [x for x in range(1)]
```
mscuthbert: is this the same as .return_value = [0] ?

music21/converter/subConverters.py
```diff
@@ -1326,11 +1326,7 @@ def testXMLtoPNGtooLong(self):
...
             xmlconverter = ConverterMusicXML()
-            try:
-                xmlconverter.findPNGfpFromXMLfp(tempfp)
-            except SubConverterFileIOException:
-                self.assertTrue(False)
-
+            xmlconverter.findPNGfpFromXMLfp(tempfp)
             mockConv.return_value = [x for x in range(0)]
```
mscuthbert: is this the same as .return_value = [] ?

mscuthbert:
Hi!  This looks great.  I'll pull no matter what by tomorrow, but if it's possible to simplify those two range() constructors or explain why they need to be the way they are, I'd appreciate that.  Thanks!

japrogramer:
@mscuthbert oh, i previously had them generate a list so i could trigger the 999 > exception. 
They can definitely be simplified for clarity. Will do that in next push.

mscuthbert:
closing and reopening to make sure that the build still works.

mscuthbert:
merging.

mscuthbert:
this commit breaks musescore (SEGFAULT) on Mac and possibly PC also -- please test on major systems before suggesting again.  Thanks!

japrogramer:
Ok, will test on other platforms soon.

jbmaxwell:
I'm having the same "SubConverterFileIOException: png file of xml not found. Is your file >999 pages?" when trying to run music21 on jupyter notebook. I'm not clear what the fix is... sorry...

japrogramer:
are you running under linux?

jbmaxwell:
yes, Ubuntu 16.04

japrogramer:
try exporting this environment variable QT_QPA_PLATFORM='offscreen'

jbmaxwell:
hmm... okay, not entirely comfortable with Linux yet (this is a deep learning box... with music21 for data prep)... I now have "export QT_QPA_PLATFORM='offscreen'" in my .bashrc file. Is that what you meant? 
My music21 environment is:
autoDownload ask
braillePath None
debug 0
directoryScratch None
graphicsPath None
ipythonShowFormat ipython.musicxml.png
lilypondBackend ps
lilypondFormat pdf
lilypondPath /home/jbmaxwell/lilypond
lilypondVersion None
localCorporaSettings {}
localCorpusPath
localCorpusSettings LocalCorpusSettings([])
manualCoreCorpusPath None
midiPath None
musescoreDirectPNGPath /usr/bin/musescore
musicxmlPath /usr/bin/musescore
pdfPath None
showFormat musicxml
vectorPath None
warnings 1
writeFormat musicxml



/tmp/music21

japrogramer:
yeah that's what i meant,
now source your bashrc.

jbmaxwell:
ah, wait. Sorry. Yes, I logged out and back in and it's working. Thanks!

japrogramer:
np.

PR #233, opened 2017-07-25
Musecore headless, with platform check.

japrogramer:
Since this is a a very specific case that only applies to unix systems running without a Xserver,
I included an if statement.
It checks for the platform and also the DISPLAY environment variable, If there is no DISPLAY set and the platform is 'linux' than QT_QPA_PLATFORM is set to 'offscreen' to hint to musescore that there is no display server.

### DIFF
--- a/music21/converter/subConverters.py
+++ b/music21/converter/subConverters.py
@@ -1328,11 +1328,11 @@ class Test(unittest.TestCase):
         os.rename(tempfp, tempfp + "-0001.png")
         tempfp += "-0001.png"
         if six.PY3:
-            from unittest import mock  # @UnusedImport # pylint: disable=no-name-in-module
+            from unittest import mock, MagicMock  # @UnusedImport # pylint: disable=no-name-in-module
         else:
-            from music21.ext import mock # @Reimport
+            from music21.ext import mock, MagicMock # @Reimport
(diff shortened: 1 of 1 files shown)


japrogramer:
In PEP324 subprocess is meant to replace os.system and os.spawn*, That is why i have decided to use it over those methods.
 QT_QPA_PLATFORM was causing the spawned process to behave differently, so that it doesn't try to launch the gui, without that environment variable set the spawned process should behave exactly as before.
At the moment i do not have a windows or mac machine.

mscuthbert:
Hi @japrogramer -- because Mac and Windows are the only officially supported systems, I'll need this to be tested on both of them before merging, especially since the last merge caused major problems on both other systems.

dhgutteridge:
May I add, it might be better if the test "if sys.platform == 'linux'" is changed to check if the code is not running under Windows or macOS. Alternately, more than just "linux" needs to be specified here. (I say that as a NetBSD user and committer who runs this software, and may be running it headless.)

mscuthbert:
Hi all -- anything that deals with OS specific changes I need tested on Mac and PC or I can't merge.  Thanks!

mscuthbert:
Closing for lack of action or owner.

PR #244, opened 2017-09-16
Update translate.py

KKONZ:
This proposed change updates the translate.py script so that it no longer adds rests to the beginning of the midi file when translating from stream. It changes the default of addStartDelay = True to addStartDelay = False.

### DIFF
--- a/music21/midi/translate.py
+++ b/music21/midi/translate.py
@@ -36,7 +36,7 @@ class TranslateException(exceptions21.Music21Exception):
 #-------------------------------------------------------------------------------
 # Durations
 
-def offsetToMidi(o, addStartDelay=True):
+def offsetToMidi(o, addStartDelay=False):
     '''
     Helper function to convert a music21 offset value to MIDI ticks,
     depends on *defaults.ticksPerQuarter* and *defaults.ticksAtStart*.


mscuthbert:
Do MIDI files play back properly on QuickTime 7 (Mac) and Windows Media Player after making the change?  If not, then we'll want to make this an optional flag on .show('midi') or .write('midi')

KKONZ:
Let me check quick!

mscuthbert:
Thanks!  Now that you've found the damn code I couldn't find before I think that I should be able to make a system that works for everyone, but if no one NEEDS the other behavior then it should be very simple!

KKONZ:
I just checked on both QuickTime and Windows media player, both start in the right place for me with the default is changed to addStartDelay = False! :) 
FYI, I am using music21 to splice up midi tracks into smaller segments to use to train a DNN, the raw_generator call in keras fails when the rests are appended to the beginning of the midi files.

mscuthbert:
Okay!  If all the tests pass then I'll merge this and if there are other problems, we'll call this the new Normal and bugs arising from this change will need to be swatted rather than vice-versa.

KKONZ:
Great! This would be my first official git merge, so I really hope everything passes! Let me know if there is anything else I can do to help and thank you for your quick responses. Hope you have a nice day!

mscuthbert:
Fantastic first git pull request.  You've done great!

Some errors in tests have already arisen.  Look at the link (https://travis-ci.org/cuthbertLab/music21/jobs/276316783) and fix them and commit again.  You might want to run the midi/*.py files yourself to make sure they work before committing again, since it takes about 20 minutes for all tests to run, so might as well make sure if you're going to need to wait anyhow, to make sure that problems that are easily fixed are done first.

KKONZ:
Thank you for the kind words! Sorry about the errors, let me look those over and I will re-commit when they are fixed.

mscuthbert:
If committing with errors was something to be sorry about, I'd be saying "sorry" to myself 10x a day! :-)

PR #245, opened 2017-09-16
Update translate.py

KKONZ:
I couldn't figure out how to run the midi/*.py files so I apologize if there are still errors. I had overlooked the tests in the code with the first pull request. Hope this works!


mscuthbert:
best to keep everything in one patch and one pull-request if you can.  (Sorry if it's frustrating; but after you've done one PR, it gets really easy and applies to thousands of projects the same way!)

PR #253, opened 2017-10-04
Transpositions and NeoRiemannian Improvements

MarkGotham:


mscuthbert:
A couple of doctest failures in the Travis-CI build. Not sure if the code or the expected value needs to change. Thanks!!

music21/analysis/neoRiemannian.py
```diff
@@ -29,12 +31,20 @@
 class LRPException(exceptions21.Music21Exception):
     pass
 
-def L(c, raiseException=False):
+def simplerEnharmonics(self):
```
mscuthbert: This isn't in a class so the variable should be called "ch" or something like that, not "self".

music21/setup.py
```diff
@@ -0,0 +1,75 @@
+##!/usr/bin/env python
```
mscuthbert: This file seems to have been copied and moved accidentally -- can it be deleted?

mscuthbert:
This is really really good -- if you'd like the make the two requested changes, that'd be great!  Otherwise I'm happy to merge and make the changes myself.  Thanks!

Could we also point pitch.simplifyMultipleEnharmonics to use the new code so we don't have two different codes doing the same thing?

btw -- there's a simpler way to do the hexatonic cycle:

```python
hexatonicList = [c]
lastChord = c
operations = [P, L, P, L, P, L]
for operation in operations:
    lastChord = operation(lastChord)
    if simplifyEnharmonics:
        lastChord = simplerEnharmonics(lastChord)
    hexatonicList.append(lastChord)
return hexatonicList
```

MarkGotham:
Thanks Myke, though wouldn't calling the new code within simplifyMultipleEnharmonics make that a redundant function? I.e. shouldn't it just be removed in that case? I'll let you make the judgment call on removing functions from the base. Finally, do you want segmentByRests added as well? In analysis?

mscuthbert:
other people may be using simplifyMultipleEnharmonics, so we need to leave it in for at least one version with a "@common.deprecated" decorator and an explanation for how to work without it.  :-)  the fun of backwards compatibility!  For SegmentByRests, etc.  best to create a new pull request for each one -- then you make a new branch for each patch and submit those separately.  It can be a little bit of a pain but it's so much easier for managing a multi-person project.

PR #256, opened 2017-10-18
add chord symbols

Ghadjeres:
ABC converter.parse is now able to handle chord symbols


music21/abcFormat/translate.py
```diff
@@ -276,6 +277,13 @@ def parseTokens(mh, dst, p, useMeasures):
             #ql += t.quarterLength
 
         elif isinstance(t, abcFormat.ABCNote):
+            # add the attached chord symbol
+            if len(t.chordSymbols) == 1:
+                cs_ascii = t.chordSymbols[0]
+                assert cs_ascii[0] == cs_ascii[-1] == '"'
```
mscuthbert: We don't do bare asserts in music21 -- can you do this as a "if ...: raise Music21Exception('...detailed information here...')" but in general with handcoded formats like ABC, if there's an error, we use environLocal.warn("information...") and continue to the next token -- it's not in general worth crashing parsing for an incorrectly formatted symbol in ABC/Humdrum, etc.  (machine coded formats such as MusicXML and MIDI are different, since it's up to the creator to produce valid code).

music21/abcFormat/translate.py
```diff
@@ -279,9 +281,12 @@ def parseTokens(mh, dst, p, useMeasures):
         elif isinstance(t, abcFormat.ABCNote):
             # add the attached chord symbol
             if len(t.chordSymbols) == 1:
-                cs_ascii = t.chordSymbols[0]
-                assert cs_ascii[0] == cs_ascii[-1] == '"'
```
mscuthbert: Ah -- ignore previous comment, since this is gone.

mscuthbert:
Hi -- looking good.  Can you make sure that you've merged the latest version -- some of the errors that are arising seem not to be your fault, but we need to pass the test suite before merging.

Also -- please include some tests and doc tests so that we can be sure that the system is working properly.  Thanks!

mscuthbert:
The remaining error check seems to be from these additions ('.S.').  Thanks for writing tests.  I'll be informed when any pushes arrive.

a-papadopoulos:
I filed a new pull request with the latest master in pull request #276.

PR #262, opened 2017-11-30
Tablature Stuff

Luke-Poeppel:
Added FretNote, FretBoard, and Fretted Instrument Objects to base. Also added m21ToXml converter.


mscuthbert:
Congrats!!!

PR #265, opened 2017-12-05
Simplify Multiple Enharmonics improvements

MarkGotham:
Replaces simplifyMultipleEnharmonics with a point to
analysis.enharmonics as requested in #253


music21/analysis/enharmonics.py
```diff
@@ -0,0 +1,187 @@
...
+class EnharmonicSimplifier:
+    '''
+    Takes any pitch list input and returns the best enharmonic respelling according to the input
+    criteria and rule weightings.
+    Those criteria and rule weightings are currently fixed, but in future the user should be able
+    to select their own combination and weighting of rules according to preferences,
+    with predefined defaults for melodic and harmonic norms.
+    Note: EnharmonicSimplifier itself returns nothing.
+    '''
+    def __init__(self, pitchList, ruleClass=EnharmonicScoreRules):
```
mscuthbert: For our documentation system, we need for all classes to be instantiable with no required arguments, so just set pitchList=None  (I'll double check that this is still true, but it might be)

music21/analysis/enharmonics.py
```diff
@@ -0,0 +1,187 @@
...
+            allSpellings.append(spellings)
+        self.allSpellings = allSpellings
+
+    def getProduct(self):
+        self.allPossibleSpellings = list(itertools.product(*self.allSpellings))
+
+    def bestPitches(self):
+        '''
+        Returns a list of pitches in the best enharmonic spelling according to the input criteria.
+        >>> pList1 = [pitch.Pitch('C'), pitch.Pitch('D'), pitch.Pitch('E')]
```
mscuthbert: add one line of space between docs and code examples

music21/analysis/enharmonics.py
```diff
@@ -0,0 +1,187 @@
...
+        >>> es = analysis.enharmonics.EnharmonicSimplifier(pList1)
+        >>> es.bestPitches()
+        (<music21.pitch.Pitch C>, <music21.pitch.Pitch D>, <music21.pitch.Pitch E>)
+        >>> pList2 = ['D--', 'E', 'F##']
+        >>> es = analysis.enharmonics.EnharmonicSimplifier(pList2)
+        >>> es.bestPitches()
+        (<music21.pitch.Pitch C>, <music21.pitch.Pitch E>, <music21.pitch.Pitch G>)
+        '''
```
mscuthbert: What happens if there's also a C# in here -- will it return C,C#,E,G or C,Db,E,G?

music21/analysis/neoRiemannian.py
```diff
@@ -5,8 +5,9 @@
 #
 # Authors:      Maura Church
 #               Michael Scott Cuthbert
+#               Mark Gotham
 #
-# Copyright:    Copyright © 2012 Michael Scott Cuthbert and the music21 Project
+# Copyright:    Copyright © 2017 Michael Scott Cuthbert and the music21 Project
```
mscuthbert: change to 2012, 2017 -- US law -- we need to put in the start and end date of modifications

music21/pitch.py
```diff
@@ -474,73 +476,23 @@ def _greedyEnharmonicsSearch(oldPitches, scoreFunc=_dissonanceScore):
...
-
-
-    >>> pitch.simplifyMultipleEnharmonics([pitch.Pitch('D--3'),
-    ...                                    pitch.Pitch('F-3'),
-    ...                                    pitch.Pitch('A--3')],
-    ...                                    keyContext=key.Key('C'))
-    [<music21.pitch.Pitch C3>, <music21.pitch.Pitch E3>, <music21.pitch.Pitch G3>]
+def simplifyMultipleEnharmonics(pitches):
     '''
+    Replaces previous function with the new 'EnharmonicSimplifier' in analysis.
```
mscuthbert: No need for this line so long as it does what happened before.  But losing key-context is an issue -- the best enharmonics for B, E-, G- are likely to be different in C major vs. in Gb major.

music21/setup.py
```diff
@@ -0,0 +1,75 @@
+##!/usr/bin/env python
```
mscuthbert: This file needs to be deleted -- there should not be a setup.py in the music21 directory, only in music21base.

music21/analysis/enharmonics.py
MarkGotham: C# in that case to minimise the aug/ dim score (it depends on the order). What we have works well. but we need to run an experiment on highly-trained musical subjects to try and pin down  good, exact, starting values for this metric. I'll see if I can recruit someone.

music21/analysis/enharmonics.py
matthiasroder: spoken like a real book nerd! ;-)

mscuthbert:
Hi Mark -- great work, but still getting failed tests on this.  thanks!

PR #270, opened 2018-01-31
Fixes #269: Expand list of decorations to be ignored in abc parse.

### LINKED ISSUE #269: Ignore more decorations in abc parse
I am trying to import an abc file with mordents defined with the "P" decoration. Here is an abc fragment:

```
X:91
T:Prelude 9 (BWV 854)
C:J. S. Bach
L:1/16
M:12/8
K:E
V:1 treble 
V:1
 E2G2B2 e2d2e2 Pc2B2c2 e6 | PB2A2B2 e6 A2B2c2 B2c2A2 |
```

But I am seeing warnings like these:

> abc: WARNING: Could not get pitch information from note:  P, assuming C

And then I have additional C notes included in my streams. This does not seem like the appropriate behavior. All such decorations should be skipped if they are not supported.

I have a fix for this that simply augments the list of decorations that are ignored.

dvdrndlph:
I can create some test cases if necessary. I am a bit concerned about the "O" (coda) decoration. But the existing tests all succeed.

### DIFF
--- a/music21/abcFormat/__init__.py
+++ b/music21/abcFormat/__init__.py
@@ -2021,22 +2021,38 @@ class ABCHandler:
                 self._tokens.append(ABCTenuto(c))
                 continue
 
-            # get the start of a note event: alpha, or
-            # ~ tunr/ornament, accidentals ^, =, - as well as ^^
+            # get the start of a note event: alpha, decoration, or accidental
             if (c.isalpha() or c in '~^=_'):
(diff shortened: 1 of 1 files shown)


mscuthbert:
Super -- test cases would be appreciated, but let's do them in a separate pull request unless the build fails here -- can you merge to the latest master and add it to this PR?  The URL for downloading lilypond changed apparently, so I just updated master to reflect that.

Issue #197, opened 2016-09-28
Invalid harmony.ChordSymbol() behavior with sus|aug


Atorich:
When i tried to ignore #196 then I have the next bug:
With `harmony.ChordSymbol('A7sus4')` it generates weird notes: `['A1', 'G#3']`

mscuthbert:
I don't think that A7sus4 is in our chord database.  It'd need to be an enhancement to add it.  (w/ a citation from a theory text of what it would mean).  Thanks!

Atorich:
I have tried to make some enhancements but it needs to go more deeper to understand the algorithm of ChordSymbol creation that I supposed before :/

Atorich:
Is there (undocumented, maybe) a way to alter A7 to A7sus4 (change step from 3 to 4) or A7sus2 (3 -> 2) ?

Atorich:
I'm sorry @mscuthbert for disturbing again about this but is there a way to get ```7th suspended``` chords with few casualties?

mscuthbert:
Can you provide a citation from a theory text about what this would mean.  Thanks!

Atorich:
I don't fluent enough in music theory to provide that. But A7sus2 & A7sus4 chords are used in guitar playing. I tried to developed yet another chords generator last year using music21 but I had some issues with this type of chords.

mscuthbert:
There are ways to add custom chord names to music21's Harmony dictionary -- look at the docs there.  That's the way you'll need to add these chords.  I'm afraid that without citations from the literature, we don't add anything to music21; it's the only way to ensure that a chord that means one thing to person X isn't meaning another thing to someone who plays in a different style or instrument.

Atorich:
It is not literature, but 
piano:
https://www.google.ru/search?newwindow=1&q=piano+asus7&oq=piano+asus7&gs_l=psy-ab.3...5558.5957.0.6155.5.5.0.0.0.0.189.189.0j1.1.0....0...1.1.64.psy-ab..5.0.0....0.wOZXrHIILLs

guitar:
[...]

PR #272, opened 2018-02-14
Add 7sus4 chords

Ghadjeres:
7sus4 chords are the only sus4 chords (sus4, 7sus4, 9sus4) that could not be parsed with 
```
music21.harmony.chordSymbolFromChord
```
even when the root note is specified.
Solves #197

### DIFF
--- a/music21/harmony.py
+++ b/music21/harmony.py
@@ -95,8 +95,7 @@ CHORD_TYPES = collections.OrderedDict([
     # other
     ('suspended-second',            ['1,2,5', ['sus2']]),                        # Y
     ('suspended-fourth',            ['1,4,5', ['sus', 'sus4']]),                 # Y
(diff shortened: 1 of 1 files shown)


mscuthbert:
Thanks.  If all checks pass, I'll merge

mscuthbert:
Hi. This pull request is going quite out of scope. Music21’s default chord symbols are based on music theory symbols not jazz symbols. Those can be housed in a different lookup library. I was happy to accept the addition of one symbol the sus7. I see now lots of “dom7dim5” and the like. These will not be accepted as default outputs in favor of existing 7b5 etc.

mscuthbert:
Closing from stale request -- please feel free to reopen, if the fixes above are implemented.  Also note that reformating the intentional spacing of the tables file will be rejected, so please have your linter on manual mode.

PR #274, opened 2018-02-21
voiceLeading update

ryaanahmed:
Fixes some issues in voiceLeading module:
- VoiceLeadingQuartet.parallelMotion() only accepted parallel motion by specific interval, rather than the normal convention of parallel motion being true when two consecutive intervals are the same generic interval. It now defaults to looking at the generic intervals, and takes an optional requiredInterval argument of either a generic or specific interval. This method also allowed octave displacement of one of the intervals, which is contrary to the normal definition of parallel motion (i.e., 3 -> 10 by similar motion is not typically considered parallel motion). This is now optional (for use mainly in the parallelFifth and similar methods).

- VoiceLeadingQuartet.isProperResolution() wasn't handling minor keys at all. Removed the requirement that m7 resolve by inward contrary motion, to allow 5-1 to resolve downwards in the bass.

### DIFF
--- a/music21/voiceLeading.py
+++ b/music21/voiceLeading.py
@@ -45,6 +45,7 @@ from music21 import pitch
 from music21 import key
 from music21 import note
 from music21 import chord
+from music21 import scale
 
 
 #from music21 import harmony can't do this either
@@ -891,6 +892,9 @@ class VoiceLeadingQuartet(base.Music21Object):
             m7:     Resolves to a third with a leap from 5 to 1 in the bass
 
 
+        We will make the examples shorter with this abbreviation:
+        >>> N = note.Note
+
         >>> n1 = note.Note('B-4')
         >>> n2 = note.Note('A4')
         >>> m1 = note.Note('E4')
@@ -976,13 +980,20 @@ class VoiceLeadingQuartet(base.Music21Object):
             return True
 
         if self.key:
-            scale = self.key.getScale()
-            n1degree = scale.getScaleDegreeFromPitch(self.v2n1)
-            n2degree = scale.getScaleDegreeFromPitch(self.v2n2)
+            keyScale = self.key.getScale(self.key.mode)
+            n1degree = keyScale.getScaleDegreeFromPitch(self.v2n1)
+            n2degree = keyScale.getScaleDegreeFromPitch(self.v2n2)
+
+            # catches case of #7 in minor
+            if self.key.mode == 'minor' and n1degree is None:
+                minorScale = scale.MelodicMinorScale(self.key.tonic)
+                n1degree = minorScale.getScaleDegreeFromPitch(self.v2n1, direction=scale.DIRECTION_ASCENDING);
+
         else:
-            scale = None
+            keyScale = None
             n1degree = None
             n2degree = None
+
 
         firstHarmony = self.vIntervals[0].simpleName
         secondHarmony = self.vIntervals[1].generic.simpleUndirected
@@ -994,25 +1005,25 @@ class VoiceLeadingQuartet(base.Music21Object):
                 return False
 
         elif firstHarmony == 'A4':
-            if scale and n1degree != 4:
+            if keyScale and n1degree != 4:
                 return True
-            if scale and n2degree != 3:
+            if keyScale and n2degree != 3:
                 return False
             return (self.outwardContraryMotion()
                         and secondHarmony == 6)
 
         elif firstHarmony == 'd5':
-            if scale and n1degree != 7:
+            if keyScale and n1degree != 7:
                 return True
-            if scale and n2degree != 1:
+            if keyScale and n2degree != 1:
                 return False
             return (self.inwardContraryMotion()
                         and secondHarmony == 3)
 
         elif firstHarmony == 'm7':
-            if scale and n1degree != 5:
+            if keyScale and n1degree != 5:
                 return True
-            if scale and n2degree != 1:
+            if keyScale and n2degree != 1:
                 return False
             return secondHarmony == 3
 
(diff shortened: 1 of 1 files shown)


music21/voiceLeading.py
```diff
@@ -605,43 +654,34 @@ def parallelInterval(self, thisInterval):
...
-        if not (self.parallelMotion() or self.antiParallelMotion()):
-            return False
-
-        if isinstance(thisInterval, str):
-            thisInterval = interval.Interval(thisInterval)
+        return (self.parallelMotion(
+                    requiredInterval=thisInterval,
+                    allowOctaveDisplacement=True
+                    )
+                or self.antiParallelMotion())
```
mscuthbert: This should probably also pass requiredInterval to antiParallelMotion() as well, since it `thisInterval` could be P5 and antiParallelMotion could be `P8`

music21/voiceLeading.py
```diff
@@ -769,6 +814,63 @@ def hiddenOctave(self):
...
+
+        >>> vl = voiceLeading.VoiceLeadingQuartet(N('A4'), N('C4'), N('F4'), N('Bb4'))
+        >>> vl.voiceOverlap() # V2 overlaps V1
+        True
+
+        >>> vl = voiceLeading.VoiceLeadingQuartet(N('A4'), N('E4'), N('F4'), N('D4'))
+        >>> vl.voiceOverlap() # V1 overlaps V2
+        True
+        '''
+        if self.v1n2.pitch < self.v2n1.pitch or self.v2n2.pitch > self.v1n1.pitch:
```
mscuthbert: This assumes that v1 is higher than v2 -- perhaps we should make that a requirement?

music21/voiceLeading.py
```diff
@@ -769,6 +814,63 @@ def hiddenOctave(self):
...
+            return True
+        else:
+            return False
+
+    def voiceCrossing(self):
+        '''
+        Returns True if either note in V1 is lower than the simultaneous note in V2.
```
mscuthbert: This assumes that v1 is (supposed to be) higher than v2 -- perhaps we should make that a requirement?

music21/voiceLeading.py
ryaanahmed: Yes! antiParallelMotion() currently doesn't have a requiredInterval flag, but this can be added.

ryaanahmed:
As this affects the whole module, I'm putting this discussion here:

@mscuthbert wrote:
> This assumes that v1 is (supposed to be) higher than v2 -- perhaps we should make that a requirement?

This remains an open question, and could lead to a broader discussion of the intended use and scope of this module.

The docstring for the class VoiceLeadingQuartet indicates that v1 should be the top voice, but some of the code in the module seems ambivalent about which voice is higher, and some of the docstring tests explicitly have examples in which V2 is higher. These examples resolve fine as much of the code works even if the input is upside down.

isProperResolution() and having any coherent notion of voice crossing depend on a required ordering of the voices. We might also think about the case of analysis of two equal voices that cross a lot, in which case the notion of a higher voice fades away.

Enforcing the correct order of voices might break backwards compatibility, as users might well have been putting the voices in backwards (as in the docstrings!) but getting correct results, since it doesn't matter which order you put the voices for parallelFifth, parallelOctave, etc.

An unrelated question about the module: what to do with isProperResolution(), opensIncorrectly(), closesIncorrectly() etc. These are all style specific, and should probably be renamed(?).

mscuthbert:
> The docstring for the class VoiceLeadingQuartet indicates that v1 should be the top voice

That's something I added today, since now there are now methods that assume that.  I'm not sure it was required or even encouraged before.  I believe I was usually doing the opposite, but it's easily changed.

> isProperResolution() and having any coherent notion of voice crossing depend on a required ordering of the voices. We might also think about the case of analysis of two equal voices that cross a lot, in which case the notion of a higher voice fades away.

I think that since we didn't enforce or give any warnings for people in the past, we might add routines for `higherVoice()`, `lowerVoice()` which simply gets whichever has a higher n1, with n2 as a tie-breaker, and use those in cases where we assume one voice is higher than the other.  In which case voiceCrossing would only be true if `hi_voice_n2 < lo_voice_n2`. But I haven't given much thought to it.

  > isProperResolution(), opensIncorrectly(), closesIncorrectly() etc. These are all style specific, and should probably be renamed(?).

I really don't like these terms because they imply we "should" be composing in/analyzing specific styles and the module is more general than that.  Usually we put in a deprecation notice for a while (1 year; remove in version 6 or at latest 7) and have the function call the correct one.  Or we leave the names alone and put in a style/genre flag:  "isProperResolution(style=styleEnum.MODAL_COUNTERPOINT)" and then at least we're making explicit what we mean and we can possibly add other styles later -- probably needing another method anyhow.

ryaanahmed:
> That's something I added today, since now there are now methods that assume that. I'm not sure it was required or even encouraged before. I believe I was usually doing the opposite, but it's easily changed.

The version of the file that I began hacking on already had the notion of V1 being higher in the docstring (see e.g., https://github.com/cuthbertLab/music21/blob/7319095d13896a64df7d2dc805c862086c121dcd/music21/voiceLeading.py - line 73), but the only functions that depend on the ordering of the voices are isProperResolution(), opensIncorrectly(), closesIncorrectly(), etc.

I'm wary of automatically detecting which voice is higher for the routines that depend on strict ordering, as it seems like there's a lot of room in that kind of scenario for user error to result in incorrect results that are difficult to catch. For example, a user might iterate through a two-voice counterpoint looking for incorrect resolutions, and instead of flagging a voice crossed segment as incorrect, an automatically-detecting-the-top-voice solution would flip the order in which it's considering the voices, potentially (and silently) giving unexpected/incorrect results.

A more aggressive solution could be that VLQ remains agnostic to which voice is the higher voice, but routines that require a strict ordering of voices throw AssertionErrors with malformed input. It's more hostile but less likely to produce unexpected bugs for users.

mscuthbert:
That's a good idea, though music21 doesn't usually throw AssertionErrors, so throw either a subclass of Music21Exception or a ValueError

PR #275, opened 2018-03-01
Bach chorale title corrections

doctor-schmidt:


mscuthbert:
Merging.  While one could argue that 133.6's title is "Wohlan! So will ich mich" instead of the melody title, "Ich freue mich in dir" it's definitely not about summer lambs.

doctor-schmidt:
Hi Myke,

We might want to use these older original Riemenschneider titles.
  63  Nun ruhen alle Wälder
  64  Freu' dich sehr, o meine Seele
  
I'll let you decide. I'm finally understang!

Norman

mscuthbert:
We're not yet close to having totally accurate metadata on the chorales, so I think it's fine as it is right now.  Eventually we should have the cantata title, the chorale melody title, and the chorale text title all in the metadata, since one sometimes wants to search on any of these fields, but we're a long ways away.  Before that, getting texts and repeats in the XML files is much more important.  Some of the analyses of our chorale set differ from other analyses because almost all repeats are expanded.  Margaret Greentree did a great job getting all the initial data in and I'd love to honor her legacy by making the music even better.

I also haven't checked any of the suspicious moments against the NBA to see if they have been emended.  Though I also don't know the NBA's editorial standards -- I would hate to have amazing moments removed simply because someone thought, "Bach wouldn't do that" when all evidence points to him having done so.  (Such as the one vocal melodic Augmented 2nd which is clearly attested to in the manuscript)

doctor-schmidt:
I understand your point, the corpus chorales should be as close as possible to Bach himself, although most of what we have is from CPE. I love your attention to detail!

Introducing lyrics in the XML files will be nice. I might start doing that in my personal collection (most transcribed by my own hand, with replats, and beaming being the main difference). Maybe in the future I could contribute to consolidating ChoraleList() and ChoraleListRKVWV().

Thanks for music21!
	
Norman

PR #277, opened 2018-03-01
Two newly implemented jSymbolic feature extractors, one corrected feature extractor

mjwalter:
Implemented DurationFeature and VariabilityOfNoteDurationFeature.

In addition, the NoteDensityFeature gave incorrect results when compared against the original jSymbolic implementation. Its approach also seemed inelegant: it involved discretizing the piece by seconds. (Doing so may potentially have allowed for finding the note density of a piece at various points, but such was not the original function of the jSymbolic version, and it was an unwieldy method.) I reimplemented NoteDensityFeature using a simpler approach.

Finally, a stray typo correction found its way into my master branch.


mjwalter:
The tests are currently working on my local repository, but I notice that the tests are failing in the build checks: `f.vector` is apparently set to be a float and not an integer. I am making sure that the value is assigned to `f.vector[0]` and not `f.vector` itself, but the problem still results. I will look to see why this might be the case.

mscuthbert:
Looks really good -- can you add tests for the two feature extractors you've implemented and also check that they're included in the get_all_extractors method (I can't remember what it's called).  THANKS!  Very happy with this.  Will merge when the Coverage percentage stays the same or increases.

mjwalter:
Working on it now! I have a bunch of other edits on this file – after this gets merged, I'll make another request.

mjwalter:
@mscuthbert How do you want to categorize DurationFeature? As rhythmic? I don't think it has an ID assigned.

mscuthbert:
Strange — I can’t find this feature in JSymbolic at all — it seems to be a helper for a number of other R features that divide X by the length of the recording.  We could call it R0, or maybe it belongs in native.py?

On Mar 1, 2018, at 15:21, Micah Walter <notifications@github.com<mailto:notifications@github.com>> wrote:


@mscuthbert<https://github.com/mscuthbert> How do you want to categorize DurationFeature? As rhythmic? I don't think it has an ID assigned.

mjwalter:
Ah, I see it here on the [jSymbolic GitHub](https://github.com/DDMAL/jSymbolic2/blob/a98340f8dfd473b3de0d02cb3641c7d20c0c10bf/src/jsymbolic2/features/DurationFeature.java). It's called R-36.

mscuthbert:
Merging.  Congrats!

mscuthbert:
One suggestion for the future:  I like the simplification from:

```
>>> f = fe.extract()
>>> round(f.vector[0])
```

to 

```
>>> f = fe.extract().vector
>>> round(f[0])
```

etc., but when doing this, please change the variable name, even "v" is good:

```
>>> v = fe.extract().vector
>>> round(v[0])
```

since `fe` is used pretty consistently to mean `FeatureExtractor` object, and `f` for `Feature` object, so using `f` to mean a vector could be confusing to people reading the docs.

mjwalter:
Got it, thanks for the suggestion!

Issue #273, opened 2018-02-15
Behavior of PitchClassDistributionFeature doesn’t match documentation


mjwalter:
The documentation for the jSymbolic class `PitchClassDistributionFeature` reads as follows:

> A feature array with 12 entries where the first holds the frequency of the bin of the pitch class histogram with the highest frequency, and the following entries holding the successive bins of the histogram, wrapping around if necessary.

This is similar to the original jSymbolic implementation. However, the actual music21 implementation doesn't reorder the array to put the highest-frequency. Furthermore, the music21 implementation normalizes the highest frequency to 1 and expresses the frequencies of the other pitch classes as fractions of that.

(I do not know which other classes, if any, have an implementation that varies from the documentation.)

mscuthbert:
Is still an issue or did the recent pull request address it?

mjwalter:
I'll be making another pull request soon that addresses this!

PR #278, opened 2018-03-05
Updates to feature extractors to match jSymbolic implementation

mjwalter:
The current music21 feature extractors "normalize" by dividing by the maximum element in an array. This keeps all values <= 1, but is not what normalization is (and is not how jSymbolic implements it). Rather, the *sum* of all values in the array should be 1.

"Most Common Pitch" and "Fifths Pitch Histogram" each had a specific difference in implementation from jSymbolic; these have also been corrected.

This pull request would address issue #273.


mscuthbert:
aha! I thought we were following JSymbolic's methods -- if they don't do the normalization of setting the largest to 1 then there's no reason we should either.  Will watch the build and then commit if it passes.  If it doesn't, feel free to reuse the method "normalize" to do the new thing -- in fact that would be better, so that if anyone has written an extractor that calls "normalize" it will still work, just with the new normalization scheme.

mjwalter:
Note that my updated test suites referred to the features as extracted by jSymbolic (and not just the result of my new code).

mjwalter:
I missed a couple test cases elsewhere in the project to update – will continue to work on this until it passes.

mscuthbert:
Thanks!

It's possible that bwv66.6 might at some point trigger a change of meter problem as Norman is updating the chorales and putting in repeats, so we can fix that to another file at that time.

PR #281, opened 2018-03-06
Corrected melodic arcs feature extractors; other minor changes

mjwalter:
The main material here is the rewritten functions for the feature extractors "size of melodic arcs" and "average duration of melodic arcs" (commit 1bd7d31). I compared the functionality with the jSymbolic implementation, and rewrote the test cases.

This pull request also eliminates `common.standardDeviation` in favor of the already-imported `statistics.pstdev`.


mjwalter:
I hope that commit 889533c, which mostly just edits the large jSymbolic.py file for formatting consistency and clarity (along with slightly enhanced documentation), doesn't pose problems. It should be independent of the more functional commits.

music21/features/jSymbolic.py
```diff
@@ -607,30 +591,45 @@ def process(self):
 
 class DurationOfMelodicArcsFeature(featuresModule.FeatureExtractor):
     '''
-    Average number of notes that separate melodic peaks and troughs in any channel.
+    Average number of notes that separate melodic peaks and troughs
+    in any part. This is calculated as the total number of intervals
+    (not counting unisons) divided by the number of times the melody
+    changes direction.
+    
+    >>> # C D E D C E D C C
```
mscuthbert: Comments are better done without ">>> #"  There's no reason to put them with the code inside a docstring unless they belong on the same line as code

music21/features/jSymbolic.py
```diff
@@ -640,70 +639,89 @@ def process(self):
...
-    Average melodic interval separating the top note of melodic peaks and the
-    bottom note of melodic troughs.
+    Average span (in semitones) between melodic peaks and troughs
+    in any part. Each time the melody changes direction begins a
+    new arc. The average size of melodic arcs is defined as the
+    total size of melodic intervals between changes of directions -
+    or between the start of the melody and the first change of
+    direction - divided by the number of direction changes.
+    
+    >>> # C D E D C E D C C
```
mscuthbert: as above

mjwalter:
Updated the doctests!

The percentage of lines covered by tests technically decreases because the total number of lines of code is smaller, but I don't think I've effectively reduced coverage.

mjwalter:
Hi @mscuthbert – just letting you know that I made the docstring correction; hopefully this update is ready to merge!

mscuthbert:
Hi! Thanks for doing that.  The docs will be improved by it.  Sorry I missed the previous merge, and btw -- I don't mind that the doc test % decreased because of the removal of redundant messages -- that was fine, so feel free to go back and reintroduce that in another pull request.  I'll break the rule against doing a merge between beta/rc1 and the actual release and include this since everything had already passed before that moritorium and I'm sure you'll want this in 5.0 release.

mjwalter:
Thanks! I'll pull the new rc and continue working :)

mscuthbert:
Thanks also for shedding standardDeviation -- it was a relic from before Py 3 was required.  However, in the future, let's mark functions like that as deprecated for a cycle and stop using them in our own code before removing them for everyone.  (I probably should not have merged that right before a release since it might be breaking code...but we're a small operation.)

mjwalter:
Yes, I see how that could be an issue. Hopefully the impact will be minimal, especially considering that it is a common function.

PR #282, opened 2018-03-07
Update chorales.py

doctor-schmidt:


mscuthbert:
great!  we can also do chorale submissions as pull requests if that's easier.  Just overwrite the files in corpus/bach!

PR #284, opened 2018-03-08
Add files via upload

doctor-schmidt:
Added lyrics


mscuthbert:
I can merge anytime but might as well keep adding to this for a little bit and then I can fix all the new errors at once.  One of the pieces seem to be a piece for the testing of graph offsets so the changed measure numbers has got to it.

doctor-schmidt:
No need to merge anytime soon. I use Musescore for encoding, maybe Finale or music21 would be better. Show me an example of what you fix.

mscuthbert:
I think this might need to be resubmitted -- apologies!!!

PR #286, opened 2018-03-21
Placeholder for base12 and base 26

MarkGotham:


mscuthbert:
Looks great to me -- I think I have a slight preference for putting all this into base40 since they're subtypes of the same general idea, but merging now and we can move later.

MarkGotham:
Great, thanks -- v happy to put it in base40 if you prefer.

Issue #285, opened 2018-03-18
Problems with show('braille')


ajirving:
Calling show('braille') on any object fails from the standard Python interpreter (rather than using a notebook).  For example:

>>> n=note.Note()
>>> n.show('braille')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\Alastair\AppData\Local\Programs\Python\Python36-32\lib\site-pac
kages\music21\base.py", line 2578, in show
    **keywords)
  File "C:\Users\Alastair\AppData\Local\Programs\Python\Python36-32\lib\site-pac
kages\music21\converter\subConverters.py", line 461, in show
    super().show(obj, fmt, app=None, subformats=None, **keywords)
  File "C:\Users\Alastair\AppData\Local\Programs\Python\Python36-32\lib\site-pac
kages\music21\converter\subConverters.py", line 182, in show
    returnedFilePath = self.write(obj, fmt, subformats=subformats, **keywords)
  File "C:\Users\Alastair\AppData\Local\Programs\Python\Python36-32\lib\site-pac
kages\music21\converter\subConverters.py", line 470, in write
    if 'ascii' in subformats:
TypeError: argument of type 'NoneType' is not iterable
>>>

Looking at the code I see two issues.  Firstly the line which causes the above failure should probably be changed to include if subformats is not None.  Fixing this the above example works for me.  Secondly, in the show method of the BrailleConverter class, when ipython is not being used, the super call should probably pass the value of subformats to the parent rather than None.  This is needed so that show('braille.ascii') works correctly.

mscuthbert:
ah confirmed.  will await a patch.

PR #287, opened 2018-03-21
Fix the show method of the Braille converter

ajirving:
Fix the show method of the Braille subconverter so it works from the standard python interpreter as well as with the .ascii subformat.  Fixes issue #285.


mscuthbert:
Looks good to me.

PR #288, opened 2018-03-22
Segment by rests

MarkGotham:


mscuthbert:
closing and reopening to get the diffs excluding base12_26, etc.

mscuthbert:
Can we get rid of the prints even in debug (switch to environLocal.debug() then it will use the main debug interface which can be activated by running test/toggleDebug or switching debug in .music21rc) and is base12_26 a mistake to include in this request?  

Is there a recurse() mode for working on a score, or does it assume a flat stream?  If so, can we put that in the docs.  Thanks!

mscuthbert:
THANKS MARK!

PR #295, opened 2018-04-20
Added support for accid.ges on accid element

raffazizzi:
Hello,

I've added a small change to support accid.ges on `<accid>` This is based on the support for accid.get on `<note>`.

Please let me know if there is anything more needed to properly contribute to the project.

We're hoping to use this fix in a workshop at the Digital Humanities Summer School at the University of Victoria this June 2018, so it would be awesome to have this integrated by then!

### DIFF
--- a/music21/mei/base.py
+++ b/music21/mei/base.py
@@ -1898,6 +1898,10 @@ def accidFromElement(elem, slurBundle=None):  # pylint: disable=unused-argument
     **Attributes/Elements Implemented:**
 
     - @accid (from att.accid.log)
+    - @accid.ges (from att.accid.ges)
+
+    .. note:: If set, the @accid.ges attribute is always imported as the music21 :class:`Accidental`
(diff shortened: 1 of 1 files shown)


mscuthbert:
Hi Raffael -- I would LOVE this change and would make the MEI import much better.  There are some problems in the tests for this that I will need to have fixed (either the change broke the tests or the expected answers were wrong).  Unfortunately, the test results files was overwhelmed by an unrelated deprecation warning triggered by Travis-CI's upgrade of matplotlib.  I've fixed the method that broke it, so if you can pull the latest music21 repository and then push again, you will be able to see the tests.  Or you can run them yourself: http://web.mit.edu/music21/doc/developerReference/testing.html

Thanks for contributing!

mscuthbert:
Tests passed -- failed test for Python 3.4 is no longer relevant.  Merging.  Thanks @raffazizzi !

PR #299, opened 2018-05-07
Significantly improve performance of midi file loading.

HalfVoxel:
This was done primarily by removing some cases were extremely inefficient quadratic algorithms were used where a simpler linear (or at least linear for all reasonable midi files which don't have all notes in the same location) algorithm would work.

There are still several cases were quadratic algorithms are still used and they definitely should be optimized.

In my tests this made midi file loading approximately 5.5x faster.

### DIFF
--- a/music21/stream/__init__.py
+++ b/music21/stream/__init__.py
@@ -9103,36 +9103,29 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
             flatStream = flatStream.sorted
         # these may not be sorted
         durSpanSorted = self._getDurSpan(flatStream)
+        # According to the above comment, the spans may not be sorted
+        # so we sort them to be sure, but keep track of their original indices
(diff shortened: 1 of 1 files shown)


mscuthbert:
Coverage decrease is acceptable since it comes only from reducing the length of well-tested code.

mscuthbert:
Merging.  Would love to have more places where quadratic speedups can be introduced.  One major culprit remaining is in the parsing of multi-staff (piano for instance) parts in musicxml.

HalfVoxel:
A good candidate for optimization is the deep copy methods. They are currently quadratic because the purgeOrphans method (which is called for every note) usually has to look through all elements in the stream that is being deep copied.

mscuthbert:
ah! I think that there was a planned item to improve purgeOrphans at some point a long time ago -- I'm not sure if we timed it in real life cases and figured it wasn't a problem, or just never got around to it.

mscuthbert:
(Sorry thought it was already merged!)

PR #301, opened 2018-05-10
Fix wrong __repr__ invocations

danlupei:


parkjh688:
I have still same problem with changed code in python3.

This one `TypeError: expected 0 arguments, got 1`

mscuthbert:
@parkjh688 can you confirm you are on the latest m21 version. What is the full traceback?

parkjh688:
@mscuthbert Mine was the latest m21 version. Yesterday I got #301 error and then I upgrade m21. 2 days ago it occurred same error like #301 but now it works well. Thanks.

PR #303, opened 2018-05-18
Fixed wrong b to - conversion (shouldn't convert lowercase pitch b)

vinzentt:
We fixed a bug where pitch 'b' was converted to a '-' (ultimately raising an exception), when the intention was to convert a flat specified with a 'b'. This bug affected the creation of a chord that contained pitch 'b' either in its root or its bass.

Incidentally, the calls to replace() in music21/harmony.py lines 249 and 255 had not effect, so the bug wasn't always manifesting itself, depending on how the chords were instantiated.

Co-authored-by: Vincent Degroote <1230888+vinzentt@users.noreply.github.com>
Co-authored-by: Alexandre Papadopoulos <32906275+a-papadopoulos@users.noreply.github.com>

### DIFF
--- a/music21/abcFormat/translate.py
+++ b/music21/abcFormat/translate.py
@@ -284,6 +284,7 @@ def parseTokens(mh, dst, p, useMeasures):
                 cs_name = t.chordSymbols[0]
                 cs_name = re.sub('["]','', cs_name).lstrip().rstrip()
                 cs_name = re.sub('[()]', '', cs_name)
+                cs_name = common.cleanedFlatNotation(cs_name)
                 try:
                     cs = harmony.ChordSymbol(cs_name)
                     dst.coreAppend(cs, setActiveSite=False)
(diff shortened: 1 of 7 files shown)


vinzentt:
We wrote tests for this bugfix which highlighted some other bugs, so we will come back to this PR next week after we fixed the tests and the associated bugs.

mscuthbert:
thanks!  I was going to ask for some specific tests and docs for what _cleanedFlatNotation does (because flat also has the meaning of .flat it's not entirely clear).  Thanks!

mscuthbert:
We can ignore the Py3.5 error in Travis-CI -- seems to be a bug with the most recent release of SciPy.

a-papadopoulos:
Fair enough, if you have a suggestion for a better name for what this method does, I will use it instead.

mscuthbert:
Is the Nottingham dataset on a Creative Commons or other free license? Just want to be sure we are not adding copyright problems.

mscuthbert:
Other than that clarification ready to merge.

a-papadopoulos:
Good question! I've actually haven't found any licence information. The oldest source I've found is this rather old page of Eric Foxley's:
http://www.chezfred.org.uk/freds/music/database.htm
It looks pretty safe to me...

mscuthbert:
They are claiming a GPL license over the conversion which is quite tricky -- would they claim that all of music21 needs to be licensed under GPL if we distribute it?  I compared the versions in http://abc.sourceforge.net/NMD/nmd/reelsa-c.txt to the versions here and the cleaning up seems light enough (mostly adding forward repeats but also a little code to Cuckoo's nest that actually looks wrong) that it would be hard to claim it's a sufficient creative work to qualify for a new copyright, but we need to be sure.  I've posted a note at https://research.jukedeck.com/releasing-a-cleaned-version-of-the-nottingham-dataset-928cdd18ec68 and we'll see what response comes.

a-papadopoulos:
To be clear, we were using the original version (http://abc.sourceforge.net/NMD/nmd/reelsa-c.txt), not the cleaned version (from jukedeck), or at least, so was our intention. I can double check that we are using the original version, if you think that's enough in terms of copyright. Alternatively, we could really just create our own simple examples, the intention was just to test that the conversion works fine with different chord symbols (although we did run it on the whole Nottingham dataset for testing purposes).

mscuthbert:
the version in the commits seems to be the cleaned version with repaired repeat signs.

mscuthbert:
Looks good!  Merging and THANK YOU!

PR #304, opened 2018-05-29
Added new procedures for pickling and unpickling

00sapo:
I added new methods to freezeThaw.py, as it was born in my master thesis. It allows to pickling and unpickling objects just like the standard `pickle` module. Now can use `from music21 import freezeThaw as pickle` to be compatible with other modules aimed at serialization. It also supports compression. Moreover, it works fine with the standard `multiprocessing` module.

I also added a bit of documentation, but I didn't manage to build it.

Line 1367 in `freezeThaw.py` is very ugly, but it is needed because `freezeThaw` is loaded before than `stream` module is ready.

This code was born from [this discussion on music21 google group](https://groups.google.com/d/msg/music21list/VTy1cA2D1yw/wLyq7IKqBQAJ)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,5 @@ documentation/source/tutorials/MersenneDemo_files/_fig_05.png
 .cache/v/cache/lastfailed
 
 .ropeproject/config.py
+music21/.ropeproject/
+music21/test/.ropeproject/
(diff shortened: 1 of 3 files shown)


mscuthbert:
@00sapo -- I didn't ever review this because the tests failed.  Is this still something you're interested in merging?  Thanks!

00sapo:
Hello, at now I don't plan to use this feature of music21 again, but if you think that it could be useful for all, I can try to solve the errors :)

mscuthbert:
Closing -- can reopen if someone is willing to maintain..

Issue #309, opened 2018-06-28
Specify no chord symbol


a-papadopoulos:
Hi,
I'd like to be able to specify in a Score that there is no ChordSymbol occurring during some time interval. Currently, the convention is that a ChordSymbol lasts until a new one is found. Of course, we can always set the duration of a particular ChordSymbol, but that's very fragile, for example, applying a music21.harmony.realizeChordSymbolDurations(piece) would break the durations.
Is there currently any way of doing this?
Otherwise, I was thinking of implementing a special type of "No Chord" or "N.C." ChordSymbol. I believe that should be pretty straightforward, but is there anything special I should look out for?

mscuthbert:
Great idea.  I think that maybe a ChordSymbol subclass called NoChordSymbol would be best?  As a subclass it would still be found with .getElementsByClass(ChordSymbol) but it would have more configuration options.  

I'm not sure what we're currently doing when we parse something from MusicXML with no-chord though... will investigate.

a-papadopoulos:
I've investigated this a bit, at the moment music21 parses them correctly. MusicXML [encodes them with the "none" kind](https://usermanuals.musicxml.com/MusicXML/Content/ST-MusicXML-kind-value.htm), and this is exactly how it gets parsed in music21.
So I was wondering if we shouldn't just keep things like that (kind == "none") instead of having a specific subclass. I'll check that a bit more to verify that this doesn't entail too many special cases that would justify subclassing.

mscuthbert:
Okay, so it seems that the only thing we need to do is make sure docs are clear about how to do this.  Happy to take a documentation edit pull-request.  Thanks!

a-papadopoulos:
Yes the doc will need to be updated (hadn't though about this), but I think there's be some more work. 

For example, parseFigure will need to be able to parse things like 'N.C.' or 'NC' or 'No Chord'. I believe this should make it possible to parse abc with NCs in it (even if that's not standard abc), but that will probably break some parts of the code that check for well-formed figures...

Also, there's little tricks to take care of when exporting back to MusicXML (e.g. hide the root of the chord which is required for all chords in MusicXML including "none" chords).

Anyway, I'll get started and come back with updates.

mscuthbert:
[...]

PR #320, opened 2018-08-09
Add support for no chords

a-papadopoulos:
This closes issue #309


a-papadopoulos:
http://lilypond.org seems to be down these days, and the check won't complete because of that. The tests look fine though on my system.

mscuthbert:
Yeah, I had noticed it was down.  I'll merge and unmerge if problems arise.  Thanks so much!

PR #323, opened 2018-08-21
Fix xml serialization of fingering articulations

hofst:
Unlike other articulations, fingers should not only be applied to the first note of a chord. Fingerings must be applied in orderd, one each to one note respectively. This commit fixes this behavior.

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -3099,7 +3099,7 @@ class MeasureExporter(XMLExporterBase):
 
         return notations
 
-    def noteToXml(self, n, addChordTag=False, chordParent=None):
+    def noteToXml(self, n, noteIndexInChord=0, chordParent=None):
         '''
         Translate a music21 :class:`~music21.note.Note` or a Rest into a
         ElementTree, note element.
@@ -3189,6 +3189,7 @@ class MeasureExporter(XMLExporterBase):
         TODO: Test with spanners...
 
         '''
+        addChordTag = (noteIndexInChord != 0)
         setb = _setAttributeFromAttribute
 
         if chordParent is None:
@@ -3363,7 +3364,7 @@ class MeasureExporter(XMLExporterBase):
 
         # TODO: staff
 
-        mxNotationsList = self.noteToNotations(n, addChordTag, chordParent)
+        mxNotationsList = self.noteToNotations(n, noteIndexInChord, chordParent)
 
         # add tuplets if it's a note or the first <note> of a chord.
         if addChordTag is False:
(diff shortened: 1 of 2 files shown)


music21/musicxml/m21ToXml.py
```diff
@@ -3858,7 +3858,7 @@ def noteheadToXml(self, n):
             mxNotehead.set('color', color)
         return mxNotehead
 
-    def noteToNotations(self, n, notFirstNoteOfChord=False, chordParent=None):
+    def noteToNotations(self, n, pitchNumber, chordParent=None):
         '''
```
mscuthbert: This needs a default value for pitchNumber.

music21/musicxml/m21ToXml.py
```diff
@@ -3099,7 +3099,7 @@ def objectAttachedSpannersToNotations(self, obj, objectSpannerBundle=None):
 
         return notations
 
-    def noteToXml(self, n, addChordTag=False, chordParent=None):
+    def noteToXml(self, n, pitchNumber=0, chordParent=None):
```
mscuthbert: pitchNumber is too vague as a variable name -- does it mean C=0, C#=1? is it midi number?  it might be "noteIndexInChord" or something like that

music21/musicxml/m21ToXml.py
mscuthbert: and again the name "pitchNumber" could mean too many different things.

music21/musicxml/m21ToXml.py
```diff
@@ -3892,15 +3891,27 @@ def noteToNotations(self, n, notFirstNoteOfChord=False, chordParent=None):
...
-                else:
-                    if mxArticulations is None:
-                        mxArticulations = Element('articulations')
-                    mxArticulations.append(self.articulationToXmlArticulation(artObj))
+        # apply all articulations apart from fingerings only to first note of chord
+        from music21.articulations import Fingering
+        applicableArticulations = []
+        fingeringNumber = 0
+        for a in chordOrNote.articulations:
+            if type(a) is Fingering:
```
mscuthbert: do not do type(x) checks.  use isinstance -- otherwise subclasses fail.  But better still is use the music21.base.Music21Object.classes or classSet interface -- it is much faster than isinstance.

music21/musicxml/m21ToXml.py
```diff
@@ -3892,15 +3891,27 @@ def noteToNotations(self, n, notFirstNoteOfChord=False, chordParent=None):
...
+                if fingeringNumber == pitchNumber:
+                    applicableArticulations.append(a)
+                fingeringNumber += 1
+            elif pitchNumber == 0:
+                applicableArticulations.append(a)
```
mscuthbert: what if there are more Fingering articulations than pitches? will they be lost?

music21/musicxml/m21ToXml.py
```diff
@@ -3870,14 +3870,13 @@ def noteToNotations(self, n, notFirstNoteOfChord=False, chordParent=None):
...
-            if chordParent is not None:
-                # get expressions from first note of chord
-                chordOrNote = chordParent
+        chordOrNote = n
+        if chordParent is not None:
+            # get expressions from first note of chord
+            chordOrNote = chordParent
 
+        # only apply expressions to notes or the first note of a chord...
+        if pitchNumber == 0:
```
mscuthbert: I still prefer setting up a separate variable that makes more clear what "pitchNumber == 0" means:
`if isSingleNoteOrFirstInChord:`  even a wordy expression like this is helpful to show what is being tested.

music21/musicxml/m21ToXml.py
mscuthbert: this is otherwise really good!

music21/musicxml/m21ToXml.py
```diff
@@ -3858,7 +3858,7 @@ def noteheadToXml(self, n):
             mxNotehead.set('color', color)
         return mxNotehead
 
-    def noteToNotations(self, n, notFirstNoteOfChord=False, chordParent=None):
+    def noteToNotations(self, n, pitchNumber, chordParent=None):
```
mscuthbert: need tests to show that this works

mscuthbert:
Do we need corresponding changes in xmlToM21?  I hate to have the capabilities of import become unbalanced from export.

hofst:
import already works as intended (the articulations of the chord are collected from every note)

music21/musicxml/m21ToXml.py
hofst: yes they will be intentionally lost. I added that behavior to doc and test

hofst:
@mscuthbert I updated the pull request and believe I have now addressed all your review comments.

mscuthbert:
there's one unused variable in there, but I can fix that.  Congrats and THANK YOU!  Approved!

PR #326, opened 2018-08-29
Added IPython .show('osmd') using OpenSheetMusicDisplay

### LINKED ISSUE #306: Easy way to view musicXML in a browser
I ran into this excellent library for rendering sheet music in-browser using MusicXML, and wondered if it could be used with music21: [Open Sheet Music Display 
](https://github.com/opensheetmusicdisplay)

Turns out it is possible with a bit of js-hackery! I've posted the script below. This makes it easy to display scores without installing any other programs (apart from a web browser).

The script requires the following file in the same directory (I can't upload .js files here, but it works fine as .txt). It's simply a compiled version of the OSMD project.
[opensheetmusicdisplay.min.js.txt](https://github.com/cuthbertLab/music21/files/2058083/opensheetmusicdisplay.min.js.txt)

This should display a random score from the corpus, enjoy!
```python3
from music21 import *
import webbrowser, os, random, pathlib

osmd_js_file = os.path.join(os.path.dirname(os.path.realpath(__file__)),"opensheetmusicdisplay.min.js.txt")

if not pathlib.Path(osmd_js_file).is_file():
	print("ERROR: need ./opensheetmusicdisplay.min.js.txt at",osmd_js_file) 
	exit()
print("found required .js")


selected_piece = random.choice(corpus.getPaths())
print('loading piece', selected_piece)
b = corpus.parse(selected_piece)


def stream_to_web(b):
	html_template = """
	<html>
	    <head>
	        <meta http-equiv="content-type" content="text/html; charset=utf-8" />
	        <title>Music21 Fragment</title>
			<script src="{osmd_js_path}"></script>
		</head>
		<body>
			<span>may take a while to load large XML...<span>
			<div id='main-div'></div>
			<button opensheetmusicdisplayClick="show_xml()">Show xml</button>
			<pre id='xml-div'></pre>
			<script>
			var data = `{data}`;
			function show_xml() {{
				document.getElementById('xml-div').textContent = data;
			}}

			  var openSheetMusicDisplay = new opensheetmusicdisplay.OpenSheetMusicDisplay("main-div");
			  openSheetMusicDisplay
			    .load(data)
			    .then(
			      function() {{
			        console.log(openSheetMusicDisplay.render());
			      }}
			    );
			</script>
		</body>

	"""

	osmd_js_path = pathlib.Path(osmd_js_file).as_uri()

	filename = b.write('musicxml')
	print("musicXML filename:",filename)
	if filename is not None:
		with open(filename,'r') as f:
			xmldata = f.read()
		with open(filename+'.html','w') as f_html:
			html = html_template.format(
				data=xmldata.replace('`','\\`'),
				osmd_js_path=osmd_js_path)
			f_html.write(html)

		webbrowser.open('file://' + os.path.realpath(filename+'.html'))


stream_to_web(b)
```

supersational:
A proper version of the hack described here: Closes #306

Within an IPython notebook you can now use `score.show('osmd')` and have the score displayed inline.

Issues:
- Empty scores sometimes cause a Javascript error
- Might want to generate a .html file if not in an IPython environment
- conditional import of IPython needs adding/testing
- Minor issue but if several scores are shown at the same time it might add an unnecessary amount of script tags in a JS race condition


supersational:
Actually there's an issue with the script path that needs fixing..

mscuthbert:
IPython is not a music21 dependency so be sure that the script runs without it and have as many tests as possible that don’t require it so that other users can test without it. That means all IPython imports should be tested and return a nice error (“install IPython via...”) if a necessary call is made in an environment without it

mscuthbert:
Can more of the logic go into an osmd.py file and out of subconverters? Trying to put that as the bare minimum to the interface. The IPython muse score etc are a bad example to follow.

supersational:
Now moved to seperate module, seemed appropriate since we have additional .js files.

I'm not sure what the convention for import errors are, but it should now ignore missing IPython until .show('osmd') is called.

supersational:
Testing is quite difficult given that it needs to run in a Jupyter environment for it to work. Unfortunately we can't test for javascript console errors, but I'll try adding a notebook for testing in the meantime.

mscuthbert:
Search for the scipy import to see how to test for installation without importing. Importing IPython takes time so it shouldn’t be done unless it’ll be used. The graph module also has a number of nice tricks for imports.

mscuthbert:
I'm concerned that coverage decreased after the tests.  We can add IPython to the testing environment if that's necessary to make tests involving IPython, but I wonder if we can't test at least that the code to generate the `<script>` tags is working properly.

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+class OpenSheetMusicDisplayException(exceptions21.Music21Exception):
+    pass
+
+hasInstalledIPython = False
+try:
+    # raise ImportError('for testing')
+    from IPython.core.display import display, HTML, Javascript
+    hasInstalledIPython = True
```
mscuthbert: This still will import upon load.  Look at base.py and graph.utilities.  The way to check for installation without importing is:

```python
    loader = importlib.util.find_spec('IPython.core.display')
    if loader is None:
        hasInstalledIPython = False
    else:
        hasInstalledIPython = True
```

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+    '''
+    Converter for displaying sheet music inline
+    in an IPython notebook.
+    Uses: https://opensheetmusicdisplay.org/ for rendering
+    '''
+    registerFormats = ('osmd',)
+    registerShowFormats = ('osmd',) 
+
+    def show(self, obj, fmt, 
+            fixPartName=True, offline=True, div_id=None,
```
mscuthbert: Feel free to add "*' before fixPartName -- the rest should be keyword only.

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+            **keywords):
+        score = obj
+        if fixPartName:
+            score = self.addDefaultPartName(score)
+
+        if div_id is None:
+            # create unique reference to output div in case we wish to update it
+            div_id = self.getUniqueDivId()
+            # div contents should be replaced by rendering
+            display(HTML('<div id="'+div_id+'">loading OpenSheetMusicDisplay</div>'))
```
mscuthbert: Please space the + operator (`"' + div_id + '"`)

music21/opensheetmusicdisplay/osmd.py
mscuthbert: `div_id` should be `divId` -- sorry that music21 was originally a Perl package and we've stuck with CamelCase variables throughout.  It's okay to use PEP8 underscore spacing in internal variables, but since it's exposed in the arguments, it should conform to the music21 convention (which has worked out well for music21j, since Javascript uses camelcase conventions)

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+    in an IPython notebook.
+    Uses: https://opensheetmusicdisplay.org/ for rendering
+    '''
+    registerFormats = ('osmd',)
+    registerShowFormats = ('osmd',) 
+
+    def show(self, obj, fmt, 
+            fixPartName=True, offline=True, div_id=None,
+            **keywords):
+        score = obj
```
mscuthbert: Does obj need to be a stream.Score()?  can it be any stream, a part, a measure, a single note, etc.?  Generally, we allow .show() to be run on any musicxml element, so if it's not a supported element, we wrap it so it can be show.

This might be a place where the code to do all this in musicxml should be factored out into the generic subconverter class so it can be used in any converter method.

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+
+    @staticmethod
+    def musicXMLToScript(xml, div_id, offline=True):
+        
+        # print('xml length:', len(xml))
+        script_dir = os.path.join(os.path.dirname(__file__), 'opensheetmusicdisplay.0.3.1.min.js')
+
+        # script that will replace div contents with OSMD display
+        script = """
+        console.log("loadOSMD()");
```
mscuthbert: these and other console.log should be wrapped in environLocal debug tests -- eventually we don't wan't debugging information in either the javascript or python consoles.

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+        function loadOSMD() { 
+            return new Promise(function(resolve, reject){
```
mscuthbert: we'll need a release note saying that IE11 is not supported.  I'm totally fine with that.  Edge isn't perfect, but IE must die.  However, I'll be the one getting complaints about it not working if it's not totally clear.  In fact I'd be fine with adding:

```
try {
    Promise;
} catch {
    alert("Sorry, OSMD is not supported in Internet Explorer or other pre-ES6 browsers");
}
```

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+                "-"+str(time.time()).replace('.','-') # '.' is the class selector
+
+    @staticmethod
+    def musicXMLToScript(xml, div_id, offline=True):
+        
+        # print('xml length:', len(xml))
+        script_dir = os.path.join(os.path.dirname(__file__), 'opensheetmusicdisplay.0.3.1.min.js')
+
+        # script that will replace div contents with OSMD display
+        script = """
```
mscuthbert: if the whole script tag is this extensive, maybe it should be a separate file and something to load in, so that we can run javascript checkers/linters on it?

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+
+    @staticmethod
+    def getUniqueDivId():
+        return "OSMD-div-"+ \
+                str(random.randint(0,1000000))+ \
+                "-"+str(time.time()).replace('.','-') # '.' is the class selector
+
+    @staticmethod
+    def musicXMLToScript(xml, div_id, offline=True):
+
```
mscuthbert: this can be tested with a minimal xml file -- there are several in the musicxml directory.

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+                };
+                {{script_command}}
+                
+            }) 
+        }
+        loadOSMD().then((OSMD)=>{
+            console.log("loaded OSMD",OSMD)
+            var div_id = "{{DIV_ID}}";
+                console.log(div_id)
+            document.querySelector('#'+div_id).innerHTML = "";
```
mscuthbert: again watch spacing of operators (including in Javascript)

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+
+        # script that will replace div contents with OSMD display
+        script = """
+        console.log("loadOSMD()");
+        function loadOSMD() { 
+            return new Promise(function(resolve, reject){
+
+                if (window.opensheetmusicdisplay) {
+                    console.log("already loaded")
+                    return resolve(window.opensheetmusicdisplay)
```
mscuthbert: missing semicolon -- please run the Javascript code through eslint or another linter.  Preference is eslint with the preferences from music21j's .eslintrc.json.

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+                function oncompleted(){
+                    window.define = _define
+                    console.log("loaded OSMD for the first time",opensheetmusicdisplay)
+                    resolve(opensheetmusicdisplay);
+                };
+                {{script_command}}
+                
+            }) 
+        }
+        loadOSMD().then((OSMD)=>{
```
mscuthbert: Arrow functions aren't supported in iOS 9 which is still a good chunk of the audience.  If they are to be used then other ES6 terms should be used, such as "let" and "const" instead of var.  I'm okay with dropping Safari 9 as a supported system if it's made explicit, but I'm a little worried that this code hasn't been tested in a variety of browsers.

music21/opensheetmusicdisplay/osmd.py
mscuthbert: I'm not a big fan of nested functions like this, better to call:

```
function loadOSMD() {
    return new Promise(loadOSMDPromise);
}

function loadOSMDPromise(resolve, reject) {
    ...
}
```

and thus making more testable functions.

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+        """ \
+        .replace('{{DIV_ID}}',div_id) \
+        .replace('{{data}}',json.dumps(xml)) \
+        .replace('{{script_dir}}',pathname2url(script_dir))
+
+        if offline is True:
+            script_content = open(script_dir).read()
+            script = script.replace('{{script_command}}',"""
+                s.type = 'text/javascript';
+                s.text="""+json.dumps(script_content)+""";
```
mscuthbert: Can you explain this bit some more.  Is this secure against maliciously coded XML?  Can script_content be something that causes the json.dumps to produce text that leaks into the rest of the code.  I don't like injecting user content into a general script tag.

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+                .load({{data}})
+                .then(
+                  function() {
+                    console.log("rendering data")
+                    openSheetMusicDisplay.render();
+                  }
+                );
+        })
+        """ \
+        .replace('{{DIV_ID}}',div_id) \
```
mscuthbert: why do these need to be replacements?  Can't they be variables passed to the javascript function?

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+        if offline is True:
+            script_content = open(script_dir).read()
+            script = script.replace('{{script_command}}',"""
+                s.type = 'text/javascript';
+                s.text="""+json.dumps(script_content)+""";
+                document.body.appendChild( s ); // browser will try to load the new script tag
+                oncompleted();
+                """)
+        else:
+            script = script.replace('{{script_command}}',"""
```
mscuthbert: forcing online use seems to be a better solution.  The URL can be set as a variable that can be passed to the function, and if someone wants to work offline, they can point to a file:/// or other local resource.

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+    def addDefaultPartName(score):
+        # If no partName is present in the first instrument, OSMD will display the ugly 'partId'
+        allInstruments = list(score.getInstruments(returnDefault=False, recurse=True))
+        
+        if len(allInstruments)==0:
+            # print("adding default inst")
+            defaultInstrument = Piano()
+            defaultInstrument.instrumentName = 'Default'
+            score.insert(None, defaultInstrument)
+            score.coreElementsChanged()
```
mscuthbert: not needed.

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+    @staticmethod
+    def addDefaultPartName(score):
+        # If no partName is present in the first instrument, OSMD will display the ugly 'partId'
+        allInstruments = list(score.getInstruments(returnDefault=False, recurse=True))
+        
+        if len(allInstruments)==0:
+            # print("adding default inst")
+            defaultInstrument = Piano()
+            defaultInstrument.instrumentName = 'Default'
+            score.insert(None, defaultInstrument)
```
mscuthbert: If "None" as a location works, it's a bug that I'll try to fix.  Use 0.0 as an offset.

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+            score.coreElementsChanged()
+        elif not allInstruments[0].instrumentName: 
+            # print("adding instrumentName")
+            # instrumentName must not be '' or None
+            allInstruments[0].instrumentName = 'Default'
+            score.coreElementsChanged()
```
mscuthbert: The elements haven't changed in this case, only the name of the instrument, so no need to call coreElementsChanged() -- coreElementsChanged() only needs to be run if another `core*` function has been called.  All general music21 stream commands will call it if necessary and not call it if not.

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+
+        return score
+
+class Test(unittest.TestCase):
+
+    @unittest.skipUnless(hasInstalledIPython, "skipping since IPython not installed")
```
mscuthbert: I LOVE THIS.  Thank you!!!!

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+    def testOpenSheetMusicDisplayRuns(self):
+        from music21 import corpus, environment
+        environLocal = environment.Environment()
+
+        s = corpus.parse('luca/gloria').measures(1, 19)
+        # s.show('osmd')
+        ConverterOpenSheetMusicDisplay().show(s, None)
```
mscuthbert: this should be in TestExternal.  Test() only should be run if there are no side effects.  TestExternal is for anything that opens up a browser, plays music, etc.  Since the general test system runs all Test() functions, it isn't needed.

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+        return score
+
+class Test(unittest.TestCase):
+
+    @unittest.skipUnless(hasInstalledIPython, "skipping since IPython not installed")
+    def testOpenSheetMusicDisplayRuns(self):
+        from music21 import corpus, environment
+        environLocal = environment.Environment()
+
+        s = corpus.parse('luca/gloria').measures(1, 19)
```
mscuthbert: This is a big score that loading takes some time.  Unless there's a feature from it to be tested, a lighter-weight score (`bwv66.6` is a short score) can be used.

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def testAddsDefaultPartId(self):
+        from music21 import converter
+        s = converter.parse("tinyNotation: 3/4 E4 r f#")
+
+        s = ConverterOpenSheetMusicDisplay.addDefaultPartName(s)
+        firstInstrumentObject = s.getInstruments(returnDefault=True, recurse=True)[0]
+        # print("firstInstrumentObject",firstInstrumentObject)
+        # print("firstInstrumentObject.instrumentName",firstInstrumentObject.instrumentName)
+        assert firstInstrumentObject.instrumentName != None
```
mscuthbert: bare asserts cause the test system to crash.  Use self.assertNotEqual(firstInstrumentObject.instrumentName, None) or better, self.assertNotNone(firstInstrumentObject.instrumentName)

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+        assert firstInstrumentObject.instrumentName != ''
```
mscuthbert: same as above.

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+
```
mscuthbert: Take a look at the coveralls report to see what is not run and what is not tested in the module.

music21/__init__.py
```diff
@@ -107,6 +107,7 @@
     'musedata',
     'musicxml',
     'noteworthy',
+    'opensheetmusicdisplay',
```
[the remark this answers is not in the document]
supersational: renamed here: 69de31944f4637d75e55fdc47b2848af03611cd0

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
+# -*- coding: utf-8 -*-
+#-------------------------------------------------------------------------------
+# Name:         opensheetmusicdisplay/osmd.py
+# Purpose:      music21 classes for displaying music21 scores in IPython notebooks
+#
+# Authors:      Sven Hollowell
+#
+# Copyright:    Copyright © 2012-14 Michael Scott Cuthbert and the music21 Project
```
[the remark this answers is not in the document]
supersational: done 021d56a4cc14b8e7d610c636ece088c9a0627417

music21/opensheetmusicdisplay/osmd.py
supersational: Should be fixed now: 90779099fe7ffcebba6afe622e080a9ac8b2ebd5. Would be great to be able to check the output (e.g. a notebook that displays all the wigets and .js options?).

music21/opensheetmusicdisplay/osmd.py
supersational: e9484aec83962b2597e46b1224e2e659db533724 fixed the other issues but as far as I can tell the coveralls doesn't run the [testOpenSheetMusicDisplayRuns](https://coveralls.io/builds/18748610/source?filename=music21%2Fopensheetmusicdisplay%2Fosmd.py#L167) function? Hopefully the next report will run it.

music21/opensheetmusicdisplay/osmd.py
supersational: Good idea, done: c83f8c91fc3b781c98d669abc3bd60cb63b4d42b

music21/opensheetmusicdisplay/osmd.py
supersational: Unfortunately this is still needed (thought it would be fixed in the most recent osmd 0.5.1). 

This can be seen in many of the folk corpus pieces such as 'ryansMammoth/NorfolkHornpipe.abc', which will convert to musicXML with part-name `<part-name />` in them. This will then display in the score as 'Instr. P6d2a255fb4b93a88d140fb66e502d885' since osmd uses the id when part-name is missing.

That being said my attempted fix doesn't seem to work! Even changing `partName` and calling `coreElementsChanged` willl not reflect in the piece when written to musicXML.

supersational:
Added some more fixes, still need to sort the JavaScript-as-a-seperate-file issue since I agree it's currently a mess and doing linting would be very helpful. 

I think now that we have multiple files involved it should stay as a separate module so we can put associated files in the music/osmd/ e.g. Javascript downloads. If you'd stil like `__init__.py` to be removed then that's also fine, I was trying to follow the example of other modules when I added it.

Some (still) broken things:
- missing part-name (as discussed above) 
- coveralls report (will see as the results come through)
- some of the corpus pieces don't actually convert to musicXML so won't work! may investigate further but it has something to do with minimum note duration.

Again thanks for the detailed comments on this! I'm still getting used to python best practices about testing and so forth, and this project has some excellent standards in place.

Should be useable in notebook currently via:
``` python
import music21
s = music21.converter.parse("tinyNotation: 3/4 E4 r f# g=lastG trip{b-8 a g} c4~ c")
s.show('osmd')
```

music21/opensheetmusicdisplay/osmd.py
supersational: An example of what I mean:
image
image

mscuthbert:
https://travis-ci.org/cuthbertLab/music21/jobs/441779288  -- ModuleNotFoundError doesn't exist in Py3.5.

supersational:
Phew seems to have passed finally, hope I didn't use too many of your Travis credits.
For the coverage report would it be necessary to add IPython as an install step to `.travis.yaml` for the extra test functions to run? Seems like scipy and numpy are already there..

mscuthbert:
Absolutely -- and don't worry about Travis credits.  :-)

music21/osmd/osmd.py
```diff
@@ -0,0 +1,156 @@
...
+    '''
+    Converter for displaying sheet music inline
+    in an IPython notebook.
+    Uses: https://opensheetmusicdisplay.org/ for rendering
+    '''
+    registerFormats = ('osmd',)
+    registerShowFormats = ('osmd',)
+    # when updating the script tag osmd will only reload in a notebook if you: Kernel -> restart and clear output, then
+    # save the notebook and refresh the page. Otherwise the script will stay on the page and not reload.
+    script_url = """https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/releases/download/0.6.3/opensheetmusicdisplay.min.js"""
```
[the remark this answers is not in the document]
supersational: @mscuthbert just tried that but getting only the 'Usage: testLint.py' message. Is this the right command? `$ python music21/test/testLint.py ./music21/osmd/osmd.py` tried in git-bash.

music21/osmd/osmd.py
```diff
@@ -0,0 +1,156 @@
...
+            script_content = open(self.osmd_file).read()
+            script = script.replace('"{{offline_script}}"',json.dumps(script_content),1 )
+
+        else:
+            script = script.replace('"{{script_url}}"',json.dumps(self.script_url), 1)
+        return script
+
+    @staticmethod
+    def addDefaultPartName(score):
+        # If no partName is present in the first instrument, OSMD will display the ugly 'partId'
```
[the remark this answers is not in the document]
supersational: Hi @mscuthbert  I'm afraid this module is running into a lot of problems with tests! 

When I try to `.show('osmd')` in a test or doctest I get: `music21.base.Music21ObjectException: cannot support showing in this format yet: osmd`

Despite that this works in regular python and in notebooks.

Any ideas?

music21/osmd/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+# Copyright:    Copyright © 2012-14 Michael Scott Cuthbert and the music21 Project
+# License:      LGPL or BSD, see license.txt
+#-------------------------------------------------------------------------------
+
+__all__ = ['ConverterOpenSheetMusicDisplay']
+
+from music21.osmd.osmd import ConverterOpenSheetMusicDisplay
+
+from music21.converter import registerSubconverter 
+registerSubconverter(ConverterOpenSheetMusicDisplay)
```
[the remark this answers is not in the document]
supersational: Ok, there seems to be a lot of logic in each subConverters.py class, should this be a simple cut and paste or more complicated?

music21/osmd/osmd.py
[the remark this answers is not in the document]
mscuthbert: this should be working -- AH! for some reason, it assumes you're in the music21 directory or fails -- cd to music21 and run 'test/testLint.py osmd/osmd.py' instead and see if that works.  Will push a fix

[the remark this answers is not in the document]
supersational: Strange tried `python test/testLint.py osmd/osmd.py` and getting the same usage info with no error.

`--long-help` doesn't seem to work either, I get `testLint.py: error: unrecognized arguments: --long-help`

music21/osmd/osmd.py
```diff
@@ -0,0 +1,156 @@
...
+    script_url = """https://github.com/opensheetmusicdisplay/opensheetmusicdisplay/releases/download/0.6.3/opensheetmusicdisplay.min.js"""
+    osmd_file = os.path.join(os.path.dirname(__file__), 'opensheetmusicdisplay.0.6.3.min.js')
+
+    def __init__(self):
+        self.display, self.HTML, self.Javascript = getExtendedModules()
+
+    def show(self, obj, fmt,
+             fixPartName=True, offline=False, divId=None,
+             **keywords):
+        score = obj
```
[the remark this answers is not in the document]
psychemedia: By which you mean use `def show(self, obj, fmt, **keywords)` and handle the other args via `keywords`?

mscuthbert:
Pinging if still interested in including?  Thanks!

psychemedia:
As an edtechie outside the music field, I would find this really useful for helping demonstrate how music related educational resources can be developed and published via Jupyter notebooks and Jupyter mediated publishing tools.

mscuthbert:

supersational:
@mscuthbert just submitted some improvements. It now also works in non-IPython environments e.g. scripts and interactive mode,  by opening a tab with the rendered sheet music, pretty sweet! 

The only major outstanding issue is `Method 'parseData' is abstract in class 'SubConverter' but is not overridden (58:0) [abstract-method]`
Which I'm not sure how to resolve.. to be honest I have no idea how the 'SubConverter' class is supposed to work, I was just trying to copy existing implementations of `.show()`

jacobtylerwalls:
All you have to do is add `def parseData(show):  raise NotImplementedError('osmd is display-only')`!

music21/osmd/osmd.py
```diff
@@ -0,0 +1,222 @@
...
+            open(tmp_path, 'w').write(f"""
+            <div id="{divId}"></div>
+            <script>{script}</script>
+            """)
+            filename = 'file:///' + tmp_path
+            webbrowser.open_new_tab(filename)
+        return divId
+
+    def parseData(self):
+        raise NotImplementedError('osmd is display-only')
```
jacobtylerwalls, line 121: This line is very easy to either exempt from coverage altogether with a `# pragma: no cover` after the signature, or else just actually test with a `self.assertRaises(NotImplementedError, myScore.show('osmd'))`

music21/osmd/osmd.py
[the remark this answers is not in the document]
jacobtylerwalls: Keyword-only can be accomplished by adding a `*` to the list, i.e. after the positional arguments `obj` and `fmt` and before `fixPartName` (the * consumes any non-keyword argument)

music21/osmd/osmd.py
```diff
@@ -0,0 +1,222 @@
...
+        # If no partName is present in the first instrument, OSMD will display the ugly 'partId'
+        allInstruments = list(score.getInstruments(returnDefault=False, recurse=True))
+
+        if not allInstruments:
+            defaultInstrument = Piano()
+            defaultInstrument.instrumentName = ' '
+            score.insert(0.0, defaultInstrument)
+        elif not allInstruments[0].instrumentName:
+            # instrumentName must not be '' or None
+            allInstruments[0].instrumentName = ' '
```
jacobtylerwalls, line 182: In v7 the new way of doing this is to call `.first()` instead of indexing into [0].

music21/osmd/osmd.py
```diff
@@ -0,0 +1,222 @@
...
+        s.show('osmd')
+
+class Test(unittest.TestCase):
+
+    def testAddsDefaultPartId(self):
+        from music21 import converter
+        s = converter.parse("tinyNotation: 3/4 E4 r f#")
+
+        ConverterOpenSheetMusicDisplay.addDefaultPartName(s)
+        firstInstrumentObject = s.getInstruments(returnDefault=True, recurse=True)[0]
```
jacobtylerwalls, line 214: same, in v7, just call `.first()`

music21/osmd/osmd.py
```diff
@@ -0,0 +1,222 @@
...
+        self.assertNotEqual(firstInstrumentObject.instrumentName, None)
```
jacobtylerwalls, line 215: nitpick, but for None I think .assertIsNotNone would be preferable

music21/osmd/osmd.py
```diff
@@ -0,0 +1,222 @@
...
+        self.assertNotEqual(firstInstrumentObject.instrumentName, '')
+
+
+if __name__ == "__main__":
+    import music21
+
+    music21.mainTest(TestExternal)
```
jacobtylerwalls, line 222: Myke will chime in if I get this wrong, but usually the command line call here runs the `Test` class instead of `TestExternal`

music21/osmd/osmd.py
```diff
@@ -0,0 +1,222 @@
...
+
+        >>> import music21
+        >>> # display two scores to demonstrate difference.
+        >>> s = music21.converter.parse("tinyNotation: 3/4 E4 r f# g=lastG trip{b-8 a g} c4~ c")
+        >>> _ = s.show('osmd',fixPartName=False)
+        >>> music21.osmd.ConverterOpenSheetMusicDisplay.addDefaultPartName(s)
+        >>> _ = s.show('osmd',fixPartName=False)
+        '''
+        # If no partName is present in the first instrument, OSMD will display the ugly 'partId'
+        allInstruments = list(score.getInstruments(returnDefault=False, recurse=True))
```
jacobtylerwalls, line 174: Casting to list is unnecessary -- getInstruments returns a stream of elements, and you can use .first()

music21/osmd/osmd.py
```diff
@@ -0,0 +1,222 @@
...
+        score = obj
+        if fixPartName:
+            self.addDefaultPartName(score)
+
+        if divId is None:
+            # create unique reference to output div in case we wish to update it
+            divId = self.getUniqueDivId()
+
+
+        xml = open(score.write('musicxml')).read()
```
jacobtylerwalls, line 102: is this file ever closed? perhaps better to use a `with open(...) as f:` context manager?

jacobtylerwalls: Actually, if you just want to get some xml into memory, you can just use the `GeneralObjectExporter` in the musicxml export class; this is a bit safer than depending on the user having permissions to write to a temp dir

```
gex = music21.musicxml.m21ToXml.GeneralObjectExporter(score)
bytesOut = gex.parse()
xml = bytesOut.decode('utf-8')
```

music21/osmd/osmd.py
```diff
@@ -0,0 +1,222 @@
...
+        script = self.musicXMLToScript(xml, divId, offline=offline)
+        if in_ipython:
+            self.display(self.HTML(f'<div id="{divId}">loading OpenSheetMusicDisplay</div>'))
+            self.display(self.Javascript(script))
+        else:
+
+            tmp = tempfile.NamedTemporaryFile(delete=False)
```
jacobtylerwalls, line 109: For consistency you can use `environLocal.getTempFile()` so that all the music21 temp files are in the same place. 
Most files have something like this at the top:
```
_MOD = 'osmd'
environLocal = environment.Environment(_MOD)
```

jacobtylerwalls (commented):
Hi, and thanks for this! Left some comments to help increase coverage and consistency with the rest of the library.

music21/opensheetmusicdisplay/osmd.py
mscuthbert: Not a problem with IE not being supported any more.

music21/opensheetmusicdisplay/osmd.py
mscuthbert: Arrow functions are now fine.

music21/osmd/osmd.py
mscuthbert: That's usually true, but here I'm okay with making it a list of [Test, TestExternal] because the main thing someone would want to check is "did the notation get generated?"  I think we do that in `graph` (or we should) and maybe flat.io

music21/osmd/osmd.py
mscuthbert: yes, please!

Though I'm in the `_MOD` killing mood, so just go ahead and do:

`environLocal = environment.Environment('osmd.osmd')`

(the _MOD should be the path after `music21.`)

mscuthbert:
Thanks for the work on this -- I agree with Jacob's comments, so let's get them in.  There are still comments above (especially not having the script in the distribution) that need addressing, but this is very exciting!

(when downloaded, the script should probably go into the environment Scratch directory, since we shouldn't assume that the music21 directory is writeable by the user).

mscuthbert:
We can work on getting the coverage up, but it's less of a concern since scripts that do external work often lower the overall coverage of m21.  It'd be good to keep it improving, but in this case, I can let it go down a little.

music21/osmd/osmd.py
supersational: used `#pragma: no cover` since seems neater, thanks.

supersational:
Fab! I think this is getting there, thanks for the detailed feedback. Some final thoughts/questions:

- `.show('osmd')` is a bit unwieldy, I imagine `.show('web')` is a more user-friendly way to call this. Would be nice to link back to the OpenSheetMusicDisplay project in another way to give them credit, maybe a link in the footer of the score?
- Since OSMD is under active development it could be nice to set the version of the loaded script in an ENV variable, so users could easily choose which version to use - would also make it easy for users to get new features and ensure new versions are compatible. Implementing this should be fairly straightforward..
- Code-wise I'm not sure about returning the id of the div element, this does allow re-using a div and choosing where the score is rendered, however I can't think of any practical applications and was mainly for debugging, this would remove the  `divId=None` keyword argument.
- Same for `fixPartName`? This fixes an issue with empty part names, would only need to be disabled if it was causing issues, which should be a code-fix not a keyword argument.
- Finally is there any place for an example notebook? Would be nice to demo the capabilities:
image

jacobtylerwalls:
Had a thought about the part name issue. Agree that it's cumbersome to build this workaround into the API. (Especially since if in a couple weeks OSMD starts supporting empty strings.)

@mscuthbert what about a new, non-empty default for streams lacking instruments in v7?

https://github.com/cuthbertLab/music21/blob/686677283e95cce4e9546baf8b2f7b6ffcc39b67/music21/defaults.py#L57

What about, drumroll...
```
    partName = 'music21 Part'
```
Then we can remove the workaround from this feature.

MusicXML export already reads from this default.  (We can also hide it by adding a condition to the last condition in this block if you're concerned it's too loud to print "music21 Part", but I like the idea of keeping it.)

https://github.com/cuthbertLab/music21/blob/f2c3dc16b7946309b33d72e2e2ddcecb7f73b332/music21/musicxml/m21ToXml.py#L2531-L2537

MIDI export doesn't observe it, but that's easy to fix:

https://github.com/cuthbertLab/music21/blob/686677283e95cce4e9546baf8b2f7b6ffcc39b67/music21/midi/translate.py#L185-L186

FWIW I didn't notice that we were inserting `Piano` objects into people's streams without making a copy first.

music21/osmd/osmd.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def getExtendedModules():
+    # pylint: disable=import-outside-toplevel
+    if hasInstalledIPython:
+        from IPython.core.display import display, HTML, Javascript
+    else:
+        def display():
+            raise OpenSheetMusicDisplayException(
+                'OpenSheetMusicDisplay requires IPython to be installed')
```
jacobtylerwalls, line 56: My fears of this being a little too fancy bore out -- doesn't do anything:

My diff to force this to run like IPython on my system and force problems:
```diff
--- a/music21/osmd/osmd.py
+++ b/music21/osmd/osmd.py
@@ -46,9 +46,9 @@ except ImportError:
 hasInstalledIPython = loader is not None
 
 
-def getExtendedModules():
+def getExtendedModules(forceProblem=True):
     # pylint: disable=import-outside-toplevel
-    if hasInstalledIPython:
+    if hasInstalledIPython and not forceProblem:
         from IPython.core.display import display, HTML, Javascript
     else:
         def display():
@@ -108,7 +108,7 @@ class ConverterOpenSheetMusicDisplay(SubConverter):
 
         # generate script to be run on page
         script = self.musicXMLToScript(xml, divId, offline=offline)
-        if in_ipython:
+        if in_ipython or True:
             self.display(self.HTML(f'<div id="{divId}"></div>'))
             self.display(self.Javascript(script))
         else:
```

Results in a lot of `TypeError`s:
```
FAIL: show (__main__.ConverterOpenSheetMusicDisplay)
Doctest: __main__.ConverterOpenSheetMusicDisplay.show
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/doctest.py", line 2204, in runTest
    raise self.failureException(self.format_failure(new.getvalue()))
AssertionError: Failed doctest test for __main__.ConverterOpenSheetMusicDisplay.show
  File "/Users/jwalls/music21/music21/osmd/osmd.py", line 82, in show

File "/Users/jwalls/music21/music21/osmd/osmd.py", line 88, in __main__.ConverterOpenSheetMusicDisplay.show
Failed example:
    fig_id1 = s.show('osmd')
Exception raised:
      File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/doctest.py", line 1336, in __run
        exec(compile(example.source, filename, "single",
      File "<doctest __main__.ConverterOpenSheetMusicDisplay.show[2]>", line 1, in <module>
        fig_id1 = s.show('osmd')
      File "/Users/jwalls/music21/music21/stream/__init__.py", line 259, in show
      File "/Users/jwalls/music21/music21/base.py", line 2644, in show
      File "/Users/jwalls/music21/music21/osmd/osmd.py", line 112, in show
        self.display(self.HTML(f'<div id="{divId}"></div>'))
    TypeError: display() takes 0 positional arguments but 1 was given


```

If the idea is to raise as last as possible, when actually calling show, then the `__init__` could just catch ImportError and set those variables to None, and then in display, check if None and raise then.

jacobtylerwalls:
Sorry for the piecemeal review -- I didn't think I would have much to contribute to this discussion, but I'm enjoying playing with it -- very cool!

doctests open browser windows now. CI doesn't complain, I guess, but someone running `python3 -m music21.test.testSingleCoreAll` is gonna get four browser windows opened. Are we okay with that?

music21/osmd/osmd.py
```diff
@@ -0,0 +1,233 @@
...
+        This is so we can update a previously used div.
+        '''
+        return ("OSMD-div-"
+                + str(random.randint(0, 1000000))
+                + "-" + str(time.time()).replace('.', '-'))  # '.' is the class selector
+
+    def musicXMLToScript(self, xml, divId, offline=False):
+        '''
+        Converts the xml into Javascript which can be injected into a webpage to display the score.
+        If divId is provided then it will be used as the container, if not a new
```
jacobtylerwalls, line 145: trailed off here... :-)

music21/osmd/osmd.py
```diff
@@ -0,0 +1,233 @@
...
+
+        # generate script to be run on page
+        script = self.musicXMLToScript(xml, divId, offline=offline)
+        if in_ipython:
+            self.display(self.HTML(f'<div id="{divId}"></div>'))
+            self.display(self.Javascript(script))
+        else:
+
+            tmp = tempfile.NamedTemporaryFile(delete=False)
+            tmp_path = tmp.name + '.html'
```
jacobtylerwalls, line 117: Still think it would be a good idea to have the temp files all in one place / observe the user's scratch directory.

```suggestion
            tmp = self.getTemporaryFile()
```

Problem is, we can't do that until we register the subconverter, I'll reply to the earlier comment about this.

```
>>> osmdc.getTemporaryFile(subformats=['html'])
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/music21/converter/subConverters.py", line 242, in getTemporaryFile
    ext = self.getExtensionForSubformats(subformats)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/music21/converter/subConverters.py", line 219, in getExtensionForSubformats
    raise SubConverterException(
music21.converter.subConverters.SubConverterException: This subconverter cannot show or write: no output extensions are registered for it
```

music21/osmd/osmd.py
```diff
@@ -0,0 +1,233 @@
...
+
+class ConverterOpenSheetMusicDisplay(SubConverter):
+    '''
+    Converter for displaying sheet music inline
+    in an IPython notebook.
+    Uses: https://opensheetmusicdisplay.org/ for rendering
+    '''
+    registerFormats = ('osmd',)
+    registerShowFormats = ('osmd',)
+    # when updating the script tag osmd will only reload in a notebook if you: Kernel -> restart and
```
jacobtylerwalls, line 70:
```suggestion
    registerOutputFormats = ('html',)
```
Then you can call `self.getTemporaryFile()`. It was staring at me when I compiled the osmd module documentation.

jacobtylerwalls: Obviously don't get rid of the comment, I guess you can't suggest insertions?

mscuthbert:
> doctests open browser windows now. CI doesn't complain, I guess, but someone running `python3 -m music21.test.testSingleCoreAll` is gonna get four browser windows opened. Are we okay with that?

No -- only TestExternal is allowed to open windows, play music, etc.  I don't want to be testing some other part of the system and have osmd tests opening.

There will be many 'web' formats in the future, so no, the format needs to be 'osmd'

supersational:
> No -- only TestExternal is allowed to open windows, play music, etc. I don't want to be testing some other part of the system and have osmd tests opening.

So best to add ` # doctest: +SKIP` to the doctests that are opening windows then?

music21/osmd/osmd.py
supersational: Ah I see how it works now, had to change the line to `registerOutputExtensions = ('html',)` and then it worked :)

music21/osmd/osmd.py
supersational: Ah so you're forcing it to fail? I agree it's well over fanciness-budget, how about just moving it all into the actual call site? Avoids all the `self.display = display` stuff and we can delete the new exception class.

```python 
        # generate script to be run on page
        script = self.musicXMLToScript(xml, divId, offline=offline)
        if in_ipython and hasInstalledIPython:
            # pylint: disable=import-outside-toplevel
            from IPython.core.display import display, HTML, Javascript
            display(HTML(f'<div id="{divId}"></div>'))
            display(Javascript(script))
        else:
```

music21/osmd/osmd.py
supersational: Yes this is working now, seems like default on windows is `%HOME%/AppData/Local/Temp/music21`

jacobtylerwalls: Yeah, there's some tooling for this in music21, you can set the music21 environment variable `directoryScratch` to put it somewhere else. For users who have done so, I'm trying to avoid "where did it go?" posts.

music21/osmd/osmd.py
jacobtylerwalls: Just trying to exercise the code, not being gleefully evil! Yes, this looks easier to maintain. I don't think we need to be too worried about performance, but if you want to avoid the import-on-every-call overhead, there's a gist here about optimizing lazy imports inside functions using a global variable:

https://wiki.python.org/moin/PythonSpeed/PerformanceTips

music21/osmd/osmd.py
```diff
@@ -0,0 +1,216 @@
...
+        '''
+        Converts the xml into Javascript which can be injected into a webpage to display the score.
+        If divId is set then it will be used as the container, if not a new div will be created
+        '''
+
+        # script that will replace div contents with OSMD display
+        script_path = os.path.join(getSourceFilePath(), 'osmd', 'notebookOSMDLoader.js')
+        script = open(script_path, 'r').read() \
+            .replace('{{DIV_ID}}', divId) \
+            .replace('"{{data}}"', json.dumps(xml))
```
jacobtylerwalls, line 135: m21 doesn't use line continuation characters: use open parens

jacobtylerwalls:
> Can you put a note in subconverter.py that this is where the OSMD converter is?


If at least the class itself and show() is moved back into subconverters, then it automatically will appear in this list, and we don't need to leave notes and such. Then there can be static fucntions in osmd.py to handle everything else. Thoughts?

https://github.com/cuthbertLab/music21/blob/3fa1eb9dd86e91594eb7588a745165c104745293/music21/converter/__init__.py#L782-L794

jacobtylerwalls:
> * Code-wise I'm not sure about returning the id of the div element, this does allow re-using a div and choosing where the score is rendered, however I can't think of any practical applications and was mainly for debugging, this would remove the  `divId=None` keyword argument.

Agreed, let's strip it. In terms of a return value, I think if we're not running in IPython, `show()` should return the tempfilepath (the other converters do this). In IPython returning None seems okay.

> So best to add # doctest: +SKIP to the doctests that are opening windows then?

By my count, removing the two unnecessary kwargs remove 3 of the 4 show() doctests. So for the remaining one, just annotate the skipped line with `#_DOCS_SHOW` , which is the m21 way of skipping but still showing the user what to do:

```
        >>> fig_id1 = s.show('osmd')  #_DOCS_SHOW
```

music21/osmd/osmd.py
mscuthbert: There's no worries about an import on every call statement on something that is only called on the order of 1-1000 times per user session.  The bottleneck only happens if you're calling the import thousands of times per minute.  

For instance, when music21 was first released, some Python systems did not have weakref module.  For a time, `import weakref` was called within `def activeSite()` and that was a major slowdown because just about everything calls .activeSite.

mscuthbert: look in the ipython21 directory for various calls to `runningUnderIPython()` -- that's the "one-place" interface to check if IPython is running.

mscuthbert:
> > No -- only TestExternal is allowed to open windows, play music, etc. I don't want to be testing some other part of the system and have osmd tests opening.
> 
> So best to add ` # doctest: +SKIP` to the doctests that are opening windows then?

Doctest +SKIP will appear in the docs and confuse people.  Look for `>>> # _DOCS_SHOW`  and `>>> [something]  # _DOCS_HIDE` in the code base or the notebook in the dev docs on documenting to see how to demonstrate what should happen without actually running anything.

jacobtylerwalls:
> @mscuthbert what about a new, non-empty default for streams lacking instruments in v7?

drafted in #935

jacobtylerwalls:
Hey @supersational , Michael had a look at my suggestion and thinks we ought to adhere to your original idea of just making a last-minute adjustment in the OSMD converter to handle empty part names.

I still think inserting a `Piano` is too presumptive and difficult to debug if we're operating on the input stream (we've had posts on "how do I change the piano sound") Should be enought to just iterate `for part in s.parts`, and if `not part.partName`, set `partName = ' '` as you've done.

Check first if there are no parts, if not--put that stream's contents in a Part, and put that in a score, set the dummy partName, and export THAT. This is what the musicxml exporter will do anyway, but you just have to do it here so you can inject a part name. The alternative would be to use ElementTree to correct the XML output, but that's kind of a drag.

Sound right, @mscuthbert ?

Notice setting partName is functional without having to create and insert Instrument objects (no instrument object shown in the text dump at the end):


```
>>> p = converter.parse('tinyNotation: c1 d1 e1 f1')
>>> p.partName
>>> p.partName = 'living large'
>>> p.show()
>>> p.show("t")
    {0.0} <music21.meter.TimeSignature 4/4>
    {0.0} <music21.note.Note C>
{4.0} <music21.stream.Measure 2 offset=4.0>
    {0.0} <music21.note.Note D>
{8.0} <music21.stream.Measure 3 offset=8.0>
    {0.0} <music21.note.Note E>
{12.0} <music21.stream.Measure 4 offset=12.0>
    {0.0} <music21.note.Note F>
```

Screen Shot 2021-03-28 at 5 17 37 PM

supersational:
Hey nice, yep that sounds like a better way to do it, thanks!

I've handled the two obvious types for this since `.show()` can be called on types: `(music21.stream.Part, music21.stream.Score)`. Not sure if there are other types I should consider?

jacobtylerwalls:
I don't think you need to check types -- you can call show on any stream.Stream subclass. `.parts` is always available, it's just a filter that can be empty if no parts exist at the next level down. So you'd need to check if the input IS a part, or if it `.parts` is truthy, and if not, take all the contents and INSERT them into a new part. In that last case you could let the musicxml exporter handle wrapping that new part you created in a score.

music21/osmd/__init__.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# -------------------------------------------------------------------------------
+# Name:         osmd/__init__.py
+# Purpose:      music21 classes for displaying music21 scores in IPython notebooks
+#
+# Authors:      Sven Hollowell
+#
+# Copyright:    Copyright © 2012-14 Michael Scott Cuthbert and the music21 Project
```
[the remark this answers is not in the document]
supersational: Haha, I've updated them, yep sure has been a while.. One crazy year for sure :)

supersational:
I think .parts isn't always available e.g. 
```python
s = converter.parse("tinyNotation: 3/4 E4 r f#")
s.parts
``` 
Gives an attributeError, but yes using `getattr` would do the job, just checking I've understood correctly?

jacobtylerwalls:
Ah, I forgot the `.iter` -- you can do `inputStream.iter.parts`. Or just handle scores and parts explicitly and then have one else case to put all the contents in a new Part.

Be careful with subclasses -- users should be able to subclass anything, and stream.PartStaff is a subclass of Part. You can call `if 'Part' in s.classes` instead.

music21/osmd/__init__.py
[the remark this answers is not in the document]
supersational: Feel like you deserve to be credited here as well for the amount of guidance you've been giving!

supersational:
Hmm just realised the following is possible as well: `note.Note('c4').show('osmd')` which also needs to work!(?) Was thinking of using `.getElementsByClass(stream.Part)` but that doesn't exist on a `Note`. Maybe the safest way is to just wrap the whole thing in a part if it doesn't have `.parts`. I don't think that could do any harm but will need to investigate.. will have to do that when I next get a chance though.. if you find something that works I've given you access to the repo, so do please feel free to make whatever changes you think suitable.

jacobtylerwalls:
Yeah, I think way at the top somewhere Michael mentioned that's why it would someday be worth factoring out the logic in the musicxml exporter that wraps objects in suitable score > part > measure > voice hierarchies. Something like fixup hierarchy, without fixing up the rest of the notation.

Some point this week I can push to your branch; something like this should be a real small change. And do absolutely nothing to people's streams!

jacobtylerwalls:
Solution for part name should be spruced up now.

Also in the last commit I had a go at moving the converter into subConverters.py, based on some of the comments at the top of the thread. Now we don't leave to leave notes about where the converter is, or register it, or have it get "cleared" when external subconverters are cleared, etc. 

Since what was left was just a few static functions, I just got rid of the osmd directory so it's just osmd.py (and put the js script in a new folder "js")

Let me know how it looks -- I only tested command line launching, not displaying inside IPython.

jacobtylerwalls:
> I think if we're not running in IPython, `show()` should return the tempfilepath (the other converters do this).

Nope, NVM -- that's write() that does that, not show(). All good.

jacobtylerwalls:
The Coveralls stats are kind of meaningless because we're insisting that the `Test` classes not open browser windows, only `TestExternal`, which we don't run on CI--so everything except the helper methods is a missed line. The code will only be exercised by running `python3 -m music21.osmd` to run TestExternal.

Honestly I think ciMain() should also run TestExternal, if it doesn't double the time of the suite, in some way that wouldn't affect multiprocessTest and the other tools developers use where we don't to window-bomb them. But can be discussed separately.

supersational:
> Let me know how it looks -- I only tested command line launching, not displaying inside IPython.

Working perfectly in my notebook, fantastic!

jacobtylerwalls (commented):
This seems super close to me. Need to document and finalize the design of offline mode, though.

music21/osmd.py
```diff
@@ -0,0 +1,135 @@
...
+            # on first use we download the file and store it locally for subsequent use
+            # TODO: catch error for no internet connection
+            urllib.request.urlretrieve(SCRIPT_URL, osmd_file)
+
+        # since we can't link to files from a notebook (security risk) inject into file.
+        script_content = open(osmd_file).read()
+        script = script.replace('"{{offline_script}}"', json.dumps(script_content), 1)
+
+    else:
+        script = script.replace('"{{script_url}}"', json.dumps(SCRIPT_URL), 1)
```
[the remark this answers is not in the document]
jacobtylerwalls: FYI @mscuthbert I think this PR is mergeable but for my question here, was there anything else you were waiting on?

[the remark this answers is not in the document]
jacobtylerwalls: @supersational my bad, I totally misunderstood offline mode. You're just trying to let folks decide whether their pages should load from CDN or not. So I don't think we need to cache the js in the else branch. Sorry for going off in wrong direction. I pushed some doc edits and reraised URLError with an explanation.

music21/opensheetmusicdisplay/osmd.py
```diff
@@ -0,0 +1,189 @@
...
+    def getUniqueDivId():
+        return "OSMD-div-"+ \
+                str(random.randint(0,1000000))+ \
+                "-"+str(time.time()).replace('.','-') # '.' is the class selector
+
+    @staticmethod
+    def musicXMLToScript(xml, div_id, offline=True):
+        
+        # print('xml length:', len(xml))
+        script_dir = os.path.join(os.path.dirname(__file__), 'opensheetmusicdisplay.0.3.1.min.js')
```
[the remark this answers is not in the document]
jacobtylerwalls: I fear getting latest would requires scraping github endpoints and be overkill.

music21/opensheetmusicdisplay/osmd.py
jacobtylerwalls: I commented them out in the .js file.

jacobtylerwalls (approved):
Thanks for the work (and patience!) here @supersational. With CI passing after the latest linting and cleanups, I think we should pull this in. A LOT of improvements went in to make it easy to use and easy to maintain!

I don't want to rush Myke, so I will hold off on merging for a few weeks in case there are concerns, but I would anticipate this to be merged as the first major feature in v.7.3 🎉

jacobtylerwalls:
Re: an earlier review from Myke, I struggled to understand "not having the script in the _distribution_" (since I took that to refer to an outdated version of this PR that vendored the min.js file), but it's possible that meant something like "not having a cached script injected into the file in offline mode", essentially https://github.com/cuthbertLab/music21/pull/326#discussion_r214218637 or https://github.com/cuthbertLab/music21/pull/326#discussion_r214218344 -- if so, then those would be good reasons to wait to hear more from Myke first. So I do want his sign-off.

***

## SQUASH this PR

to avoid getting the .min.js from 9a3ebd0 in the VCS

mscuthbert:
I will get on reviewing this this month.  Thanks!

mscuthbert:

mscuthbert:
Closing for lack of activity.  In the past 6 years, the project philosophy has evolved to where modules like this are encouraged to be separate pip projects that interface with music21.  PRs that make it easier to connect to the OSMD project are welcome but not a whole OSMD converter.

supersational:
Hi Mike,

I've extracted the core idea here into a separate package: https://github.com/supersational/showscore

It is also on pip which should make it quick to start using alongside music21: https://pypi.org/project/showscore/

I think this can make it substantially easier to work with scores - it works with no installation of other programs or setting environment variables, both which can be time consuming for people just wanting to visualise scores! 

Works both inline in Jupyter notebooks, and also in non-jupyter environments by opening the score in a new tab. 

Any comments on the API appreciated - I was considering adding a method to make it the default for `.show()` calls from music21 by monkey-patching the methods, but maybe that can be added in a future version.

mscuthbert:
@supersational -- just posted on music21 my official endorsement of this direction and all the cool things it's doing.  Hope I have time someday to contribute back.

Issue #350, opened 2018-12-05
music21 doesn't respect harmony offset when importing musicxml


a-papadopoulos:
I think we've found a bug in the way music21 imports harmony objects from musicxml. In the musicxml file [pasted here](https://gist.github.com/a-papadopoulos/f4d4bdee859706a8b690c158a9da0ed0), the first measure contains only a whole note, but two chord symbols C and G at beat 1 and 3 of the measure respectively. However, when imported in music21, the ChordSymbol object corresponding to the G chord has an offset of 0 too.

I can dive into this a bit further, but my intuition is that the `offset` element of the G chord isn't taken into account by music21, and that music21 expects a chord symbol to occur at the start of a new note. Can you confirm this?

mscuthbert:
Fixed in #361

PR #361, opened 2019-01-17
Chord offset

a-papadopoulos:
This solves #350. I didn't have a lot of data to test this, but there's a few scores in wikifonia with `<harmony>` objects that have an `<offset>` component. Obviously I can't include them here, but things look fine.
(edit: escaped xml tags)

### DIFF
--- a/music21/musicxml/xmlToM21.py
+++ b/music21/musicxml/xmlToM21.py
@@ -4333,10 +4333,10 @@ class MeasureParser(XMLParserBase):
         '''
         h = self.xmlToChordSymbol(mxHarmony)
         chordOffset = self.xmlToOffset(mxHarmony)
-        if chordOffset > 0:
-            self.insertCoreAndRef(chordOffset, mxHarmony, h)
-        else:
-            self.insertCoreAndRef(self.offsetMeasureNote, mxHarmony, h)
+        if chordOffset != 0:
+            print(chordOffset)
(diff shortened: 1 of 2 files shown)


music21/musicxml/xmlToM21.py
```diff
@@ -6475,7 +6477,19 @@ def testNoChordImport(self):
...
                                          1].chordKindStr))
 
+    def testChordOffset(self):
+        import pathlib
+        from music21 import converter
+        from music21 import musicxml
+
+        thisDir = common.getSourceFilePath() / 'musicxml'
+        testFp = thisDir / 'testChordOffset.xml'
+        s = converter.parse(testFp, forceSource=True)
```
mscuthbert: If you can remove the "forceSource=True" I can merge this.  Thank you!

mscuthbert:
The only tests where it should appear are tests of the caching system.  forceSource makes the testing system run about 3x slower -- it's good to do while developing, but once a test is stable, it just slows down running tests for everyone else.  I know I've discussed this elsewhere and it's not something I'm reconsidering at this point.  If I find it elsewhere in the code where it's not testing the cache system, I'll be sure to remove it.

mscuthbert:
Thanks! Looks great

mscuthbert:
And thanks especially for the great MusicXML test file

PR #364, opened 2019-01-22
Checks whether two chords are related by a single L, P, or R transform

MarkGotham:
… complementing the existing options to generate that change.

### DIFF
--- a/music21/analysis/neoRiemannian.py
+++ b/music21/analysis/neoRiemannian.py
@@ -7,7 +7,7 @@
 #               Michael Scott Cuthbert
 #               Mark Gotham
 #
-# Copyright:    Copyright © 2017-18 Michael Scott Cuthbert and the music21 Project
+# Copyright:    Copyright © 2017-19 Michael Scott Cuthbert and the music21 Project
 # License:      LGPL or BSD, see license.txt
 # ------------------------------------------------------------------------------
 '''
@@ -94,8 +94,10 @@ def L(c, raiseException=True):
             raise LRPException('Cannot perform L on this chord: not a major or minor triad')
         return c
 
-    return _LRP_transform(c, transposeInterval, changingPitch)
+    outChord = _singleNoteTransform(c, transposeInterval, changingPitch)
+    outChord.quarterLength = c.quarterLength
 
+    return outChord
 
 def P(c, raiseException=True):
     '''
@@ -129,8 +131,11 @@ def P(c, raiseException=True):
         if raiseException is True:
             raise LRPException('Cannot perform P on this chord: not a Major or Minor triad')
         return c
-    return _LRP_transform(c, transposeInterval, changingPitch)
 
+    outChord = _singleNoteTransform(c, transposeInterval, changingPitch)
+    outChord.quarterLength = c.quarterLength
+
+    return outChord
 
 def R(c, raiseException=True):
     '''
@@ -165,11 +170,14 @@ def R(c, raiseException=True):
             raise LRPException('Cannot perform R on this chord: not a Major or Minor triad')
         return c
 
-    return _LRP_transform(c, transposeInterval, changingPitch)
+    outChord = _singleNoteTransform(c, transposeInterval, changingPitch)
+    outChord.quarterLength = c.quarterLength
 
-def _LRP_transform(c, transposeInterval, changingPitch):
+    return outChord
+
+def _singleNoteTransform(c, transposeInterval, changingPitch):
     '''
-    Performs a neoRiemannian transformation on c that involves transposing `changingPitch`
+    Performs a neoRiemannian transformation on c that involves transposing `changingPitch` by
     `transposeInterval`.
     '''
     changingPitchCopy = copy.deepcopy(changingPitch)
@@ -181,7 +189,7 @@ def _LRP_transform(c, transposeInterval, changingPitch):
 
 def isNeoR(c1, c2):
     '''
-    Tests if two chords are related by a single L, P, or R transformation,
+    Tests if two chords are related by a single L, P, R, or chromatic mediant transformation,
     and returns that transform if so (otherwise, False).
 
     >>> c1 = chord.Chord('C4 E4 G4')
@@ -204,14 +212,39 @@ def isNeoR(c1, c2):
     pc1 = P(c1)
     rc1 = R(c1)
 
+    UFM = chromaticMediants(c1, transformation='UFM')
+    USM = chromaticMediants(c1, transformation='USM')
+    LFM = chromaticMediants(c1, transformation='LFM')
+    LSM = chromaticMediants(c1, transformation='LSM')
+
+    Slide = S(c1)
+    Neben = N(c1)
+
     if lc1.normalOrder == c2.normalOrder:
         return 'L'
     elif pc1.normalOrder == c2.normalOrder:
         return 'P'
     elif rc1.normalOrder == c2.normalOrder:
         return 'R'
+
+    elif UFM.normalOrder == c2.normalOrder:
+        return 'UFM'
+    elif USM.normalOrder == c2.normalOrder:
+        return 'USM'
+    elif LFM.normalOrder == c2.normalOrder:
+        return 'LFM'
+    elif LSM.normalOrder == c2.normalOrder:
+        return 'LSM'
+
+    elif Slide.normalOrder == c2.normalOrder:
+        return 'Slide'
+    elif Neben.normalOrder == c2.normalOrder:
+        return 'Neberverwandt'
+
     else:
         return False
+
+# ------------------------------------------------------------------------------
 
 def LRP_combinations(c,
                      transformationString,
@@ -293,7 +326,6 @@ def LRP_combinations(c,
         return _simplerEnharmonics(c)
         ## Previously:
         # newPitches = pitch.simplifyMultipleEnharmonics(c.pitches, keyContext=key.Key('C'))
-
 
 def completeHexatonic(c, simplifyEnharmonics=False, raiseException=True):
     '''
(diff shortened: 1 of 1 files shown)


music21/analysis/neoRiemannian.py
```diff
@@ -181,7 +189,7 @@ def _LRP_transform(c, transposeInterval, changingPitch):
 
 def isNeoR(c1, c2):
     '''
-    Tests if two chords are related by a single L, P, or R transformation,
+    Tests if two chords are related by a single L, P, R, or chromatic mediant transformation,
```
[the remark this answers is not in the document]
mscuthbert: In fact, I think that all the editions should be in a different function "isExtendedNeoR" that tests for these things.
(should also have a doctest showing what happens if two chords are not neoR related; or if one is not a triad.

music21/analysis/neoRiemannian.py
```diff
@@ -391,9 +423,158 @@ def hexatonicSystem(c):
...
+            newChord = c.transpose(-4)
+        elif transformation == 'LSM':
+            newChord = c.transpose(-3)
+
+    outChord = _simplerEnharmonics(newChord)
+    outChord.quarterLength = c.quarterLength
+
+    return outChord
+
+def disjunctMediants(c, UpperOrLower='Upper', raiseException=True):
```
mscuthbert: why is UpperOrLower capitalized?

mscuthbert: Also the value should be lowercase.

music21/analysis/neoRiemannian.py
```diff
@@ -391,9 +423,158 @@ def hexatonicSystem(c):
...
+    >>> Cmaj = chord.Chord('C5 E5 G5')
+    >>> UpChrom = analysis.neoRiemannian.disjunctMediants(Cmaj, UpperOrLower='Upper')
+    >>> [x.name for x in UpChrom.pitches]
+    ['E-', 'G-', 'B-']
+    >>> DownChrom = analysis.neoRiemannian.disjunctMediants(Cmaj, UpperOrLower='Lower')
+    >>> [x.name for x in DownChrom.pitches]
+    ['A-', 'C-', 'E-']
+    '''
+
+    if raiseException is True:
```
mscuthbert: raiseException is unnecessary -- if the keyword is wrong it should raise an exception no matter what.  raiseException type work is only used when the data might be corrupt or not.

music21/analysis/neoRiemannian.py
```diff
@@ -391,9 +423,158 @@ def hexatonicSystem(c):
...
+        if transformation == 'UFM':
+            newChord = c.transpose(3)
+        elif transformation == 'USM':
+            newChord = c.transpose(4)
+        elif transformation == 'LFM':
+            newChord = c.transpose(-4)
+        elif transformation == 'LSM':
+            newChord = c.transpose(-3)
+
+    outChord = _simplerEnharmonics(newChord)
```
mscuthbert: what is simplifyEnharmonics is False?

music21/analysis/neoRiemannian.py
```diff
@@ -391,9 +423,158 @@ def hexatonicSystem(c):
...
+    outChord.quarterLength = c.quarterLength
```
mscuthbert: this line seems unnecessary.  transpose definitely does and _simplerEnharmonics SHOULD (untried) preserve all other attributes.

music21/analysis/neoRiemannian.py
```diff
@@ -391,9 +423,158 @@ def hexatonicSystem(c):
...
+def disjunctMediants(c, UpperOrLower='Upper', raiseException=True):
+    '''
+    Transforms a chord into the upper or lower disjunct mediant.
+    These transformations involve:
+    root motion by a non-diatonic third;
+    mode-change;
+    no common-tones.
+
+    >>> Cmaj = chord.Chord('C5 E5 G5')
+    >>> UpChrom = analysis.neoRiemannian.disjunctMediants(Cmaj, UpperOrLower='Upper')
```
mscuthbert: why are these variable names capitalized?  They look like classes.

music21/analysis/neoRiemannian.py
```diff
@@ -391,9 +423,158 @@ def hexatonicSystem(c):
...
+        return outChord
+
+    elif c.isMajorTriad():
+
+        if UpperOrLower=='Upper':
+            transposeChordIntv = 3
+        elif UpperOrLower=='Lower':
+            transposeChordIntv = -4
+
+        newChord = _simplerEnharmonics(c.transpose(transposeChordIntv))
```
mscuthbert: this is a lot of redundant code from 505-512; simplify.

music21/analysis/neoRiemannian.py
```diff
@@ -391,9 +423,158 @@ def hexatonicSystem(c):
...
+    '''
+    Slide transform connecting major and minor triads with the third as the single common tone
+    (i.e. with the major triad root a semi-tone below the minor). So:
+    root motion by a semi-tone;
+    mode-change;
+    one common-tones.
+    Slide is equivalent to 'LPR'.
+
+    >>> Cmaj = chord.Chord('C5 E5 G5')
+    >>> SlideUp = analysis.neoRiemannian.S(Cmaj)
```
mscuthbert: more capitalization issues

music21/analysis/neoRiemannian.py
```diff
@@ -204,15 +212,40 @@ def isNeoR(c1, c2):
...
+        return 'UFM'
+    elif USM.normalOrder == c2.normalOrder:
+        return 'USM'
+    elif LFM.normalOrder == c2.normalOrder:
+        return 'LFM'
+    elif LSM.normalOrder == c2.normalOrder:
+        return 'LSM'
+
+    elif Slide.normalOrder == c2.normalOrder:
+        return 'Slide'
```
mscuthbert: why "Slide" and not "S"?  same for N

music21/analysis/neoRiemannian.py
```diff
@@ -204,15 +212,40 @@ def isNeoR(c1, c2):
     pc1 = P(c1)
     rc1 = R(c1)
 
+    UFM = chromaticMediants(c1, transformation='UFM')
+    USM = chromaticMediants(c1, transformation='USM')
+    LFM = chromaticMediants(c1, transformation='LFM')
+    LSM = chromaticMediants(c1, transformation='LSM')
+
+    Slide = S(c1)
+    Neben = N(c1)
```
mscuthbert: we are now making nine different chords, but in the case of L, only one will be used.  Can we make it a loop so that there is only one `if x.normalOrder == c2.normalOrder` test?  Pylint is going to complain that the number of if statements in the loop exceeds the maximum.  Note also that computing `c2.normalOrder` adds run time but it will always be the same for each comparison.  Cache it once.

music21/analysis/neoRiemannian.py
mscuthbert: newChord is undefined if not a major or minor chord and program will crash.

music21/analysis/neoRiemannian.py
```diff
@@ -438,32 +619,76 @@ def testNeoRiemannianCombinations(self):
...
+        c8 = chord.Chord('C-4 E-4 A-4')
+        ans6 = isNeoR(c1, c8)
+        self.assertEqual(ans6, False) # disjunct mediants not currently included
+
+    def testMediants(self):
+
+        c9 = chord.Chord('C5 E5 G5')
+
+        UFMCmaj = chromaticMediants(c9, transformation='UFM')
+        self.assertEqual([x.name for x in UFMCmaj.pitches], ['E-', 'G', 'B-'])
```
mscuthbert: test octaves at least once also to ensure that octaves are preserved.

music21/analysis/neoRiemannian.py
```diff
@@ -438,32 +619,76 @@ def testNeoRiemannianCombinations(self):
...
+
+        LSMCmaj = chromaticMediants(c9, transformation='LSM')
+        self.assertEqual([x.name for x in LSMCmaj.pitches], ['A', 'C#', 'E'])
+
+        UpChrom = disjunctMediants(c9, UpperOrLower='Upper')
+        self.assertEqual([x.name for x in UpChrom.pitches], ['E-', 'G-', 'B-'])
+
+        DownChrom = disjunctMediants(c9, UpperOrLower='Lower')
+        self.assertEqual([x.name for x in DownChrom.pitches], ['A-', 'C-', 'E-'])
+
```
mscuthbert: test chords farther from C to ensure that enharmonic simplification is working.

music21/analysis/neoRiemannian.py
```diff
@@ -467,70 +480,56 @@ def chromaticMediants(c, transformation='UFM', simplifyEnharmonics=True, raiseEx
...
+    if upperOrLower not in options:
+        raise ValueError('upperOrLower must be one of %s' %options)
 
     if c.isMinorTriad():
-
-        if UpperOrLower=='Upper':
-            transposeChordIntv = 4
-        elif UpperOrLower=='Lower':
+        transposeChordIntv = 4 # Upper, default
+        if upperOrLower=='lower':
```
mscuthbert: linting -- spaces around ==  (will still merge, but will need to fix later)

music21/analysis/neoRiemannian.py
```diff
@@ -213,24 +221,41 @@ def isNeoR(c1, c2):
...
+    transformList = [x for x in transforms]
 
-    else:
-        return False
+    for i in transformList:
+        if i == 'L':
+            c = L(c1)
+            if c.normalOrder == c2NO:
+                return 'L'
+                break
```
mscuthbert: Style: break is unnecessary after a return, since return ends everything.

mscuthbert:
Looks great!  Thanks for the changes -- note some of the pylint style errors that I'll easily fix post merge.  THANKS!

PR #367, opened 2019-02-02
Key signature for multi voice score being exported

pconerly:
When a score has multiple voices (i.e. two notes played at the same time and not combined into a Chord) then the KeySignature is lost when the score is run through 

```
        GEX = m21ToXml.GeneralObjectExporter(s)
        wellFormedObject = GEX.fromGeneralObject(s)
```

The bug presented as a missing key signature when I was using `score.show()` and looking at the output in MuseScore, or when I was exporting to `.musicxml` file.

My fix works, but it feels a tad hacky. I don't quite understand how or why the keySignature gets destroyed in a multi-voice situation. Code review is be appreciated!

I included tests-- one that started passing, and one that started broken.

### DIFF
--- a/music21/stream/makeNotation.py
+++ b/music21/stream/makeNotation.py
@@ -497,6 +497,8 @@ def makeMeasures(
         # creating Measures; this clef is from getClefs, called above
         if measureCount == 0:
             m.clef = clefObj
+            if voiceCount > 0 and s.keySignature is not None:
+                m.insert(0, s.keySignature)
             # environLocal.printDebug(
             #    ['assigned clef to measure', measureCount, m.clef])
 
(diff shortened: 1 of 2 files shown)


mscuthbert:
Ah -- this is entirely possible;  testing automatic voice creation during `makeNotation()` is not where it could be.  However, the bug is in `stream.makeNotation.makeMeasures()` or somewhere in that file -- the musicxml export, etc. is just a consumer of that function.  So the best thing to do is to move the tests closer to the problem, and test just the stream before and after makeNotation.  Also to test is that a stream that already has voices but not measures works properly.

Note the four failing tests after the change. 

Thanks for looking into this, and I would appreciate a patch.

pconerly:
@mscuthbert I moved the tests to be around `makeNotation`, that makes more sense. I figured out how to resolve those broken tests too. Thanks for looking at my previous code so quickly!

mscuthbert:
Sorry that I forgot to review this.  It looks really good.  Waiting for tests to run again and will merge if they pass.

PR #369, opened 2019-02-12
Access to XML IDs for individual Notes in a Chord.

boncata:
Added the property method 'noteIds'. It exposes the individual XML IDs
of Note objects inside a Chord object. These are the XML IDs given in a
MEI file.

This method was implemented by Werner Goebl and Martin Bonev on
the 7th February 2019.

Co-authored-by: Werner Goebl <wergo@users.noreply.github.com>

### DIFF
--- a/music21/chord/__init__.py
+++ b/music21/chord/__init__.py
@@ -252,7 +252,7 @@ class Chord(note.NotRest):
         True if the it passes all `super()`
         equality tests and the pitches are the same
         (possibly in a different order)
-        
+
         >>> c1 = chord.Chord('C4 E4 G4')
         >>> c2 = chord.Chord('E4 C4 G4')
         >>> c1 == c2
@@ -268,7 +268,7 @@ class Chord(note.NotRest):
         False
         >>> c1 != c2
         True
-        
+
         '''
         if not super().__eq__(other):
             return False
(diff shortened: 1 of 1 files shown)


mscuthbert:
Your commit deletes the entire file and then rewrites it -- I suspect that you've changed the line encodings on the file or something like that.  Can you make a new push that only changes the relevant lines for reviewing?

Without seeing the changed lines, can you explain what it does that: `[n.id for n in ch]` would not do?  If it's not complex, it seems a bit of a niche method to include as a core music21 chord property.

boncata:
Hi, thank you for the quick response!

What our method does is exactly the same: return [n.id for n in self._notes], but we access the notes from the chord itself. 

In our project, we are reading MEI files and extracting the score information from them. What we need is the XML IDs of individual notes. To be honest, I was not aware that I could iterate over the notes inside a chord that way: [n.id for n in ch] (in hindsight, it is not a large leap of logic, given that the rest of the library functions in a similar way). That is why we added the functionality to the Chord class.

With the command you proposed we solve the problem we had and we are happy to use music21 that way.

 If you think the method we added can be useful for music21 keep it, but it only provides an alternative way to access individual note ids. In that sense, you can say it is niche. I did another push to my branch, adding only our method.

mscuthbert:
What we need is to make `.notes` return `._notes` cast to a tuple -- then it's more obvious how it works and how to do things like this.

boncata:
Yes, this would probably make it more obvious how to access individual notes in a chord.

mscuthbert:
`.notes` is added as a property in the current Git master and will be part of the next release.

PR #373, opened 2019-02-25
Fix the name in the corpus.parse call

namin:


mscuthbert:
Why this change?  bach/bwv66.6 should work perfectly well.  Does it not for you?

namin:
Oops, it does now. Sorry for the noise.

mscuthbert:
no worries!

PR #376, opened 2019-03-15
 Converter for tabular representations of harmonic analyses

MarkGotham:


mscuthbert:
will review this weekend, but be sure to check build failure first.

PR #385, opened 2019-04-15
Update MuseScore from v2 to v3

YoWenqin:
The latest version of MuseScore is v3. Running the configuration script will set "musicxmlPath
 correctly, but leave "musescoreDirectPNGPath" set to a path for MuseScore 2. Maybe it's better to dynamically update the "musescoreDirectPNGPath" to the path of the version of MuseScore that the user installs.


mscuthbert:
Sorry I forgot about this PR.  Thanks!

PR #389, opened 2019-05-06
abc accidental propagation, three flavors (for notes anyway)

dvdrndlph:
I have implemented and tested code to support the three `@@propagate-accidentals` alternatives in the latest (2.1) abc standard, but only for files explicitly identified as `@abc-2.1` files or for callers requesting this type of parsing.

I have not done anything to change how ties are processed. This is left for future discussion/work.

Chords are another topic for investigation.


music21/abcFormat/__init__.py
```diff
@@ -1235,11 +1235,13 @@ class ABCNote(ABCToken):
     # given a logical unit, create an object
     # may be a chord, notes, bars
 
-    def __init__(self, src=''):
+    def __init__(self, src='', carriedAccidental=None):
```
mscuthbert: Can you add information about what type of object a `carriedAccidental` is -- is it a string (abc accidental) or a pitch.Accidental object?

mscuthbert: If there is an accidental explicitly set, does this override carriedAccidental?

music21/abcFormat/__init__.py
```diff
@@ -1639,16 +1656,16 @@ def parse(self, forceKeySignature=None, forceDefaultQuarterLength=None):
...
 class ABCHandler:
-
     # divide elements of a character stream into objects and handle
     # store in a list, and pass global information to compontns
-    def __init__(self):
-        # tokens are ABC objects in a linear stream
+    def __init__(self, abcVersion=None):
+        # tokens are ABC objects import n a linear stream
+        self.abcVersion = abcVersion # At this point, only "2.1" and None are valid settings.
+        self.abcDirectives = dict()
```
mscuthbert: please use literal `{}` instead of `dict()`

mscuthbert: Can abcVersion be a tuple like other version information (2, 1) -- or converted to such.  We should be doing > and < with versions.

music21/abcFormat/__init__.py
```diff
@@ -1766,11 +1783,18 @@ def barlineTokenFilter(token):
         else: # append unaltered
             post.append(ABCBar(token))
         return post
-
-
+    
     # --------------------------------------------------------------------------
     # token processing
 
+    def _accidentalPropagation(self):
```
mscuthbert: Can this be tested here with a doctest and explained in a doc string (required even for "private" functions)?  What about %abc-2.0?

mscuthbert: (I just don't want to fix a weird bug when I'm 60 or because the version numbering system of abc changes and ABC 10.0 is released)

music21/abcFormat/__init__.py
```diff
@@ -1811,9 +1835,15 @@ def tokenize(self, strSrc):
         collect = []
         lastIndex = len(strSrc) - 1
         skipAhead = 0
+        accidentalsAndDecorations = '.~^=_HLMOPSTuv'
+        accidentals = '^=_'
 
         activeChordSymbol = '' # accumulate, then prepend
-
+        accidentalized = dict()
```
mscuthbert: as above

mscuthbert: should accidentalized be stored in `self.` so it can be used in sub functions?

music21/abcFormat/__init__.py
```diff
@@ -1811,9 +1835,15 @@ def tokenize(self, strSrc):
...
+        accidental = None
+        pit = None
```
mscuthbert: what is `pit` here?  We try to avoid abbreviations that go too far.  Is it `pitchStr` or `abcPitch` or `lastPitch`?

music21/abcFormat/testFiles.py
```diff
@@ -612,8 +676,116 @@ def testBasic(self):
...
+            title = ah.getTitle()
+            environLocal.printDebug([title])
+            s = translate.abcToStreamScore(ah)
+            try:
+                unused_out = gex.parse(s)
+            except UnicodeDecodeError as ude:
+                environLocal.warn('About to fail on ABC file #{}'.format(i))
+                raise ude
+            if title == 'Through Measure':
+                notes = s.flat.getElementsByClass(note.Note)
```
mscuthbert: wrap `s.flat....` in a `list()` call to make this run faster (just a good practice -- otherwise we're using an expensive iterator)

music21/abcFormat/testFiles.py
```diff
@@ -612,8 +676,116 @@ def testBasic(self):
...
+    def testAbc21(self):
+        from music21 import abcFormat, note
+        from music21.abcFormat import translate
+        from music21.musicxml import m21ToXml
 
+        af = abcFormat.ABCFile(abcVersion="2.1")
+
+        gex = m21ToXml.GeneralObjectExporter()
         
+        files = [carryThrough]
```
mscuthbert: If we're only doing one file, go ahead and read it without the for loop, which was confusing.

music21/abcFormat/testFiles.py
```diff
@@ -612,8 +676,116 @@ def testBasic(self):
...
+        from music21 import abcFormat, note
+        from music21.abcFormat import translate
+        from music21.musicxml import m21ToXml
+
+        af = abcFormat.ABCFile()
+
+        gex = m21ToXml.GeneralObjectExporter()
+        
+        files = [directiveCarryPitch, directiveCarryOctave, directiveCarryNot]
+        for i, tf in enumerate(files):
```
mscuthbert: instead of doing this with a for loop, do a separate test method for each file please.

music21/abcFormat/testFiles.py
```diff
@@ -612,8 +676,116 @@ def testBasic(self):
...
+
+        gex = m21ToXml.GeneralObjectExporter()
         
+        files = [carryThrough]
+        for i, tf in enumerate(files):
+            ah = af.readstr(tf)
+            title = ah.getTitle()
+            environLocal.printDebug([title])
+            s = translate.abcToStreamScore(ah)
+            try:
```
mscuthbert: are we worried about stream to musicxml conversion?  If not, cut this test which will slow down testing of what we actually want to test.

music21/abcFormat/testFiles.py
```diff
@@ -612,8 +676,116 @@ def testBasic(self):
...
+        gex = m21ToXml.GeneralObjectExporter()
+        
+        files = [directiveCarryPitch, directiveCarryOctave, directiveCarryNot]
+        for i, tf in enumerate(files):
+            ah = af.readstr(tf)
+            title = ah.getTitle()
+            environLocal.printDebug([title])
+            s = translate.abcToStreamScore(ah)
+            try:
+                unused_out = gex.parse(s)
```
mscuthbert: as above about not needing to test musicxml output

music21/abcFormat/translate.py
```diff
@@ -279,6 +278,7 @@ def parseTokens(mh, dst, p, useMeasures):
             #ql += t.quarterLength
 
         elif isinstance(t, abcFormat.ABCNote):
+            pit = t.pitchName
```
mscuthbert: revert please -- this change takes something that is explicit (t.pitchName) into something that I need to look up where `pit` comes from

mscuthbert:
Can you run through the abc directory test/testLint.py file to see if everything passes linting?  THanks!

music21/abcFormat/__init__.py
dvdrndlph: New comments will clarify.

music21/abcFormat/__init__.py
dvdrndlph: Done.

music21/abcFormat/__init__.py
dvdrndlph: Agreed. I have added the future-proofing.

Also, 2.0 is included now. For some reason, I thought this was only ever a draft standard.

music21/abcFormat/__init__.py
dvdrndlph: {} are used throughout to initialize dictionaries. Not sure about storing accidentalized in self. It really only has utility within tokenize().

music21/abcFormat/__init__.py
dvdrndlph: I went with `abcPitch` (the abc that defines a pitch if present in the token being identified)

music21/abcFormat/testFiles.py
dvdrndlph: Okay, I will add more test methods whenever I deal with ties and chords.

music21/abcFormat/testFiles.py
dvdrndlph: Agreed.

music21/abcFormat/testFiles.py
dvdrndlph: Sorry. Just cutting, pasting, and not thinking. The gex stuff is gone now.

music21/abcFormat/testFiles.py
dvdrndlph: Agreed.

music21/abcFormat/translate.py
dvdrndlph: Roger that.

dvdrndlph:
The abc code now generates no messages from pylint.

music21/abcFormat/__init__.py
```diff
@@ -1788,7 +1796,18 @@ def barlineTokenFilter(token):
     # token processing
 
     def _accidentalPropagation(self):
-        if not self.abcVersion or self.abcVersion[0] == "1":
+        '''
+        Determine how accidentals should "carry through the measure."
+        
+        >>> ah = abcFormat.ABCHandler(abcVersion=(1,3,0))
+        >>> ah._accidentalPropagation()
+        not
```
mscuthbert: these will need to be in single quotes.

music21/abcFormat/__init__.py
```diff
@@ -1788,7 +1796,18 @@ def barlineTokenFilter(token):
     # token processing
 
     def _accidentalPropagation(self):
-        if not self.abcVersion or self.abcVersion[0] == "1":
+        '''
+        Determine how accidentals should "carry through the measure."
+        
+        >>> ah = abcFormat.ABCHandler(abcVersion=(1,3,0))
```
mscuthbert: spaces after comma in tuple same w/ line 1805, 1809

mscuthbert:
See https://travis-ci.org/cuthbertLab/music21/jobs/528975356 for fix

PR #403, opened 2019-05-28
an attempt at a fix of context sites

mscuthbert:
there were many inner functions that shadowed names from outer functions.  Hopefully this should fix it.


mscuthbert:
Three runs and no errors -- merging.

PR #414, opened 2019-07-07
Refactor four methods in sorting.py.

alexandermorgan:
Make modest performance improvements in the shortRepr, modify, add,
and sub methods in sorting.py.

### DIFF
--- a/music21/sorting.py
+++ b/music21/sorting.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-#------------------------------------------------------------------------------
+# -----------------------------------------------------------------------------
 # Name:         sorting.py
 # Purpose:      Music21 class for sorting
 #
@@ -8,7 +8,7 @@
 # Copyright:    Copyright © 2014-2015 Michael Scott Cuthbert and the music21
 #               Project
 # License:      LGPL or BSD, see license.txt
-#------------------------------------------------------------------------------
+# -----------------------------------------------------------------------------
 '''
 This module defines a single class, SortTuple, which is a named tuple that can
 sort against bare offsets and other SortTuples.
@@ -32,7 +32,7 @@ from music21 import exceptions21
 
 INFINITY = float('inf')
 
-_attrList = ('atEnd', 'offset', 'priority', 'classSortOrder', 'isNotGrace', 'insertIndex')
+_attrList = ['atEnd', 'offset', 'priority', 'classSortOrder', 'isNotGrace', 'insertIndex']
 
 class SortingException(exceptions21.Music21Exception):
     pass
@@ -105,7 +105,7 @@ class SortTuple(namedtuple('SortTuple', _attrList)):
             elif self.atEnd == 1:
                 return True
             else:
-                return (self.offset == other)
+                return self.offset == other
         except ValueError:
             return NotImplemented
 
@@ -116,7 +116,7 @@ class SortTuple(namedtuple('SortTuple', _attrList)):
             if self.atEnd == 1:
                 return False
             else:
-                return (self.offset < other)
+                return self.offset < other
         except ValueError:
             return NotImplemented
 
@@ -129,7 +129,7 @@ class SortTuple(namedtuple('SortTuple', _attrList)):
             elif self.atEnd == 1:
                 return False
             else:
-                return (self.offset > other)
+                return self.offset > other
         except ValueError:
             return NotImplemented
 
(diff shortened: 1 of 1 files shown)


mscuthbert:
Can you give a before and after on the test speed improvements?  I'd like to keep the changes to add, sub, modify, since those are performance critical, but my preference is to revert the changes to shortRepr since the current version is to me much more readable and it is not in any way performance critical (it's only used for documentation and testing).

music21/sorting.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# ------------------------------------------------------------------------------
+#------------------------------------------------------------------------------
```
mscuthbert: Revert these changes since they interfere with PyCharm linting.

music21/sorting.py
```diff
@@ -105,7 +105,7 @@ def __eq__(self, other):
             elif self.atEnd == 1:
                 return True
             else:
-                return self.offset == other
+                return (self.offset == other)
```
mscuthbert: revert since it doesn't change output and we're trying to reduce extraneous parentheses.

music21/sorting.py
```diff
@@ -116,7 +116,7 @@ def __lt__(self, other):
             if self.atEnd == 1:
                 return False
             else:
-                return self.offset < other
+                return (self.offset < other)
```
mscuthbert: as above.

music21/sorting.py
```diff
@@ -129,10 +129,13 @@ def __gt__(self, other):
             elif self.atEnd == 1:
                 return False
             else:
-                return self.offset > other
+                return (self.offset > other)
         except ValueError:
             return NotImplemented
 
+    def __ne__(self, other):
```
mscuthbert: thanks!

music21/sorting.py
```diff
@@ -153,21 +156,16 @@ def shortRepr(self):
         >>> st.shortRepr()
         'End <4.7.[Grace].200>'
         '''
-        reprParts = []
```
mscuthbert: revert since not performance critical and later method is much less readable.

music21/sorting.py
```diff
@@ -194,16 +192,15 @@ def modify(self, **kw):
 
         Changing offset, but nothing else, helps in creating .flat positions.
         '''
-        outList = []
-        for attr in _attrList:
-            outList.append(kw.get(attr, getattr(self, attr)))
-        return self.__class__(*tuple(outList))
+        outList = [kw.get(attr, getattr(self, attr)) for attr in _attrList]
+        return self.__class__(*outList)
```
mscuthbert: thanks!

music21/sorting.py
```diff
@@ -246,30 +241,29 @@ def sub(self, other):
...
 ZeroSortTupleLow = SortTuple(atEnd=0, offset=0.0, priority=float('-inf'), classSortOrder=0,
                           isNotGrace=1, insertIndex=0)
 
-ZeroSortTupleHigh = SortTuple(atEnd=0, offset=0.0, priority=float('inf'), classSortOrder=0,
+ZeroSortTupleHigh = SortTuple(atEnd=0, offset=0.0, priority=INFINITY, classSortOrder=0,
                           isNotGrace=1, insertIndex=0)
 
 
-# -----------------------------------------------------------------------------
+#------------------------------------------------------------------------------
```
mscuthbert: revert for linting.

music21/sorting.py
alexandermorgan: I didn't add this method, or make some of the changes that produced the linting errors. I'm not sure where some of these diff lines came from. I do understand all your comments though, so I'm making the necessary changes now.

music21/sorting.py
alexandermorgan: I disagree about the readability, but no problem to revert this chunk.

alexandermorgan:
Using the code snippits in the doc strings as examples, and clocking with cProfile, I got the following reductions in runtime:
modify: 24%
add: 20%
sub: 20%

And for what it's worth, the shortRepr change would have reduced runtime by 80%, but I understand that method isn't important.
The main reason for these speed improvements is that the dot syntax in Python is surprisingly slow, so avoiding .append wherever possible is always going to be a win.

alexandermorgan:
At the moment, I'm working on getting handle on a couple of optimization strategies for loops in particular. So if there are some methods you'd like to speed up, please let me know. I'd love to give it a try.

alexandermorgan:
@mscuthbert is there anything else you'd like to see in this PR? I believe I've incorporated all of your change requests, but if there's something I missed I can make more edits. The coverage decrease seems to just be because there are a couple fewer lines in code that was already tested. I'm kind of surprised coveralls doesn't automatically detect that.

mscuthbert:
Looks great to me.  THANK YOU!

PR #417, opened 2019-07-17
Audio search freq hotfix

mscuthbert:
New servers at Travis-CI do not truncate numbers in print like before


mscuthbert:
apparently duplicate -- was merged.

PR #428, opened 2019-08-04
updated Chorales

doctor-schmidt:


mscuthbert:
I think I'll need to run this through the measure renumbering script I used once before.  Let me know when all the chorale.mxl files are in and I'll dig up that script and run it across the new corpus.

Since you've gotten really good at Github now, one suggestion is to make a new branch off of master just before any pull request and then do a PR from that branch.  That way you can make any other changes to master that you want without affecting this PR. It also makes it easier for me to modify your PR without affecting any other changes you've made.  But for now this will be great and enough on its own.

doctor-schmidt:
The PR and the zip contain all the chorale.mxl files I have worked with over the past year.
Also, I always run them through the measure renumbering script, so they should be fine as they are.

Thanks so much for the suggestion, my next PR will be from a branch off of master. Music21 has taught me a lot!

mscuthbert:
Ah! if they're already been renumbered could you look at the output of our test program and see if the new results look valid for any of the tests that you can follow fully (the features ones don't worry about -- those have likely changed) -- it looks like there are some repeat signs added or thing of that sort -- just want to be sure that the last measure of BWV66.6 should now be 10 not 9.  Thanks!

doctor-schmidt:
I didn’t edited bwv66.6, it's not in the Riemenschneider collection. So that error was made by someone else 2 years ago.

doctor-schmidt:
I just looked at bwv66.6. To me it looks like it should be 9.
To see the last two measures I do this:
>>> c = corpus.parse('bwv66.6')
>>> m = c.measures(8,9)
>>> m.show()

doctor-schmidt:
I just ran bwv66.6 through my work flow which gives me this



first measure (pickup) is 0 and last measure is 9

mscuthbert:
Wait? Why close? I wanted to merge soon.

doctor-schmidt:
I thought I messed up something. If you are able to straighten things out, great!

PR #442, opened 2019-09-22
Fixing example to match following comment about double parens.

ivarley:
Noticed that there's a comment in this chapter about ensuring you have the double parentheses, but the example actually doesn't have double parens because it's got a parameter. This parameter is the default, so the example stays the same if it's excluded, and then it matches the comment.

### DIFF
--- a/documentation/source/usersGuide/usersGuide_02_notes.ipynb
+++ b/documentation/source/usersGuide/usersGuide_02_notes.ipynb
@@ -1379,7 +1379,7 @@
    },
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


mscuthbert:
ah! got it.  Will fix.  Yes -- I'll need to also change the image, etc.

mscuthbert:
Got it -- I didn't like that it was showing as a whole rest (technically a full measure rest) even though it's just one beat long.  But it's better to show the basic case.  I'm closing this request since I've regenerated the image also in the repo.  Will go up w/ next doc update.

PR #443, opened 2019-09-23
meter.py: fix testBasic() test case

dhgutteridge:
Fix a reference so that the test case executes. (Or, to put it another way,
make this reference to TimeSignature() consistent with the rest of the test
cases here.)

Noted while running the test suite:

ERROR: testBasic (music21.meter.TestExternal)
  File "/home/disciple/pkgsrc/wip_dhg/py-music21/work/music21-5.7.0/music21/meter.py", line 4444, in testBasic
    ts = music21.meter.TimeSignature('%s/%s' % (meterStrNumerator,
NameError: name 'music21' is not defined


mscuthbert:
THANKS!

PR #449, opened 2019-10-04
Handle <staff-type> from musicxml

maxrothman:
When ingesting musicxml, store the contents of \<staff-type> in a new
"custom" metadata field so it can be discovered downstream, for example
to be distinguished from other parts and turned into a variant via
variant.mergePartAsOssia


mscuthbert:
This needs tests to see that it works -- what is the problem that is being solved?  Note that the PartParser attribute is one shared by the entire class -- this means that one parse will bleed into another.

Note also music21 naming conventions (camelCase not underscore).  We're not likely to have 'custom' as something that goes with 'title', etc. -- that seems like a general metadata convention and not something that should be used for staffTypes, etc.

maxrothman:
I tried to describe the problem to be solved in the PR body and commit message, are there parts of that that aren't clear? What questions can I answer?

I'm happy to write some tests and modify the approach. This was definitely a bit of a hack job, but I thought it'd be a good way to start a conversation. Ultimately my priority is to make `<staff-type>` information available inside music21, so if there's a better way to approach the problem I'm all ears.

mscuthbert:
If the information is staff-type then it should be a new attribute on the current PartStaff (or Part?) called `.staffType` which makes sure that the only allowable values are "ossia", "cue", "editorial", "regular", "alternate" or None.  All contributions to music21 need documentation and tests -- how will someone know how to set the staff-type information or get it from the parsing.  If you're just setting up a hacking script for yourself, then it's no problem to not have docs and tests and ways to ensure that illegal values don't break other things, but if you'd like to contribute this back into the project it needs all of these things.  Thanks!

maxrothman:
> All contributions to music21 need documentation and tests

No argument there, as I said, my goal was to start a discussion, and a discussion has been started. Success! I figured that there would be feedback on the approach, and once that's settled I'll handle docs and tests.

> If the information is staff-type then it should be a new attribute on the current PartStaff (or Part?) called `.staffType` which makes sure that the only allowable values are "ossia", "cue", "editorial", "regular", "alternate" or None.

I was hesitant to couple music21's internal representation to the musicxml format because I thought other input formats might have completely different metadata that one might want to preserve, so I thought that using a more general shape (i.e. a dict) might be a better approach. I also thought about simply converting staves marked with certain `<staff-type>`s directly into variants, but I figured it was better to just preserve the info and allow music21 users to make their own decisions about how to handle the data. I'm agnostic between all three approaches: `.staffType`, free-form metadata, and going directly into a variant, do you have a preference?

mscuthbert:
The most helpful thing for getting this through would be some tests (they can fail) that provide an XML file or fragment with staff-type and some tests to make sure that it parses and converts back.  The coding will not take me much time.  The writing of tests is what slows down the project.

maxrothman:
I've added a commit to my branch containing a couple of tests that use a musicxml file containing `<staff-type>ossia</staff-type>` that I also checked in. You can verify that the musicxml file is correct by unzipping it and uploading it to https://www.truugo.com/xml_validator/, which will validate it against the internally-linked DTD. The tests currently pass on my branch because they use the method of accessing stafftype metadata that was added in the branch. I thought that would be helpful to you, since you can easily modify the access method in the tests to see if your changes work.

So far I've just got a test verifying that the musicxml file deserialized correctly and that the stafftype metadata can be accessed. Are there any other tests that would be helpful to have?

maxrothman:
After a little more thinking, I also added a serialization test checking that `<staff-type>` makes it back into the musicxml after calling `score.write()`. It currently fails since I didn't add that functionality to the PR.

mscuthbert:
hi!  sorry -- busy times.  working this weekend on m21 will try to get to it.

music21/musicxml/xmlToM21.py
```diff
@@ -1407,6 +1407,8 @@ def rightsToCopyright(self, rights):
 
 # -----------------------------------------------------------------------------
 class PartParser(XMLParserBase):
+    staff_type_map = {}
```
mscuthbert, line 1410: This presumably needs to be inside the __init__ since it shouldn't be shared among PartParsers?

music21/test/musicXMLTest.py
```diff
@@ -0,0 +1,48 @@
+from pathlib import Path
+from tempfile import NamedTemporaryFile
+import unittest
+from xml.etree import ElementTree as ET
+
+import music21 as m21
+
+DATA_FILE = Path(__file__).parent / 'data/Test1.mxl'
```
jacobtylerwalls, line 8: `xmlToM21` already has a `Test` class, so I would add these three methods there unless @mscuthbert is of the opinion that class is getting too large.

also if your external file is reasonably small (not a bad idea anyway!) you can just block quote it as text in `musicxml.testPrimitive`. otherwise there are .xml and .mxl files loose in the `musicxml` directory which you could add this to, without creating a new `music21.test.data` directory.

figured this would keep this easy to understand for the next person who comes after us!

music21/test/musicXMLTest.py
```diff
@@ -0,0 +1,48 @@
...
+
+class Test(unittest.TestCase):
+    def testDeserialization(self):
```
jacobtylerwalls, line 11: these method names could be more descriptive since (de)serialization is the point of the entire module! :-)

music21/test/musicXMLTest.py
```diff
@@ -0,0 +1,48 @@
...
+        """Sample score deserializes properly"""
+        score = m21.converter.parse(str(DATA_FILE))
+        self.assertEqual(
+            [str(p.getInstrument()) for p in score.parts],
+            ['P1: Piano: Piano',
+             'P1: Piano: Piano',
+             'P2: Voice: Voice',
+             'P2: Voice: Voice']
+        )
+        self.assertTrue(all(len(p.getElementsByClass('Measure')) == 8 for p in score.parts))
```
jacobtylerwalls, line 21: if this were to fail, IIRC the test runner would just show the truthiness of the expression, not the specific incorrect values. I suggest
`self.assertEqual([len(p.getElementsByClass('Measure')) for p in score.parts], [8, 8, 8, 8])`

maxrothman:
Sorry for the delay in my response, I'll be able to look at this during the
week of the 24th.

maxrothman:
@mscuthbert Sorry for the delay, I've just caught up on this. From your [October 8 comment](https://github.com/cuthbertLab/music21/pull/449#issuecomment-539775755), my understanding was that you were looking for me to proof-of-concept some tests, then you'd take over and implement the requested feature the way you wanted. If that's true, then does it make sense to review the code as-is, and if not, what do you need from me to move this PR forward? 

Honestly I've barely touched music21 since making this PR, so I am no longer at all familiar with the codebase. If your time is the limited resource here, then it very well might be faster for you to take the reins. That said, I'm still interested in seeing this feature added, so I'm willing to help in whatever way makes the most sense.

jacobtylerwalls:
Hi @maxrothman, thanks for the PR. I think the best way to get this feature in is for you to polish it based on the feedback so far, and then I can give you another review if I think it needs one. So far the feedback includes:
- moving `staff_type_map`
- a new attribute `staff_type` on `Part`, probably an enum
- conforming the tests to the surrounding ones
- mark the failing test with an "x_" like `def x_test_...` so that they won't fail on CI until such time someone implements the feature in the other direction.

Speaking of which, having a polished PR makes it faster/more likely that someone else (me? @MarkGotham ?) can hop in and provide the missing code to make the test pass. Good luck, and reach out if you have questions.
Cheers,
Jacob

mscuthbert:

mscuthbert:
Need to keep closing and reopening -- apparently need fewer concurrent runs.

mscuthbert:
Closing in favor of #735 which puts the information in the StaffLayout object.

PR #466, opened 2019-11-04
solved issue with float/int conversion

SandroLuck:
The issue is(I assume due to changes of how python3 handels division), that windowCount in the 'noOverlap' case will be a float (even in the case where maxWindowCount / windowSize is divisible). This will then create an issue in line 149,       
`  data = [0] * windowCount`
`  File "/usr/local/lib/python3.7/site-packages/music21/analysis/windowed.py", line 146, in analyze
    data = [0] * windowCount
TypeError: can't multiply sequence by non-int of type 'float'`

This Pr solves this by casting(I assume this way is most consistent with prior versions of python), also adds a warning, since i belive user should eb made aware

### DIFF
--- a/music21/analysis/windowed.py
+++ b/music21/analysis/windowed.py
@@ -124,7 +124,7 @@ class WindowedAnalysis:
         >>> len(wa._windowedStream)
         36
         >>> a, b = wa.analyze(1)
-        >>> len(a) , len(b)
+        >>> len(a), len(b)
         (36, 36)
 
         >>> a, b = wa.analyze(4)
--- a/music21/analysis/windowed.py
+++ b/music21/analysis/windowed.py
@@ -130,6 +130,7 @@ class WindowedAnalysis:
         >>> a, b = wa.analyze(4)
         >>> len(a), len(b)
         (33, 33)
+
         >>> a, b = wa.analyze(1, windowType='noOverlap')
         >>> len(a), len(b)
         (37, 37)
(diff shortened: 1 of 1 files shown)


SandroLuck:
side note on the warning: this will stop exeution, can be solved by either introducing warnings module, or simply removing the warning(not sure who decides on this)

mscuthbert:
Can you fix the failing test, and also make sure that there are spaces before the commas.  Thanks!

SandroLuck:
fixed, I assume that you ment space after the comma (from what i can see in the rest of the file format wise)

SandroLuck:
Found an additional issue, with adj average. When the windowSize is bigger 1, I get the error 
`music21.exceptions21.StreamException: the object (<music21.stream.Measure 2 offset=1.0>, id()=4979511120) is already found in this Stream (<music21.stream.Stream 0x12e1a81d0>, id()=5068456400)
` not entirely sure how and where that fix would be

mscuthbert:
> Found an additional issue, with adj average. When the windowSize is bigger 1, I get the error
> `music21.exceptions21.StreamException: the object (<music21.stream.Measure 2 offset=1.0>, id()=4979511120) is already found in this Stream (<music21.stream.Stream 0x12e1a81d0>, id()=5068456400) ` not entirely sure how and where that fix would be. Maybe an issue for a diffrent PR

Yes -- let's make that a different issue, though related.

mscuthbert:
I'm going to merge because this does everything that the title of the issue suggests/requires, but if you're interested in still working on this, I'd appreciate docs on what can be used as `windowType` and probably making a "StrEnum" type class (look for `StrEnum` at https://docs.python.org/3/library/enum.html) for WindowType.ADJACENT  WindowType.OVERLAP etc. so it's easier for people to see what are valid values for it  (we'd make it a subclass of `str` so that old code that uses strings would still work).

PR #469, opened 2019-11-05
Fix beaming issues

AlmogCohen:
Using music21 we've noticed various beaming bugs in the past. We've originally just kept the fixes internally but now as we move to python3 we take this opportunity to contribute the fixes back to the master repo.

As you can see, I've added "TestMakeBeams" class to demonstrate some proper beaming, as well as demonstrate some of the errors we encountered.
The unit tests that originally failed were: "test_1_e_a_after_16th_note" and "test_repeated_1_e_a". Not the best unit test names...I know...but we are working on drummers project here :)

The fix was simple and you can see our changes in the "getBeams" function.
For each failing unit test I've provided the wrong representation today and a proper representation as well that I was aiming for.

I'd be very happy to see these changes come into the product. Please let me know if there is anything I should change and I'd be happy to do so (e.g. remove the notes for example, once you approve them).

### DIFF
--- a/music21/meter.py
+++ b/music21/meter.py
@@ -3687,7 +3687,6 @@ class TimeSignature(base.Music21Object):
         else:
             return []
 
-
         if len(srcList) <= 1:
             return [None for _ in srcList]
 
@@ -3717,7 +3716,6 @@ class TimeSignature(base.Music21Object):
 
             beamNext = beamsList[i + 1] if not isLast else None
             beamPrevious = beamsList[i - 1] if not isFirst else None
-
 
             # get an archetype of the MeterSequence for this level
             # level is depth, starting at zero
@@ -3786,25 +3784,19 @@ class TimeSignature(base.Music21Object):
                 else:
                     beamType = 'start'
 
-
-            # last beams was active, last beamNumber was active,
-            # and it was stopped or was a partial-left
             elif (beamPrevious is not None
                   and beamNumber in beamPrevious.getNumbers()
-                  and beamPrevious.getTypeByNumber(beamNumber) in ['stop', 'partial-left']
-                  and beamNext is not None):
-                beamType = 'start'
+                  and beamPrevious.getTypeByNumber(beamNumber) in ['stop', 'partial-left']):
+                    # last beams was active, last beamNumber was active,
+                    # and it was stopped or was a partial-left
+                    if beamNext is not None:
+                        beamType = 'start' if beamNumber in beamNext.getNumbers() else 'partial-right'
 
-
-            # last note had beams but stopped, next note cannot be beamed to
-            # was active, last beamNumber was active,
-            # and it was stopped or was a partial-left
-            elif (beamPrevious is not None
-                  and beamNumber in beamPrevious.getNumbers()
-                  and beamPrevious.getTypeByNumber(beamNumber) in ['stop', 'partial-left']
-                  and beamNext is None):
-                beamType = 'partial-left'  # will be deleted later in the script
-
+                    # last note had beams but stopped, next note cannot be beamed to
+                    # was active, last beamNumber was active,
+                    # and it was stopped or was a partial-left
+                    else:
+                        beamType = 'partial-left'  # will be deleted later in the script
 
             # if no beam is defined next (we know this already)
             # then must stop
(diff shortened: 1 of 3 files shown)


AlmogCohen:
sorry for the commit mess, for some reason it happen when I merged master back into my branch. I hope eventually you'll squash all commits when merging to master

music21/meter.py
```diff
@@ -3786,25 +3784,19 @@ def fixBeamsOneElementDepth(i, el, depth):
...
+                    # last beams was active, last beamNumber was active,
+                    # and it was stopped or was a partial-left
+                    if beamNext is not None:
+                        beamType = 'start' if beamNumber in beamNext.getNumbers() else 'partial-right'
+
+                    # last note had beams but stopped, next note cannot be beamed to
+                    # was active, last beamNumber was active,
+                    # and it was stopped or was a partial-left
+                    else:
+                        beamType = 'partial-left'  # will be deleted later in the script
```
AlmogCohen: Here I simply did 2 things:

1. merge the two identical long "if"s into one with additional nested "if-else". No code was changed here.
2. Fix the source of the beaming issues: change `beamType = 'start'` into `beamType = 'start' if beamNumber in beamNext.getNumbers() else 'partial-right'`

music21/test/testStream.py
```diff
@@ -284,6 +285,168 @@ def testBeamsMeasure(self):
...
+        self.assertEqual(second_note_beams, beams[1])
+        self.assertEqual(third_note_beams, beams[2])
+        self.assertEqual(forth_note_beams, beams[3])
+
+    def test_1_e__after_16th_note(self):
+        """
+        Test that a 16th+8th notes after a 16th notes have proper beams.
+
+        Note: MuseScore repr (auto-corrected): https://cl.ly/90ce7b
+        Currently the forth note has 2nd "start" beam which has no "end" anywhere after it.
```
mscuthbert: correct typo "fourth" for forth -- here and lower.

music21/test/testStream.py
```diff
@@ -284,6 +285,168 @@ def testBeamsMeasure(self):
         bMeasure.show()
 
 
+class TestMakeBeams(unittest.TestCase):
```
mscuthbert: These tests need to go within the single "class Test" to run in the music21 testing system.

music21/test/testStream.py
```diff
@@ -14,14 +14,15 @@
...
+from music21.note import GeneralNote
 
 from music21.stream import Stream
 from music21.stream import Voice
 from music21.stream import Measure
 from music21.stream import Score
 from music21.stream import Part
 
-from music21 import bar
+from music21 import bar, beam
```
mscuthbert: one import per line please.

music21/test/testStream.py
```diff
@@ -284,6 +285,168 @@ def testBeamsMeasure(self):
         bMeasure.show()
 
 
+class TestMakeBeams(unittest.TestCase):
+    """Group together unittests to verify proper beams creation."""
+    @staticmethod
+    def get_beams_from_stream(stream):
+        """Helper function to return beam list for all notes and rests in the stream."""
+        return [note.beams for note in stream if isinstance(note, GeneralNote)]
```
mscuthbert: Should be `NotRest` instead of `GeneralNote` -- rests don't have beams
Please don't use `stream` or `note` as variable names because they can shadow imports.

AlmogCohen: Here, I would like to keep the `GeneralNote` as I usually verify in the unit tests the beaming for all the notes to keep it clear. Rests do have the `beams` attribute which always returns an empty `Beams` object. That's fine as my unit tests are looking for that too.

AlmogCohen:
@mscuthbert All changes have been made :) Thank you for the quick response!

The only thing I've kept is the `GeneralNote` in the helper function, as I do expect the empty `Beams` object coming from rests when verifying the beams structure for the measures.

music21/test/testStream.py
AlmogCohen: I will think I'll actually remove all the Notes with the screenshots, as all issues has been fixed now. I'll keep a short note with a screenshot of how certain cases should look like

music21/test/testStream.py
mscuthbert: Ah!  Well, that's a MISTAKE for rests!  It shouldn't be there!  They don't have stems for sure. But okay!

mscuthbert:
Looks great to me -- just note that rests will soon lose their beams attribute -- it shouldn't be there.  But there will be a deprecation cycle before they go away.

PR #470, opened 2019-11-05
REALLY BIG changes to pass linting

mscuthbert:


.flake8
```diff
@@ -0,0 +1,15 @@
+[flake8]
+ignore=
+  E127,  # over indented
+  E128,  # under indented
+  E302,  # blank lines
+  E303,  # blank lines
```
AlmogCohen: Just a curious question following my latest PR...Why do we ignore these line linting errors? I imagine that this one is a very easy one to fix while making the code look much better and pep8 complaint...

mscuthbert: because I'm going to soon turn on the linter on PRs and I need all my existing code to pass first.  There's a lot that won't pass right now, so just getting a minimum number of tests in -- already has taken 6 hours to get the code up to that level.

AlmogCohen:
Beautiful work, thank you. I believe it is very important for the project
in the long run to signal of it's maturity. PEP8 and style might look super
technical, but it means a lot for experienced open source developers.

Sometimes I can try and work with some project and be scared away just by
it's style and structure. Many times bad style signals low project quality
even if this is not true. Thankfully I ignored this in music21 case and
realized how great and useful this package is.

This project enabled us to do things which would have not been easily done
otherwise. Keep the good work.

If you need some help with the linting down the line I would actually be
happy to help. Let me know if I can help with some simple linting stuff
(such as the spaces etc). I'm sure you are changing things which are not
only style related when you refactor now so much files.

Cheers!

mscuthbert:
> Beautiful work, thank you. I believe it is very important for the project in the long run to signal of it's maturity. PEP8 and style might look super technical, but it means a lot for experienced open source developers. Sometimes I can try and work with some project and be scared away just by it's style and structure. Many times bad style signals low project quality even if this is not true. Thankfully I ignored this in music21 case and realized how great and useful this package is. This project enabled us to do things which would have not been easily done otherwise. Keep the good work. If you need some help with the linting down the line I would actually be happy to help. Let me know if I can help with some simple linting stuff (such as the spaces etc). I'm sure you are changing things which are not only style related when you refactor now so much files. Cheers!
> […](#)

Yes -- I want this project to become more standardized so more people feel that they can contribute, and I wanted automated testing of style also so that people know when they put in a PR what I'm expecting.  I hope to make it more standard -- it'll never go to underscore_case etc., because it's just too far along to do that, but I'm hoping to get it going better.  Thanks for the offer to help.  Some of the ignore things we can turn on one by one and fix them once this is up to date.  All best

AlmogCohen:
@mscuthbert Just wanted to say in case you don't know about it - There is a way in GitHub to "Squash and Merge" instead of just merging PRs. When using that feature, you can commit into master all the changes of the PR as just one neat commit.

In the case of this PR for example, once all the linting is in place, you could end up with only one commit to master. In that way, the commits on master seem clearer and make more sense. Some people like this practice.

For example, this PR maybe added 100 commits to the master branch, which adds lots of noise and not much value of their own. I think you got the point, if that interests you.

There is a value to lower the number of commits as well not just from clarity - less commits also mean smaller git repo. The size of this repo is already huge :) Once the branch was squashed and merged into master, it can be deleted and repo stays nice and clean.

All of this is just a recommendation anyway. Thanks for everything!

AlmogCohen:
Squash and Merge is a nice practice, especially when a PR is full of so much "trial and test" commits which will add no much value on master

PR #471, opened 2019-11-06
Allow dotted tuplet notes when looking for duration candidates

AlmogCohen:
Before this change music21 did not support dotted notes within tuplets when constructing notes based on their quarterLength. This change allow up to 1 dot when looking for the best represation candidate.

One unittest was added to demonstrate that the original issue was indeed fixed. I'd be happy to make any changes to get these changes into music21.

Thanks!

### DIFF
--- a/music21/duration.py
+++ b/music21/duration.py
@@ -51,7 +51,6 @@ import unittest
 import copy
 
 from collections import namedtuple
-from fractions import Fraction
 
 from music21 import prebase
 
@@ -70,7 +69,7 @@ environLocal = environment.Environment(_MOD)
 DENOM_LIMIT = defaults.limitOffsetDenominator
 
 POSSIBLE_DOTS_IN_TUPLETS = [0, 1]
-DOT_LENGTH_MULTIPLAYER = Fraction(3, 2)
+DOT_LENGTH_MULTIPLAYER = fractions.Fraction(3, 2)
 
 _inf = float('inf')
 
--- a/music21/duration.py
+++ b/music21/duration.py
@@ -456,11 +456,10 @@ def quarterLengthToTuplet(qLen,
             qLenBase = opFrac(typeValue / float(i))
             # try multiples of the tuplet division, from 1 to max-1
             for m in range(1, i):
-                for number_of_dots in POSSIBLE_DOTS_IN_TUPLETS:
-                    dot_multiplayer = fractions.Fraction(common.dotMultiplier(number_of_dots))
-                    qLenCandidate = qLenBase * m * dot_multiplayer
+                for numberOfDots in POSSIBLE_DOTS_IN_TUPLETS:
+                    qLenCandidate = qLenBase * m * fractions.Fraction(common.dotMultiplier(numberOfDots))
                     if qLenCandidate == qLen:
-                        tupletDuration = durationTupleFromTypeDots(typeKey, number_of_dots)
+                        tupletDuration = durationTupleFromTypeDots(typeKey, numberOfDots)
                         newTuplet = Tuplet(numberNotesActual=i,
                                            numberNotesNormal=m,
                                            durationActual=tupletDuration,
--- a/music21/duration.py
+++ b/music21/duration.py
@@ -51,6 +51,7 @@ import unittest
 import copy
 
 from collections import namedtuple
+from fractions import Fraction
 
 from music21 import prebase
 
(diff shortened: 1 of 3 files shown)


music21/duration.py
```diff
@@ -68,6 +68,9 @@
 
 DENOM_LIMIT = defaults.limitOffsetDenominator
 
+POSSIBLE_DOTS_IN_TUPLETS = [0, 1]
+DOT_LENGTH_MULTIPLAYER = fractions.Fraction(3, 2)
```
mscuthbert: spelling:  DOT_LENGTH_MULTIPLIER
but see below also...

music21/test/testStream.py
```diff
@@ -8009,6 +8009,15 @@ def testGetElementAfterElement(self):
     #     cLast = m10.notes[-1]
     #     self.assertEqual(cLast.expressions, [])
 
+    def testDottedTupletNotes(self):
```
mscuthbert: this test should be in duration.Duration not here, there is no need to create a Measure object to test it.

music21/duration.py
```diff
@@ -454,15 +457,17 @@ def quarterLengthToTuplet(qLen,
...
-                if qLenCandidate == qLen:
-                    tupletDuration = durationTupleFromTypeDots(typeKey, 0)
-                    newTuplet = Tuplet(numberNotesActual=i,
-                                       numberNotesNormal=m,
-                                       durationActual=tupletDuration,
-                                       durationNormal=tupletDuration,)
-                    post.append(newTuplet)
-                    break
+                for number_of_dots in POSSIBLE_DOTS_IN_TUPLETS:
+                    dot_multiplayer = DOT_LENGTH_MULTIPLAYER ** number_of_dots
```
mscuthbert: spelling, but this is not the correct formula for dot length multiplier -- for instance 2 dots would give 9/4 instead of correct 7/4 -- use the existing function in `duration` to get the multiplier for an arbitrary number of dots.  Feel free to check up to 2 or 3 dots.

music21/test/testStream.py
AlmogCohen: I'll just note here that I'm a bit confused regarding the structure of the unit test classes and their locations. I'm used to projects with separate file/folder for unit tests.

I say that while saying that the 91% coverage now is already impressive.

Thank you for everything, hope that the feedback is taken in a good way

AlmogCohen: I'll put this test in the currently right place anyway

music21/duration.py
AlmogCohen: I used the `common.dotMultiplier` function for dots calculation. Added support double 0,1,2 dots. Didn't had any unit test example for a triple dotted tuplet note in mind, so I kept the 3 dots out of the picture for now.

AlmogCohen:
Changes made. I've moved the unit test to the proper place and added more cases to make sure we cover everything well. Added support for 2 dots as well (with a unit test case demonstrating that!)

Thanks!

AlmogCohen:
It seems that the hunch I was following was right. Allowing double dot for tuplet notes pretty much broke a million other tests across the package (As you can see on Travis for previous commits). For now, I simply remove the double dot tuplet support until someone will be willing to take it on :)

AlmogCohen:
I've also added the `Septuplet` term to the code and fixed some tests that assumed another name for it. Please also make sure if you can that the `braille` unit test I've changed is still correct.

Thanks! Sorry for all the comments here :)

music21/test/testStream.py
mscuthbert: A lot of standard testing and documenting tools available today didn't exist when music21 was started, so this is why it does some things idiosyncratically -- but generally test cases called Test are what are run when testing the system.  TestExternal is where anything that has an external effect (playing sound, opening musescore, etc.) goes, and TestSlow is something that's too slow to test with every push but that needs to pass before a release.

music21/duration.py
mscuthbert: 3 dots is too much for people to expect that we'll get it right, so that's fine.

music21/duration.py
```diff
@@ -458,7 +457,7 @@ def quarterLengthToTuplet(qLen,
             # try multiples of the tuplet division, from 1 to max-1
             for m in range(1, i):
                 for number_of_dots in POSSIBLE_DOTS_IN_TUPLETS:
-                    dot_multiplayer = DOT_LENGTH_MULTIPLAYER ** number_of_dots
+                    dot_multiplayer = fractions.Fraction(common.dotMultiplier(number_of_dots))
```
mscuthbert: still have this typo "multiplier" not "multiplayer" * and in next line.

mscuthbert:
Closing and reopening to get new checks run.

AlmogCohen:
Spelling issues fixed, some underscore style changed to camleCase

I'd recommend "Squashing and Merging" this PR into master, as most of the commits here has no special meaning. This whole PR can be accepted as one-unit change. One commit to master (Which can also have a meaningful name, not just "merge blablbablabla")

Thank you!!!

PR #478, opened 2019-11-12
Tsv roman tsv

MarkGotham:


mscuthbert:
Have you merged in the latest master?  There are additional github actions added last week that aren't being run in this PR.  Thanks!

MarkGotham:
As far as I can see, I have the latest master and am 'ahead by two commits' [not behind by any]. Are actions handled differently (i.e. not transferred with the fork / clone?). Happy to follow any new instructions there might be.

mscuthbert:
Ah! I had the actions set up to run only on pushes.  They now run on PRs also.  I will close and reopen this request and hopefully they'll run.  Thanks!

music21/romanText/tsvConverter.py
```diff
@@ -6,7 +6,7 @@
 # Authors:      Mark Gotham
 #
 # Copyright:    Copyright © 2019 Michael Scott Cuthbert and the music21 Project
-# License:      LGPL or BSD, see license.txt
+# License:      BSD, see license.txt
 # ------------------------------------------------------------------------------
 '''
 Convertor for parsing the tabular representations of harmonic analysis such as the
```
mscuthbert: Spelling error on Converter

mscuthbert:
and then mysteriously, it started running again.  Will give it a bit longer to see if it straightens itself out.

MarkGotham:
Some lint fails but no specific instructions. PyCharm flags up single quotes and camel case which both contravene PEP8, but are music21 standard.

music21/romanText/tsvConverter.py
MarkGotham: Isn't that American English? Here I am trying to fit in ;)

mscuthbert:
> Some lint fails but no specific instructions. PyCharm flags up single quotes and camel case which both contravene PEP8, but are music21 standard.

could you fix the lint and flake errors and then we'll be able to get in.  thanks!  sorry for delay -- i only get informed on commits that pass tests.

MarkGotham:
Done and done. Thanks for the reminder!

PR #479, opened 2019-11-12
Trill Fixer for OMR 

jcsands123:
Trill Fixer detects where a trill should be in music generated using OMR when compared against music generated from MIDI using the new Trill Recognizer. And inserts trill on appropriate omr note. 

Changes hasher to no longer split ties and only use note references in source and target stream, not root reference. This enables not in place fixing with Ornament Fixer.

### DIFF
--- a/music21/alpha/analysis/__init__.py
+++ b/music21/alpha/analysis/__init__.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-__all__ = ['aligner', 'fixer', 'hasher', 'search',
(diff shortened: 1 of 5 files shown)


mscuthbert:
Hi Janelle -- could you update from master one more time and push to this branch so I can see it with our new testing system which wasn't properly configured a minute ago.  Thanks!

mscuthbert:
Never mind!  It's working now, but there are some style-guide errors, etc. that need to be fixed in the unsuccessful checks.  Thanks!

PR #492, opened 2019-11-24
Turn Fixer

jcsands123:
Turn Fixer detects where a turn should be in music generated using OMR when compared against music generated from MIDI using the new Turn Recognizer. And inserts turn on appropriate omr note.

### DIFF
--- a/music21/alpha/analysis/fixer.py
+++ b/music21/alpha/analysis/fixer.py
@@ -10,6 +10,7 @@
 # ------------------------------------------------------------------------------
 import unittest
 from copy import deepcopy
+from typing import Optional, TypeVar
 
 from music21.alpha.analysis import aligner
 from music21.alpha.analysis import ornamentRecognizer
@@ -20,6 +21,8 @@ from music21 import interval
 from music21 import note
 from music21 import pitch
 from music21 import stream
+
+_T = TypeVar('_T')
 
 
 class OMRMidiFixer:
@@ -316,7 +319,6 @@ class EnharmonicFixer(OMRMidiFixer):
             return True
         return False
 
-
 class OrnamentFixer(OMRMidiFixer):
     '''
     Fixes missed ornaments in OMR using expanded ornaments in MIDI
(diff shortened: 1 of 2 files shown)


music21/alpha/analysis/fixer.py
```diff
@@ -316,17 +319,111 @@ def intervalTooBig(self, midiRef, omrRef, setint=5):
...
+        :param show: Whether to show results
+        :param inPlace: Whether to make changes to own omr stream or
+            return a new OrnamentFixer with changes
+        '''
+        changes = self.changes
+        sa = None
+        omrNotesLabeledOrnament = []
+        midiNotesAlreadyFixedForOrnament = []
+
+        if not inPlace:
```
mscuthbert: Let's think through this in a bit more depth -- is a OMRFixer of any sort the best place to do Stream Alignment?  I can understand some of the pros and con.   A big pro -- the fixer will know exactly what it needs to hash in order to do its job most efficiently.  A big con -- what is the system that takes two scores and aligns and improves them called?  Is it an XXAligner, a XXFixer, a XXDetector?  Generally one class of system should be running the whole show and I can't figure out at this point which one that is.  I thought it was an Aligner, but now I'm not sure.

mscuthbert: (that was a comment on the whole system in general, not on the refactor to move out of TrillFixer, which I 100% agree with).

music21/alpha/analysis/ornamentRecognizer.py
```diff
@@ -150,24 +146,243 @@ def recognize(self) -> Union[bool, expressions.Trill]:
...
+        so A goes down to G, G#, G flat, G##, etc
+
+        Returns: False if not possible or the Turn/Inverted Turn Expression
+        '''
+        # number of notes/ duration of notes ok
+        if len(busyNotes) != 4:
+            return False
+
+        if simpleNotes:
+            eps = .1
```
mscuthbert: leading zero 0.1 (fix next push, don't make a new one just for this)

music21/alpha/analysis/ornamentRecognizer.py
```diff
@@ -150,24 +146,243 @@ def recognize(self) -> Union[bool, expressions.Trill]:
...
+                return False
+
+        # pitches ok
+        if busyNotes[1].pitch.midi != busyNotes[3].pitch.midi:
+            return False
+        if simpleNotes and simpleNotes[0].pitch.midi != busyNotes[1].pitch.midi:
+            return False
+
+        # intervals ok
+        acceptableIntervals = [interval.Interval("M2"), interval.Interval("M-2"),
```
mscuthbert: This creates 6 new interval objects for each possible set of notes.  Run `%timeit` on creating an interval object and see if this might be a problem or not.  Is there a more efficient way of doing this?

music21/alpha/analysis/fixer.py
jcsands123: Yes agreed to give this some more thought!

music21/alpha/analysis/ornamentRecognizer.py
jcsands123: Couldn't figure out how to run timeit successfully, but I realized this is inefficient. Code is changed to initialize acceptable intervals only when fixer instance initialized. Is this enough of an improvement or were you thinking something even better?

jcsands123:
Changed code to address all 👍reacted comments.

mscuthbert:
Looks good -- in the future can you not make force-pushing the default?  It overwrites history so I can't see what changed in just the last two commits.  Thanks!

PR #509, opened 2020-01-26
Fix Chord.removeRedundantPitchClasses and Chord.removeRedundantPitchNames docs

ryaanahmed:
These two docstrings said that if `inPlace` is True, `None` is returned, but both methods actually return the list of deleted pitches.


mscuthbert:
Thanks!

PR #511, opened 2020-01-27
MANIFEST.in: specify the correct path to prune .mypy_cache

dhgutteridge:
Source distributions and subsequent downstream packaging results were including .mypy_cache contents.


mscuthbert:
Thank you!

PR #513, opened 2020-02-04
remove corpus, moved to music21-corpus

gesellkammer:
This PR splits the corpus from the main distribution of music21 as a subpackage `music21-corpus`. When both packages are installed, it can still be accessed via `from music21 import corpus`. 
The address of the new subpackage is https://github.com/gesellkammer/music21-corpus. If accepted, that repo could be transferred here.


mscuthbert:
Already rejected as noted.  Make another fork if you want it.  Having batteries included is a core music21 value.

PR #528, opened 2020-04-18
Fix formatting in music21.analysis.neoRiemannian docs

aspiers:
A few blank lines were missing in between documentation prose and the following code samples, which was causing the code samples to be incorrectly rendered as part of the prose.

An example of this can currently be seen here:

https://web.mit.edu/music21/doc/moduleReference/moduleAnalysisNeoRiemannian.html#music21.analysis.neoRiemannian.isNeoR


mscuthbert:
Thank you!

Issue #348, opened 2018-12-01
show() doesn't work in Windows 10


danshorstein:
show() is not working. Using Windows 10. Getting "SubConverterFileIOException: png file of xml not found. Or file >999 pages?".

I can see the xml tmep files are created, but no png temp files are being created. 

I've installed MuseScore using the microsoft app store, which is the MuseScore recommended way to install. However this puts the musescore.exe in a hidden folder under c:\program files\windowsapps. I was able to locate the path and set "musicxmlPath" and "musescoreDirectPNGPath" to the path, and am still getting an error when trying to run show(). I also installed lilypad and when I run show('lily'), that works.

Wind4Greg:
I'm seeing the same issue. Brand new Music21 and MuseScore install, using Python 3.6.6 (Anaconda), with Jupyter Notebooks. Ran some tests with `show()` and saw the following for this test case:

```python
temp = "text" # works
# "midi" works
#  "musicxml" opens up and shows in MuseScore
# "vexflow" showed it in a new tab, 
# "musicxml.png" gets the error condition mentioned above
f = note.Note("F5")
f.show(temp)
```

Note that temporary music XML files are being created in the *scratch* directory specified in:

```
<preference name="directoryScratch" value="C:\tmp" />
```
After `f.show("musicxml.png")` or `f.show()` is called a new MusicXML file shows up in the *scratch* directory. If I run MuseScore in command line mode show below it correctly turns the MusicXML file into a PNG.

```
$ "C:\Program Files (x86)\MuseScore 2\bin\MuseScore.exe" tmp4hkcsxd7.xml -o temp.png
```

If you can point me to the Python file that calls MuseScore I can try to figure out what's going wrong.

Cheers

Greg B.

LucasLoerch:
I am also having trouble with the show() function. I have Windows 10 64bit, Anaconda3/Python 3.7 and MuseScore3 all installed recently. I already set the musicxmlPath to the correct folder containing MuseScore. When I run the show() command I get the following error:
OSError: [WinError 123] Die Syntax für den Dateinamen, Verzeichnisnamen oder die Datenträgerbezeichnung ist falsch: 'C:\\Users\\Lucas Loerch\\C:\\Program Files (x86)\\MuseScore 2\\MuseScore.exe'
(in english: the syntax for the file name, directory name or volume name are wrong).

I also tried to uninstall MuseScore and install Finale. I adapted the musicxmlPath accordingly, but the error remains the same.

[...]

PR #532, opened 2020-04-29
Use subprocess.run instead of os.system for PNG generation

vug:
On Windows 10, inside Jupyter Notebook environment, `show` method with PNG output does not work because of the double quotes placed around the arguments of the called command.

```
import music21
s = music21.corpus.parse('bach/bwv65.2.xml')
s.show('ipython.musicxml.png')
```

throws `SubConverterFileIOException: png file of xml not found. Or file >999 pages?` error. Because PNG file was not generated in the scratch folder. This problem and its solution were mentioned in issue #348

The reason I found is that the command for PNG generation puts double quotes around arguments, and `os.system` does not like them. Below is the system call made by above `show` method, stored in `musescoreRun` variable..

```
os.system('"C:\\Program Files\\MuseScore 3\\bin\\MuseScore3.exe" "c:\\tmp\\tmpsu4ako8g.xml" -o "c:\\tmp\\tmpsu4ako8g.png" -T 0  -r 200')
```

If I remove the double quotes manually and run the same command manually, PNG generation happens without a problem. ^_^

```
os.system('"C:\\Program Files\\MuseScore 3\\bin\\MuseScore3.exe" c:\\tmp\\tmpsu4ako8g.xml -o c:\\tmp\\tmpsu4ako8g.png -T 0  -r 200')
```

Instead of changing the content of `musescoreRun` variable by removing the double quotes around arguments, the easier and better solution is to switch using `subprocess` to do the system calls, instead of using the old `os.system`.

Replacing the call for PNG conversion done via `os.system` with `subprocess` fixes the  problem. (Also this solution was suggested by GaetanBaert in the issue). Below is a screenshot after I made the changes in this pull request to my local copy of music21 and ran same cell to produce PNG output.

music21-png-output-fix-screenshot

### DIFF
--- a/music21/converter/subConverters.py
+++ b/music21/converter/subConverters.py
@@ -900,16 +900,12 @@ class ConverterMusicXML(SubConverter):
         fpOut = fp[0:len(fp) - 3]
         fpOut += subformatExtension
 
-        musescoreRun = '"' + str(musescorePath) + '" "' + fp + '" -o "' + fpOut + '" -T 0 '
+        musescoreRun = [str(musescorePath), fp, '-o', fpOut, '-T', '0']
         if 'dpi' in keywords:
-            musescoreRun += ' -r ' + str(keywords['dpi'])
+            musescoreRun.extend(['-r', str(keywords['dpi'])])
 
         if common.runningUnderIPython():
-            musescoreRun += ' -r ' + str(defaults.ipythonImageDpi)
-
-        platform = common.getPlatform()
-        if platform == 'win':
-            musescoreRun = '"' + musescoreRun + '"'
+            musescoreRun.extend(['-r', str(defaults.ipythonImageDpi)])
 
         storedStrErr = sys.stderr
         fileLikeOpen = io.StringIO()
--- a/music21/converter/subConverters.py
+++ b/music21/converter/subConverters.py
@@ -22,6 +22,7 @@ import base64
 import io
 import os
 import pathlib
+import subprocess
 import sys
 import unittest
 
@@ -913,7 +914,7 @@ class ConverterMusicXML(SubConverter):
         storedStrErr = sys.stderr
         fileLikeOpen = io.StringIO()
         sys.stderr = fileLikeOpen
-        os.system(musescoreRun)
+        subprocess.run(musescoreRun)
         fileLikeOpen.close()
         sys.stderr = storedStrErr
 
--- a/music21/converter/subConverters.py
+++ b/music21/converter/subConverters.py
@@ -914,7 +914,7 @@ class ConverterMusicXML(SubConverter):
         storedStrErr = sys.stderr
         fileLikeOpen = io.StringIO()
         sys.stderr = fileLikeOpen
-        subprocess.run(musescoreRun)
+        subprocess.run(musescoreRun, check=False)
         fileLikeOpen.close()
         sys.stderr = storedStrErr
 


vug:
I've never contributed to this project before. This is going to be my first time. If I have to run certain unit tests before submitting a pull request please let me know! This change was so short, only two lines, hence I did it via the GitHub online editor on the browser.

vug:
The linter has some complaints, but looks like they are not related to the line that are changed in this request. (There was a related one and I fixed it in a second commit.) ^_^

mscuthbert:
Thank you -- this is something I've wanted to have in since dropping Python 2 support.  I think that the PyLint errors are from the default version of Pylint being upgraded.  I have PR #533 fixing them; so I'll merge that first and then look at this.

mscuthbert:
Inside Jupyter, does a simple `.show()` still generate the image?

mscuthbert:
This is great -- but shouldn't subprocess.run take in a list of arguments rather than a single one with everything in a string; I thought that this would have been a larger refactor.

vug:
> Inside Jupyter, does a simple `.show()` still generate the image?

Yes. I reset to default settings, and just set my MuseScore path only.

```
import music21

user_settings = music21.environment.UserSettings()
user_settings.restoreDefaults()
MS_PATH = 'C:/Program Files/MuseScore 3/bin/MuseScore3.exe'
user_settings['musicxmlPath'] = MS_PATH
user_settings['musescoreDirectPNGPath'] = MS_PATH
s = music21.corpus.parse('bach/bwv65.2.xml')
```

Then call `s.show()`. And I get the rendered image shown in the Notebook.

music21-simple-show-example

I also tried `s.show('musicxml')` which opens the score in MuseScore and `s.show('midi')` puts the mudcube MIDIPlayer.

vug:
> This is great -- but shouldn't subprocess.run take in a list of arguments rather than a single one with everything in a string; I thought that this would have been a larger refactor.

https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess.args Here is what the official document for `subprocess.run(args)` says about `args`: "The arguments used to launch the process. This may be a list or a string.".

Providing a list as argument to `subprocess` call is a more common choice I see in the wild. Here, I went for the least amount of code change. :-) I didn't even touch the other `os.system` call in the same file, following the principle "if it isn't broken, don't fix it" :-P

mscuthbert:
I guess my thought is that if we're trying to fix the problem of spaces and quotation marks in file paths, don't we need to do the list version to fix it?

If we're going for least changes to help most cases, then do nothing, -- my system has no spaces.  :-)

I'm going to merge this, don't worry, because I want to get subprocess.run() in for every place we're doing os.system, but let's generalize a little bit more so the GNU/Linux people don't bite us.

vug:
> I guess my thought is that if we're trying to fix the problem of spaces and quotation marks in file paths, don't we need to do the list version to fix it?

My MuseScore path, `C:/Program Files/MuseScore 3/bin/MuseScore3.exe`, has a space and providing a string to `subprocess` works fine on my machine. I guess, because the subconverter code first wraps the executable path and each of its the arguments with double quotes. (See examples in my first comment. It was those wrapping double quotes on the arguments that `os.system` doesn't like and looks like `subprocess.run` doesn't mind them.)

> If we're going for least changes to help most cases, then do nothing, -- my system has no spaces. :-)

That's a clever solution :-P
 
> I'm going to merge this, don't worry, because I want to get subprocess.run() in for every place we're doing os.system, but let's generalize a little bit more so the GNU/Linux people don't bite us.

If you want me to replace both `os.system` calls with `subprocess.run`, and provide a list argument instead of a string, I can do that too! But I don't have a Linux machine to test. I work on Windows. So please test it on Linux and OSX before merging my change. ^_^

(Let me figure out how to download my patch to my machine and update this pull request from there now...)

mscuthbert:
Don't change any other place but here to `subprocess` we'll do more than one small change so that if it breaks, it affects the fewest number of people.  I think it'll work/  I'm testing on MacOS.

vug:
OK! I'll switch to subprocess with list for PNG rendering only, and will update the PR.

vug:
OK! (Looks like one need to push a commit to their own fork to update a PR)
I switched to giving a list to `subprocess.run` function and it worked fine on my Windows setup with MuseScore executable path that has a space in it. (I got the same screenshot as above.) ^_^

PR #533, opened 2020-04-29
Fix musicxml png output

mscuthbert:
Fix some bugs in musicxml.png output -- especially the >999 pages warning

Fix new pylint 1.50 issues 
Also update some Humdrum code


mscuthbert:
(Removing redundant code that is well tested can lower coverage)

PR #539, opened 2020-05-14
Add error messages for invalid chord roots and abbreviations

viknat:
Hey all, I noticed that error messages aren't helpful in cases where an invalid string is passed to ChordSymbol so I added a couple in this PR. It took me some debugging to figure out why ChordSymbol('Bb') was not working because the error I was getting was not informative -- so it took a while for me to realize that music21 expects - for flats in this case.

Do let me know your thoughts

### DIFF
--- a/music21/harmony.py
+++ b/music21/harmony.py
@@ -1806,7 +1806,7 @@ class ChordSymbol(Harmony):
                 st = prelimFigure.replace(m1.group(), '')
             else:
                 raise ValueError(f'Chord {prelimFigure} does not begin '
-                                  'with a valid root note')
+                                 + 'with a valid root note.')
 
         if root:
             self.root(pitch.Pitch(root))
@@ -1854,9 +1854,9 @@ class ChordSymbol(Harmony):
             try:
                 justInts = int(justInts)
             except ValueError:
-                raise ValueError(f'Invalid chord abbreviation "{st}"; see '
-                                  'music21.harmony.CHORD_TYPES for valid '
-                                  'abbreviations or specify all alternations')
+                raise ValueError(f'Invalid chord abbreviation {st!r}; see '
+                                 + 'music21.harmony.CHORD_TYPES for valid '
+                                 + 'abbreviations or specify all alterations.')
             if justInts > 20:   # MSC: what is this doing?
                 skipNext = False
                 i = 0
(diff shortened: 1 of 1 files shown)


music21/harmony.py
```diff
@@ -1804,8 +1808,8 @@ def _parseFigure(self):
                 # remove the root and bass from the string and any additions/omissions/alterations/
                 st = prelimFigure.replace(m1.group(), '')
             else:
-                raise ValueError  # This means that the given argument wasn't
-                # a proper chord name.
+                raise ValueError(f"Chord {prelimFigure} does not begin "
+                    f"with a valid root note")
```
mscuthbert, line 1812: single quotes; no f-string on second , and ensure that strings aline on consecutive lines.  e,g, : 

```
                raise ValueError(
                      f'Chord {prelimFigure} does not begin '
                      + 'with a valid root note'
                )
```

music21/harmony.py
```diff
@@ -1736,8 +1739,9 @@ def _getKindFromShortHand(self, sH):
             sH = sH[0:sH.index('omit')]
         if '#' in sH and sH[sH.index('#') + 1].isdigit():
             sH = sH[0:sH.index('#')]
-        if 'b' in sH and sH[sH.index('b') + 1].isdigit() and 'ob9' not in sH and 'øb9' not in sH:
-            # yuck, special exception
+        if ('b' in sH and sH.index('b') < len(sH) - 1 and
```
mscuthbert, line 1742: break before `and`

music21/harmony.py
```diff
@@ -113,6 +113,9 @@
     ('Tristan', ['1,#4,#6,#9', ['tristan']]),  # Y
 ])
 
+VALID_ABBRS = [abbr for chord, (notes, abbrs) 
+                    in CHORD_TYPES.items() for abbr in abbrs]
+
```
mscuthbert, line 118: This logic is obscure to me.  abbr is being used for two different variables at the same time.  nested comprehensions are something I avoid.  please rewrite as for loops.

music21/harmony.py
```diff
@@ -1853,7 +1857,8 @@ def _parseFigure(self):
             try:
                 justInts = int(justInts)
             except ValueError:
-                raise ValueError  # Not a properly formatted chord, ignore it
+                raise ValueError(f"Invalid chord abbreviation '{st}'; must be one of "
+                    f"{VALID_ABBRS} or specify all alterations")
```
mscuthbert, line 1861: as above.

But I don't think that printing out VALID_ABBRS is the right thing to do.  it'll be far too long.  Maybe tell users:  "See music21.harmony.VALID_ABBRS for a list of all abbreviations."  ?

viknat: Makes sense -- I removed VALID_ABBRS as the info is already in CHORD_TYPES anyway and changed the error message to reference CHORD_TYPES instead.

PR #542, opened 2020-05-17
Add ipython SVG support + fix no show images

bsdz:
This PR adds SVG support in Jupyter and fixes an issue where images would not display when generated from Musescore due to a SubConverterFileIOException.

The fix required switching to subprocess module and supplying environment variable QT_QPA_PLATFORM being set to "offscreen".

SVG support in ipython requires the latest version of musescore which is in alpha currently. This is because there was a bug in the SVG rendering was was resolved in recently merged PR for https://github.com/musescore/MuseScore/pull/5970.

The following is possible with SVG support:

image

The images can be scaled to any size to assist in readability.

With SVG color support one can use Jupyter dark theme like so:

image

### DIFF
--- a/music21/converter/subConverters.py
+++ b/music21/converter/subConverters.py
@@ -24,6 +24,9 @@ import os
 import pathlib
 import sys
 import unittest
+import subprocess
+import locale
+import re
 
 from music21 import common
 from music21 import defaults
@@ -344,7 +347,7 @@ class ConverterIPython(SubConverter):
         helperConverter.setSubconverterFromFormat(helperFormat)
         helperSubConverter = helperConverter.subConverter
 
-        from IPython.display import Image, display, HTML  # @UnresolvedImport
+        from IPython.display import Image, display, HTML, SVG  # @UnresolvedImport
 
         if helperFormat in ('musicxml', 'xml', 'lilypond', 'lily'):
             # hack to make musescore excerpts -- fix with a converter class in MusicXML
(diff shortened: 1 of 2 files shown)


music21/converter/subConverters.py
```diff
@@ -364,11 +367,30 @@ def show(self, obj, fmt, app=None, subformats=None, **keywords):  # pragma: no c
...
+                            # musescore appears to scale svg as 3.6 x png scale
+                            svgScale = keywords.get("svgScale", float(environLocal["ipythonSVGScale"]))
+                            if svgScale:
+                                def _scale(mo):
+                                    begin, w, h, end = mo.groups()
+                                    w, h = float(w)*svgScale, float(h)*svgScale
+                                    return f'{begin}width="{w}px" height="{h}px"{end}'
+                                svgData = re.sub(
+                                    r'^(<svg[^>]*)width\s*=\s*"([\d.]+)px"\s+height\s*=\s*"([\d.]+)px"([^>]*>)',
+                                    _scale, svgData, flags=re.MULTILINE)
```
mscuthbert, line 385: All this is too fragile -- it depends on width preceding height, etc.  -- it should use ElementTree or another built in xml parser to extract and change the values.  Please don't define a function inside an if statement.

bsdz: I admit it does look fragile but that said the musescore C++ always generates the attributes in that order. See

https://github.com/musescore/MuseScore/blob/9e456031cf1e843aab2a96322d6c3efbd92816a2/mscore/svggenerator.cpp#L1036

That said, I can pick those attributes out one-by-one although I won't use a full blown xml parser as this would slow things down and potentially cause headaches with namespaces etc. Let me know if you will be happy with that?

I'm happy to move the function out of the "if" scope although that will mean it will become a partial function.

music21/converter/subConverters.py
```diff
@@ -364,11 +367,30 @@ def show(self, obj, fmt, app=None, subformats=None, **keywords):  # pragma: no c
                 fp = helperSubConverter.write(s, helperFormat,
                                               subformats=helperSubformats, **keywords)
 
-                if helperSubformats[0] == 'png':
+                if helperSubformats[0] in ('png', 'svg'):
                     if not str(environLocal['musescoreDirectPNGPath']).startswith('/skip'):
                         ipo = ipythonObjects.IPythonPNGObject(fp)
                         # noinspection PyTypeChecker
```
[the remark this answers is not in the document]
bsdz: I don't quite understand your point here. Perhaps you can elaborate?

music21/converter/subConverters.py
bsdz: @mscuthbert just awaiting your reply to my comment

mscuthbert: It doesn't matter what MuseScore does today -- if they change in the future or there are successor tools, etc., I'm the one who is going to need to read this regexp and figure out how to fix it 8 years from now.  I've learned from experience, I only put into music21 code I can read and I can fix, because eventually I'll have to.  :-)

music21/converter/subConverters.py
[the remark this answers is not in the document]
mscuthbert: What was a few lines inside of an if statement is now a large and complex nested code that isn't fully tested.  Please factor out and show that the tests work.

music21/converter/subConverters.py
bsdz: Then I'm sorry about this. I don't really have time to work an xml parser into this. I'll leave the PR here for now for someone else to pick up if interested.

Best wishes on your project.

PR #548, opened 2020-05-27
Respect quantizePost=False when parsing raw MIDI

### LINKED ISSUE #192: Provide easy way to disable MIDI quantization
In processing a string quartet, I get the same offset values for the C and Bb in measure 2, beat 3 of the first violin in the snapshot below. 
image
I'm running into this problem in music21 3.1.0, but not in 2.1.2. The way I use music21 to determine the offset of a note is with the following logic where `event` has already been filtered to make sure it has a 'Note', 'Rest', or 'Chord' type:

```
for y in event.contextSites():
    if y[0] is part:
        return y[1]
```

When I make a pandas.Series of these note, rest, or chord objects with their offsets as determined by the code above as the index, these are the first few values for the first violin:

```
0.000000               <music21.note.Rest rest>
3.000000                  <music21.note.Note E>
4.000000                  <music21.note.Note F>
5.500000                  <music21.note.Note A>
6.000000                  <music21.note.Note C>
6.000000                 <music21.note.Note B->
6.333333               <music21.note.Rest rest>
6.500000                  <music21.note.Note A>
7.000000                 <music21.note.Note B->
7.500000                  <music21.note.Note D>
```

Should I be using contextSites() differently in 3.1.0, or is there something else I'm missing? Just let me know if you'd like me to send the file. Thanks.

### LINKED ISSUE #546: Let MIDI converter read keywords when parsing raw data
Whether keywords are respected in a call to converter.parse() depends on whether the input is a filepath or raw data. One downstream effect is in MIDI parsing.

To reproduce:

```
dirLib = common.getSourceFilePath() / 'midi' / 'testPrimitive'
fp = dirLib / 'test09.mid' # quantizePost must be False to catch the 12th note in m.9
with fp.open('rb') as f:
  data = f.read()

streamFp = converter.parse(fp, quantizePost=False)
m9Fp = streamFp.measures(9,9)
len(m9Fp.elements[0].elements[-1].elements[0].elements) # Expected result 12
>>> 12

streamData = converter.parse(data, quantizePost=False)
m9Data = streamData.measures(9,9)
len(m9Data.elements[0].elements[-1].elements[0].elements) # Expected result 12
>>> 11
```

Suggested fix:
Adding **keywords here and in the subsequent calls in the above test case:
https://github.com/cuthbertLab/music21/blob/f65327fae4a8064743d2776fc32a49f9d39f2bf7/music21/converter/__init__.py#L608

I was able to pass **keywords all the way down to midi.translate.midiStringToStream where the specific keyword in this case converted to an argument, and got this test case passing. But each subconverter would have to be tested to make sure they don't crash on unexpected keyword arguments. I may have time to put this together soon. This would lend itself nicely to tests.

jacobtylerwalls:
Fixes #546, fixes #192 and adds test method converter.testParseMidiNoQuantize()

New example MIDI file containing a 1/4 measure with exactly 3 16th notes and 2 32nd notes


mscuthbert:
Closing and reopening to rerun tests since there were potentially MIDI keyword conflicts with a recently merged PR.

mscuthbert:
Looks really great, but couple of comments:

* forceSource and not use pickle slow down the test suite considerably.  Have you gotten the parsing down to the minimum level of using forceSource to not cause a problem?  I imagine that only the skipping quantize parse needs it.

* How will anyone know that they can pass in this new keyword?  Documentation needs to be updated as well.  Don't worry about the User's Guide, but at least the MIDI Converter and a few other places need it.

Thanks!

jacobtylerwalls:
Sure thing, will definitely get on the documentation. With luck it could cut down inquires about 32nd notes.

RE: pickles -- so long as I am testing the same file and parsing it two ways, I think I would have to force the source each time. Otherwise if the first call is forceSource=False, quantizePost=True, upon running the test a second time, wouldn't it just load in the result of the second parsing?

The motivation was to avoid a false positive test result if behavior changed elsewhere, but I could check for that more directly by just asserting that every value in defaults.quantizationQuarterLengthDivisors is less than 8, so if that behavior ever changes then this test will complain that it must be rewritten. Then I could just call parse once and let it use a pickle.

jacobtylerwalls:
Documentation fixes documentation request in #592

jacobtylerwalls:
On second thought I definitely still need to forceSource even when parsing the test file one time, otherwise when the test passes once it will pass forever -- need to ensure I'm testing that the pipeline for **keywords remains intact all the way down from converter.parse()

jacobtylerwalls:
• Noticed MIDI file had an incorrect note_off value that was causing M21 to fill a gap with an extraneous rest. Didn't affect my tests, but want to contribute a clean test file.

• Noticed two more places to document the quantization keywords.

mscuthbert:
Thanks -- good explanation for why you need forceSource in both these cases.  Thanks for trying it.  Will review new changes today!

mscuthbert:
This is a model PR.  Thank you Jacob!  Will definitely make MANY people happy!

PR #553, opened 2020-06-01
Find mid-measure and end of measure hidden rests from <forward> elements

### LINKED ISSUE #444: Hidden rests from <forward> tags not filled with rests by xmlToM21
I have an xml source with hidden rests implied from forward tags in Voice 2, and the import parsing leaves me with gaps if the forward tag occurs midmeasure. See the example below: Voice 2 consists of **forward/note/forward**. The result of parsing Voice 2 yields **rest, note** instead of **rest, note, rest**.

https://github.com/cuthbertLab/music21/blob/74b342816e6da686f711b6ad334fdbaf05812ce0/music21/musicxml/xmlToM21.py#L2317

Is there a reason fillGaps is not set to True here? Is it so that the "filling" of incomplete final measures is avoided?

I can loop through every measure with voices and call makeRests(fillGaps=True, hideRests=True) on every voice, but I am curious why xmlToM21 is making its current assumptions.

Example measure tag attached
[hiddenRestsVoice.txt](https://github.com/cuthbertLab/music21/files/3648729/hiddenRestsVoice.txt)

jacobtylerwalls:
Resolves #444

Summary: leading hidden rests were being found because makeRests() creates them in all cases, but mid-measure and end of measure hidden rests from `<forward>` elements were not created.

fillGaps handles the midmeasure case, and using refStreamOrTimeRange to compare with the entire measure stream handles the end case.

### DIFF
--- a/music21/musicxml/testPrimitive.py
+++ b/music21/musicxml/testPrimitive.py
@@ -17808,6 +17808,76 @@ tremoloTest = """<?xml version="1.0" encoding="UTF-8" standalone="no"?>
 </score-partwise>
 """
 
+hiddenRests = """<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!DOCTYPE score-partwise PUBLIC "-//Recordare//DTD MusicXML 3.1 Partwise//EN" "http://www.musicxml.org/dtds/partwise.dtd">
+<score-partwise version="3.1">
+  <part-list>
+    <score-part id="P1">
+      <part-name print-object="no">MusicXML Part</part-name>
+    </score-part>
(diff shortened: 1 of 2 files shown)


mscuthbert:
gad -- i should've merged this first.  My apologies.  I think I get this fully now what it does, how it fixes something ,etc.  so if you could fixx the new conflicts, will happily merge.

jacobtylerwalls:
should be good now! (and there would have been merge conflicts either way, so no worries!)

mscuthbert:
Wonderful .  Thanks for the amazing work!

PR #562, opened 2020-06-16
parse <sound> and <staff> associated with <direction-type><words>

### LINKED ISSUE #159: Parse <sound> tag in MusicXML
Both in <measure> and <direction>

peterg98:
Partially fixes #159 

Issue: parser does not parse `<sound>` and `<staff>` in imported `<direction><direction-type><words>` elements. This fixes the issue by passing `staffKey` and the `<sound>` Element itself to `TextExpression`. The exporter then places these Elements inside `<direction>`.

### DIFF
--- a/music21/expressions.py
+++ b/music21/expressions.py
@@ -319,7 +319,7 @@ class TextExpression(Expression):
     classSortOrder = -30
     _styleClass = style.TextStyle
 
-    def __init__(self, content=None):
+    def __init__(self, content=None, staffKey=None, mxSound=None):
         super().__init__()
         # numerous properties are inherited from TextFormat
         # the text string to be displayed; not that line breaks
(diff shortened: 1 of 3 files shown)


music21/expressions.py
```diff
@@ -333,6 +333,10 @@ def __init__(self, content=None):
 
         # this does not do anything if default y is defined
         self.positionPlacement = None
+        # staff number of this TextExpression
+        self.staffKey = staffKey
+        # musicXML <sound> element that is in <direction>, if it exists
+        self.mxSound = mxSound
```
mscuthbert, line 339: These are MusicXML specific notions, so they'll need to be translated into a platform-agnostic format.  What will this do with, say, MIDI or MEI?

peterg98:
Indeed it's a quick fix for a small project I'm working on. I just got into this library so I'm not too familiar with the codebase.

PR #565, opened 2020-06-21
Let stripTies() recurse when deleting and searching

### LINKED ISSUE #266: music21.midi.realtime.play does not play ties correctly on OS X Sierra
First things first: This project is absolutely incredible. A couple of other developers and I have been building an application for the past couple months using music21 as the primary backend technology, and I cannot thank all of you enough for the enormous amount of work you all have put into this massive project. 

That said, I came across an issue recently that is a bit problematic for our particular use case. We've been using music21.midi.realtime.play in order to play back pieces that our users compose, and we ran into a strange problem involving playback of notes connected with ties. This simple python program is enough to replicate the issue: 
```
import music21

# Create the stream
s = music21.stream.Stream()
# Make two notes with a tie between them
note1 = music21.note.Note("C#4", 1.0)
note1.tie = music21.tie.Tie("start")
note2 = music21.note.Note("C#4", 2.0)
note2.tie = music21.tie.Tie("stop")
# Insert the notes in the stream
s.insert(0.0, note1)
s.insert(1.0, note2)
# Attempt to playback the stream
music21.midi.realtime.StreamPlayer(s).play() # 2 notes play instead of 1
# Check the score in a GUI to make sure the tie is present
s.show()
```

Correct me if I'm wrong, but I'm pretty sure the whole point of ties is to ensure a note keeps sounding beyond what a single note head can notate. Fortunately, it appears that Musescore and other GUI applications display the tie correctly, which means that the tie is indeed interpretable. However, this does not appear to be the case on playback - where only 1 continuous note should sound, 2 are played. 

I've only poked a little bit into the music21 source code - there's quite a lot of it - but this seems like something that should be doable with a little tooling around. One potential workaround could be storing semantic note duration (as it appears on the page) and playback note duration (as one would hear the piece when played) as separate values. I imagine this could get a little hairy when trying to create ties across separate Stream objects, though. I don't know how feasible it is in all cases, but it may be something to consider implementing. 

It also may very well be that this issue is local to my machine environment, and if so it's definitely a low priority bug that we can probably hack around. But if it's not unique to my architecture, it seems like something that could probably be investigated and improved upon. 

Thanks again for all your hard work! :)

Edit: If it helps, we're using the most recent version of music21, version 4.1.0.

### LINKED ISSUE #491: Stream.melodicIntervals() extra interval to start
I am trying to get the melodic interval sequence for a part in a Bach chorale, for example:
```
for chorale in music21.corpus.chorales.Iterator(1, 3):
  chorale.parts[0].melodicIntervals().show('txt')
  chorale.show()
```
Here I just want the melodic intervals for the top soprano line.
It seems that there is an extra interval to start – i.e. there's an extra P1 for the first two chorales and an extra P5 for the third one in the soprano line for this code.

What's happening? How can this be fixed? (there's the obvious workaround of skipping the first one if this is a known problem). Otherwise, please let me know if I'm misunderstanding something. Thanks!

Update: I see, the first interval seems to be the interval from going from the last note of the part to the first note. This seems a little unintuitive from looking at the examples from the documentation.

### LINKED ISSUE #514: Incorrect midi for ties
Hi! Using the a fresh music21 installation from PyPi I find that midi is generated incorrectly for ties.

Please see this example:
```
    s1 = stream.Stream()

    n = note.Note("C4", quarterLength=1.0)
    n.tie = tie.Tie("start")

    n2 = note.Note("C4", quarterLength=1.0)
    n2.tie = tie.Tie("stop")

    n3 = note.Note("C4", quarterLength=1.0)
    n4 = note.Note("C4", quarterLength=1.0)


    s1.append(n)
    s1.append(n2)
    s1.append(n3)
    s1.append(n4)

    # s1.show()
    s1.show('midi')
```

The output midi is incorrect and all notes are played one-by-one as quarters. This is how the midi looks like in MuseScore: WhatsApp Image 2020-02-04 at 14 16 13

This is how the output looks like in LogicX: WhatsApp Image 2020-02-04 at 14 16 36

**NOTE:** In the LogicX output I've shifted the first note up so you could see how they overlap. Originally all notes are C. In the LogicX output it seems like the first note is indeed generated with the tie, but the second one is generated as well while it shouldn't!

What do you think guys? Thanks in advance!

### LINKED ISSUE #568: stripTies() doesn't find tied notes consecutive in a voice but nonconsecutive in parent stream
`stripTies()` depends on tied notes being consecutive in the flattened version of the parent container, which is usually a `Part`:

https://github.com/cuthbertLab/music21/blob/ed66817fdbb4430f9de10dd63c32d088eb01ab0b/music21/stream/__init__.py#L6205-L6206

`stripTies()` can yield unexpected behavior when ties are consecutive in a voice but not in the part. For instance, another voice might initiate a tie before the first one concluded.

https://github.com/cuthbertLab/music21/blob/ed66817fdbb4430f9de10dd63c32d088eb01ab0b/music21/stream/__init__.py#L6213-L6216

To reproduce:
```
s = stream.Score()
p = stream.Part()
v1 = stream.Voice()
v2 = stream.Voice()

n1 = note.Note('A4')
n1.quarterLength = 2.0
n1.tie = tie.Tie('start')
n2 = note.Note('A4')
n2.quarterLength = 2.0
n2.tie = tie.Tie('stop')
n3 = note.Rest()
n3.quarterLength = 1
n4 = note.Note()
n4 = note.Note('D4')
n4.quarterLength = 1.0
n4.tie = tie.Tie('start')  # tie starts in v2 before tie in v1 stops
n5 = note.Note()
n5 = note.Note('D4')
n5.quarterLength = 2.0
n5.tie = tie.Tie('stop')

v1.append([n1,n2])
v2.append([n3, n4, n5])
p.append(v1)
p.insert(0, v2)
s.append(p)

stripped = s.stripTies()
stripped.show('t')
```
```
{0.0} <music21.stream.Part 0x7fc6a29cb250>
    {0.0} <music21.stream.Voice 0x7fc6a38d6050>
        {0.0} <music21.note.Note A>  // quarterLength = 2.0 **should be 4.0**
    {0.0} <music21.stream.Voice 0x7fc6a38d6210>
        {1.0} <music21.note.Note D> // quarterLength = 3.0
        {2.0} <music21.note.Note D> //quarterLength = 2.0 **should be deleted**
```

Wrapping this test case in a score forces `retainContainers=True`, but just using a Part and accepting the default `retainContainers=False` to get back a single stream without measures yields similar incorrect behavior, too.

Have a patch for this that uses `voicesToParts()`.

jacobtylerwalls:
Fixes #266 fixes #514 

Summary: when `stripTies(retainContainers=True)` deletes notes, it only searches `Measure` containers. Notes in other containers or no containers will stick around and end up as duplicates.

Fix: use `Stream.remove(recurse=True)` rather than popping notes from `Measure` containers.

Update: fixes #568 

Summary: `stripTies()` depends on tied notes being consecutive in the flattened parent container, which when voices are involved, may not obtain, and will yield unexpected results.

Fix: make recursive calls to `stripTies()` for each voice, modeling on how this is currently done for part-like substreams.

Update: Fixes #491 by filtering non-`GeneralNote` objects from `findConsecutiveNotes()` output.

### DIFF
--- a/music21/stream/__init__.py
+++ b/music21/stream/__init__.py
@@ -6193,14 +6193,19 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
         remove the formerly-tied notes.
 
         This method can be used on Stream and Stream subclasses.
-        When used on a Score, Parts and Measures are retained.
+        When used on a stream containing Part-like substreams, as with many scores,
+        :class:`~music21.stream.Part`, :class:`~music21.stream.Measure`, and other
+        Stream subclasses are retained.
 
-        If `retainContainers` is False (by default), this method only
-        returns Note objects; Measures and other structures are stripped
-        from the Stream. Set `retainContainers` to True to remove ties
-        from a :class:`~music21.part.Part` Stream that contains
-        :class:`~music21.stream.Measure` Streams, and get back a
-        multi-Measure structure.
+        Otherwise, if `retainContainers` is False (by default), this method only
+        returns a Stream of Notes; Measures and other structures are stripped
+        from the Stream. Set `retainContainers=True` to preserve substreams
+        when Part-like substreams are not present. (The value of `retainContainers` is not
+        observed if a Stream has part-like substreams.)
+
+        Although `inPlace=True` will avoid making deep copies of the input Stream, the keyword
+        will not determine the return object if `retainContainers=False`, which will always
+        return a new Stream of Note objects.
 
         Presently, this only works if tied notes are sequential; ultimately
         this will need to look at .to and .from attributes (if they exist)
@@ -6376,18 +6381,9 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
                 # environLocal.printDebug(['removing note', notes[i]])
                 # get the obj ref
                 nTarget = notes[i]
-                # go through each container and find the note to delete
-                # note: this assumes the container is Measure
-                # TODO: this is where we need a recursive container Generator out
-                for sub in reversed(returnObj.getElementsByClass('Measure')):
-                    try:
-                        i = sub.index(nTarget)
-                    except StreamException:
-                        continue  # pass if not in Stream
-                    junk = sub.pop(i)
-                    # get a new note
-                    # we should not continue searching Measures
-                    break
+                # Recurse rather than depend on the containers being Measures
+                # https://github.com/cuthbertLab/music21/issues/266
+                returnObj.remove(nTarget, recurse=True)
             returnObj.coreElementsChanged()
             return returnObj
 
--- a/music21/stream/__init__.py
+++ b/music21/stream/__init__.py
@@ -6185,7 +6185,7 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
     def stripTies(self,
                   inPlace=False,
                   matchByPitch=False,
-                  retainContainers=False):
+                  retainContainers=True):
         '''
         Find all notes that are tied; remove all tied notes,
         then make the first of the tied notes have a duration
@@ -6198,16 +6198,16 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
         Stream subclasses are retained.
 
         `inPlace` controls whether the input stream is modified or whether a deep copy
-        is made. Independently of what stream is operated upon, `retainContainers=False`
-        provides the option to have returned a flat Stream of Notes where Measures
-        and other structures have been stripped. This keyword only controls the return
-        object, not the input stream, which may have been edited `inPlace`.
+        is made. `retainContainers=False` returns a flattened stream where Measures
+        and other structures have been stripped.
 
-        If `retainContainers=True`, which is enforced for all streams with part-like
-        substreams without regard to the keyword supplied by the user, the Stream
-        operated upon is returned, which may have been edited `inPlace` or deep copied.
-        `retainContainers=False` is the default keyword accepted by the method, but to
-        reiterate, it will only have an effect if the stream has no part-like substreams.
+        N.B.: `retainContainers=False` will have no effect on streams with part-like
+        substreams, such as a :class:`~music21.stream.Score`.
+
+        Changed in v.6 -- `retainContainers` defaults to True.
+        Changed in v.6 -- `retainContainers=False` now only flattens the return
+        stream, rather than also calling `.notesAndRests`.
+        TODO: retainContainers TO BE DEPRECATED in v.7 (just call `.flat`)
 
         Presently, this only works if tied notes are sequential in the same voice; ultimately
         this will need to look at .to and .from attributes (if they exist)
@@ -6390,7 +6390,7 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
         if retainContainers:
             return returnObj
         else:
-            return returnObj.flat.notesAndRests.stream()
+            return returnObj.flat
 
     def extendTies(self, ignoreRests=False, pitchAttr='nameWithOctave'):
         '''
--- a/music21/stream/__init__.py
+++ b/music21/stream/__init__.py
@@ -6197,28 +6197,25 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
         :class:`~music21.stream.Part`, :class:`~music21.stream.Measure`, and other
         Stream subclasses are retained.
 
-        Otherwise, if `retainContainers` is False (by default), this method only
-        returns a Stream of Notes; Measures and other structures are stripped
-        from the Stream. Set `retainContainers=True` to preserve substreams
-        when Part-like substreams are not present. (The value of `retainContainers` is not
-        observed if a Stream has part-like substreams.)
+        `inPlace` controls whether the input stream is modified or whether a deep copy
+        is made. Independently of what stream is operated upon, `retainContainers=False`
+        provides the option to have returned a flat Stream of Notes where Measures
+        and other structures have been stripped. This keyword only controls the return
+        object, not the input stream, which may have been edited `inPlace`.
 
-        Although `inPlace=True` will avoid making deep copies of the input Stream, the keyword
-        will not determine the return object if `retainContainers=False`, which will always
-        return a new Stream of Note objects.
+        If `retainContainers=True`, which is enforced for all streams with part-like
+        substreams without regard to the keyword supplied by the user, the Stream
+        operated upon is returned, which may have been edited `inPlace` or deep copied.
+        `retainContainers=False` is the default keyword accepted by the method, but to
+        reiterate, it will only have an effect if the stream has no part-like substreams.
 
-        Presently, this only works if tied notes are sequential; ultimately
+        Presently, this only works if tied notes are sequential in the same voice; ultimately
         this will need to look at .to and .from attributes (if they exist)
 
         In some cases (under makeMeasures()) a continuation note will not have a
         Tie object with a stop attribute set. In that case, we need to look
         for sequential notes with matching pitches. The `matchByPitch` option can
         be used to use this technique.
-
-        Note that inPlace=True on a Stream with substreams currently has buggy behavior.
-        Use inPlace=False for now.
-
-        TODO: Fix this.
 
         >>> a = stream.Stream()
         >>> n = note.Note()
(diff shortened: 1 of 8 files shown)


jacobtylerwalls:
I looked into the failing tests in the feature extractor, centering on Schoenberg op19 no6. The original change in the PR made `stripTies()` able to actually delete the notes it wants to delete, which means it could make other failures like #568 more prevalent. I was able to isolate #568 by looking at the Schoenberg excerpt.

If you'd like to compare how this was handled
  - in production
  - in original PR
  - after the fix to #568 using `voicesToParts()`, which handles the Schoenberg very well
I threw together some PDFs with greenboxing and redboxing of the problem spots.

[1 - schoenberg_parse.pdf](https://github.com/cuthbertLab/music21/files/4842712/1.-.schoenberg_parse.pdf)
[2 - schoenberg_stripTiesProduction.pdf](https://github.com/cuthbertLab/music21/files/4842713/2.-.schoenberg_stripTiesProduction.pdf)
[3 - schoenberg_stripTiesBranch.pdf](https://github.com/cuthbertLab/music21/files/4842714/3.-.schoenberg_stripTiesBranch.pdf)
[4 - schoenberg_stripTiesBranch_voicesToParts.pdf](https://github.com/cuthbertLab/music21/files/4842715/4.-.schoenberg_stripTiesBranch_voicesToParts.pdf)

jacobtylerwalls:
The remaining feature extractor doctest failures are off by small magnitudes. My impulse is that if debugging op.19/6 showed an improvement, the doctest expected results ought to change, assuming they continue to run `stripTies`.

jacobtylerwalls:
Suppressed PyLint warnings for cell variables defined in a for loop. The way the algorithm is written, `updateEndMatch()` is only ever called when the value of `posConnected` has been updated appropriately.

jacobtylerwalls:
There is another effect of this latest refactoring which is worth mentioning, which is that the keyword combination `inPlace=True, retainContainers=False`--which is the default if a user asks for `inPlace=True` on a stream without part-like substreams--actually returns something reasonable now: it edits the stream in place and then, in addition, provides a return object with containers removed, for convenience/compatibility with existing code.

Before, the stream was sort of edited in place, which is to say, deep copies weren't made, and note prolongations happened on the stream in place, but note deletions only happened on the returned flattened stream, leaving the input stream half-edited (same problem as fixed in #266/#514 when retainContainers=True)

jacobtylerwalls:
I'm going to squash this to just two commits representing the two issues being fixed.

I also checked that I can simplify the handling of the return object if `retainContainers=False` without all the `None` checking.

Then I will move "Improve feature extractor handling of streams without notes" to a separate ticket/PR, since I can reproduce that in 5.7.2, and I shouldn't need it to get the test suite to pass. It was a red herring from looking into the doctest failure for `features.base.allFeaturesAsList()`.

That doctest started passing as soon as the two feature extractors that strip ties a *second* time with `retainContainers=False` no longer did that, which is good, because the code comments suggested this was unexpected. And given that change, I then need the fix to `findConsecutiveNotes()`, since it is now more likely that streams with extraneous objects like `Key` make it into the feature extractors, given that the first call of stripTies retains containers. That second go of calling stripTies with `retainContainers=False` was guaranteeing keys etc. would be stripped , but `findConsecutiveNotes()` should filter.

Letting the *first* stripTies call in `features.base._prepareStream()` operate `inPlace` didn't improve or degrade the results in the feature extractor tests, so I will leave that in. I will update the expected results to the feature extractor tests in a third commit on the assumption that if this gets merged the expected results have to change.

Happy also to hear feedback on the approach :-)

jacobtylerwalls:
I see now that it's too presumptive to change whether `_prepareStream()` operates `inPlace`, so I reverted that change.  I let a comment in the existing code tempt me.

However, holy unrelated NumPy import failures o_O

mscuthbert:
If it has a `_` in front of it, we can change defaults, etc., without a deprecation cycle.  Closing and Reopening to avoid these annoying matplotlib errors -- they were bugging me mid-day but they were find this evening.

jacobtylerwalls:
True, but I don't think I have enough experience using the extractors to call the trade-off between speed (avoiding deepcopies) and messing with people's streams who might expect to keep using them after calling an extractor. stripTies() still isn't perfect, and a round trip with makeTies() won't always restore the original, so I don't know if it's appropriate that the feature extractor would leave side effects on the input. And `inPlace` seems to be defaulting False everywhere else, so that was my thought. But if all people do, generally, is to parse files from disk and then discard the Python objects immediately after extraction, I could see it going the other way.

jacobtylerwalls:
Indentation was getting out of control, so I solved that and avoided the call to `voicesToParts` by just modeling this change on the existing recursive calls to `stripTies` if there are part-like substreams.

Fewer expected results have to change in the feature extractor doctests now, which is good. The only ones that will change are for the Schoenberg op19/6, which is the very case I was setting out to address in #568.

music21/stream/__init__.py
```diff
@@ -6367,36 +6376,21 @@ def updateEndMatch(n):
...
-        # in result object may not be the same
         posDelete.reverse()  # start from highest and go down
 
+        for i in posDelete:
+            # environLocal.printDebug(['removing note', notes[i]])
+            # get the obj ref
+            nTarget = notes[i]
+            # Recurse rather than depend on the containers being Measures
+            # https://github.com/cuthbertLab/music21/issues/266
+            returnObj.remove(nTarget, recurse=True)
```
mscuthbert, line 6387: This is so great.  There wasn't a `recurse=True` when we wrote this routine, and it now not only simplifies but also speeds up (probably) and removes bugs from it.  THANKS!  There are probably lots of other places that could depend on this.

music21/stream/__init__.py
```diff
@@ -6367,36 +6376,21 @@ def updateEndMatch(n):
...
-            returnObj.coreElementsChanged()
             return returnObj
-
         else:
-            for i in posDelete:
-                # removing the note from notes
-                junk = notes.pop(i)
-            notes.coreElementsChanged()
-            return notes
+            return returnObj.flat.notesAndRests.stream()
```
mscuthbert, line 6393: WOW.  and that's it!  :-)

mscuthbert: Actually, check me on this -- why are we doing `.notesAndRests.stream()` ?  I would have thought that StripTies with retainContainers=False returned a Stream that was identical to the original flat stream except that the tied notes have been removed.  I wouldn't want the Clefs, etc. removed -- is that what we've been doing for a long time?  Uh oh.  I think just `return returnObj` should be good, and people can do `.notesAndRests` themselves.

if `retainContainers` default was True, then I would remove it completely as an option, since `.flat` is simple to call later, but since it exists with a True default, we should keep that.

music21/stream/__init__.py
```diff
@@ -8936,6 +8930,9 @@ def findConsecutiveNotes(self,
 
         for v in vGroups:
             for e in v:
+                # Filter out all but notes and rests
+                if not isinstance(e, note.GeneralNote):
+                    continue
```
mscuthbert, line 8935: Thank, you.  This is great.

btw -- we use `if 'GeneralNote' not in e.classes:` or `not in e.classSet` a lot in the code, because it's actually an order of magnitude faster than isinstance (or was 5 years ago last time I did a lot of speed profiling).  findConsecutiveNotes isn't a performance critical routine, so no-big-deal, but something to know for the future.

music21/stream/__init__.py
jacobtylerwalls: Yep, that was the existing behavior for `retainContainers=False`, strip clefs, everything--however, the complex defaults made it difficult to reach. The keyword supplied or the default value from the method signature is _ignored_ if the input stream has part-like substreams. So all scores are forced to run under `retainContainers=True`. Otherwise, we read the keyword, which defaults False but a user can set True. Nasty to explain in the docs.

Everywhere else in the m21 codebase sets `retainContainers=True`. (Well, now that I commented out those two lines in the two feature extractors that were causing trouble.) There is one example in the user's guide that uses `retainContainers=False`:
```
One thing to notice is that the note E extends now beyond the end of the first 2/4 measure. 
The second D, in measure 2, by contrast, does not begin at the beginning of the measure, 
but instead halfway through the first measure. This is why it's sometimes most helpful 
to run stripTies() without retaining containers:

stripped2 = littlePiece.stripTies(retainContainers=False)
stripped2.show('text', addEndTimes=True)
{0.0 - 1.5} <music21.note.Note D>
{1.5 - 3.0} <music21.note.Note E>
{3.0 - 4.5} <music21.note.Note D>
{4.5 - 6.0} <music21.note.Note F>
```

I think it would be a big win to remove the keyword. Single responsibility principle. No difficult to explain keyword-overwriting.

jacobtylerwalls: It's curious, even some of the calls that pass no keywords to `stripTies` just end up calling `.notes` later anyway, probably because of the fickle keyword behavior:
https://github.com/cuthbertLab/music21/blob/6e5b01d1852097c6da94e4cbb492c36105d49902/music21/analysis/discrete.py#L1287

jacobtylerwalls: Actually, now that I look at this, the only reason I derived the return object this way was because a few commits ago I was trying to run `voicesToParts` and indented most of the algorithm under that, but now I don't have to re-derive the return object, I can just return `notes` from near the top of the function.

jacobtylerwalls: Nope, doesn't work. Still need to re-derive it because I'm only removing notes from `returnObj` now. Re-deriving at the end of the function is extra overhead -- this all points to removing the keyword, I think. Deprecating the keyword and rewriting some docs could be handled in another PR down the line, no?

music21/stream/__init__.py
```diff
@@ -6247,6 +6249,13 @@ def stripTies(self,
...
+        if returnObj.hasVoices():
+            for v in returnObj.voices:
+                # already copied if necessary; edit in place
+                v.stripTies(inPlace=True, matchByPitch=matchByPitch,
+                            retainContainers=retainContainers)
+            return returnObj  # exit
+
         # need to just get .notesAndRests, as there may be other objects in the Measure
         # that come before the first Note, such as a SystemLayout object
         f = returnObj.flat
```
jacobtylerwalls, line 6261: the line below is where were stripping clefs, etc., when `notes` was returned at the end of the function

music21/stream/__init__.py
mscuthbert: I would say, let's do this -- remove the `.notesAndRests` and make a note saying that non-Notes are retained now.  And change retainContainers to default to True, and put a note saying that retainContainers has defaults to `True` now, and that the keyword will be removed completely in a future version around 2022 -- this gives people who need to support both music21 v5 and v6 a way of keeping the containers (via `retainContainers=True`) but gives two years as a transition time.

I think that this is the best way to help people support v5 and v6 and future versions, but not to commit to keeping retainContainers forever!

jacobtylerwalls:
merge ho!

jacobtylerwalls:
Somehow a stray double-tap on the Github app effected the "close button", so reopening

mscuthbert:
Looks great! Merging! :-)

PR #574, opened 2020-07-11
Update Humdrum links in FAQ

jacobtylerwalls:
The Humdrum links in the FAQ were outdated. The Ohio State link no longer leads to Humdrum info, and the .net yielded a spam site.


mscuthbert:
thanks!! :-)

PR #576, opened 2020-07-14
Propagate time sig, key sig, barline in voicesToParts()

### LINKED ISSUE #575: voicesToParts
voicesToParts seemingly doesn't transfers initial time and key signatures if the first measure doesn't contain a second voice (even if subsequent measures do).

The images show how this looks when opened in MuseScore, (some other notation softwares will 'correct' the issue, masking the problem).

Before
After

jacobtylerwalls:
Fixes #575 

Before: time sig and key sig were only propagated to lower parts by `voicesToParts()` if the source measure for those elements had voices. Barlines also weren't propagated even in measures with voices due to a typo.

Now: time sig and key sig and barlines all propagate whether the source measure had voices or not.

Current behavior for clef is maintained: don't propagate, and let `bestClef()` run at the end.

### DIFF
--- a/music21/stream/__init__.py
+++ b/music21/stream/__init__.py
@@ -9970,6 +9970,7 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
                 ...
                 {3.5} <music21.note.Note A>
             {8.0} <music21.stream.Measure 3 offset=8.0>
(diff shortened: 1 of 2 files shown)


mscuthbert:
Looks wonderful, THANKS!

PR #578, opened 2020-07-16
Let makeAccidentals() recurse

### LINKED ISSUE #577: Making voices can disrupt makeAccidentals()
Based on case reported by Igor Praznik, music21 list 3/3/20.

```
s = stream.Stream()
s.insert(key.Key('Gb'))
s.insert(0, note.Note('D-5'))
s.insert(0, note.Note('D-4'))
s.show()  # During export to xml, makeNotation(), makeVoices(), makeMeasures()
```
Screen Shot 2020-07-14 at 9 19 28 PM

`makeNotation()` calls `makeAccidentals()` per `Measure` object, but `makeAccidentals()` only inspects `measure.notesAndRests`, which is empty in this case because the notes have been put in voices already:

https://github.com/cuthbertLab/music21/blob/93bea7cc87bdb9fb2bb0ab0cb3faef1717b24004/music21/stream/__init__.py#L5867

<s>Need to grab the voice objects somewhere, possibly just recursively calling `makeAccidentals()`, but there are going to be corner cases here with accidentals tied from previous measures.</s> Update: I think this is only affecting the code path where measures have not been created yet. In that case, probably easier to just call `makeAccidentals()` before `makeVoices()` and `makeMeasures()`, since without measures to worry about, those corner cases go away.

jacobtylerwalls:
Fixes #577. Easy patch, so I just forged on.

Before: calling `makeAccidentals()` on a Measure with substreams would have no effect. Such calls are made by `makeNotation()` and xml export.

Now: just let `makeAccidentals()` recurse over the input Stream.

Question: is there a valid use case for not recursing over substreams when making accidentals?  If so, this needs a couple quick lines of documentation in `makeAccidentals()`, possibly a keyword `recurse=`, and a decision about whether the default should be True or False in v.6 and whether it should move to True in v.7, etc.

I'm having trouble imagining the use case for NOT wanting accidentals made on substreams, so I didn't create a keyword arg. If a user doesn't want the new behavior, well, `makeAccidentals()` wasn't doing anything to substreams before, so such a user could just call `makeAccidentals()` directly on the desired substreams, if any.

`makeNotation()` now does what it advertises in more cases, so I didn't see a specific reason to update the docs, but happy to if you think wise.

### DIFF
--- a/music21/stream/__init__.py
+++ b/music21/stream/__init__.py
@@ -8783,7 +8783,7 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
         {0.0} <music21.note.Note B>
         {2.0} <music21.chord.Chord A B->
         {3.0} <music21.note.Note D>
-        
+
         (Technical note: All elements of class NotRest are being found
         right now.  This will eventually change to also filter out
         Unpitched objects, so that all elements returned by
--- a/music21/stream/__init__.py
+++ b/music21/stream/__init__.py
@@ -5865,18 +5865,15 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
 
         # need to move through notes in order
         # recurse to capture notes in substreams: https://github.com/cuthbertLab/music21/issues/577
-        noteStream = returnObj.recurse().notesAndRests
+        noteIterator = returnObj.recurse().notesAndRests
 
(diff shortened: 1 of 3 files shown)


mscuthbert:
thank you!

PR #581, opened 2020-07-20
Fixes #580: ScoreExporter no longer resets offsets on measures not contained in parts

### LINKED ISSUE #580: ScoreExporter resets measure offsets if measures aren't contained in Parts
Based on a case from music21 list, Jan 2019:

```
s = stream.Score()
s.append(meter.TimeSignature('1/4'))
s.append(note.Note())
s.append(note.Note())
s.show()
```

Screen Shot 2020-07-20 at 4 36 45 PM

`ScoreExporter.setPartsAndRefStream()` resets offsets on substreams and transfers those offsets to the contained elements, but here, the substreams are Measures, not Parts--meaning this stream isn't totally well-formed. Still, it would be simple to handle Measures differently in this routine.

jacobtylerwalls:
Fixes #580 

Before: `ScoreExporter` will eventually create a `Part` for a `Score` without one, but `setPartsAndRefStream()` is called first, causing problems when it clears offsets on substreams (which, until Parts are created, are Measures).

Now: don't clear/transfer offsets on Measures; do calculate `self.highestTime`, though.

NB: this is a case where `parseFlatScore()` indeed runs (code comment wonders if it ever does).

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -1500,9 +1500,14 @@ class ScoreExporter(XMLExporterBase):
         streamOfStreams = s.getElementsByClass('Stream')
         for innerStream in streamOfStreams:
             # may need to copy element here
-            # apply this streams offset to elements
-            innerStream.transferOffsetToElements()
-            ht = innerStream.highestTime
+            # apply this stream's offset to elements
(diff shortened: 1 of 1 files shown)


mscuthbert:
CI Failure is unrelated to this code and is fixed in another PR.  Merging!

PR #582, opened 2020-07-25
Bug report template

jacobtylerwalls:
Loosely modeled on the template used at Supercollider. Take as much or as little of this as you want!

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -6,12 +6,13 @@ labels: ''
 assignees: ''
 
 ---
+music21 version: **5.7.2**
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,29 @@
+---
+name: Bug report
+about: Report an issue with music21
+title: ''
+labels: ''
+assignees: ''
+
+---
+
```
mscuthbert, line 9: asking for music21 version is probably important too -- we still get bug reports for 1.0 sometimes... and definitely for 5.7.

mscuthbert:
Thank you!

PR #592, opened 2020-08-24
Allow element to lack duration when coinciding with last measure's end offset

### LINKED ISSUE #515: makeMeasures error "cannot place element within any measures"
When music21 parses midi files, it seems not to create automatically the measures. If I run the `makeMeasures()` function, however, I get often the error "StreamException('cannot place element <music21.meter.TimeSignature 3/4> with start/end 76.0/76.0 within any measures')". I noticed that this happens when the offset specified in start/end is exactly the entire duration of the part. 

Now, in some cases I can just open the MIDI file with musescore, re-export it and everything is fine (I have no idea why). But there is at least one trickier case: when the music is multi-instrument and one instrument has a time signature change after the last note it plays. In that case it seems to me that music21 cuts the length of the part to the last object it finds -- the time signature -- and raises the exception.

I'm sorry I can not provide a MWE right now, I don't really understand how to do that. I hope that what I wrote can help you identify the issue (or me understand what I'm doing wrong, in case this is the expected behavior).

### LINKED ISSUE #534: music21.exceptions21.StreamException: cannot place element
Hi, 

**How to reproduce:**
I reproduced the problem using:

```
m21mid = converter.parse("1-Midi.mid")
measures = m21mid.measures(0, None)
```
**Midi file example:**
I simplified the Midi received by my colleague as much as poss (I'm not MuseScore expert)

[1-Midi.mid.zip](https://github.com/cuthbertLab/music21/files/4552127/1-Midi.mid.zip)

**Result:**
> Traceback (most recent call last):
> [...]
>   File ".../venv/lib/python3.7/site-packages/music21/stream/makeNotation.py", line 577, in makeMeasures
>     'within any measures' % (e, start, end))
> music21.exceptions21.StreamException: cannot place element <music21.tempo.MetronomeMark adagio Quarter=56.0> with start/end 3.0/3.0 within any measures

Thanks for Music21 !! 👍🏻

jacobtylerwalls:
Fixes #534 Fixes #515 (duplicates)

**Before**:  `makeMeasures()` would choke on a duration-less element coinciding with the end offset of the final measure.

**Now**: `makeMeasures()` places a duration-less element coinciding with the end offset of the final measure <s>in that final measure</s> at the end of the stream.

### DIFF
--- a/music21/stream/__init__.py
+++ b/music21/stream/__init__.py
@@ -2031,7 +2031,7 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
         This method is useful for putting things such as
         right bar lines or courtesy clefs that should always
         be at the end of a Stream no matter what else is appended
-        to it
+        to it.
 
         As sorting is done only by priority and class,
         it cannot avoid setting isSorted to False.
--- a/music21/stream/makeNotation.py
+++ b/music21/stream/makeNotation.py
@@ -566,7 +566,6 @@ def makeMeasures(
             mEnd = postMeasureInfo['mEnd']
             m = postMeasureInfo['measure']
 
-            # also allow zero duration element at last measure's end offset
             if mStart <= start < mEnd:
                 match = True
                 # environLocal.printDebug([
--- a/music21/stream/makeNotation.py
+++ b/music21/stream/makeNotation.py
@@ -567,16 +567,19 @@ def makeMeasures(
             m = postMeasureInfo['measure']
 
             # also allow zero duration element at last measure's end offset
-            if mStart <= start < mEnd or mEnd == oMax == start == end:
+            if mStart <= start < mEnd:
                 match = True
                 # environLocal.printDebug([
                 #    'found measure match', i, mStart, mEnd, start, end, e])
                 break
 
         if not match:
-            raise stream.StreamException(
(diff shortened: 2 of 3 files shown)


jacobtylerwalls:
Probably could get a slight performance improvement if I broke this out into a second conditional after `for i in range(postLen):` finishes.

mscuthbert:
Closing and reopening to get a fresh pylint.

mscuthbert:
I think that in this case, it might be better to use `storeAtEnd()` instead of insert() or append() -- I'm trying to think of a zero-length element at the end of the piece that wouldn't want to stay at the end of the measure if the measure changed.

jacobtylerwalls:
Coveralls shows the `else: raise stream.StreamException` as a new missed line ... I could be persuaded to strip it out entirely, and then just have:
```
if not match:
    post.storeAtEnd(e)
```

and let `storeAtEnd` raise the `StreamException` if the duration is > 0, which it already does.

mscuthbert:
I think that the exception might still run if an element is placed beyond the end of the last measure or something like that?  In any case, I don't worry about missing exception coverage.  I'd be happy to put pragma: no cover over any that aren't run that we don't think should ever occur.  Merging, with my thanks!

PR #595, opened 2020-08-25
Create score-instrument when all we have is .midiProgram

### LINKED ISSUE #594: Invalid midi-instrument id produced when instrument has midi program but no instrument name
**music21 version**

`6.0.5.a1`

**Problem summary**

An `Instrument` object with a `.midiProgram` but no `.instrumentName` produces upon .xml export a `<midi-instrument>` id matching no other id in the document. This causes validation issues in Finale:

id_validation

Such instruments can arise from parsing musicXML into m21, since m21 will set `.midiProgram` but only create generic `Instrument` instances without names, etc. In other words, this is an issue on a roundtrip back to .xml.

**Steps to reproduce**

```
from music21 import converter, musicxml
from music21.musicxml import testFiles

mxString = testFiles.ALL[5]  # contains Electric Guitar midi program
scoreObj = converter.parseData(mxString)  # midiProgram stored; no instrumentName stored
scoreObj.show()

scex = musicxml.m21ToXml.ScoreExporter(scoreObj)
tree = scex.parse()
mxScoreInstrument = tree.findall('.//score-instrument')[0]
mxPart = tree.findall('.//part')[0]
```

```
>>> print(mxMidiInstrument.get('id')
I8f4138faff93acfe295ec826023c467a
>>> print(mxPart.get('id')
P0
```
Expected equality with `<part>` id or with a `<score-instrument>` id.

**More information**

Spec says the id of `<midi-instrument>` should match that of the part or the `<score-instrument>`. My intended patch consists of creating a `<score-instrument>` when we only have a `.midiProgram`. That turns out to be enough to get a matching id with the `<midi-instrument>`.

jacobtylerwalls:
Fixed #594

**Before:** On .xml export, schema validation issues occurred when `Instrument` instances with merely a `.midiProgram` and no `.instrumentName` never had a corresponding `<score-instrument>` created alongside the `<midi-instrument>`.

**Now:** we create a `<score-instrument>` element in this case so that the `<midi-instrument>` id matches something.

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -2545,7 +2545,8 @@ class PartExporter(XMLExporterBase):
         # TODO: unbounded...
         i = self.firstInstrumentObject
 
-        if i.instrumentName is not None or i.instrumentAbbreviation is not None:
+        if (i.instrumentName is not None or i.instrumentAbbreviation is not None
+                or i.midiProgram is not None):
             mxScorePart.append(self.instrumentToXmlScoreInstrument(i))
 
         # TODO: midi-device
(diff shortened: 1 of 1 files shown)


mscuthbert:
Thank you!

PR #596, opened 2020-08-26
Write compressed xml directly from .write('mxl')

### LINKED ISSUE #456: Using the extension mxl for compressed musicxml when writing to file
I can not open MusicXML files generated from imported krn files. Whenever I try to do so, Musescore says _Cannot read file [path]: bad format_  . I tried to import those MusicXML files back into music21 and, while it eventually processes them, it gives lots of warnings:

> musicxml.xmlToM21: WARNING: Cannot put in an element with a missing voice tag when no previous voice tag was given.  Assuming voice 1... 
> musicxml.xmlToM21: WARNING: Cannot find voice 1; putting outside of voices.
> musicxml.xmlToM21: WARNING: Current voiceIds: ['139734311396297', '139734311398537']
> musicxml.xmlToM21: WARNING: Current voices: [<music21.stream.Voice 139734311396297>, <music21.stream.Voice 139734311398537>]

This same warning is repeated identical several times for each score.

The code I'm using is the following:
```
from music21 import converter

path_krn = 'score.krn'
path_mxl = 'score.mxl'
converter.parse(path_krn).write('mxl', fp=path_mxl)
```

The following krn file of [a Bach chorale](https://kern.humdrum.org/cgi-bin/ksdata?file=chor001.krn&l=users/craig/classical/bach/371chorales&format=kern) would do just fine to test.

jacobtylerwalls:
Fixes #456

### DIFF
--- a/music21/converter/__init__.py
+++ b/music21/converter/__init__.py
@@ -1048,9 +1048,6 @@ def parse(value: Union[bundles.MetadataEntry, bytes, str, pathlib.Path],
 
     `format` specifies the format to parse the line of text or the file as.
 
-    `compress` specifies whether to immediately compress the .xml output to
-    a .mxl file. The intermediate .xml file will not be retained.
-
     `quantizePost` specifies whether to quantize a stream resulting from MIDI conversion.
     By default, MIDI streams qre quantized to the nearest sixteenth or triplet-eighth
     (i.e. smaller durations will not be preserved).
--- a/music21/converter/__init__.py
+++ b/music21/converter/__init__.py
@@ -1048,6 +1048,9 @@ def parse(value: Union[bundles.MetadataEntry, bytes, str, pathlib.Path],
 
     `format` specifies the format to parse the line of text or the file as.
 
+    `compress` specifies whether to immediately compress the .xml output to
+    a .mxl file. The intermediate .xml file will not be retained.
+
     `quantizePost` specifies whether to quantize a stream resulting from MIDI conversion.
     By default, MIDI streams qre quantized to the nearest sixteenth or triplet-eighth
     (i.e. smaller durations will not be preserved).
--- a/music21/converter/subConverters.py
+++ b/music21/converter/subConverters.py
@@ -952,7 +952,7 @@ class ConverterMusicXML(SubConverter):
         return fp
 
     def write(self, obj, fmt, fp=None, subformats=None, **keywords):  # pragma: no cover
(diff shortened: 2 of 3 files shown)


music21/converter/__init__.py
```diff
@@ -1048,6 +1048,9 @@ def parse(value: Union[bundles.MetadataEntry, bytes, str, pathlib.Path],
 
     `format` specifies the format to parse the line of text or the file as.
 
+    `compress` specifies whether to immediately compress the .xml output to
+    a .mxl file. The intermediate .xml file will not be retained.
+
```
mscuthbert, line 1053: Not sure what compress on parse does for the user -- can you explain?

music21/converter/subConverters.py
```diff
@@ -952,7 +952,11 @@ def writeDataStream(self, fp, dataBytes: bytes):  # pragma: no cover
         return fp
 
     def write(self, obj, fmt, fp=None, subformats=None, **keywords):  # pragma: no cover
-        from music21.musicxml import m21ToXml
```
mscuthbert, line 955: Add `compress=False` to the signature so people know what is available here, I *think* -- it violates some subclassing paradigms but fits with what we have done in the rest of the system.

music21/converter/subConverters.py
```diff
@@ -982,6 +986,10 @@ def write(self, obj, fmt, fp=None, subformats=None, **keywords):  # pragma: no c
                 and ('png' in subformats or 'pdf' in subformats)
                 and not str(environLocal['musescoreDirectPNGPath']).startswith('/skip')):
             outFp = self.runThroughMusescore(xmlFp, subformats, **keywords)
+        elif 'compress' in keywords and keywords['compress']:
```
mscuthbert, line 989: Then this just needs to be `elif compress:`

music21/converter/__init__.py
jacobtylerwalls: Once I take your suggestion below to get compress into the method signature, then I don't need to expose this here at all. As it was, `compress` was hanging out in `**keywords`, so a user _could_ have passed it in here, setting format '.xml' and then setting compress=True, but there's no reason to support that when `format='mxl'` will do.

jacobtylerwalls: Haha nope, I definitely cut and pasted this into the wrong method when I was looking for more places to document. Duh.

mscuthbert:
Great!  Thank you!

PR #598, opened 2020-08-27
Support multi-digit endings in xml export; bump MuseScore version paths

jacobtylerwalls:
- Now with support for multi-digit endings from m21 -> xml (before I only did xml -> m21)

- Tried .write('musicxml.png') for the first time and needed to download MuseScore, which now downloads as MuseScore 3.5.app, so updated the default paths.


mscuthbert:
LGTM!  Merging!

PR #603, opened 2020-09-02
Conductor track

### LINKED ISSUE #152: slightly unexpected behavior in writing midi from score
from music21.note import Note  
from music21.stream import Score, Part  
from music21.tempo import MetronomeMark                                          
# if I create a score with

s = Score()                                                                      
# and then insert a tempo marker with...

s.insert(MetronomeMark(referent=1, number=85))                                   
# and then append to the score with...

part1 = Part()  
part1.append(Note())  
s.append(part1)                                                                  

part2 = Part()  
part2.append(Note())  
s.append(part2)                                                                  
# and then finally look at the midi file...

s.write('midi', '/tmp/mytest.midi')                                              

"""  
looking at the midifile...  
0, 0, Header, 1, 2, 1024  
1, 0, Start_track  
1, 0, Title_t, ""  
1, 0, Pitch_bend_c, 0, 8192  
1, 0, Tempo, 705882  
1, 0, Note_on_c, 0, 60, 90  
1, 1024, Note_off_c, 0, 60, 0  
1, 1024, End_track  
2, 0, Start_track  
2, 0, Title_t, ""  
2, 0, Pitch_bend_c, 0, 8192  
2, 0, Note_on_c, 0, 60, 90  
2, 1024, Note_off_c, 0, 60, 0  
2, 1024, End_track  
0, 0, End_of_file                                                                

Above the tempo event is attached to track 1,  
when it was originally added to the Score instance,  
not a Part instance. I suppose this is totally cool,  
but is somewhat inconsistent with what I would  
expect if score is supposed to directly relate  
to the multitrack midifile, and track is supposed to directly  
relate to a track on the midifile. Abjad/Lilypond work a bit  
differently, with a separate track in the beginning of the file  
without note events to hold things like an initial global tempo.  
"""

### LINKED ISSUE #569: converter.parse fails on zero-track MIDIs
When calling `music21.converter.parse` on a zero-track MIDI, an `IndexError` is returned because there is no 0th Stream.

```
>>> import music21
music21: Certain music21 functions might need these optional packages: matplotlib, numpy, scipy;
                   if you run into errors, install them by following the instructions at
                   http://mit.edu/music21/doc/installing/installAdditional.html
>>> music21.converter.parse('blank.mid')
  File "/home/pi/.local/lib/python3.7/site-packages/music21/converter/__init__.py", line 1129, in parse
    forceSource=forceSource, **keywords)
  File "/home/pi/.local/lib/python3.7/site-packages/music21/converter/__init__.py", line 1009, in parseFile
    v.parseFile(fp, number=number, format=format, forceSource=forceSource, **keywords)
  File "/home/pi/.local/lib/python3.7/site-packages/music21/converter/__init__.py", line 544, in parseFile
    self.parseFileNoPickle(fp, number, format, forceSource, **keywords)
  File "/home/pi/.local/lib/python3.7/site-packages/music21/converter/__init__.py", line 478, in parseFileNoPickle
    self.subConverter.parseFile(fp, number=number, **keywords)
  File "/home/pi/.local/lib/python3.7/site-packages/music21/converter/subConverters.py", line 982, in parseFile
    midiTranslate.midiFilePathToStream(fp, self.stream, **keywords)
  File "/home/pi/.local/lib/python3.7/site-packages/music21/midi/translate.py", line 1975, in midiFilePathToStream
    return midiFileToStream(mf, inputM21, **keywords)
  File "/home/pi/.local/lib/python3.7/site-packages/music21/midi/translate.py", line 2115, in midiFileToStream
    **keywords)
  File "/home/pi/.local/lib/python3.7/site-packages/music21/midi/translate.py", line 1905, in midiTracksToStreams
    p = s.getElementsByClass('Stream')[0]
  File "/home/pi/.local/lib/python3.7/site-packages/music21/stream/iterator.py", line 328, in __getitem__
    e = fe[k]
IndexError: list index out of range
```

jacobtylerwalls:
Fixed #152 Fixed #569

**Before:** music21 produced multitrack MIDI files in format 1 (multi-track) without a conductor track. This isn't standard practice and could lead to the failure to propagate tempo, meter, and key to subsequent tracks when loaded into sequencers or notation programs.

**Now:** every produced MIDI file begins with a conductor track (Track 0) before the first music track (Track 1).

**Update:** On MIDI _import_, A `MetronomeMark` in the conductor track will now propagate to every m21 part, to fix a bug but more importantly so that `.seconds` will be correct on notes in subsequent parts.

I would appreciate some testers beating this up before merge. I'm not a MIDI expert.

### DIFF
--- a/music21/midi/translate.py
+++ b/music21/midi/translate.py
@@ -1847,15 +1847,20 @@ def conductorStream(s: stream.Stream) -> stream.Part:
     from music21 import tempo, meter
     out = stream.Part()
 
-    condTrkEvents = s.flat.getElementsByClass((
-        'MetronomeMark', 'TimeSignature', 'KeySignature'))
-    for el in condTrkEvents:
-        o = condTrkEvents.elementOffset(el)
-        s.remove(el, recurse=True)
-        # Don't overwrite anything at this offset
-        if not out.getElementsByOffset(o).getElementsByClass(el.classes[0]):
-            out.insert(o, el)
+    for klass in ('MetronomeMark', 'TimeSignature', 'KeySignature'):
+        events = s.flat.getElementsByClass(klass)
+        lastOffset = -1
+        for el in events:
+            o = events.srcStream.elementOffset(el)
+            s.remove(el, recurse=True)
+            # Don't overwrite an event of the same class at this offset
+            if o > lastOffset:
+                out.coreInsert(o, el)
+            lastOffset = o
 
+    out.coreElementsChanged()
+
+    # Defaults
     if not out.getElementsByClass('MetronomeMark'):
         out.insert(tempo.MetronomeMark(number=120))
     if not out.getElementsByClass('TimeSignature'):
@@ -2020,6 +2025,7 @@ def streamHierarchyToMidiTracks(
     2. we make a list of all instruments that are being used in the piece.
 
     Changed in v.6 -- acceptableChannelList is keyword only.  addStartDelay is new.
+    Changed in v.6.5 -- Track 0 (tempo/conductor track) always exported.
     '''
     # makes a deepcopy
     s = _prepareStreamForMidi(inputM21)
--- a/music21/midi/translate.py
+++ b/music21/midi/translate.py
@@ -2113,26 +2113,16 @@ def midiTracksToStreams(
                               quantizePost,
                               inputM21=conductorTrack, **keywords)
     # environLocal.printDebug(['show() conductorTrack elements'])
-    # if we have time sig/key sig elements, add to each part
+    # if we have time sig/key sig/tempo elements, add to each part
 
-    # TODO: this would be faster if we iterated in the other order.
-    for p in s.getElementsByClass('Stream'):
-        for e in conductorTrack.getElementsByClass(
-                ('TimeSignature', 'KeySignature')):
+    for e in conductorTrack.getElementsByClass(
+            ('TimeSignature', 'KeySignature', 'MetronomeMark')):
+        for p in s.getElementsByClass('Stream'):
             # create a deepcopy of the element so a flat does not cause
             # multiple references of the same
             eventCopy = copy.deepcopy(e)
             p.insert(conductorTrack.elementOffset(e), eventCopy)
 
-    # if there is a conductor track, add tempo only to the top-most part
-    # MSC: WHY?
-
-    p = s.getElementsByClass('Stream')[0]
-    for e in conductorTrack.getElementsByClass('MetronomeMark'):
-        # create a deepcopy of the element so a flat does not cause
-        # multiple references of the same
-        eventCopy = copy.deepcopy(e)
-        p.insert(conductorTrack.elementOffset(e), eventCopy)
     return s
 
 
@@ -3046,10 +3036,10 @@ class Test(unittest.TestCase):
         fp = dirLib / 'test11.mid'
         s = converter.parse(fp)
         self.assertEqual(len(s.parts), 3)
-        # metronome marks end up only on the top-most staff
+        # metronome marks end up on every staff
         self.assertEqual(len(s.parts[0].getElementsByClass('MetronomeMark')), 4)
-        self.assertEqual(len(s.parts[1].getElementsByClass('MetronomeMark')), 0)
-        self.assertEqual(len(s.parts[2].getElementsByClass('MetronomeMark')), 0)
+        self.assertEqual(len(s.parts[1].getElementsByClass('MetronomeMark')), 4)
+        self.assertEqual(len(s.parts[2].getElementsByClass('MetronomeMark')), 4)
 
     def testMidiExportConductorA(self):
         '''Export conductor data to MIDI conductor track.'''
--- a/music21/midi/translate.py
+++ b/music21/midi/translate.py
@@ -2148,9 +2148,9 @@ def streamToMidiFile(
     >>> n.quarterLength = 0.5
     >>> s.repeatAppend(n, 4)
     >>> mf = midi.translate.streamToMidiFile(s)
-    >>> len(mf.tracks)
-    1
-    >>> len(mf.tracks[0].events)
+    >>> len(mf.tracks)  # Includes conductor track
+    2
+    >>> len(mf.tracks[1].events)  # First music track
     22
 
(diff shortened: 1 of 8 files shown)


mscuthbert:

music21/midi/translate.py
```diff
@@ -1787,27 +1795,53 @@ def _prepareStreamForMidi(s):
...
+    '''
+    Strip the given stream of any events that belong in a conductor track
+    rather than in a music track, and return a :class:`~music21.stream.Part`
+    containing just those events, without duplicates.
+    '''
+    from music21 import tempo, meter
+    out = stream.Part()
+
+    condTrkEvents = s.flat.getElementsByClass((
+        'MetronomeMark', 'TimeSignature', 'KeySignature'))
```
mscuthbert, line 1831: This should have a .stream() at the end, otherwise the .elementOffset(el) is very inefficient.  Or you can use below `o = condTrkEvents.srcStream.elementOffset(el)`

music21/midi/translate.py
```diff
@@ -1787,27 +1795,53 @@ def _prepareStreamForMidi(s):
...
+    for el in condTrkEvents:
+        o = condTrkEvents.elementOffset(el)
+        s.remove(el, recurse=True)
+        # Don't overwrite anything at this offset
+        if not out.getElementsByOffset(o).getElementsByClass(el.classes[0]):
+            out.insert(o, el)
```
mscuthbert, line 1837: If this gets slow (like working with a Huge score), this is exactly what tree.trees.OffsetTree() is meant to do extremely quickly. OR since o is sorted and out starts out empty, you can keep track of lastOffset and just do the check if o is the same as lastOffset.  If you keep track of elements at the lastOffset (or just the classes) then you can skip the whole .getElementsByOffset() and .getElementsByClass() and therefore use .coreInsert() with a .coreElementsChanged() afterwards and get a huge speed increase.

music21/midi/translate.py
```diff
@@ -1833,10 +1867,17 @@ def channelInstrumentData(s: stream.Stream,
     substreamList = []
     if s.hasPartLikeStreams():
         for obj in s.getElementsByClass('Stream'):
-            substreamList.append(obj)
+            # _prepareStreamForMidi() supplies defaults for these
+            if obj.getElementsByClass(('MetronomeMark', 'TimeSignature')):
+                # Conductor track: don't consume a channel
+                continue
```
mscuthbert, line 1873: Maybe make this an optional flag in the argument OR update the documentation, so that no one uses this on another stream besides one immediately being parsed.

music21/midi/translate.py
```diff
@@ -1966,13 +2011,15 @@ def streamHierarchyToMidiTracks(
...
-    # store streams in uniform list
+    # store streams in uniform list: _prepareStreamForMidi() ensures there are substreams
     substreamList = []
-    if s.hasPartLikeStreams():
-        for obj in s.getElementsByClass('Stream'):
+    for obj in s.getElementsByClass('Stream'):
+        # _prepareStreamForMidi() supplies defaults for these
+        if obj.getElementsByClass(('MetronomeMark', 'TimeSignature')):
+            # Ensure conductor track is first
+            substreamList.insert(0, obj)
```
mscuthbert, line 2020: As above, should be a flag that this is supporting conductor tracks OR we will need to expose `_prepareStreamForMidi()` and note that this is assumed to already have happened.  I prefer the former.

music21/scale/__init__.py
```diff
@@ -309,7 +309,7 @@ def buildNetworkFromPitches(self, pitchList):
         if pitchList[-1].name == pitchList[0].name:  # the completion of the scale has been given.
             # print('hi %s ' % pitchList)
             # this scale is only octave duplicating if the top note is exactly
-            # 1 octave above the bottom; if it spans more thane one active,
+            # 1 octave above the bottom; if it spans more than one active,
```
mscuthbert, line 312: `than one octave`, I believe

mscuthbert (changes requested):
Super great -- I think that I finally understand everything.  The only requested changes are places where I think the docs need to explain a little more that the stream to be passed in should already have a top Part that is the conductor track.  (unless I'm misunderstanding)

music21/midi/translate.py
jacobtylerwalls: Thanks for having a look! I'm just diving in first with the bit I'm having trouble following.

_prepareStreamForMidi() is called earlier in this method, and now _prepare...() will always strip meta events from the (deepcopy of the) inputted parts and collect them in a conductor stream.Part; so there is no way to opt out of exporting a conductor track (nor should there be IMO). Could you elaborate on what you had in mind here?

music21/midi/translate.py
jacobtylerwalls: I've made this check more agnostic so that it's just `if not obj.notesAndRests`; then it's safe to assume it shouldn't consume a channel--whether it was run through _preapreStreamForMidi() or not. What do you think?

jacobtylerwalls: Haha nope that caused a bunch of test failures. It's in the commit history if you want to see it.

music21/midi/translate.py
mscuthbert: I mostly just mean that the docs just say (if I remember) that it takes in a Stream, but it's not just any Stream, it's a Stream that already has had `._prepareStreamForMidi()` run on it, so since this is a public function, we probably need to expose `._prepareStreamForMidi()` as a public function and tell people to run that first and probably need to show running that first in the docs.  If the function can't run on a normal Stream, then that's fine, but neither of us want to answer questions on the mailing list that the function didn't work, because the asker didn't run `.prepareStreamForMidi()` first.

jacobtylerwalls: Got it, then only the other place is an issue; this place can accept any stream and will always have _prepare run.

music21/midi/translate.py
jacobtylerwalls: My instinct I think was good; I just needed to check `not obj.flat.notesAndRests` and now it's agnostic to any stream input.

mscuthbert:
I'm going to merge because this is AMAZING -- I think it's easier for me to implement the little doc changes I'm looking for than to do another back and forth on something that's been waiting for over two months and is crucial to people.  THANKS!

PR #607, opened 2020-09-03
Catch UnicodeDecodeError

### LINKED ISSUE #606: Midi event to Instrument shouldn't assume utf-8
**music21 version**
`6.1`

**Problem summary**
My recent contribution to get instrument objects from `MetaEvents.SEQUENCE_TRACK_NAME` assumes utf-8 input and might crash.

**Steps to reproduce**

Found this while poking around on the m21 list looking for juicy posts. Also because I wanted to beat up my conductor track PR. The midi file on a post from 4/21/17 has the following midi event that crashes when the copyright symbol can't be decoded:
```
<MidiEvent SEQUENCE_TRACK_NAME, t=0, track=5, channel=None, data=b'Copyright \xa9 2008 by Bernd Krueger'>```
```

### LINKED ISSUE #610: UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa9 in position 10: invalid start byte
I am trying to use music21 to parse some midi files. I can use mido to open these files but I get the aforementioned UnicodeDecodeError when I try with music21.

Any clues on how I can work around this ?

Thanks,
Saif

jacobtylerwalls:
Fixed #606 Fixed #610 (duplicates)

### DIFF
--- a/music21/midi/translate.py
+++ b/music21/midi/translate.py
@@ -700,7 +700,7 @@ def midiEventsToInstrument(eventList):
             i = instrument.fromString(event.data.decode('utf-8'))
         else:
             i = instrument.instrumentFromMidiProgram(event.data)
-    except instrument.InstrumentException:  # pragma: no cover
+    except (instrument.InstrumentException, UnicodeDecodeError):  # pragma: no cover
         i = instrument.Instrument()
     return i
 


mscuthbert:
Oh, that's simple -- yes -- I thought this PR was going to involve seeing what other encodings could be used, etc., but this is the best simple result.

Maybe best that in future we catch the exception as (e) and explain why data is being dropped in a warning.

PR #612, opened 2020-09-06
Fixed string formatting issue in EnvironmentException when scratch directory doesn't exist

### LINKED ISSUE #342: exists(env['DirectoryScratch']) raises unintended error
on lines 540-542 of music21/environment.py

**Unexpected Behaviour:** the non-existence of environment['directoryScratch'] on the filesystem raises an unintended and uninformative error:
*** TypeError: non-empty format string passed to object.__ format __ ()
rather than the intended, more informative 
*** EnvironmentException("user-specified scratch directory({}) does not exist" ... )

This [StackOverflow answer](https://stackoverflow.com/questions/24170519/python-typeerror-non-empty-format-string-passed-to-object-format) explain the issue through a similar error (summarized below). They suggest converting the value with str(), which works: `{!s:s}` throws no errors. Though, ':s' is default behaviour for strings, so I'm starting to believe the original intention of this line was just '!s' in the first place. At any rate, I think it's better for the correct exception to be raised.

The issue with ':s'
The argument to format() (in this case, a PosixPath() object) has no __ format __ function, so format() defaults to object. __ format __
object.__ format __ () doesn't support any format specification, which is why it's expecting an empty format string.. hence the TypeError throws before returning a string to the EnvironmentException constructor.

Here's the documentation for [string.format()](https://docs.python.org/3.4/library/string.html#format-specification-mini-language)

Please let me know if there's anything else I can do or if I missed something (I made a pull-request with the single-character change)

jacobtylerwalls:
Thanks to David Garfinkle for the report and patch!

fixes #342

### DIFF
--- a/music21/environment.py
+++ b/music21/environment.py
@@ -557,7 +557,7 @@ class _EnvironmentCore:
 
         if not refDir.exists():
             raise EnvironmentException(
-                'user-specified scratch directory ({:s}) does not exist; '
+                'user-specified scratch directory ({!s}) does not exist; '
                 'remove preference file or reset Environment'.format(
                     refDir))
         return refDir


mscuthbert:
Looks great -- btw., I want to switch from `.format()` and `%` string interpolation to `f''` strings throughout the code (slowly -- or maybe quickly) since it's so much easier to see the issues like this that arise in f-strings.

PR #617, opened 2020-09-08
cleanup Braille files and fix one wrapping place

mscuthbert:


mscuthbert:
Drop in coverage expected because several tested routines were unnecessary and removed.  Actually increased total coverage.

PR #618, opened 2020-09-10
Favor pathlib objects instead of str(fp)

jacobtylerwalls:
Simplifications possible now that 3.6 is the minimum Python version.

Also, `Environment.getTempFile()` and `_EnvironmentCore.getTempFile()` default `returnPathlib=True`.

Also, updated path to Preview on failing tests in environment.py's `testToSettings()`


music21/environment.py
```diff
@@ -1462,7 +1465,7 @@ def testToSettings(self):
   <preference name="braillePath" />
   <preference name="debug" value="0" />
   <preference name="directoryScratch" />
-  <preference name="graphicsPath" value="/Applications/Preview.app" />
+  <preference name="graphicsPath" value="/System/Applications/Preview.app" />
```
ahankinson, line 1468: I don't think this is the correct path for macOS Preview.app

ahankinson: Actually, nevermind. I'm completely wrong. I hadn't noticed that Apple changed this around. Sorry.

jacobtylerwalls: No problem! In 6.1 there is a new function to derive the path to Preview based on the macOS version, but these tests had not been updated, so they would happen to pass on Travis CI but fail on my Mac.

mscuthbert:
Had a chance to review this and it looks great!

mscuthbert:
Thanks for finding all the 3.6 references.  V.7 is likely to be Python 3.7+ only, though I'm not so dying to use dataclasses or other Py3.7 features except `from __future__ import annotations`, so music21 v.7 might still work (unofficially) on Py3.6.  When we get to Py 3.8 as the minimum though, the breakage will happen fast, since there are a lot of features of Py3.8 that I really love esp. for Typing (TypedDict!).  But that won't happen for another 2 years.

PR #621, opened 2020-09-15
Augmented sixths

### LINKED ISSUE #585: RomanNumeral w/o key given has problems
**music21 version**

`master 6.0.5a`   

**Problem summary**

Not giving a Key to a RomanNumeral object that includes an Augmented 6th chord should still work and have an implied Key -- it is possible that Scales in general don't work

**Steps to reproduce**

```
roman.RomanNumeral('It6')
```

Expected:

`<music21.roman.RomanNumeral It6>`

Actual:
```
.../music21/roman.py in _parseRNAloneAmidstAug6(self, workingFigure, useScale)
   1909 
   1910         if self._augmentedSixthRegex.match(workingFigure):
-> 1911             if useScale.mode == 'major':
   1912                 useScale = key.Key(useScale.tonic, 'minor')
   1913                 self.impliedScale = useScale
AttributeError: 'MajorScale' object has no attribute 'mode'
```

### LINKED ISSUE #609: Augmented sixths missing chromatic modification when applied / secondary
**music21 version**
Dev copy

**Problem summary**
Augmented sixths missing chromatic modification when applied / secondary

**Steps to reproduce and Expected vs. actual behavior**
```
a = roman.RomanNumeral('Ger65', 'F').pitchNames
print(a)
>>> ['D-', 'F', 'A-', 'B']
b = roman.RomanNumeral('Ger65/IV', 'C').pitchNames
print(b)
>>> ['D', 'F', 'A', 'B']
```
b should match a

jacobtylerwalls:
Fixes #585
**Before:** `RomanNumeral` objects without keys would pass a `Scale` object to `_parseRNAloneAmidstAug6()`, which only contemplated `Key` objects for `useScale`, even though elsewhere `useScale` is a scale or key.
**Now:** check the appropriate attribute on scales vs. keys to get to `'major'`.

Fixes #609 
**Before:** augmented sixths must be built from minor scales, but this mode correction was never set in the context of secondary Roman numeral keys.
**Now:** set `RomanNumeral.secondaryRomanNumeralKey` to the parallel minor key when building augmented sixths.

Also! Was poking around in the test directory and ran findNonUTF8.py and testPerformance.py and found some small fixes.

### DIFF
--- a/music21/features/outputFormats.py
+++ b/music21/features/outputFormats.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from music21 import exceptions21
 from music21 import environment
 
--- a/music21/roman.py
+++ b/music21/roman.py
@@ -1921,6 +1921,10 @@ class RomanNumeral(harmony.Harmony):
                 useScale = key.Key(useScale.tonic, 'minor')
                 self.impliedScale = useScale
                 self.useImpliedScale = True
+                # Set secondary key to minor, if any
+                if self.secondaryRomanNumeralKey is not None:
+                    self.secondaryRomanNumeralKey = key.Key(
+                        self.secondaryRomanNumeralKey.tonic, 'minor')
             rm = self._augmentedSixthRegex.match(workingFigure)
             romanNumeralAlone = rm.group(1)
             if romanNumeralAlone in ('It', 'Ger'):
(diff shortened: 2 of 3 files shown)


mscuthbert:
Thank you!  Great fixes!

Btw -- findNonUtf8 was really needed when we were in a Py2/Py3 context, it's not something needed to run anymore really, BUT thank you!

PR #623, opened 2020-09-20
Corrected octaves on pitches derived in IntervalNetwork.nextPitch()

### LINKED ISSUE #319: Scale().next(p) octave shift if p not in Scale
```
>>> s = scale.MajorScale("d")
>>> s.next('C#2')
<music21.pitch.Pitch D2>
>>> s.next('C2')
<music21.pitch.Pitch C#3>
```
Note the error in octave.

reported by Juan (juayuji) on music21list.

jacobtylerwalls:
Fixed #319 

**Before:** next pitches derived from a node in `IntervalNetwork.nextPitch()` would have their octave copied from the node in question, which comes out wrong if stepping to the next pitch crosses an octave boundary.

**Now:** increment or decrement octaves as necessary until upper neighbors are above `pitchOrigin` or lower neighbors are below it. Only one octave adjustment should be necessary in most cases, but I didn't see a reason to rule out compound intervals -- lmk if those aren't supported/necessary to worry about here -- that's what the `while` is there for.

### DIFF
--- a/music21/scale/intervalNetwork.py
+++ b/music21/scale/intervalNetwork.py
@@ -396,7 +396,7 @@ class IntervalNetwork:
                  octaveDuplicating=False,
                  deterministic=True,
                  pitchSimplification='maxAccidental'):
-        # store each edge with and index that is incremented when added
+        # store each edge with an index that is incremented when added
         # these values have no fixed meaning but are only for reference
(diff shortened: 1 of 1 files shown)


mscuthbert:
Looks great!  Thanks!

PR #625, opened 2020-09-21
Transferred element offset to derived TextExpression upon export to XML

### LINKED ISSUE #624: Fine exported to xml with negative offset
**music21 version**
`6.1`

**Problem summary**
A `RepeatExpression` such as a Fine exports to xml with a negative offset, placing it at the measure start.

**Steps to reproduce**

```
s = converter.parse('tinynotation: 4/4 c1')
s.measure(1).storeAtEnd(repeat.Fine())
s.write('xml')
```

**Actual behavior**
```
      <direction>
        <direction-type>
          <words default-y="20" enclosure="none" justify="right">fine</words>
        </direction-type>
        <offset>-40320</offset>
      </direction>
```

**More information**
Issue with offset not being transferred to the replacement `TextExpression` in `MeasureExporter.textExpressionToXml()`.

jacobtylerwalls:
Fixed #624

`Coda` and `MetronomeMark` affected in the same way, although those are far less likely to matter than `Fine`. Still, I applied the same patch.

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -6228,8 +6228,8 @@ class Test(unittest.TestCase):
         s.measure(1).storeAtEnd([c, f, mm])
 
         tree = self.getET(s)
-        direction = tree.find('.//direction')
-        self.assertIsNone(direction.find('offset'))
+        for direction in tree.findall('.//direction'):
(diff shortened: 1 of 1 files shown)


mscuthbert:
Thank you!  Good to have!

PR #630, opened 2020-09-27
Made pickles contemplate quantization keywords

### LINKED ISSUE #530: Converting midi files failed in 32nd notes
When I used converter.parse to load the midi file as below,
image
The result was weird cause some durations could be 0:
image
Then I converted it back to midi file, it looked like
image
Don't know what happened...

jacobtylerwalls:
Fixes #530, or better, refers to it, but I don't think we identified any other TODOs there except this one, so marking it for closing.

**Now**: add `-noQtz` or `-qld8`, etc., to a pickle path to capture any non-default quantization keywords.


mscuthbert:
Wonderful work.  Thank you!

PR #631, opened 2020-09-29
Change instrumentLookup to use simple names

MarkGotham:


mscuthbert:
Thanks.  Fixing the lint problems and making sure that the English name is authoritative by putting it last in the dict comprehension.

Now that Python 2 is not supported, all the `\xf3` etc. can be replaced by ö etc. but a task for another time.

mscuthbert:
Hi -- can you please back out the commits that add many more features to instruments and make them a separate PR w/ justification and timing tests -- adding a lot more pitches to startup will slow down music21 startup -- should probably be reimplemented as a property that is generated upon first read.  and I'm not sure I like the meanPitch attribute.  In any case, does not belong anymore on a branch called tiny_tweaks.  (In general, it is counterproductive to introduce substantitive new features late in a PR review).

music21/instrument.py
```diff
@@ -928,7 +975,6 @@ def __init__(self):
         self.instrumentSound = 'wind.reed.saxophone.baritone'
         self.midiProgram = 67
 
-        self.lowestNote = pitch.Pitch('B-3')
         self.transposition = interval.Interval('M-13')
```
jacobtylerwalls, line 938: hey there Mark, chiming in to say I consider A3 to be the lowest note of the bari sax, which has definitely evolved over time into become more standard than nonstandard. More common IMO than the low B foot on a flute. What say you?

music21/instrument.py
```diff
@@ -128,11 +129,33 @@ class Instrument(base.Music21Object):
...
     * inGMPercMap (bool -- if it uses the GM percussion map)
     * soundfontFn (filepath to a sound font, optional)
+
+    Finally, a few clarifications on matters of range and transposition.
+
+    First, on transposing instruments, the lowestNote is for written pitch
+    (e.g. B3 for the 'English Horn')
+
+    Second, the 'hardLimit' variable is a boolean range constraint indicating that the instrument
+    cannot go beyond this pitch without non-standard modifications such as scordatura.
```
jacobtylerwalls: I get the sense here, but I suggest "extended techniques" because to play a pedal tone on a brass instrument or a subharmonic on a string instrument doesn't require a "modification" per se.

music21/instrument.py
```diff
@@ -128,11 +129,33 @@ class Instrument(base.Music21Object):
...
+    Most instruments have a hard limit on the low end
+    (e.g. a keyboard instrument's lowest key,
+    or the standard tuning of a lowest string on a string instrument)
+    but no such limit at the high side.
+    Exceptions include voices (which do not have a clearly defined lower limits) and
+    keyboard instruments (which do have clearly defined upper limits).
```
jacobtylerwalls: Perhaps something to clarify here that this is still with respect to standard instrument construction? Low B foot on the flute doesn't jeopardize `hardLimitOnLowest` on your construction, nor does a Bösendorfer grand, etc...

music21/instrument.py
```diff
@@ -128,11 +129,33 @@ class Instrument(base.Music21Object):
     * instrumentAbbreviation
     * midiProgram
     * midiChannel
-    * lowestNote (a note object or a string)
-    * highestNote (a note object or a string)
+    * lowestNote (a note object or a string for _written_ pitch)
+    * highestNote (a note object or a string for _written_ pitch)
```
jacobtylerwalls, line 133: Interesting, I don't see that strings are used anywhere for these -- do you?

music21/languageExcerpts/instrumentLookup.py
```diff
@@ -15,169 +16,141 @@
...
-    'a.': 'alto',
     'ac b': 'acoustic bass',
     'ac gtr': 'acoustic guitar',
     'acc': 'accordion',
-    'acc.': 'accordion',
     'accdn': 'accordion',
-    'arp.': 'harp',
+    'arp': 'harp',
     'b dr': 'bass drum',
-    'b.': 'bass',
```
jacobtylerwalls, line 25: is there a reason this wasn't replaced by `'b': 'bass'` -- in your view, is it too ambiguous with `baritone`?

music21/languageExcerpts/instrumentLookup.py
```diff
@@ -15,169 +16,141 @@
...
+    'kl': 'clarinet',
+    'mand': 'mandolin',
     'mar': 'marimba',
-    'mar.': 'marimba',
     'mdln': 'mandolin',
     'mez': 'mezzo-soprano',
-    'mez.': 'mezzo-soprano',
-    'mezz.': 'mezzo-soprano',
+    'mezz': 'mezzo-soprano',
     'ob': 'oboe',
```
jacobtylerwalls, line 92: one vote for `'mz'` while we're here :-)

music21/instrument.py
```diff
@@ -2095,11 +2095,27 @@ def fromString(instrumentString):
...
+    >>> t10 = instrument.fromString('Cl in B-flat')
+    >>> t10
+    <music21.instrument.Clarinet 'Cl in B-flat'>
+    >>> t9.transposition
+    <music21.interval.Interval M-2>
+
+    This should work with or without a terminal period (for both 'Cl' and 'Cl.'):
+
+    >>> t11 = instrument.fromString('Cl. in B-flat')
+    >>> t10.__class__ == t10.__class__
```
jacobtylerwalls, line 2062: ah, we lost `t11`, didn't we?  even easier to read would be `isinstance(t11, instrument.Clarinet)`

music21/instrument.py
jacobtylerwalls: whoops, outdated already, nvm

music21/instrument.py
jacobtylerwalls: also outdated, nvm!

music21/instrument.py
MarkGotham: Thanks @jacobtylerwalls . You're totally right that ranges change over time and will always be hard to pin down across the board. Probably the best approach for questions like this would be to take one (named) source as our guide deflect the questions on to them! In this case, most sources that I've seen still put written Bb as a standard lowest for all saxes.

music21/instrument.py
MarkGotham: Yeah, another tricky one. I'd like to see a variable of this kind in there, but as you say, the distinction between 'standard' and 'extended' is problematic. We could have 'lowestPitchExtended' or similar for B3 on the flute and the like but that still fails to deal properly with the cases you mention and others like the double bass for which the 'lowest' could be
- E2 (sounding E1) where there are four stings, no extension;
- C2 (sounding C1) for the case of four stings plus extension; and 
- B1 (sounding B0) for five stings instruments).

None of this will get into this PR, but we can mull over options for a more substantial contribution in future.

music21/instrument.py
MarkGotham: As above

music21/instrument.py
MarkGotham: Indeed! 

Clearly the description should match the object, though I'm inclined to think that the object should change. Myke's concern over the creation of pitches as part of start-up time would be improved by initialising this variable to pitch name stings only and maybe having a separate method for converting them to pitch objects when needed. I wonder how many people rely on lowestPitch at all, how many of them actually want a pitch and not simply the string, and whether any of those people would have an issue with adding such a line to their code for that conversion.

Apart from the type change, we should also re-name 'lowestNote' as 'lowestPitch'. This is a more accurate description, but again, that would need checking given existing usage.

Again, this is perhaps one for a separate PR.

music21/languageExcerpts/instrumentLookup.py
MarkGotham: Fine by me!

music21/instrument.py
MarkGotham: Thanks! I'll fix that typo and one in the test before. Very happy for the style of this testing run to change, though for consistency's sake many of the existing test above should also be edited in that case.

music21/languageExcerpts/instrumentLookup.py
MarkGotham: Cleaning up this process exposed some more major issues with the method for getting pitch name from string. This is as far as I've got with unpicking it:
https://github.com/cuthbertLab/music21/pull/631/commits/4323fa22980e810dc6c0855e06b109dd42e74c38
Basically the previous 'b.' abbreviation was never called and 'b' (as well as 'a') don't work alongside the handling of transposition. So this PR is an improvement (not a trade-off) but only an incremental one for now. A more substantial fix in needed to move this further along.

music21/instrument.py
jacobtylerwalls: Cool. Out of scope for this PR I suppose.

music21/instrument.py
mscuthbert: All this discussion and we all missed until now -- it's a pitch object not a note object!  :-)  

These will need to become cached-auto-convert strings to pitch objects in the future.  Will switch that at some point (probably for v7 since it'll be a breaking change)

mscuthbert:
Thanks Mark! And Jacob

PR #632, opened 2020-09-30
Fixed #464: Expand repeats when writing MIDI

### LINKED ISSUE #464: converter fails to write midi files correctly
For those who have repeats and changes in `TimeSignature`, converter fails to write them to correct midi files.
Try use case below
```python
    it = corpus.chorales.Iterator(currentNumber=11)
    c = next(it.__iter__())
    c.show() # This score has repeat notation and various TimeSignature, 4/4 -> 3/4 -> 4/4
    ts = c.getTimeSignatures()
    print(len(ts)) # Got 1, expect 2
    c.save('somefile.mid') # Can open somefile.mid from MuseScore 3 directly, but it fails to duplicate the repeating part
    another_c = converter.parse('somefile.mid')
    c.show() # MuseScore 3 fails to show the score
```

jacobtylerwalls:
Now, repeats are expanded when writing MIDI.

### DIFF
--- a/music21/midi/translate.py
+++ b/music21/midi/translate.py
@@ -1775,17 +1775,18 @@ def midiTrackToStream(mt,
 
 def _prepareStreamForMidi(s):
     '''
-    Given a score, prepare it for midi processing.
-    In particular, place MetronomeMark objects at
+    Given a score, prepare it for midi processing. In particular,
+    expand repeats, and place MetronomeMark objects at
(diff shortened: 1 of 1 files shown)


mscuthbert:
Great!  We used to have problems with Expanding repeats if there were none, but I think YOU fixed that earlier :-)  so now this is great!.

jacobtylerwalls:
Hah, I don't think that was me, but good to know!

PR #634, opened 2020-10-04
Join PartStaffs using <staff>

### LINKED ISSUE #613: PartStaffs exported to musicXML as distinct <part> elements
**Version**
`6.1`

**Problem summary**

Two `PartStaff` objects representing a single part will be written to musicXML files as distinct `<part>` elements rather than  a single `<part>` element where `<staff>1</staff>` and `<staff>2</staff>`, etc. Importers will fail to collect them together as a grand staff.

**Steps to reproduce**
```
from music21.musicxml import testPrimitive
s = converter.parse(testPrimitive.pianoStaff43a)
exporter = musicxml.m21ToXml.ScoreExporter(s)
tree = exporter.parse()
exporter.dump(tree.findall('part-list')[0])
```

**Observed result:**
```
<part-list>
  <score-part id="P1">
    <part-name>MusicXML Part</part-name>
  </score-part>
  <score-part id="Pb14b43e5a3a10a087215097122972f89">
    <part-name>MusicXML Part</part-name>
  </score-part>
</part-list>
```

partStaffs

Expected one `<part>` in the `<part-list>` and ultimately `<staff>` tags under the notes.

**More information**

`fromPart()` is going to have to become `PartStaff`-aware, or better yet, a new method `fromPartStaff()` to handle this.

jacobtylerwalls:
Fixes #613 

**Before:** music for grand staff instruments represented by multiple `PartStaff` objects in music21 would be exported to .musicxml as distinct `<part>` elements and would lack a brace to connect them. There was also no link associating the `PartStaffs` save for the part name formula "{originalPart}-Staff{n}".

**Now:**
- On .xml import, a `layout.StaffGroup` object is created to group the entire grand staff
- On .xml export, such `StaffGroup`s are used to merge the `PartStaff` objects back to a single `<part>` tag where child `<note>` elements have a child `<staff>` element, and `<clef>` elements have the attribute `number="2"`, etc.


mscuthbert:
Sorry I let this drop off -- it looks like a very promising method!  I think that we shouldn't rely on the .id element except to be helpful -- people should be allowed to rename their ids whatever they want (.id = 'leftHand', etc.) and some of the dictionaries that keep track of part info, should probably be properties of the ScoreExporter, so that the routine can be broken up into smaller, more testable parts (i.e., that I can still read and understand 10 years from now), but as a general approach, I like and want to endorse continuing with!

jacobtylerwalls:
Awesome. I'm intending to get rid of those dicts entirely, now that I've read up on `PartGroup`. I was thrown off the track originally because on xml import `PartGroup`s aren't created. (The line creating the `PartGroup` was commented out in 3695cbb.)

Hopefully I can get those created on import again without breaking any tests.

jacobtylerwalls:
Added the commit from #674 so that the test for gapful streams would work correctly (without having to use GeneralObjectExporter to dump the root xml to a string and grep through it for forward and backup durations).

music21/musicxml/m21ToXml.py
```diff
@@ -1672,6 +1743,186 @@ def setScoreHeader(self):
...
+                            initialRootCursor += 1
+                            continue
+                        # No gaps found ...
+                        ScoreExporter.moveElements(thisMeasure, initialMeasure, staffNumber)
+                        initialRootCursor += 2  # comment & measure
+
+        # Pass 2: set number on initial clefs, measure attributes
+        # Need the earliest mxAttributes, which may not exist in initialRoot
+        # until moved there in Pass 1, e.g. RH of piano doesn't appear until m. 40
+        for group in joinableGroups:
```
mscuthbert, line 1821: Do any of the joinableGroups affect the parts in other joinableGroups?  Could these be within the same loop as before but still with the two-part workflow?

mscuthbert (commented):
More soon.  This is really great!  Still have more to read.  Thanks!

music21/musicxml/m21ToXml.py
jacobtylerwalls: I suppose what I need to think through is whether there could be a Frankenstein set of `layout.StaffGroup` objects that would enclose the same `PartStaff` within multiple `StaffGroups` (and where those `StaffGroup`s, unlike orchestral scores, consist of all and only `PartStaffs`.) In essence, overlapping piano braces. I'm willing to think that through.

jacobtylerwalls: Aaaand I got all of four seconds before realizing I've just articulated a scenario I can short-circuit and raise an exception for.

jacobtylerwalls: On further thought, those kinds of overlapping StaffGroups might not be Frankenstein-y if handled carefully in some algorithmic composition use cases, so rather than rule them out, I can just check for errors closer to a problem source by raising exceptions in `addStaffTags()`, see https://github.com/cuthbertLab/music21/pull/634/commits/ce7039c0572cf1222c0e154278b581bf442f557c.

music21/musicxml/m21ToXml.py
```diff
@@ -1672,6 +1743,186 @@ def setScoreHeader(self):
...
+                    initialRootCursor = 0
+                    for mNum in range(highestMeasureNumber + 1):
+                        initialMeasure = initialRoot.find(f"measure[@number='{mNum}']")
+                        thisMeasure = root.find(f"measure[@number='{mNum}']")
+                        if thisMeasure is None and initialMeasure is None:
+                            # Gap in both measure sequences
+                            continue
+                        if thisMeasure is None:
+                            # Gap in this measure sequence
+                            initialRootCursor += 2  # Advance for comment & measure in initial seq.
```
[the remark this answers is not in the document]
jacobtylerwalls: I probably need to refactor this to use three iterators, for element in initial, for element in this, for i in range(highestMeasureNum + 1), since I probably shouldn't depend on measure numbers being strictly numeric (suffixes, etc.)

mscuthbert:

music21/musicxml/m21ToXml.py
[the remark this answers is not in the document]
jacobtylerwalls: Gritted my teeth and decided to walk through the <measure> tags in the target <part> and the music21 elements in the PartStaffs being merged while also accounting for measure suffixes. This ballooned the complexity but is hopefully going to catch more cases. Needs some more code comments but wanted to push it up so you could see where this was headed. LMK how it looks; I'm all ears for suggestions to simplify it.

jacobtylerwalls:
I think I need to look more closely at whether notes in staff 2 should automatically be in voice 2. I was probably modeling on a bad example--I don't think this should necessarily be the case. Shouldn't be complex to fix up.

music21/musicxml/m21ToXml.py
```diff
@@ -1672,7 +1779,577 @@ def setScoreHeader(self):
...
+        # Joinable groups must consist of only PartStaffs with Measures
+        for sg in staffGroups:
+            if len(sg) <= 1:
+                continue
+            if not all(isinstance(p, stream.PartStaff) for p in sg):
+                continue
+            if not all(p.getElementsByClass('Measure') for p in sg):
+                continue
+            joinableGroups.append(sg)
+        return joinableGroups
```
jacobtylerwalls, line 1920: Now that this is factored out (awesome), what about using `yield` so that we don't store a list?

mscuthbert:
On last point -- a list of 4-10 items is simpler and faster and more memory efficient than a generator, so I don't think we need to do this.  I started a larger refactor today -- moving out some routines to two new files because reducing m21ToXml file size was already a TODO -- will do another push when I fix the bugs it introduces.  My goal is to introduce pylint restriction on 5,000 lines -- but gotta get Stream down first.  :-)

mscuthbert:
Thanks @jacobtylerwalls  -- huge and amazing work!

PR #636, opened 2020-10-04
WIP: Test on Python 3.8 -- MEI errors!

mscuthbert:
The MagicMock used in MEI tests in Py3.8 iterates over all properties.  This causes an error because .beat raises an exception if there is no TS in sites.

Solution is either try to quickly patch that out.  OR stop raising an exception on beat if no TS in sites and either assume a default of 4/4 OR return -1 or something like that.


jacobtylerwalls:
-1 makes sense to me for a property.

mscuthbert:
I want to go with float('nan') but then people have to remember that float('nan') != float('nan') which is always confusing.  But with -1 I worry that people will just blindly take the average of all the beat's in a corpus (or the beatStrength) and then end up with bogus results.

jacobtylerwalls:
What about `None`?

mscuthbert:
Decided to go with nan and just bit the bullet and added documentation.  This way the return value is still always a float `(isinstance(float('nan'), float) == True)` but it still will bite any software that tries to blindly average numbers.

Issue #629, opened 2020-09-26
problem importing midifile


laurentVeliscek:
**music21 version**

`6.1.0`   

**Problem summary**

Something seems to be broken when importing multi-part midifiles

When I try to show() the imported score, I **always** get warnings about meter (?):
```
meter: WARNING: Found a messed up beam pair <music21.beam.Beams <music21.beam.Beam 1/start>/<music21.beam.Beam 2/partial/right>>, <music21.beam.Beams <music21.beam.Beam 1/stop>/<music21.beam.Beam 2/stop>>, at index 2 of 
[None, None, <music21.beam.Beams <music21.beam.Beam 1/start>/<music21.beam.Beam 2/partial/right>>, <music21.beam.Beams <music21.beam.Beam 1/stop>/<music21.beam.Beam 2/stop>>]
```

Then, I get a MuseScore error message "the xxx.xml file is not a valid MusicXML file".
(here, the error message is in french, but that's the idea...)

If I try to open the file in museScore, it crashes, or if I'm lucky, I can display something, but the score seems to be corrupted.

**Steps to reproduce**
```
## Here's my code 

from music21 import *
score = converter.parse("test.mid")
score.show()
```

**Expected vs. actual behavior**

I wish I could import and manipulate the score transparently.

**More information**

Mac OS Mojave, python 3.7, all modules freshly installed and updated

I tried using many midifiles downloaded from different locations, I tried loading and saving with museScore, but the same results.

In fact, I couldn't find any multi-part midifile that can be imported and showed properly.

To illustrate this, I join 2 screen copies of the file before and after import, using museScore , so you can compare
(notes timing are just messed up...)

I attached the test.mid file (zipped).

Did I miss something ?

Thanks !

![before import](https://user-images.githubusercontent.com/19191214/94345740-8bdadb80-0028-11eb-83ca-3bbb32c305e2.png)
![after import](https://user-images.githubusercontent.com/19191214/94345741-8d0c0880-0028-11eb-973a-abc8deefc25f.png)

[test.mid.zip](https://github.com/cuthbertLab/music21/files/5287031/test.mid.zip)

![warnings](https://user-images.githubusercontent.com/19191214/94346024-2ab40780-002a-11eb-9a2b-4530353e536c.png)

jacobtylerwalls:
heya @laurentVeliscek, thanks for the report.

There are MIDI messages in your file I don't understand. The following is from the second track of music:

```
>>> mf = midi.MidiFile()
[...]

PR #637, opened 2020-10-05
Workaround for null-terminated instrument name

jacobtylerwalls:
Came up in the discussion in #629. Workaround for behavior of MuseScore in writing MIDI files:
https://musescore.org/en/node/310158


mscuthbert:
thanks!

PR #641, opened 2020-10-09
Enforce nonzero duration on non-grace notes when quantized

### LINKED ISSUE #629: problem importing midifile

jacobtylerwalls:
Fixes #629 

**Before:** `Stream.quantize(processDurations=True)` would allow durations of 0 on non-grace notes. By contrast, sequencers and notation programs such as Logic Pro and Finale enforce a minimum nonzero duration when quantizing MIDI:

> For example, a note with a short note value, such as a 32nd note, can only be displayed at its original length if Quantize is set to 32 or shorter. If Quantize is set to 8, the 32nd note is displayed as an eighth note (the shortest note value displayed at that Quantize setting). 
https://support.apple.com/guide/logicpro/quantize-lgcp8535c422/mac

**After:** Unless a note's duration `isGrace`, `Stream.quantize(processDurations=True)` uses the largest value of `quarterLengthDivisors` to derive and set a nonzero duration.

### DIFF
--- a/music21/midi/translate.py
+++ b/music21/midi/translate.py
@@ -2987,7 +2987,7 @@ class Test(unittest.TestCase):
         dirLib = common.getSourceFilePath() / 'midi' / 'testPrimitive'
         # a file with three tracks and one conductor track
         fp = dirLib / 'test11.mid'
-        s = converter.parse(fp, forceSource=True)
+        s = converter.parse(fp)
         self.assertEqual(len(s.parts), 3)
         # metronome marks end up only on the top-most staff
         self.assertEqual(len(s.parts[0].getElementsByClass('MetronomeMark')), 4)
@@ -3227,7 +3227,7 @@ class Test(unittest.TestCase):
 
         dirLib = common.getSourceFilePath() / 'midi' / 'testPrimitive'
         fp = dirLib / 'test16.mid'
-        s = converter.parse(fp, forceSource=True)
+        s = converter.parse(fp)
         self.assertEqual(len(s.parts[0].voices), 2)
         els = s.parts[0].flat.getElementsByOffset(0.5)
         self.assertSequenceEqual([e.duration.quarterLength for e in els], [0, 1])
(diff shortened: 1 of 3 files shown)


mscuthbert:
Looks great -- see note on not using forceSource in tests (one or two are no problem, but if we do it everywhere eventually the whole suite will slow to a crawl) -- waiting for new merges to pass then merging.

Issue #633, opened 2020-10-01
Instrument name abbreviations


MarkGotham:
Better handling of instrument name abbreviations with a regex that can parse typical structures like '{instrument} in {key}' for transposing instruments (in multiple languages) allowing the inclusion of abbreviations like 'a' for 'alto' as distinct from the 'a' in 'Clarinet in A'.

jacobtylerwalls:
Yep, and the reverse like '{key} {instrument}'. However, I'm not sure that we're prevented from handling 'A' -> `Alto` with the abbreviation list because I'm not sure why this comparison is not using `strip()[0]`. As far as I can tell this will always be True for one-word instrument names:

https://github.com/cuthbertLab/music21/blob/942366e46fadca3a73827a360313ab8e0a3d940b/music21/instrument.py#L2090-L2096

jacobtylerwalls:
If we use `strip[0]` we should be able to put `A` in the abbreviations list without ruining Clarinet in A.

jacobtylerwalls:
Never mind, I now see how multiple word substrings are handled by this condition. My suggestion would degrade them.

mscuthbert:
I'm not sure where we are on this issue -- what is the problem and proposed solution?

jacobtylerwalls:
Problem: the abbreviation "A" for Alto can't be added because "Clarinet in A" will normalize to "Alto", since each word of the input string is tested, and the last usually wins:
```
>>> i = instrument.fromString('Clarinet in A')
>>> i
<music21.instrument.Alto 'Clarinet in A'>
```

Solution: None! 

@MarkGotham's proposal is a new regex for instrument keys; I mentioned it will have to handle being inverted.

My first proposed solution interpreted the code comments to be suggesting there was an attempt to measure the number of letters, the longer results being more likely to be correct (clarinet instead of alto) and ties of equal letters breaking for the later one (trumpet in piccolo trumpet). This was wrong--it was measuring the number of words, to ensure that "Tenor Saxophone" beats "Tenor", which my patch broke.

mscuthbert:
I think that "A." and "A" only means Alto [voice] when it's used alone, and even then it's unclear unless one of S, or T appears as another Part name, since people might name their parts A, B, C, D (but likely not to S)  

We are also expecting that in general we're parsing Part Names, not Part Abbreviations.

[...]

PR #643, opened 2020-10-17
Refs #633: Give precedence to longer words; Allow Alto and Bass abbrevs

jacobtylerwalls:
The code comments surrounding this change suggest to me the original intent was to give precedence to longer words when parsing instrument names--or at least a subsequent word with the same number of letters (Piccolo Trumpet). As written, the check `len(bestName.split())` is pointless; it always returns 1 because it's only fed substrings already split.

This means we can support `A` and `B` in the instrument lookups without failing `Clarinet in A`.

~~Finally, and this was a head scratcher, `instrument.fromString('I <3 music saxofono tenore go beavers')` fails for me locally, and I can't figure out why it's not failing on CI or on multiprocessTest.py. Is this cached somewhere? The existing behavior was to normalize to `Tenor`, and with this patch it normalizes to `Saxophone`, hurrah, but I never saw how it could normalize to `TenorSaxophone`. Any leads?~~ Helps if you spell correctly at a shell prompt.

### DIFF
--- a/music21/instrument.py
+++ b/music21/instrument.py
@@ -2006,7 +2006,7 @@ def fromString(instrumentString):
 
     >>> t4 = instrument.fromString('I <3 music saxofono tenore go beavers')
     >>> t4
-    <music21.instrument.TenorSaxophone 'I <3 music saxofono tenore go beavers'>
+    <music21.instrument.Saxophone 'I <3 music saxofono tenore go beavers'>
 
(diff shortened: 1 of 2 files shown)


jacobtylerwalls:
Closing because working through the tenor saxophone example helped me understand how this check is currently working and DOES feed multiple word substrings to `len(bestName.split())`

PR #651, opened 2020-10-26
Create writeRoman.py

MarkGotham:
Writes Roman text files given a music21 stream of Roman numerals


MarkGotham:
I'm not sure why lines 212, 4 fall foul of lint's W0106(expression-not-assigned). All the same, I'm very happy for this to be re-cast some other way that lint prefers

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,316 @@
...
+            self.prepList()
+
+        if not fileName:  # Never an empty string: placeholders set by prepPreamble as needed.
+            fileName = f'{self.composer}_-_{self.title}'
+            fileName = fileName.replace('.', '')
+            fileName = fileName.replace(':', '')
+            fileName = fileName.replace(' ', '_')
+
+        text_file = open(os.path.join(outPath, f'{fileName}.txt'), "w")
+        [text_file.write(entry + "\n") for entry in self.preamble]
```
jacobtylerwalls, line 212: Hi Mark. Pylint will be happier if you don't wrap this in a list comprehension. It's seeing you created a list with `[]` and didn't assign it to a variable.

It's not really more verbose to just break this out into:
```
for entry in self.preamble:
    text_file.write(entry + '\n')
```

MarkGotham: Thanks! I'll do that.

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,332 @@
...
+    if type(beat) not in options:
+        raise ValueError(f'Beat, (currently {beat}) must be one of {options}.')
+
+    if type(beat) in [str, fractions.Fraction]:
+        beat = float(beat)
+
+    if int(beat) == beat:
+        return int(beat)
+    else:
+        return round(float(beat), roundValue)
```
jacobtylerwalls, line 265: Already did `float()` once.

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,332 @@
...
+                 analyst: str = '',
+                 proofreader: str = '',
+                 note: str = ''
+                 ):
+
+        self.score = score
+
+        self.romanNumerals = self.score.parts[0].recurse().getElementsByClass('RomanNumeral')
+        self.firstMeasureNumber = self.romanNumerals[0].measureNumber
+        self.lastMeasureNumber = self.romanNumerals[-1].measureNumber
```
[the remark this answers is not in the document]
jacobtylerwalls: On the issue ticket Myke was worried about measure numbers with suffixes, so it won't be quite as terse as this, probably need a function for the sorting key.

jacobtylerwalls (commented):
Hi Mark! I'll let Myke comment on the approach, but I left you some comments that I think will make this easier to review, test, and maintain. Maybe @napulen has thoughts too? :-)

music21/romanText/writeRoman.py
mscuthbert: Exactly.  And Myke would be happier too.  It creates a list of Nones that is assigned to nothing so it's memory inefficient.

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,316 @@
...
+
+# ------------------------------------------------------------------------------
+
+class WriteRomanException(exceptions21.Music21Exception):
+    pass
+
+
+# ------------------------------------------------------------------------------
+
+class RnWriter:
```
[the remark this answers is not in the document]
mscuthbert: If it's a writer or otherwise a non-musical element that's optional.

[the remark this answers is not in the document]
MarkGotham: Noted!

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,332 @@
...
+                 composer: str = '',
+                 title: str = '',
+                 analyst: str = '',
+                 proofreader: str = '',
+                 note: str = ''
+                 ):
+
+        self.score = score
+
+        self.romanNumerals = self.score.parts[0].recurse().getElementsByClass('RomanNumeral')
```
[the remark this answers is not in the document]
mscuthbert: I like that.

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,332 @@
...
+
+# ------------------------------------------------------------------------------
+
+    # To write
+
+    def writeRomanText(self,
+                       outPath: str = '.',
+                       fileName: str = ''):
+        '''
+        Writes the combined information to a 'Roman text' file (using the extension .txt).
```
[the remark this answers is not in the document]
MarkGotham: Yeah. The rntxt extension only really serves to confuse operating systems ;)

music21/romanText/writeRoman.py
jacobtylerwalls: How many `None`s could a NoneCheck check if a `None`...

music21/romanText/writeRoman.py
MarkGotham: Oops!

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,332 @@
...
+
+    options = [str, int, float, fractions.Fraction]
+
+    if type(beat) not in options:
+        raise ValueError(f'Beat, (currently {beat}) must be one of {options}.')
+
+    if type(beat) in [str, fractions.Fraction]:
+        beat = float(beat)
+
+    if int(beat) == beat:
```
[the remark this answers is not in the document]
MarkGotham: As in `if type(beat) == int` -- I find that fails, leaving a trail of b3.0 cases

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,332 @@
...
+            if x in self.analysisDict.keys():
+                self.combinedList.append(self.analysisDict[x])
+
+# ------------------------------------------------------------------------------
+
+    # To write
+
+    def writeRomanText(self,
+                       outPath: str = '.',
+                       fileName: str = ''):
```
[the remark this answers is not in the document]
MarkGotham: Fair enough. I'm happy to adjust if necessary, though prefer it this way given the option.

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,332 @@
...
+        '''
+
+        if self.composer:
+            self.preamble.append(f'Composer: {self.composer}')
+        else:  # default unless set
+            if self.score.metadata.composer:
+                self.composer = self.score.metadata.composer  # overwrite
+                self.preamble.append(f'Composer: {self.composer}')
+            else:
+                self.composer = 'Unknown'
```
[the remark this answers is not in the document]
MarkGotham: So `composer: str = 'Unknown'` and then change if available?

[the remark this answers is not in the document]
jacobtylerwalls: I was thinking set the fallback in the constructor (either Unknown or '', whichever you're want in the output), which means any user input overrides it, and then do:
self.composer = self.container.metadata.composer or composer
(where "or composer" falls back to the keyword arg if the metadata entry is None)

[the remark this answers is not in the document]
MarkGotham: Ok thanks, but reversed so the user entry takes priority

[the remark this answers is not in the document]
jacobtylerwalls: Ah, I see, yes. In that case the keyword argument would really have to default to `''` and not Unknown for this to work in one line, but if it needs to be Unknown, you get the drift. (See how even after reviewing it I was still a little shaky about how the priority worked!)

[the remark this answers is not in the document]
MarkGotham: Thanks -- yup I'll get on with tightening this all up a bit.

MarkGotham:
> Hi Mark! I'll let Myke comment on the approach, but I left you some comments that I think will make this easier to review, test, and maintain. Maybe @napulen has thoughts too? :-)

Thanks for the review @jacobtylerwalls! And yes please @napulen -- do weigh in. (Our conversation at ISMIR prompted all this).

music21/romanText/writeRoman.py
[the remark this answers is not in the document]
jacobtylerwalls: Oh got it, I see what you're doing now. You're accepting a string above, casting to float, then checking if casting to int is ==, and the casting to int again. It's not wrong, it's just hard to pick up what it's doing IMO as it's currently factored. Could probably get `if int(beat) == beat` up first, not sure.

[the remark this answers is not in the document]
MarkGotham: True -- ints end up going on a round about journey only end up back as ints. Now handled at the start.

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,332 @@
...
+        self.romanNumerals = self.score.parts[0].recurse().getElementsByClass('RomanNumeral')
+        self.firstMeasureNumber = self.romanNumerals[0].measureNumber
+        self.lastMeasureNumber = self.romanNumerals[-1].measureNumber
+
+        self.timeSignatures = []
+        self.timeSigMeasureDict = {}
+        self.getTSs()
+
+        # Metadata / preamble
+        self.composer = composer
```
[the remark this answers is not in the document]
MarkGotham: Done except for analysts and proofreaders which are currently unprocessed in the roman text parser, so at present the only way to have this value is to enter it (but I've left in a comment for adding this in future)

music21/romanText/writeRoman.py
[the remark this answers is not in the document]
MarkGotham: Now handling measures sequentially and using measureNumberWithSuffix

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,348 @@
...
+                                note='None of this metadata is accurate')
+        self.assertEqual(monteAltered.composer, 'Montius Claudeverdi')
+        self.assertEqual(monteAltered.title, 'Don Giovanni')
+        self.assertEqual(monteAltered.analyst, 'I, Claudius')
+        self.assertEqual(monteAltered.proofreader, 'Berg')
+        self.assertEqual(monteAltered.note, 'None of this metadata is accurate')
+
+# ------------------------------------------------------------------------------
+
+    def testMinimialCase(self):
```
napulen, line 287: "testMinimialCase" -> "testMinimalCase"?

music21/romanText/writeRoman.py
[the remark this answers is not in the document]
napulen: Interesting. Wearing my user "hat", my first reaction was to do `rna.writeRomanText(fileName='something.rntxt')`. 

It kind of surprised me to get `something.rntxt.txt`.

May I ask what issues have you experienced with OSes and the `rntxt` extension?

music21/romanText/writeRoman.py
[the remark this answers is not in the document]
napulen: +1 on the `fp` for the same reason @jacobtylerwalls pointed out, and because I think that some validation should be done on `fileName` if the user already provides an extension (e.g., `fileName=output.txt` should probably result in `output.txt` on the filesystem and not `output.txt.txt`). As this gets very tricky in practice, maybe `fp` is an easier solution.

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,348 @@
...
+                self.combinedList.append(f'\nTime Signature: {firstTS.ratioString}')
+                if len(tsThisMeasure) > 1:
+                    msg = 'further time signature change(s) unprocessed: ' \
+                          f'{[x.ratioString for x in tsThisMeasure[1:]]}'
+                    self.combinedList.append(f'Note: {msg}')
+
+            # RomanNumerals
+            measureString = ''  # Clear for each measure
+            rnsThisMeasure = [r for r in thisMeasure.getElementsByClass('RomanNumeral')]
+            if rnsThisMeasure:  # probably redundant
```
napulen, line 129: Definitely redundant

```python
for rn in rnsThisMeasure:
```
handles this as it is

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,348 @@
...
+                for rn in rnsThisMeasure:
+                    chordString = self.getChordString(rn)
+                    measureString = rnString(measure=m,
+                                             beat=rn.beat,
+                                             chordString=chordString,
+                                             inString=measureString,  # Creating update
+                                             )
+
+            self.combinedList.append(measureString)
```
napulen, line 138: maybe
```python
if measureString:
    self.combinedList.append(measureString)
```
I noticed that when there is no RNA annotation for a specific measure, a blank line is generated, but that doesn't seem to be the convention at When-in-Rome/Corpus

napulen:
Left a few comments here and there. Thanks @MarkGotham for this code and @jacobtylerwalls for the review :-).

mscuthbert:
Hi Mark -- can you use descriptive commit messages?  Update writeRoman doesn't tell me what has been changed in each commit.  Thanks!

music21/romanText/writeRoman.py
mscuthbert: definitely redundant.

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,332 @@
...
+            roundValue: int = 2):
+    '''
+    Converts beats to integers if possible, and otherwise to rounded decimals.
+    Accepts input as string, int or float.
+    '''
+
+    options = [str, int, float, fractions.Fraction]
+
+    if type(beat) not in options:
+        raise ValueError(f'Beat, (currently {beat}) must be one of {options}.')
```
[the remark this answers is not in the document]
mscuthbert: Also, this should be `if not isinstance(beat, options)` -- this should have triggered a pylint unidiomatic-typecheck error.  A subclass of any of the classes in options is definitely acceptable.

music21/romanText/writeRoman.py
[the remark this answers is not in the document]
mscuthbert: I'd like to get all the metadata information from the Stream.metadata tag -- users should not be using a different set of commands to add metadata for romanText writing than they would for MusicXML/MIDI/MEI/etc. writing.  These initial values should be removed.

Adding processing of analysts and proofreaders in RomanText parsing would be welcome and make for a Separate PR

music21/romanText/writeRoman.py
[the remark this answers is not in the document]
mscuthbert: Generally we make the initial writer be a general tool that if it's a non-Stream, we put it in a Measure, if it's a Measure (or generic Stream) we wrap it in a Part, and then if it's a Part, we wrap it in a Score, so that any music21 object can try to call "el.show('romanText')"  [btw -- does `show()` work, in addition to write?]

music21/romanText/writeRoman.py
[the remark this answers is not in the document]
mscuthbert: Should use `fp` like all other writer options.

One thing I've noticed with my students -- who are probably more like general music21 users -- is that more and more they're not really knowing about the encodings they're using; they just assume that an encoding is an encoding is an encoding, so the format information needs to be as much the same for each as possible.

music21/romanText/writeRoman.py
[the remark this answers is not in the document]
mscuthbert: The default output filename should be `.rntxt` -- this is what music21 expects to be able to parse.  It should be possible to do converter.parse(sc.show('romanText')) and get back the original score (or something very close).  If people want to use a different extension for fp that's fine.  In the same way the musicxml has gone away from the generic .xml suffix in favor of .musicxml, we should be using the suffix to specify that this isn't any old txt file but a romanText file.

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,348 @@
...
+    Tests for both main analysis cases - one full, one partial - and for a template.
+    Additional test for smaller static functions.
+    '''
+
+    def testTwoCorpusPiecesAndTwoCorruptions(self):
+
+        from music21 import corpus
+        from music21 import meter
+
+        scoreBach = corpus.parse('bach/choraleAnalyses/riemenschneider001.rntxt',
```
mscuthbert, line 236: How much additional time is this parsing adding to the music21 test suite?  I believe that RNTXT parsing is very slow.  There are much smaller test files that could be used.

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,348 @@
...
+                                 format='RomanText')
```
mscuthbert, line 237: format is unnecessary with an .rntxt extension

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,348 @@
...
+
+        wonkyBach = RnWriter(scoreBach)
+        wonkyBach.prepList()
+        self.assertEqual(wonkyBach.combinedList[8], '\nTime Signature: 10/8')
+        self.assertEqual(wonkyBach.combinedList[9],
+                         'Note: further time signature change(s) unprocessed: [\'5/8\']')
+
+        # --------------------
+
+        fullMonte = corpus.parse('monteverdi/madrigal.3.1.rntxt',
```
mscuthbert, line 259: similarly, too that this is likely too slow -- need timing.

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,348 @@
...
+        from music21 import meter
+
+        scoreBach = corpus.parse('bach/choraleAnalyses/riemenschneider001.rntxt',
+                                 format='RomanText')
+
+        rnaBach = RnWriter(scoreBach)
+        rnaBach.prepList()
+
+        self.assertEqual(rnaBach.combinedList[0], 'Composer: J. S. Bach')
+        self.assertEqual(rnaBach.combinedList[4], '\nTime Signature: 3/4')
```
mscuthbert, line 243: This test is fragile against adding additional elements such as Analyst, Proofreader, etc.  And order does not fully matter.  Should just check for it in the system.

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,348 @@
...
+                                composer='Montius Claudeverdi',
+                                title='Don Giovanni',
+                                analyst='I, Claudius',
+                                proofreader='Berg',
+                                note='None of this metadata is accurate')
+        self.assertEqual(monteAltered.composer, 'Montius Claudeverdi')
+        self.assertEqual(monteAltered.title, 'Don Giovanni')
+        self.assertEqual(monteAltered.analyst, 'I, Claudius')
+        self.assertEqual(monteAltered.proofreader, 'Berg')
+        self.assertEqual(monteAltered.note, 'None of this metadata is accurate')
```
mscuthbert, line 283: Remove and get from metadata

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,348 @@
...
+# Name:         romanText/writeRoman.py
+# Purpose:      Writes Roman text files given a music21 stream of Roman numerals
+#
+# Authors:      Mark Gotham
+#
+# Copyright:    Copyright © 2020 Michael Scott Cuthbert and the music21 Project
+# License:      BSD, see license.txt
+# ------------------------------------------------------------------------------
+'''
+Writes Roman text files given a music21 stream of Roman numerals
```
mscuthbert, line 12: RomanText -- should have a reference to the specification format (which I believe we know the authors...)

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,348 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+# Name:         romanText/writeRoman.py
+# Purpose:      Writes Roman text files given a music21 stream of Roman numerals
```
mscuthbert, line 4: RomanText

mscuthbert (changes requested):
This is coming along well, but the documentation needs examples which function also as doctests, the CI Coverage needs to increase (for new code, we're looking for 95% or higher coverage), and the system needs to be integrated with the rest of music21's converter/show/write methods in the converter/subConverters package.  Currently RomanText is an output only format; converting it to input output format.  THANKS!

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,348 @@
...
+        self.fileName = self.fileName.replace(' ', '_')
+
+        self.combinedList = [f'Composer: {self.composer}',
+                             f'Title: {self.title}',
+                             f'Analyst: {self.analyst}',
+                             f'Proofreader: {self.proofreader}']
+        if self.note:  # optional
+            self.combinedList.append(f'Note: {self.note}')
+
+        self.currentKey = ''
```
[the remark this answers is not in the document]
MarkGotham: It's a string, so I've renamed the attributed as well as adding a type hint

music21/romanText/writeRoman.py
MarkGotham: Replaced with last line (more robust than any number counted up). I don't see an internal system option now that we're going straight to string lists without storing intermediate steps (as advised in first round review from @jacobtylerwalls)

music21/romanText/writeRoman.py
MarkGotham: Replaced cases with smallest files in the Bach/Monteverdi collections. Now the test (9 of them including the new ones run in < 0.3 seconds). How's that? Is there a benchmark?

music21/romanText/writeRoman.py
MarkGotham: + tests for the new normal

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,348 @@
...
+        if self.score.metadata.movementName:
+            if self.score.metadata.movementName != self.score.metadata.title:
+                workingTitle.append(self.score.metadata.movementName)
+
+        if len(workingTitle) > 0:
+            self.title = ' '.join(workingTitle)
+
+# ------------------------------------------------------------------------------
+
+    def prepList(self):
```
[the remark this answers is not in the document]
MarkGotham: Gone

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,348 @@
...
+        Prepares a sequential list of text lines, with timeSignatures and romanNumerals
+        adding this to the (already prepared) metadata preamble ready for printing.
+        '''
+
+        for m in self.measureNumbers:
+
+            thisMeasure = self.container.measure(m)
+
+            # TimeSignatures
+            tsThisMeasure = [t for t in thisMeasure.getElementsByClass('TimeSignature')]
```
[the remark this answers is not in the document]
MarkGotham: To include a warning for any not processed. Unlikely eventuality in most plausible cases, (especially as rntxt currently only supports start of measure TS changes), but still

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,348 @@
...
+            if tsThisMeasure:
+                firstTS = tsThisMeasure[0]
+                self.combinedList.append(f'\nTime Signature: {firstTS.ratioString}')
+                if len(tsThisMeasure) > 1:
+                    msg = 'further time signature change(s) unprocessed: ' \
+                          f'{[x.ratioString for x in tsThisMeasure[1:]]}'
+                    self.combinedList.append(f'Note: {msg}')
```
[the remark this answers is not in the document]
MarkGotham: OK. Currently -- changed to
+ line breaks between measure entries -- clearly needed, so kept
+ additional line break before a time signature change -- neater but not essential, so removed)

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,348 @@
...
+            thisMeasure = self.container.measure(m)
+
+            # TimeSignatures
+            tsThisMeasure = [t for t in thisMeasure.getElementsByClass('TimeSignature')]
+            if tsThisMeasure:
+                firstTS = tsThisMeasure[0]
+                self.combinedList.append(f'\nTime Signature: {firstTS.ratioString}')
+                if len(tsThisMeasure) > 1:
+                    msg = 'further time signature change(s) unprocessed: ' \
+                          f'{[x.ratioString for x in tsThisMeasure[1:]]}'
```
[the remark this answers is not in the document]
MarkGotham: Fair enough; one thing at a time

music21/romanText/writeRoman.py
MarkGotham: Replaced cases with smallest files in the Bach/Monteverdi collections. Now the test (9 of them including the new ones run in < 0.3 seconds).

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,348 @@
...
+                       fileName: str = ''):
+        '''
+        Writes the combined information to a 'Roman text' file (using the extension .txt).
+        '''
+
+        if not self.combinedList:
+            self.prepList()
+
+        f = fileName or self.fileName
+        text_file = open(os.path.join(outPath, f'{f}.txt'), "w")
```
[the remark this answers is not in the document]
MarkGotham: Noted and moved. I'll commit that file separately

MarkGotham:
Thanks for the comments @napulen and this review @mscuthbert.

**Coverage**
Absolutely. I'm aware of the coverage drive and aiming high. The part that's been missing in coverage is the final stage of actually writing out the file (now moved as requested).

**Commit messages**
Sure thing. For single file changes I've taken the 'update' format and put details in the description but I'll promote them to the commit message itself in future.

**Test and speed**
Replaced cases with smallest files in the Bach/Monteverdi collections. Now the test (9 of them including the new ones) run in < 0.3 seconds.

music21/romanText/writeRoman.py
MarkGotham: Ah thanks! Missed this before.

napulen:
> This is coming along well, but the documentation needs examples which function also as doctests, the CI Coverage needs to increase (for new code, we're looking for 95% or higher coverage), and the system needs to be integrated with the rest of music21's converter/show/write methods in the converter/subConverters package. Currently RomanText is an output only format; converting it to input output format. THANKS!

I think the last sentence here should read: *Currently RomanText is an **input**-only format; converting it to input-output format*.

I agree that integration into `music21.stream.Stream.write(fmt='rntxt')` would be great!

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,367 @@
...
+from music21 import stream
+
+
+# ------------------------------------------------------------------------------
+
+class RnWriter:
+    '''
+    Extracts the relevant information from a stream of Roman numeral objects for
+    writing to text files in the 'RomanText' format.
+
```
napulen, line 32: I am not very good at writing, but maybe something like this would be good in here:

---

`Part` and `Score` streams are supported. When the stream is a `Part`, its inner Roman numeral objects are used to generate the output `RomanText`. When the stream is a `Score`, the Roman numeral objects within the *first* `Part` of the `Score` are used to generate the output `RomanText`.

---

MarkGotham: Good call! More explanation and demonstrations coming right up for the class itself.

MarkGotham:
Hi @mscuthbert and all! 

**The short version:** new code works and is covered; integration with subConverters still an issue.

**The not-so-short version:** Back to passing tests including coverage because writeRoman itself is fully covered. The addition to subConverter is not covered nor quite right as it stands (and btw this is not the only area of subConverter that's not covered). I'm happy to work more on this further but I'd need pointers on what exactly is needed / best practice in this case.

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,408 @@
...
+        # (Re-)assign metadata in the normal way
+        scoreMonte.metadata.title = 'Fake title'
+        scoreMonte.metadata.movementNumber = 123456789
+        scoreMonte.metadata.movementName = 'Fake movementName'
+
+        adjustedMonte = RnWriter(scoreMonte)
+        self.assertEqual(adjustedMonte.title, 'Fake title - No.123456789: Fake movementName')
+
+    @unittest.expectedFailure
```
jacobtylerwalls, line 349: `unittest.expectedFailure` is not quite right for testing expected behavior. It's more for when the code is broken and hasn't been fixed yet, or there is some dependency causing trouble. Here, you're testing expected behavior!

Better to use:
```
with self.assertRaises(TypeError):
    doStuffThatWillRaise()
```

jacobtylerwalls (commented):
Thanks, Mark! I'll be glad to have a look at the subconverter integration soon, in hopes I can be helpful. Meantime here's a tip for you on the tests.

MarkGotham:
Thanks for the tip @jacobtylerwalls – duly noted!

music21/converter/subConverters.py
```diff
@@ -1143,6 +1144,22 @@ def parseFile(self, fp, number=None):
         rtf.close()
         romanTextTranslate.romanTextToStreamScore(rtHandler, self.stream)
 
+    def write(self, obj, fp=None):
```
jacobtylerwalls, line 1147: To match the other `write()` signatures in other writers, you can use:
```
def write(self, obj, fmt, fp=None, subformats=None, **keywords):  # pragma: no cover
```
and that should take care of the `TypeError`s I was seeing. MSC might chime in on whether you can help yourself to the `# pragma: no cover` like elsewhere.

jacobtylerwalls (commented):
Very close: all I had to do was change one line, and the `.write('rntxt')` performed nicely!

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,408 @@
...
+                self.container = obj.parts[0]
+            else:  # score with no parts
+                self.container = obj
+        elif isinstance(obj, stream.Part):
+            self.container = obj
+        elif isinstance(obj, stream.Measure):
+            self.container = stream.Part()
+            self.container.insert(0, obj)
+        else:
+            msg = 'This class must be called on a must be stream (Score, Part, or measure)'
```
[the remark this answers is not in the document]
MarkGotham: ... oops ...

music21/converter/subConverters.py
MarkGotham: Thanks @jacobtylerwalls! Duly replicated and committed. I've taken the liberty of including the `# pragma: no cover` on the basis that it affects a lot of `converter.subConverter` and is perhaps one for a separate PR.

music21/romanText/writeRoman.py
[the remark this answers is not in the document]
mscuthbert: Hi -- as I said before, we don't restrict the type of Music21Object that can call .show() in a particular format (except that the testing for Opus objects leaves something to be desired -- but that's a bug).  For instance:

```
In [1]: n = note.Note()                                                         

In [2]: n.show('text')                                                          
<music21.note.Note C>

In [3]: n.show('lily')                                                          
GNU LilyPond 2.20.0
...
Success: compilation successfully completed
```

The solution is to first check, "if not obj.isStream" and put it into a Measure.  Then check if it is a generic Stream, if not, put all its elements in a Measure.  Then check if it's a Measure and put it in a Part.  Then check if it's a Part, and then put it in a Score.  Then you're ready to show a standard Score.   Again, look at how musicxml writing works -- best always to start with a check for the largest deficiencies (like it's a Note object) and then move up from there.  

But note that RomanText allows more than one score to be defined in an `.rntxt` file and are parsed as an Opus (see `romanText.translate.romanTextToStreamOpus`) so the output format should be Opus-aware also.  

These are things that I'd have to do later when inevitably someone files a bug report about it, so trying to head it off in advance.

Thanks!

music21/converter/subConverters.py
```diff
@@ -1144,7 +1144,7 @@ def parseFile(self, fp, number=None):
         rtf.close()
         romanTextTranslate.romanTextToStreamScore(rtHandler, self.stream)
 
-    def write(self, obj, fp=None):
+    def write(self, obj, fmt, fp=None, subformats=None, **keywords):  # pragma: no cover
```
mscuthbert, line 1147: Nit-pick: Single quotes around "w" and "\n" below

mscuthbert:
Hi y'all -- the documentation tests in the newer additions still are not helpful to let a user (i.e., ME!) know what is being returned by functions.  In the __init__ I see

```
    >>> s = stream.Score()
    >>> rnScoreOnly = romanText.writeRoman.RnWriter(s)
    >>> p = stream.Part()
    >>> rnPart = romanText.writeRoman.RnWriter(p)
    >>> s.insert(p)
    >>> rnScoreWithPart = romanText.writeRoman.RnWriter(s)
    >>> m = stream.Measure()
    >>> rnMeasure = romanText.writeRoman.RnWriter(m)
```

What is rnScoreOnly?  Is it a list?  a string?  something else?  What can I do with it?  Are these likely to be blank since we're passing in empty strings?  Please look at how docstrings are used in the rest of music21 and in the music21 docs to guide you on what we're going to need in the future.  Thanks.

(note that the type signature for RnWriter is incorrect -- it does not show that a Measure can be passed in).

music21/romanText/writeRoman.py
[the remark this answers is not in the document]
jacobtylerwalls: should be `class RnWriter(prebase.ProtoM21Object):`

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,397 @@
...
+                firstTS = tsThisMeasure[0]
+                self.combinedList.append(f'Time Signature: {firstTS.ratioString}')
+                if len(tsThisMeasure) > 1:
+                    unprocessedTSs = [x.ratioString for x in tsThisMeasure[1:]]
+                    msg = f'further time signature change(s) unprocessed: {unprocessedTSs}'
+                    self.combinedList.append(f'Note: {msg}')
+
+            # RomanNumerals
+            measureString = ''  # Clear for each measure
+            rnsThisMeasure = [r for r in thisMeasure.getElementsByClass('RomanNumeral')]
```
jacobtylerwalls, line 160: It's subtle, but you'll see a performance improvement from not using a list comprehension. `getElementsByClass` returns a generator, so you can get right to this:
`rnsThisMeasure = thisMeasure.getElementsByClass('RomanNumeral')`

You'll save the memory overhead of storing the entire list as well as the generator overhead from iterating an extra time: once to apply `r for r` and once to do `for rn in myListComprehension`

FYI @mscuthbert there's a pylint check we could consider turning back on that checks for unnecessary list comprehensions.

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,397 @@
...
+        # --------------------
+
+        scoreBach.parts[0].measure(3).insert(0, meter.TimeSignature('10/8'))
+        scoreBach.parts[0].measure(3).insert(1, meter.TimeSignature('5/8'))
+
+        wonkyBach = RnWriter(scoreBach)
+
+        self.assertEqual(wonkyBach.combinedList[9], 'Time Signature: 10/8')
+        self.assertEqual(wonkyBach.combinedList[10],
+                         'Note: further time signature change(s) unprocessed: [\'5/8\']')
```
jacobtylerwalls, line 327: To address Myke's concern with it being brittle to index specific lines (in case analyst, proofreader, etc. supported later), wouldn't it just be sufficient to do:
```
self.assertIn('Note: further...', wonkyBach.combinedList)
```
You might need to add a trailing newline.

music21/romanText/writeRoman.py
mscuthbert: I think it's turned off because there is one side effect of iterating over a generator -- all the RNs in thisMeasure now have their activeSites set to thisMeasure so if later you do, thisMeasure._elements[20] it will have the activeSite set.  BUT we're definitely not doing it here, and probably we should have the pylint setting turned on and just disable it the one or two places that it is actually not unnecessary.

There's another pylint thing with statement-has-no-effect, that I think I had disabled because of how filters used to work that pylint could not detect, that I think I can turn on.

There are a bunch of pylint settings that I want to turn back on, BUT haven't had time to refactor enough places to do it.  Mostly relating to statement complexity, etc. -- I really want to encourage code that my students can read.

music21/romanText/writeRoman.py
mscuthbert: Yes, and then it would also be possible to check that the TS Note is on the line after the Time Signature setting.  

Five years ago I would have accepted this, but since then I've learned more about what makes me have to return to the code later when I can't get a hold of the original writers, so I have gotten pickier over time.  :-)

music21/romanText/writeRoman.py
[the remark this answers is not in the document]
MarkGotham: Ok thanks for this detail. I've had a go at implementing this in the latest round. Grateful as always for the check over.

Three issues that are probably worth raising here:

1.
Currently the self.container variable is aiming for a part (sic, not score) with measures containing RNs. With metadata handled separately, is there any need for a score object?

2.
Despite all this object flexibility, I'm thinking of a user who's trying to write rntxt from what they think is a score or part but is actually some non-steam object. Wouldn't that user be better served by a type error flagging up the mistake than by putting the object in a fake measure and obligingly writing nonsense as we now would? Given the consistency issue for .show(), this is a wider question of music21's philosophy I suppose.

3.
Wrt Opus, are we 
- sure that supporting Opus within a single rntxt file is a good idea?
- perhaps already irreversibly invested in it anyway?
- open to creating Opus objects instead by parsing multiple rntxt files explicitly assembled by a user.
It's not obvious that Opus rntxt files are worth additional dev. -- does this contribute anything more than copying the shared metadata from one rntxt to another?

MarkGotham:
Summary of the above: all requests implemented and good except for:
- coverage of Opus (see discussion above)
- back to some brittle tests in the docs ('assertIn' not application there. Suggestions welcome!)
Thanks!

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,481 @@
...
+# ------------------------------------------------------------------------------
+
+class RnWriter(prebase.ProtoM21Object):
+    '''
+    Extracts the relevant information from a stream of Roman numeral objects for
+    writing to text files in the 'RomanText' format (rntxt).
+    Writing rntxt is handled externally in the subConverters module so
+    most users will never need to call this class directly, and will only invoke
+    it indirectly through .write('rntxt').
+    Possible exceptions include users who want to convert Roman numberals into rntxt type
```
jacobtylerwalls, line 38: typo

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,481 @@
...
+    strings and want to work with those strings directly, without writing to disc.
```
jacobtylerwalls, line 39: Not to be too American about it, but I think for computers it's always dis**k**.

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,481 @@
...
+
+    For consistency with .show() across music21, this class theoretically callable on
```
jacobtylerwalls, line 41: this class **is**

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,481 @@
...
+    any type of Music21Object.
+    Within this wide range, more plausibly use cases including calling on
```
jacobtylerwalls, line 43: plausible ... include

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,481 @@
...
+
+    >>> s = stream.Score()
+    >>> s.insert(0, m)
+    >>> rnWriterFromScoreWithoutPart = romanText.writeRoman.RnWriter(p)
+    >>> rnWriterFromScoreWithoutPart.combinedList == rnWriterFromMeasure.combinedList
+    True
+
+    In all cases, handling of composer and work metadata is
+    extracted from score metadata wherever possible.
+    A metadata.composer entry will register directly as will any enties for
```
jacobtylerwalls, line 99: entities

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,481 @@
...
+    workTitle, movementNumber, and movementName (see the prepTitle method for details).
```
jacobtylerwalls, line 100: you can add `:meth:`~music21.romanText.writeRoman.RnWriter.prepTitle` to provide a link

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,481 @@
...
+
+    def _makeContainer(self,
+                       objs: list):
+        '''
+        Makes a placeholder container for the usual cases where this class is called on
+        generic- or non-stream object.
+        '''
+        m = stream.Measure()
+        for x in objs:
+            m.insert(0, x)
```
jacobtylerwalls, line 160: This will prepend the element each time, essentially reversing the order. What about just `.append()`?

jacobtylerwalls (commented):
About the Opus thing, I think there are two options:
- support it -- extract everything in the init after you determine the object type to a new function, that call that for score in opus
- don't support it -- log a warning after simply taking the first score in the opus, just like you currently take only the first part in the score. this still provides some feedback and prevents the bug ticket MSC is worried about ("it crashed...")

About the brittle tests: which ones are you referring to? I think the [-1] indexing is okay. It's just the specific indices in the middle that might move left and right.

music21/romanText/writeRoman.py
MarkGotham: Thanks @jacobtylerwalls . 

By potentially brittle tests, yes I meant the 0 and -1 cases – glad you agree that they should be fine. 

The next commit includes opus, without separating off a new function. Grateful for your views on this solution.

Thanks again.

music21/romanText/writeRoman.py
```diff
@@ -174,8 +173,10 @@ def __init__(self,
     def _makeContainer(self,
                        objs: list):
         '''
-        Makes a placeholder container for the usual cases where this class is called on
-        generic- or non-stream object.
+        Makes a placeholder container for the unusual cases where this class is called on
+        generic- or non-stream object as opposed to
+        a :class:~music21.stream.Score, :class:~music21.stream.Part,
```
jacobtylerwalls, line 178: Everything after the `:class:` needs to be wrapped in backticks: `

(When I (often) need a refresher, I just pop `:class:` into my search bar in my code editor to pull up examples.)

MarkGotham: Thanks! Does the Opus handling look alright to you?

jacobtylerwalls: I'll take a look this week. At a glance looks good!

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,536 @@
...
+
+class RnWriter(prebase.ProtoM21Object):
+    '''
+    Extracts the relevant information from a source
+    (usually a :class:`~music21.stream.Stream` of
+    :class:`~music21.roman.RomanNumeral` objects) for
+    writing to text files in the 'RomanText' format (rntxt).
+
+    Writing rntxt is handled externally in the
+    :meth:~music21.converter.subConverters.WriteRoman so
```
jacobtylerwalls, line 39: More backticks here and below needed. There are some instructions in the music21 developer's guide on building documentation, and if they aren't clear that's a brilliant opportunity for another PR. Here is what I see when I build (see the links not rendering):

links-not-rendered

MarkGotham: Sorry! Always proof-read last – no edits after the final checks.

jacobtylerwalls (commented):
I looked at the opus handling and discovered that opus has its own 'write' method that just calls `write` on each score in `self.scores`:

https://github.com/cuthbertLab/music21/blob/09b76989b24c8636e91c61632cf320a1489f59a2/music21/stream/__init__.py#L13240-L13243

I think this means you can remove the special handling for `Opus`, because if we implement any improvements to `Opus.write()` (I spot some things), we don't want to cause more refactoring in your feature. Want to just verify an Opus parsed from rntxt will write back out without explicit extra handling for it?

jacobtylerwalls:
What do we want `.show('rntxt')` to do?  Print to the console?

```
>>> opus.show('rntxt')
No application knows how to open /var/folders/76/_fy63x8s7nbf7r8__9t1wq78ws9vrr/T/music21/tmp5b9tj93l.rntxt.
No application knows how to open /var/folders/76/_fy63x8s7nbf7r8__9t1wq78ws9vrr/T/music21/tmpgukqxqr4.rntxt.
No application knows how to open /var/folders/76/_fy63x8s7nbf7r8__9t1wq78ws9vrr/T/music21/tmp2q77r4cg.rntxt.
```

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,536 @@
...
+                self.container = obj
+
+            elif isinstance(obj, stream.Measure):
+                self.container = stream.Part()
+                self.container.insert(0, obj)
+
+            else:  # A stream, but not a measure, part, or score
+                self._makeContainer(obj)
+
+            if obj.metadata:  # sic, obj not container for metadata, and only if it's stream
```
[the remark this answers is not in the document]
MarkGotham: Thanks. I'll clarify

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,536 @@
...
+        # Note: blank analyst and proof reader entries until supported within music21 metadata
+
+        if not self.container.getElementsByClass('TimeSignature'):
+            self.container.insert(0, meter.TimeSignature('4/4'))  # Placeholder
+
+        self.currentKeyString: str = ''
+        self.prepSequentialListOfLines()
+
+    def _makeContainer(self,
+                       objs: list):
```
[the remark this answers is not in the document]
MarkGotham: This is a casualty/beneficiary (?!) of the newly implemented lint check on list comprehensions. The `isIterable` route is a great solution – thanks!

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,536 @@
...
+    else:
+        return round(beat, roundValue)
+
+
+# ------------------------------------------------------------------------------
+
+class Test(unittest.TestCase):
+    '''
+    Tests for two analysis cases (the smallest rntxt files in the music21 corpus)
+    along with two test by modifying those scores.
```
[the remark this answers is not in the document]
MarkGotham: Perfectly happy to do so – I'll let Myke rule on that. (I get the impression with documentation that 'more is more'.)

music21/romanText/writeRoman.py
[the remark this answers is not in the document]
MarkGotham: Actually no, I'm reminded that RomanNumeral objects are iterable (returning notes) so that won't quite do it here.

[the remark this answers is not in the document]
MarkGotham: Instead I'll loose the type but only slightly to `obj: Union[stream.Stream, list]`

MarkGotham:
Hi @jacobtylerwalls . Thanks for the latest review! The .show() option works for me as expected:
- `show.('rntxt')` creates rntxt files
- `show('t')` prints the stream to the console
- `show()` creates xml files that open (for me) in MuseScore.

The situation you set out looks external to music21. My system gave messages like that until I told it to open all `.rntxt` files in a simple text editor. I initially suggested writing to `.txt` for this reason but Myke pointed out that music21 would prefer the files to be in `.rntxt` so that it knows how to parse them without explicitly instruction (`fmt=`).

MarkGotham:
As for Opus, thanks for flagging the separate .write() override.
Given what you've pointed out, Opus objects won't make it through to this module in the majority of simple .write() uses cases, but we may still need Opus handling within this module for those (exceptional) cases of using this module without .write (as discussed in the init docs).

All the same, I'm happy to remove this functionality (as I've said, I don't personally anticipate using the Opus option) - whatever's best for music21.

MarkGotham:
No problem. Any more changes needed / requests on this?

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,552 @@
...
+             inString: Optional[str] = ''):
+    '''
+    Creates or extends a string of RomanText such that the output corresponds to a single
+    measure line with one or more pairs of beat to Roman numeral.
+
+    If the inString is not given, None, or an empty string then this function starts a new line.
+
+    >>> lineStarter = romanText.writeRoman.rnString(14, 1, 'G: I')
+    >>> lineStarter
+    'm14 b1 G: I'
```
mscuthbert, line 310: looking at the spec, "b1 " is much better removed -- all chords at the beginning of the measure are "b1" chords.

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,552 @@
...
+        >>> m.insert(0, roman.RomanNumeral('V', 'G'))
+        >>> p.insert(0, m)
+        >>> testCase = romanText.writeRoman.RnWriter(p)
+        >>> testCase.combinedList[-1]  # Last entry, after the metadata
+        'm0 b1 G: V'
+        '''
+
+        for thisMeasure in self.container.getElementsByClass('Measure'):
+            # TimeSignatures
+            tsThisMeasure = thisMeasure.getElementsByClass('TimeSignature')
```
mscuthbert, line 242: Key Signatures are one of the things missing here -- at least make a TODO for it here.  (But not a block)

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,552 @@
...
+            if tsThisMeasure:
+                firstTS = tsThisMeasure[0]
+                self.combinedList.append(f'Time Signature: {firstTS.ratioString}')
```
mscuthbert, line 245: The initial Time Signature should go into the header section.  My sense is that the header and the body should be stored separately and only combined later, so that the blank line between the two is consistent.

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,552 @@
...
+        # Note: blank analyst and proof reader entries until supported within music21 metadata
+
+        if not self.container.getElementsByClass('TimeSignature'):
+            self.container.insert(0, meter.TimeSignature('4/4'))  # Placeholder
+
+        self.currentKeyString: str = ''
+        self.prepSequentialListOfLines()
+
+    def _makeContainer(self,
+                       obj: Union[stream.Stream, list]):
```
mscuthbert, line 176: For Python pre-3.9 (i.e., music21 currently), list needs to be List and imported from typing module.  typing.List is different from the builtin list until Py3.9

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,552 @@
...
+
+        # --------------------
+
+        scoreBach.parts[0].measure(3).insert(0, meter.TimeSignature('10/8'))
+        scoreBach.parts[0].measure(3).insert(1, meter.TimeSignature('5/8'))
+
+        wonkyBach = RnWriter(scoreBach)
+
+        tsString1 = 'Time Signature: 10/8'
+        tsString2 = 'Note: further time signature change(s) unprocessed: [\'5/8\']'
```
mscuthbert, line 465: These are places where it's better to use double quotes on the outside to avoid backslashes internally.

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,552 @@
...
+        self.assertEqual(wonkyBach.combinedList.index(tsString2),
+                         wonkyBach.combinedList.index(tsString1) + 1)
+
+        # --------------------
+
+        scoreMonte = corpus.parse('monteverdi/madrigal.3.8.rntxt')  # Smallest file
+        rnMonte = RnWriter(scoreMonte)
+
+        self.assertEqual(rnMonte.composer, 'Monteverdi')
+        self.assertEqual(rnMonte.title, 'La piaga c\'ho nel core')
```
mscuthbert, line 479: as above -- double quote to avoid backslash  (i.e., how Python's repr() shows it)

music21/romanText/writeRoman.py
```diff
@@ -0,0 +1,552 @@
...
+            else:  # A stream, but not a measure, part, or score
+                self._makeContainer(obj)
+
+            if obj.metadata:  # Check the obj (not container) for metadata if obj is a stream
+                self.prepTitle(obj.metadata)
+                if obj.metadata.composer:
+                    self.composer = obj.metadata.composer
+
+        else:  # Not a stream
+            self._makeContainer([obj])
```
mscuthbert, line 160: WOW!  The code coverage of this module is SPECTACULAR!  This is actually the only uncovered line -- not sure if you want to write a unittest (better not doctest) to cover it to have the (first?) long module in music21 that is 100% tested?  If not, no problem, but WOW!

mscuthbert (commented):
Great work -- there are a few small things here, but the only thing I think is really crucial is getting rid of the b1s, putting the Time Signature into the header section, and removing measures that are entirely the previous measure's chord.  (I think that the last one we discussed in a previous review, no?)  Other than that, congrats!  Sorry to take so long to review.  A lot of work happening!

> What do we want `.show('rntxt')` to do? Print to the console?

I think it's okay for now.  But yeah, the best choice would be to print to the console.

mscuthbert (commented):
(Github problems -- might be the same comments twice?)

music21/romanText/writeRoman.py
MarkGotham: Thanks for raising this.

Earlier versions had this kind of header vs body separation but (if I remember right) @jacobtylerwalls and I decided to get rid of it as part of streamlining the process and avoid unnecessary storage/duplication (and the potential for confusion and errors that invites).

For instance, the handling of Opus objects (approved in this review) would have to change. Should those cases be stored as Header 1, Body 1, Header 2, Body 2, etc.?

All in all, I'd vote to leave this as it is (one continuous, unseparated list) but am happy to be overruled and implement an alternative.

music21/romanText/writeRoman.py
MarkGotham: Thanks, and hurrah!

I'm happy to add another test to this effect but both the `RnWriter` docs and the `testTypeParses` unittest include calling this class directly on a RomanNumeral object (which is obviously not a stream) so it's not clear why this line isn't registering as covered.

mscuthbert:
CONGRATS!  Merged!  In time to make the 6.5 release for new years!

MarkGotham:
Thanks @mscuthbert and also @jacobtylerwalls and @napulen! Happy New Year!

PR #652, opened 2020-11-01
Calculate key confidence for every pitch class in case of tie

### LINKED ISSUE #463: KeyAnalyzer.smoothInterpretationByMeasure key miss
In `music21.analysis.floatingKey.KeyAnalyzer`, we have method `smoothInterpretationByMeasure` to predict tones. In floatingKey.py line 129, we have `newInterpretations[k]` to get a value in a dict using a key. However, the dict generated from `getInterpretationByMeasure` cannot guarantee that the generated keys are complete.

In discrete.py, there is a method called `process` in line 598 which give us the prediction result and confidence by keys generated by `self._likelyKeys(sStream)`. In `_getLikelyKeys` method in line 395, the library uses `someList.index(someEntry)` method to look up the original position of the sorted list, which is inappropriate, especially in condition that the confidence has a tie.

E.g., construct a bar whose `pcDistribution` is [2.0, 2.0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], say, 'C C D- D-'.
The confidence of this bar being of g minor is exactly equal to that of being of g# minor. So the `_getLikelyKeys` will return a result with 2 g minor prediction, rather than that with 1 g minor prediction and 1 g# minor prediction, triggering error in `smoothInterpretationByMeasure` method.

jacobtylerwalls:
Fixes #463 
Thanks @ZM-J for the report and original patch!

### DIFF
--- a/documentation/source/usersGuide/usersGuide_23_romanNumerals.ipynb
+++ b/documentation/source/usersGuide/usersGuide_23_romanNumerals.ipynb
@@ -308,7 +308,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Note that in the last case, the scale degree is the uninflicted degree -- the flat sign isn't included in the `.scaleDegree` tag.  For that, use `.scaleDegreeWithAlteration`"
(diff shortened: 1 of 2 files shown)


mscuthbert:
Ah! I get the problem and the solution is great.  Merging.  Thanks @jacobtylerwalls  and @ZM-J

PR #653, opened 2020-11-09
Pass fp and subformats to formatWriter.write() by keyword

jacobtylerwalls:
`fp` and `subformats` are keyword arguments on every writer, so it seems safer to me to pass them by keyword than by position.


mscuthbert:
yep.  makes perfect sense!

PR #654, opened 2020-11-12
Doc edits and basic clean up

MarkGotham:


mscuthbert:
Hi -- please restore the preferSecondaryDominants -- it doesn't do anything right now, but other systems might depend on it being there, and there is a plan to make it work in the future.  You can add a note that it doesn't currently do anything, but mid-release we don't make backwards incompatible changes.  Thanks!

music21/roman.py
```diff
@@ -748,83 +690,66 @@ def romanNumeralFromChord(chordObj,
...
-    These two are debatable -- is the harmonic minor or the natural minor used as the basis?
+    >>> roman.romanNumeralFromChord(
+    ...     chord.Chord(['B-4', 'D5', 'F5']),
+    ...     key.Key('c'),
+    ...     )
+    <music21.roman.RomanNumeral bVII in c minor>
 
-    # >>> roman.romanNumeralFromChord(chord.Chord('F4 A4 C5 E-5'), key.Key('c'))
-    # <music21.roman.RomanNumeral IVb753 in c minor>
-    # <music21.roman.RomanNumeral IV75#3 in c minor>
```
jacobtylerwalls, line 823: What if we uncomment this doctest? IV75#3 is the correct result, it seems to me, and it passes.

music21/roman.py
```diff
@@ -748,83 +690,66 @@ def romanNumeralFromChord(chordObj,
...
+    Finally, diminished and half-diminished seventh chords can
+    omit the third and still be diminished
+    (note also from this example that chords can be created from a string):
 
-    # >>> roman.romanNumeralFromChord(chord.Chord('F4 A4 C5 E5'), key.Key('c'))
-    # <music21.roman.RomanNumeral IV7 in c minor>
-    # <music21.roman.RomanNumeral IV#75#3 in c minor>
```
jacobtylerwalls, line 827: Same

music21/roman.py
```diff
@@ -748,83 +690,66 @@ def romanNumeralFromChord(chordObj,
...
 
-    >>> romanNumeral9 = roman.romanNumeralFromChord(
-    ...     chord.Chord(['C4', 'E5', 'G5', 'C#6']),
-    ...     key.Key('C'),
+    >>> roman.romanNumeralFromChord(
+    ...     chord.Chord(['A-4', 'C5', 'E-5']),
+    ...     key.Key('c'),
     ...     )
-    >>> romanNumeral9
-    <music21.roman.RomanNumeral I#853 in C major>
```
jacobtylerwalls, line 799: this one looks fine, can we keep it?

music21/roman.py
```diff
@@ -748,83 +690,66 @@ def romanNumeralFromChord(chordObj,
...
     ...     chord.Chord('E-4 G4 C#5'),
     ...     key.Key('C')
     ...     )
     <music21.roman.RomanNumeral #io6b3 in C major>
 
+    As always, the case of sixth and seventh degrees in minor pose
+    additional problems given the range of widely adopted conventions.
+    Music21 generally handles this with the :class:`music21.roman.Minor67Default`
+    as explained in the :class:`music21.roman.romanNumeral` and especially the
+    :meth:`~music21.roman.RomanNumeral.adjustMinorVIandVIIByQuality`.
```
jacobtylerwalls, line 714: Just FYI if #657 gets merged first we should strike the reference to this method and only point folks to `Minor67Default`.

music21/roman.py
```diff
@@ -628,105 +628,47 @@ def correctRNAlterationForMinor(figureTuple, keyObj):
...
-    >>> romanNumeral10
-    <music21.roman.RomanNumeral #iiiø7 in d minor>
+    >>> roman.romanNumeralFromChord(
+    ...     chord.Chord(['C4', 'E-4', 'G4']),
+    ...     )  # NB: no key
+    <music21.roman.RomanNumeral i in c minor>
 
+    >>> roman.romanNumeralFromChord(
+    ...     chord.Chord(['C4', 'E4', 'G4']),
+    ...     )  # NB: no key and nor major third
```
jacobtylerwalls, line 667: I don't understand this comment.

music21/roman.py
```diff
@@ -628,105 +628,47 @@ def correctRNAlterationForMinor(figureTuple, keyObj):
...
-
-    >>> romanNumeral2 = roman.romanNumeralFromChord(
-    ...     chord.Chord(['E3', 'C4', 'G4', 'B-4', 'E5', 'G5']),
-    ...     key.Key('F'),
-    ...     )
-    >>> romanNumeral2
-    <music21.roman.RomanNumeral V65 in F major>
-
-    Note that vi and vii in minor signifies what you might think of
-    alternatively as #vi and #vii:
```
jacobtylerwalls, line 664: Is there a reason these docs were deleted?

jacobtylerwalls (commented):
Hi Mark, thanks for getting moving on some improvements here. I noticed a lot of tests were deleted--is there a motivation for that? In general, we should be adding tests, but if there's a readability reason some of them should be hidden from the user, we can stash some under `OMIT_FROM_DOCS`.

music21/roman.py
mscuthbert: I don't know about this -- should #3 be there if IV is already capitalized?  Should it just be IV7?  I'm not sure if the `#3` applies after the sharpening of IV or if it's implied.  In any case, the comment should not be removed -- it either is still a bug in which case it should be kept in so we know to fix it, or it's fixed in which case it should be documented.

music21/roman.py
mscuthbert: Same here -- do not remove tests that show bugs.

music21/roman.py
mscuthbert: Yes, why remove?

music21/roman.py
mscuthbert: Me neither

mscuthbert: Also, better to define the chord in the line before as a variable (cMaj) and then pass it to roman.romanNumeralFromChord(cMaj) .  Inside doctests, comments shouldn't be necessary, since they can be put as descriptions of the test before.

music21/roman.py
mscuthbert: as with JTW

music21/roman.py
```diff
@@ -628,105 +628,47 @@ def correctRNAlterationForMinor(figureTuple, keyObj):
     return FigureTuple(figureTuple.aboveBass, alter, rootAlterationString)
 
 
-def romanNumeralFromChord(chordObj,
+def romanNumeralFromChord(chordObj: chord.Chord,
                           keyObj: Union[key.Key, str] = None,
-                          preferSecondaryDominants=False):
+                          preferSecondaryDominants: bool = True):
```
mscuthbert, line 633: Why has the default here been changed?  It's beyond the scope of the PR.

music21/roman.py
```diff
@@ -628,105 +628,47 @@ def correctRNAlterationForMinor(figureTuple, keyObj):
...
     # noinspection PyShadowingNames
+    # TODO: preferSecondaryDominants option
     '''
-    Takes a chord object and returns an appropriate chord name.  If keyObj is
-    omitted, the root of the chord is considered the key (if the chord has a
-    major third, it's major; otherwise it's minor).  preferSecondaryDominants does not currently
-    do anything.
+    Takes a chord object (chordObj) and returns a Roman numeral.
```
mscuthbert, line 637: as long as we are cleaning up, should say:  
```
Takes a :class:`~music21.chord.Chord` object and an optional :class:`~music21.key.Key` object (or a string representing a key) and returns a :class:`~music21.roman.RomanNumeral` object.
```

I don't know what "are key-relative" means.  And this doctest removes the instructioin that "preferSecondaryDominants does not currently do anything".

music21/roman.py
mscuthbert: This is too complex for a basic user.

music21/roman.py
```diff
@@ -748,83 +690,66 @@ def romanNumeralFromChord(chordObj,
...
+    but robust handling is a work in progress.
+    Here are some examples of behaviour as currently expected:
 
-    Former bugs:
-
-    Should be iii7
-
-    >>> romanNumeral11 = roman.romanNumeralFromChord(
-    ...     chord.Chord(['E4', 'G4', 'B4', 'D5']),
-    ...     key.Key('C'),
```
mscuthbert, line 776: Why are these being deleted?

mscuthbert (changes requested):
The proposed PR was to add some additional test cases.  Instead this deletes many existing test cases and rewrites the documentation.  Cannot approve as is.

MarkGotham:
Ok, thank you both for the comments – duly noted.

Clearly this PR has already got bogged down. I'll close this, watch #657 and re-submit parts of this if relevant in smaller, unequivocal chunks, taking note of these comments.

As for tests, there are some that have been deleted (e.g. 'former bugs' – presumably these go at some point) but many have been re-worked as part of what I think is a smoother explanation of the component parts. Smaller chunks will make the difference (and discussion) clearer.

PR #657, opened 2020-11-14
Make adjustMinorVIandVIIByQuality() private; unsharpen root in figure when sharpening root

### LINKED ISSUE #437: RomanNumeral and romanNumeralFromChord alignment
roman.RomanNumeral and roman.romanNumeralFromChord differ in their default handling of 6th and 7th degrees in minor (so round trips are spirals but not circles ... so to speak).

Should roman.Minor67Default and correctRNAlterationForMinor merge?

For instance ... 

```
ps = ['D', 'G#', 'B']
chd = chord.Chord(ps)
rn = roman.romanNumeralFromChord(chd, 'a')
print(rn.figure)

```
... returns 'viio6#4', but making a Roman numeral from that figure ...

```
rn = roman.RomanNumeral('viio6#4', 'a')
print([x.name for x in rn.pitches])

```
... returns ['D#4', 'G##', 'B#']

This can be overridden with seventhMinor=roman.Minor67Default.FLAT, so ...

```
rn = roman.RomanNumeral('viio6#4', 'a', seventhMinor=roman.Minor67Default.FLAT)
print([x.name for x in rn.pitches])

```
... returns ['D4', 'G#', 'B']
... but the defaults should probably align, right?

jacobtylerwalls:
Fixes #437

The subroutine `sharpen()` was not unsharpening roots from figures representing inversions when sharpening the root. This led to the need to pass in and return `workingFigure`, which led me to question whether this method serves any purpose being public, save for its useful documentation about `Minor67Default`, since this is extremely low-level work. So I moved the documentation to `Minor67Default` and made the method private for consistency with its neighbors, and provided a backwards compatible public stub.

### DIFF
--- a/music21/roman.py
+++ b/music21/roman.py
@@ -1150,6 +1150,11 @@ class RomanNumeral(harmony.Harmony):
     >>> [str(p) for p in minoriiiB.pitches]
     ['G4', 'B-4', 'D5']
 
+    `caseMatters=False` will prevent `sixthMinor` or `seventhMinor` from having effect.
+    >>> vii = roman.RomanNumeral('viio', 'a', caseMatters=False, seventhMinor=roman.Minor67Default.QUALITY)
+    >>> [str(p) for p in vii.pitches]
+    ['G5', 'B-5', 'D-6']
+
     Can also take a scale object, here we build a first-inversion chord
     on the raised-three degree of D-flat major, that is, F#-major (late
     Schubert would be proud.)
--- a/music21/roman.py
+++ b/music21/roman.py
@@ -2039,15 +2039,14 @@ class RomanNumeral(harmony.Harmony):
 
         return workingFigure, useScale
 
-    def _adjustMinorVIandVIIByQuality(self, workingFigure, useScale):
+    def adjustMinorVIandVIIByQuality(self, useScale):
         '''
-        fix minor vi and vii to always be #vi and #vii if `.caseMatters`.
+        Fix minor vi and vii to always be #vi and #vii if `.caseMatters`.
 
         >>> rn = roman.RomanNumeral()
         >>> rn.scaleDegree = 6
         >>> rn.impliedQuality = 'minor'
-        >>> rn._adjustMinorVIandVIIByQuality('', key.Key('c'))
-        ''
+        >>> rn.adjustMinorVIandVIIByQuality(key.Key('c'))
         >>> rn.frontAlterationTransposeInterval
         <music21.interval.Interval A1>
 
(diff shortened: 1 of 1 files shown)


music21/roman.py
```diff
@@ -2094,34 +2118,34 @@ def changeFrontAlteration(intV, alter):
...
+                return sharpen(workingFigure)
         else:  # CAUTIONARY
             if not self.frontAlterationAccidental or self.frontAlterationAccidental.alter == 0:
                 # same as QUALITY in this case
                 if not normallyRaised:
-                    return
+                    return workingFigure
                 else:
-                    sharpen()
+                    return sharpen(workingFigure)
```
jacobtylerwalls, line 2138: This is the only change to how early the function might exit. I overlooked at first that I was altering the control flow, but when I looked closer, upon exiting this branch,`frontAlter` below will be guaranteed to be 1, and `normallyRaised` is guaranteed to be `True`, so we will enter the branch that immediately returns without recalling `sharpen()`, so I don't see harm in returning here.

jacobtylerwalls:
For backwards compatibility I suppose I could leave a stub for the public function that passes a dummy (empty) working figure to the new private function and returns nothing.

music21/roman.py
mscuthbert: :-)

music21/roman.py
```diff
@@ -1980,86 +2063,45 @@ def adjustMinorVIandVIIByQuality(self, useScale):
...
+
+        Made private in v.6.4 when `workingFigure` was added to the signature
+        and returned.
+
+        Altering `workingFigure` became necessary to handle these chromatic figures:
+        https://github.com/cuthbertLab/music21/issues/437
+
+        >>> rn = roman.RomanNumeral('viio#6', 'a')
+        >>> ' '.join([p.name for p in rn.pitches])
+        'B D G#'
```
mscuthbert, line 2082: What does viio6 give in this case? is it still B D G# as before? just want to be sure.

music21/roman.py
```diff
@@ -1980,86 +2063,45 @@ def adjustMinorVIandVIIByQuality(self, useScale):
...
+        >>> ' '.join([p.name for p in rn.pitches])
+        'D G# B'
+        >>> rn = roman.RomanNumeral('viio4#2', 'a')
+        >>> ' '.join([p.name for p in rn.pitches])
+        'F G# B D'
+        >>> rn = roman.RomanNumeral('viio#853', 'a')
+        >>> ' '.join([p.name for p in rn.pitches])
+        'G# B D'
+        '''
+        def sharpen(workingFigure):
```
mscuthbert, line 2093: this is going to give some errors in my PyCharm linter which doesn't like internal variables with the same name as enclosing variables.  But I can fix it later.

music21/roman.py
mscuthbert: AH! Yes, I see below.

music21/roman.py
```diff
@@ -1980,86 +2063,45 @@ def adjustMinorVIandVIIByQuality(self, useScale):
...
             changeFrontAlteration(interval.Interval('A1'), 1)
+            # Unsharpen the root for inverted figures to avoid double-sharpening
+            if '#2' in workingFigure:
+                workingFigure = workingFigure.replace('#2', '2')
+            elif '#4' in workingFigure:
+                workingFigure = workingFigure.replace('#4', '4')
+            elif '#6' in workingFigure:
+                workingFigure = workingFigure.replace('#6', '6')
+            else:
+                workingFigure = workingFigure.replace('#8', '')
```
mscuthbert, line 2103: In theory, should check first for '##8' to replace with '#8'.

music21/roman.py
```diff
@@ -917,6 +917,84 @@ def romanNumeralFromChord(chordObj,
 
 
 class Minor67Default(enum.Enum):
+    '''
+    Showing how sixthMinor affects the interpretation of `vi`
```
mscuthbert, line 921: Probably needs to have a little introduction like, "Enumeration that can be passed into RomanNumeral creation to alter how roman numerals on the sixth and seventh scale degrees are parsed in minor."

mscuthbert (commented):
I think that this looks really good, but just wanting to check on ##8 and a few other little things.  I think good to go as is, but just want to be sure.

mscuthbert:
Amazing.  You rock!

MarkGotham:
Hear hear! Thanks @jacobtylerwalls for this contribution -- much appreciated!

PR #658, opened 2020-11-14
Handles RomanText entries like 'm2 b1' with no Roman numeral (issue 655)

MarkGotham:


mscuthbert:
Per 2.2 of Spec:

```
Each line of analysis starts with the symbol ‘m’ followed immediately by a measure
number with no space; it then proceeds with pairs of beat numbers (preceded by ‘b’) and their corresponding Roman numerals.
```

Beat symbols and Roman Numerals always come in pairs.  The system is correct to reject this usage.

MarkGotham:
Thanks for this @mscuthbert.

You're right that we don't offer to do anything with this case though about two sentences later we explicitly accept lines omitting 'b1' so it's not quite true that beats and chords always come in pairs.

Anyway, I follow your reasoning for establishing that the entry type 'm2 b1' without a chord is 'wrong', though I think the logical consequence of this should be either:
- analysis fails to parse as badly formatted, or (preferably);
- lines like 'm2 b1' are ignored in the same way as 'm2' (perhaps with a warning in the log).

Currently the code does neither of the above, but rather creates a measure of no duration which unnecessarily messes things up by bypassing the nice 'fillToMeasureToken' method you or Chris have included. So if a user makes this error in an encoding (easy enough to do) then they get a problematic score back but no warning that anything's up.

Personally, I'd like to tighten up these kinds of cases more decisively. Other examples include roman.RomanNumeral accepting figures like 'IIV' without complaint (in that case, making a 'II' chord). Maybe you'd prefer a 'sanity checker' script, separate from the parser that would flag up these case? For the feedback on Roman numeral analysis script that I offer elsewhere ([code here](https://github.com/MarkGotham/When-in-Rome/blob/master/Code/romanUmpire.py) and [app here](https://fourscoreandmore.org/working-in-harmony/)), I'll need some solution of that kind to these cases.

Ok. That's my 2 cents. Naturally, other users may feel differently and it's your prerogative to make all final calls.

All best,
Mark

PR #659, opened 2020-11-14
Change the name of the default scratch directory.

tanchihpin0517:
Hi,

I find in the multi-user environment such as a server, all users share the same scratch directory (/tmp/music21 for Linux) by default will cause permission problems because the default permission of a newly created directory doesn't allow "write" from others.

I know music21 provides a way which let users custom their scratch directory. But I think this problem can be avoided by just making few changes.

Hence, I change the name of the default scratch directory to music21-$UID to avoid different users sharing the same directory.
For example, the directory /tmp/music21-1000 will be created for the user with UID of 1000.

I test and make sure that this way works. Hope you guys agree with me and accept this code.

Thanks,
Chih-Pin

### DIFF
--- a/music21/environment.py
+++ b/music21/environment.py
@@ -484,7 +484,7 @@ class _EnvironmentCore:
     def getDefaultRootTempDir(self):
         # noinspection SpellCheckingInspection
         '''
-        returns whatever tempfile.gettempdir() returns plus 'music21'.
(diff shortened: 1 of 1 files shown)


tanchihpin0517:
I don't notice this code can't pass the travis test. I will correct it and re-send a pull request.

PR #660, opened 2020-11-14
Change the name of the default scratch directory.

tanchihpin0517:
Hi,

I find in the multi-user environment such as a server, all users share the same scratch directory (/tmp/music21 for Linux) by default will cause permission problems because the default permission of a newly created directory doesn't allow "write" from others.

I know music21 provides a way which lets users custom their scratch directory. But I think this problem can be avoided by just making few changes.

Hence, I change the name of the default scratch directory to music21-$USER to avoid different users sharing the same directory.
For example, the directory /tmp/music21-tanch will be created for the user with $USER of tanch.

I test and make sure that this way works. Hope you guys agree with me and accept this code.

Thanks,
Chih-Pin

### DIFF
--- a/music21/environment.py
+++ b/music21/environment.py
@@ -28,6 +28,7 @@ import pathlib
 import sys
 import tempfile
 import unittest
+import getpass
 
 from typing import Union
 
@@ -484,7 +485,8 @@ class _EnvironmentCore:
     def getDefaultRootTempDir(self):
         # noinspection SpellCheckingInspection
         '''
-        returns whatever tempfile.gettempdir() returns plus 'music21'.
+        returns whatever tempfile.gettempdir() returns plus 'music21-$USER'.
+        (compatible with windows)
 
         Creates the subdirectory if it doesn't exist:
 
@@ -495,12 +497,13 @@ class _EnvironmentCore:
         '/var/folders/x5/rymq2tx16lqbpytwb1n_cc4c0000gn/T'
 
         >>> import os
+        >>> import getpass
         >>> e = environment.Environment()
-        >>> e.getDefaultRootTempDir() == pathlib.Path(t) / 'music21'
+        >>> e.getDefaultRootTempDir() == pathlib.Path(t) / 'music21-{}'.format(getpass.getuser())
         True
         '''
         # this returns the root temp dir; this does not create a new dir
-        dstDir = pathlib.Path(tempfile.gettempdir()) / 'music21'
+        dstDir = pathlib.Path(tempfile.gettempdir()) / 'music21-{}'.format(getpass.getuser())
         # if this path already exists, we have nothing more to do
         if dstDir.exists():
             return dstDir


mscuthbert:
I'm okay with this change (so long as it doesn't break existing users), but I need a report that all tests pass on Windows -- not just an assurance that it would but actual testing of the PR on a Windows machine.  Windows is a 100% supported music21 environment, while multi-user Linux is not.

tanchihpin0517:
Thank you for reminding me that music21 also runs on Windows platform. And I found my original method doesn't work on Windows.
So I make this change: change default scratch directory from music21-$UID to music21-$USER. This way works on both Unix and Windows.

However, I tried to test environment.py on Windows, but failed on some tests which doesn't have any relation with this PR.
I tried the old version without any change from this PR, so I'm sure even the original version can't pass the test either.
Here is one of those failures:
```
FAIL: __setitem__ (__main__.Environment)
Doctest: __main__.Environment.__setitem__
  File "C:\Users\scream\AppData\Local\Programs\Python\Python37\lib\doctest.py", line 2204, in runTest
    raise self.failureException(self.format_failure(new.getvalue()))
AssertionError: Failed doctest test for __main__.Environment.__setitem__
  File "C:\Users\scream\workspace\music21\music21\environment.py", line 911, in __setitem__

File "C:\Users\scream\workspace\music21\music21\environment.py", line 920, in __main__.Environment.__setitem__
Failed example:
    a['graphicsPath']
Expected:
    PosixPath('/test&amp;Encode')
Got:
    WindowsPath('/test&amp;Encode')
```
I have test that music21 works normally for the new scratch directory on Windows.
But how to get a report that all tests pass on Windows? Even the original one can't pass all tests.

tanchihpin0517:
I compared the log of testing environment.py (which is changed at this PR) on this PR with on the original one. Both of them failed on the unit tests 3, 12, 20, 21 and 22. Those are not related to the scratch directory. 

I think these failures are caused by that some unit tests were not written for Windows. Take the unit test 3 for example. The output "PosixPath('/test&amp;Encode')" was expected, but "WindowsPath('/test&amp;Encode')" was got while running on Windows.

I have made sure that there is no difference between the result of testing environment.py with and without this PR. Those tests of failure are caused by that they are not written for Windows.
I hope this is enough to show that this PR is stable. If not, please give me some hints about where I did wrong and what I could do next. Thanks.

Here is the full log of testing for environment.py: [log.txt](https://github.com/cuthbertLab/music21/files/5556960/log.txt)

tanchihpin0517:
Thanks to @jacobtylerwalls for #671 .
@mscuthbert I merged #671 to this PR and re-tested. It passed all tests of environment.py on Windows.

mscuthbert:
I'm thinking through this and it seems both good and fragile at the same time.  I wonder if we shouldn't add `appdirs` as a dependency (https://pypi.org/project/appdirs/ ) and use user_cache_dir() instead.  Then we're moving to someone else the problem of what to do on (say) iOS where there isn't a username associated with it, or some other future OS breakage.   I'm still a little scared to change a system that has many moving parts but which has worked until now except for this one case.

jacobtylerwalls:
I have a feeling that `environment.getTempFile()` should catch `PermissionError` and then in the except clause do something like what this patch suggests.

tanchihpin0517:
@jacobtylerwalls I have another idea. Use `tempfile.mkdtemp()` (https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp) instead of hardcoding temp directory in `environment.getDefaultRootTempDir()`.
```
def getDefaultRootTempDir():
    ...
    # tempdir = None by default
    if tempdir == None:
        tempdir = tempfile.mkdtemp()
    return tempdir
   ...
```
This way will avoid any problem currently we face, and it is robust, cross-platfrom, little code changing and behavior like the original routine.

mscuthbert:
But it doesn’t put all the music21 files in one directory for easy cleaning without deleting everything else.  Or are you saying as a backup if the current dir does not exist or is not writable?  If so, we should warn the user to update their .music21rc

tanchihpin0517:
I mean cleaning the temp directory is the side effect of using tempfile.TemporaryDirectory(). Music21 doesn't delete temp files, but the root temp directory will be delete if using tempfile.TemporaryDirectory() to replace the original temp directory (/tmp/music21).

> But it doesn’t put all the music21 files in one directory for easy cleaning without deleting everything else.

The object of using tempfile.TemporaryDirectory() is not to clean all temp files generated by music21. So it's okay for music21 not to put all temp files in on directory because we are not trying to clean them. It's just a side effect.

What I worry about is whether music21 is designed to reuse or share old temp files saved in the temp directory from different python sessions. If yes, we shouldn't use this way because the temp directory will be delete in the end of a python session. If no, it will be safe to use tempfile.TemporaryDirectory().
 
> Or are you saying as a backup if the current dir does not exist or is not writable?  If so, we should warn the user to update their .music21rc

No, I want to replace /tmp/music21 with tempfile.TemporaryDirectory() directly. It's also okay to make it as a backup. But if there is no concern mentioned above, replace it directly will make the code and the structure more clean and safe because it's python official library.

I think we don't need to warn users about this change. As I know (but it may be wrong), users won't see then temp files by default. If they want, they will set it in .music21rc. What we do is change the default temp directory, so it won't be affected if users custom the temp directory in .music21rc.

mscuthbert:
I think we're going to go with #680 and #679 but let me know if I'm wrong by reopening.

Issue #656, opened 2020-11-14
Platform Version Retrieval Broken on macOS Big Sur


dnlfrst:
# What is the issue?

After updating from macOS Catalina (10.15.*) to macOS Big Sur (10.16), I am no longer able to import `music21`. Doing so fails with the following stacktrace:

```python
>>> import music21
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/music21/__init__.py", line 169, in <module>
    from music21 import base  # noqa: E402
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/music21/base.py", line 66, in <module>
    from music21 import environment
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/music21/environment.py", line 816, in <module>
    _environStorage['instance'] = _EnvironmentCore()
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/music21/environment.py", line 175, in __init__
    self._loadDefaults(forcePlatform=forcePlatform)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/music21/environment.py", line 405, in _loadDefaults
    versionTuple = common.macOSVersion()
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/music21/common/misc.py", line 117, in macOSVersion
    major, minor, maintenance = tuple(int(v) for v in platform.mac_ver()[0].split('.'))
ValueError: not enough values to unpack (expected 3, got 2)
```

Indeed, retrieving the version of macOS Big Sur in Python 3.8 with `platform` yields:

```python
>>> import platform
>>> platform.mac_ver()
('10.16', ('', '', ''), 'x86_64')
```

This does no longer adhere to the assumption in `music21`'s implementation:

https://github.com/cuthbertLab/music21/blob/70a8dd272ba63a44cc4e0e1e8199c08b95612cbb/music21/common/misc.py#L107-L118

Instead, as seen above, Python 3.8 refers to macOS Big Sur as macOS version 10.16 which yields a tuple, not a triple:

```python
>>> import platform
>>> platform.mac_ver()
('10.16', ('', '', ''), 'x86_64')
>>> platform.mac_ver()[0].split('.')
['10', '16']
```

# What could be a solution?

[...]

PR #661, opened 2020-11-14
Correctly Retrieve Version of macOS Big Sur

dnlfrst:
As described in #656, retrieving the version of macOS Big Sur does currently not work. This effectively prevents using _music21_ on macOS Big Sur as `import music21` fails.

This pull request introduces a future-proof fix that handles unexpected macOS versions that only contain a major version (e.g., `11`) or a major and minor version (e.g., `10.16`).


mscuthbert:
Thank you!  LGTM!

PR #663, opened 2020-11-15
Remove runTest()

jacobtylerwalls:
`coverage` was treating the line `pass` as a missed line each time.


mscuthbert:
I think that the way of using `runTest(self)` in these cases is actually a mistake -- we should not be implementing it, but instead just defining the textXYZ(self) methods.  I believe that it would be better to remove them all.  I think that Ariza (who set up our testing system) intended runTest to be what setUp(self) actually does.

mscuthbert:
THANK YOU!  Appreciated!

PR #667, opened 2020-11-16
Fix #665 by altering pitches in place

### LINKED ISSUE #665: MusicXML parsing doesn't use altered degrees that contradict chord kinds
**music21 version**

`6.1.0`   

**Problem summary**

When parsing a MusicXML chord with kind "dominant" but a lowered fifth degree, this is not reflected in the chord pitches. I haven't tested if this is the case for other kinds and degree alterations.

**Steps to reproduce**

```
>>> from xml.etree.ElementTree import fromstring as EL
>>> MP = music21.musicxml.xmlToM21.MeasureParser()

>>> elStr = (r'<harmony><root><root-step>C</root-step></root><kind text="7b5">dominant</kind>' + 
... '<degree><degree-value>5</degree-value><degree-alter>-1</degree-alter>' +
... '<degree-type>alter</degree-type></degree></harmony>')
>>> mxHarmony = EL(elStr)

>>> cs = MP.xmlToChordSymbol(mxHarmony)
>>> cs.figure
'C7 alter b5'
>>> cs.pitches
(<music21.pitch.Pitch C3>, <music21.pitch.Pitch E3>, <music21.pitch.Pitch G3>, <music21.pitch.Pitch B-3>) # Should be G-3

### PR DESCRIPTION
**Pull request by pettter**
_updatePitches() is also added to the end of addChordStepModification. I can redo the pull request without that if that's something that shouldn't happen.

This fixes #665

### DIFF
--- a/music21/harmony.py
+++ b/music21/harmony.py
@@ -456,6 +456,7 @@ class Harmony(chord.Chord):
                 'cannot add this object as a degree: {0}'.format(degree))
 
         self.chordStepModifications.append(degree)
+        self._updatePitches()
 
     def findFigure(self):
         return 'No Figure Representation'
@@ -1518,6 +1519,14 @@ class ChordSymbol(Harmony):
     >>> cs.chordKindStr = 'M'
     >>> cs
     <music21.harmony.ChordSymbol E-/B->
+
+    Alterations should also affect the pitches
+
+    >>> h = harmony.ChordSymbol('C')
+    >>> mod = harmony.ChordStepModification('alter', 5, -1)
+    >>> h.addChordStepModification(mod)
+    >>> h.pitches
+    (<music21.pitch.Pitch C3>, <music21.pitch.Pitch E3>, <music21.pitch.Pitch G-3>)
     '''
 
     # INITIALIZER #
--- a/music21/harmony.py
+++ b/music21/harmony.py
@@ -448,6 +448,15 @@ class Harmony(chord.Chord):
         >>> h.addChordStepModification('juicy')
         Traceback (most recent call last):
         music21.harmony.HarmonyException: cannot add this object as a degree: juicy
+        
+        In case alterations should also affect the pitches, you need to call _updatePitches()
+
+        >>> h = harmony.ChordSymbol('C')
+        >>> mod = harmony.ChordStepModification('alter', 5, -1)
+        >>> h.addChordStepModification(mod)
+        >>> h._updatePitches()
+        >>> h.pitches
+        (<music21.pitch.Pitch C3>, <music21.pitch.Pitch E3>, <music21.pitch.Pitch G-3>)
         '''
         if not isinstance(degree, ChordStepModification):
             # TODO: possibly create ChordStepModification objects from other
@@ -456,7 +465,6 @@ class Harmony(chord.Chord):
                 'cannot add this object as a degree: {0}'.format(degree))
 
         self.chordStepModifications.append(degree)
-        self._updatePitches()
 
     def findFigure(self):
         return 'No Figure Representation'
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jacobtylerwalls** [comment]
The method being altered is already called from `_updatePitches()`, so I wouldn't call it again. (Assuming the test still passes.)

**Comment by jacobtylerwalls** [comment]
https://github.com/cuthbertLab/music21/blob/9d72217dc8fc5674e4e73286a2937548f29dbd9f/music21/harmony.py#L2030

**Comment by pettter** [comment]
Yes, but _updatePitches is not called from addChordStepModification, so just adding the modification is not going to change the pitches, per the example you posted in the issue. It'd have to be

```
....
>>> h.addChordStepModification(mod)
>>> h._updatePitches()
>>> h.pitches
...
```

Which is also fine - want me to update it to reflect that way of doing things?

**Comment by jacobtylerwalls** [comment]
It does raise the question of why `_updatePitches()` needs to be called, especially if the other function is public, but the places that call it (xml parsing) do call `_updatePitches()` afterward, so it would require more regression testing and increases the scope of the PR. I would worry about breaking `C7alter5` (even though right now it's under-tested!)  I suggest adding the `_updatePitches()` call to the doctest for now.

Also, let's add the original test from your ticket as a new test function in `xmlToM21.Test`, since if we revisit this decision about the update call, we won't want to break that workflow. Thanks!

**Comment by jacobtylerwalls** [comment]
And on this idea, I would suggest moving the doctest directly under `addChordStepModification()` to take this low-level example and push it further away from the main harmony documentation. Sweeping the fact that we are having to call a private function a little deeper under the rug, as it were.

**Comment by pettter** [comment]
Gotcha, sounds like a plan. I noticed a comment on _updatePitches saying it was "really slow", so that may be the reason why it's not called from ```addChordStepModification()```...

**Comment by pettter** [comment]
Hm, I'm not sure I understand what you want the default value of ```updatePitches``` to be? 

I would assume that there's quite a few places within the codebase that need to be updated if the default is supposed to be ```True```, or performance is going to potentially suffer - and with that we're getting into territory I'm hesitant to go into, as somewhat of a novice to the codebase. 

Conversely, for a default of ```False``` that would be somewhat of a counterintuitive semantics of a public function...

**Comment by pettter** [comment]
I added it with the semantically reasonable ```True``` default for now.

**Comment by jacobtylerwalls** [comment]
Myke may have been suggesting you check for the existing keyword on the `harmony` object in addition to the new keyword you just created, and only update if both are True. Is that so, @mscuthbert ?

https://github.com/cuthbertLab/music21/blob/d29c806aa613d558b3f84060b06bfc9c1f038244/music21/roman.py#L1272-L1278

**Comment by jacobtylerwalls** [comment]
First of all @pettter , thanks for being game to dive into this today. Disregard my comment about grabbing `updatePitches` from the `harmony` object for now. Setting it to False is a shortcut, so it's irrelevant for a user calling a public function to do more stuff. Plus, we don't save it to the instance.

We want to default `updatePitches=False` so that we don't break backward compatibility mid-release, unless @mscuthbert thinks this is so minor we can just steamroll ahead. That said, we can leave a TODO for `updatePitches will default True in v.7` to fully realize the convenience we are adding here.

When that switch comes it's just four places in `harmony.py` and one place in `xmlToM21.py` that need to have `updatePitches=False`. Not too bad. Would you be up for setting those places to `updatePitches=False` now? And adding your XML parsing test case?

We're getting a bugfix plus some good test cases PLUS making an existing public function easier to understand (even if gradually). Thank you!

**Comment by pettter** [comment]
Hey, added a test in what I hope is the right place. I can do the other changes as well, sure.

**Comment by pettter** [comment]
Well, with even the tests passing I'm starting to feel rather comfortable with the pull request as it stands now :)

File: music21/harmony.py [hunk 1]
```diff
@@ -1701,7 +1714,7 @@ def typeAlter(hD):
                 degree = degree.replace('A', '')  # A is for 'Altered'
                 if hD.degree == int(degree):
                     # transpose by semitones (positive for up, negative for down)
-                    p = p.transpose(hD.interval)
+                    p = p.transpose(hD.interval, inPlace=True)
```
mscuthbert, line 1717: No `p = ` for inPlace=True -- returns nothing.

mscuthbert (approved):
One minor change -- so minor, I can fix it in master.  THANK you all!  Merging!

PR #669, opened 2020-11-17
Cleanup pre 6.3

mscuthbert:
Deals with issues seen in #636 by making all properties (found so far) return sensible values rather than crashing on retrieval.  Fixes mock and other issues in Python 3.8.  

Also should fix corpus loading and converter.freeze()/converter.thaw() incompatibility across Mac and PC.

Pre 6.3 release

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -22,6 +22,7 @@
       <option name="ignoredNames">
         <list>
           <option value="type" />
+          <option value="format" />
         </list>
       </option>
     </inspection_tool>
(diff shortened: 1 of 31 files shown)


jacobtylerwalls:
Is there a reason we're skipping 6.2?

mscuthbert:
> Is there a reason we're skipping 6.2?

I've generally made even numbered minor versions be developer-in-progress releases and odd numbered ones be released versions, so if I see that someone reports an odd version number I know that they're on an official release and if I see that they're on an even number I know they're on a github pull.  :-)  It's mainly because I had doing 5.99.9-pre-6.0-alpha1 etc. version numbers, so that I'll never release a 7.0 -- only 7.1

Perl used to do this (though backwards, with evens being releases and odds being dev.) and I sort of liked that.

jacobtylerwalls:
Cool, I learned something!

PR #671, opened 2020-11-18
Skip Mac platform tests failing on Windows

jacobtylerwalls:
Patches the failures on Windows reported by @tanchihpin0517 in #660 . This is a pretty trivial approach. If `forcePlatform` doesn't/can't work any more, fix or deprecate?

### DIFF
--- a/music21/environment.py
+++ b/music21/environment.py
@@ -911,11 +911,14 @@ class Environment:
 
         Must call write() to make permanent:
 
+        >>> from pathlib import Path
         >>> a = environment.Environment()
         >>> a['debug'] = 1
         >>> a['graphicsPath'] = '/test&Encode'
-        >>> a['graphicsPath']
-        PosixPath('/test&amp;Encode')
+        >>> 'test&amp;Encode' in str(a['graphicsPath'])
+        True
+        >>> isinstance(a['graphicsPath'], Path)
+        True
 
         >>> a['autoDownload'] = 'asdf'
         Traceback (most recent call last):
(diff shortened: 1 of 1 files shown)


jacobtylerwalls:
Well, allegedly patches. @tanchihpin0517, are you up for checking out this branch on windows and reporting back to us? It would be a great help!

tanchihpin0517:
@jacobtylerwalls, there is still one failure:
```
File "C:\Users\scream\workspace\music21\music21\environment.py", line 1317, in __main__.UserSettings.__setitem__
Failed example:
    us['localCorpusPath'] = '/path/to/local'
Expected:
    music21.environment.UserSettingsException: attempting to set a value
        to a path that does not exist: /path/to/local
Got:
      File "C:\Users\scream\AppData\Local\Programs\Python\Python37\lib\doctest.py", line 1337, in __run
        compileflags, 1), test.globs)
      File "<doctest __main__.UserSettings.__setitem__[2]>", line 1, in <module>
        us['localCorpusPath'] = '/path/to/local'
      File "C:\Users\scream\workspace\music21\music21\environment.py", line 1334, in __setitem__
        value))
    music21.environment.UserSettingsException: attempting to set a value to a path that does not exist: \path\to\local
```

Here is the full log: [log.txt](https://github.com/cuthbertLab/music21/files/5557359/log.txt)

jacobtylerwalls:
Great, thanks, I think this should do it now.

tanchihpin0517:
I tested it, and there is no failure found.

mscuthbert:
Thank you!  There's a way to do substitutions in the expected value of doctests before comparing -- music21 (when it supported Py2 and Py3) used to do a lot of these.  Perhaps I should set it up to substitute `PosixPath\((.*))` with WindowsPath and the backslashes going in the opposite direction.  Then we can have PosixPath in output but pass on windows.

Issue #666, opened 2020-11-16
Second lyric in a note is not retrieved


DIDONEproject:
**music21 version** 
`6.1.0`  

**Problem summary**
Second lyric in a note is not retrieved.

![image](https://user-images.githubusercontent.com/52449822/99278752-bf1f3700-282f-11eb-87f7-68124eb3abe1.png)

In the example above, 'to' and 'o' are in different text annotations. This makes LyricSearcher to only retrieve 'to'.

**Steps to reproduce**

```python
music21.search.lyrics.LyricSearcher(self.score).indexText
```

**Expected vs. actual behavior**
The expected result is "ma to, o"
The actual result is "ma to,"

**Additional information**
In the documentation is indicated that currently it searches only the first lyric (http://web.mit.edu/music21/doc/moduleReference/moduleSearchLyrics.html) but I would sugest this to be a TODO instead of a restriction.

jacobtylerwalls:
Agreed. (And I think the very first TODO listed underneath that restriction is making your point.) Would you like to investigate and propose an approach?

mscuthbert:
This is an enhancement request, not a bug.  It's a known restriction -- would be glad to take a PR that fixes it.

DIDONEproject:
We have proposed an approach to solve this problem. We would be glad if you can consider it :)

#673

mscuthbert:
Fixed in #673

mscuthbert:
After thinking through the consequences of this issue a bit more, I've decided to change the implementation a bit in #794

PR #673, opened 2020-11-18
Issue #666: "Second lyric in a note is not retrieved" solved

DIDONEproject:
**Added the possibility of having more than one lyric annotation in a note.**

Works using musicxml subconverter (ConverterMusicXML):
```python
music21.converter.parse(xml_file_path, forceSource=True)
```

The modified files are:
- musicxml/xmlToM21.py
- sesarch/lyrics.py

Now, when executing:
```python
music21.search.lyrics.LyricSearcher(self.score).indexText
```
The whole text is retrieved, whitout missing any lyric :)


music21/musicxml/xmlToM21.py
```diff
@@ -4026,16 +4026,19 @@ def updateLyricsFromList(self, n, lyricList):
...
+            if lyricObj.text is not None:
+                if isinstance(lyricObj.text, list):
+                    text_list = lyricObj.text
+                    syllabic_list = lyricObj.syllabic
+                    for i, t in enumerate(text_list):
+                        uniqueLyricObj = copy.copy(lyricObj)
+                        uniqueLyricObj.text = t
+                        if syllabic_list and len(syllabic_list) > i:
+                            uniqueLyricObj.syllabic = syllabic_list[i]
+                        n.lyrics.append(uniqueLyricObj)
```
DIDONEproject, line 4038: If it is a list: parse it lyric by lyric. Otherwhise: introduce the lyric in the object.

jacobtylerwalls:
Thanks for this contribution. Could you include a test? At first glance there is a test file you could use in the repo already:
```
def testMultipleLyrics(self):
    from music21 import converter

    dir = common.getSourceFilePath() / 'musicxml' / 'lilypondTestSuite'
    fp = dir / '61b-MultipleLyrics.xml'
    s = converter.parse(fp)

    # assertions...
```


If I'm wrong, and that file does not exercise your code, there are other similar tests that block quote some xml either inline with the test or in testPrimitive.py. Thank you!

DIDONEproject:
Test included. As the case that I'm fixing was not in '61b-MultipleLyrics.xml', I built a new xml, '61L-MultipleLyricsPerNote.xml'. I hope that is okay!

jacobtylerwalls:
Awesome. If @mscuthbert wants to move the test file out of the Lilypond suite, he'll advise.

Would you be able to strip the two lines (the "Restriction" and "TODO:") that no longer apply after this patch in the docstring for `LyricSearcher`? Thanks!

DIDONEproject:
Lines in the docstring erased!

mscuthbert:
Hi!  Looking really good -- can you take the current test which does two things and split it:

1. In xmlToM21 we should not be testing if lyricSearcher works but rather if the parser successfully creates multiple Lyric objects in a stream on parsing.  So could the tests focus on that aspect, and also make sure that two lyrics with different syllabic values still parse properly.
2. In the lyric searcher, create a stream with notes with multiple lyrics by hand and then make sure that the lyric searching is working on that hand created stream?

The reason for this is that in the future, certainly something will break and it'll be much easier for me to diagnose where the breakage is if we're running separate tests on the parsing of XML lyrics and the searching of music21 lyrics.  (I've gotten good at predicting where my headaches will be 5 years from now).

DIDONEproject:
That has a lot of sense. I have divided the test in musicxml/xmlToM21 and search/lyric. I hope this changes will save you from the headache in the following years ;)

mscuthbert:
Paula and DIDONE project -- thank you so much for this contribution.  You've made the world of text/music relations a better place.

DIDONEproject:
Thank you for the great software you are developing! I think we can close the issue related to this ;)

mscuthbert:
After thinking through the consequences of this issue a bit more, I've decided to change the implementation a bit in #794

PR #674, opened 2020-11-21
Fix ScoreExporter's trouble with streams lacking measures

jacobtylerwalls:
Before, `ScoreExporter` would attempt to fixup un-measured notation, but that fixed up notation was then ignored by the branch of code responsible for exporting measures.

Curiously, `GeneralObjectExporter` tended to do just fine.


mscuthbert:
Wonderful!  Thank you!

PR #679, opened 2020-11-25
Refactor getTempFile() and catch PermissionError when unable to write to scratch directory

jacobtylerwalls:
Came up while discussing #660:

First commit refactors `getTempFile()`. Second commit attempts to address #660 by writing to a user-specific dir only when needed (when catching `PermissionError`). Second commit has lower utility if #660 or #680 is merged, but `PermissionError`s could still theoretically arise, even if those PRs make it less likely to arise accidentally.

### DIFF
--- a/music21/environment.py
+++ b/music21/environment.py
@@ -1628,7 +1628,8 @@ class Test(unittest.TestCase):
 
     @unittest.skipUnless(
         os.access(Environment().getDefaultRootTempDir(), stat.S_IRWXU),
-            'test will programmatically set read/write/exec permissions on this dir')
+        'test will programmatically set read/write/exec permissions on this dir'
+    )
     def testGetTempFile(self):
         import getpass
         import stat
--- a/music21/environment.py
+++ b/music21/environment.py
@@ -611,15 +611,6 @@ class _EnvironmentCore:
         >>> import pathlib
         >>> isinstance(e.getTempFile(), pathlib.Path)
         True
-
-        >>> oldScratchDir = e['directoryScratch']
-        >>> e['directoryScratch'] = None
-        >>> import getpass, os, stat
-        >>> os.chmod(e.getDefaultRootTempDir(), stat.S_IREAD)  # wipe out write, exec permissions
-        >>> 'music21-' + getpass.getuser() in e.getTempFile(returnPathlib=False)
-        True
-        >>> os.chmod(e.getDefaultRootTempDir(), stat.S_IREAD|stat.S_IWRITE|stat.S_IEXEC)  # restore
-        >>> e['directoryScratch'] = oldScratchDir  # restore
         '''
         # get the root dir, which may be the user-specified dir
         rootDir = self.getRootTempDir()
@@ -1460,6 +1451,7 @@ def get(key):
 # -----------------------------------------------------------------------------
 
 class Test(unittest.TestCase):
+    import stat
 
     def stringFromTree(self, settingsTree):
         etIndent(settingsTree.getroot())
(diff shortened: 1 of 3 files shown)


jacobtylerwalls:
I have another thought after putting this together: the status quo seems like a lot of overhead just to replicate what `NamedTemporaryFile` does by itself.

What if we refactored everywhere that needed to write to temp files to just:

```
with NamedTemporaryFile(dir=environLocal.getRootTempDir(), suffix=suffix, delete=False):
    # do the actual writing
```

instead of passing strings around.

And the reporter's problem goes away because we're no longer creating directories that end in `music21` on demand that other users might lack write access to.

jacobtylerwalls:
The "convenience" of having all the temp files in a given directory ('...music21') seems like a power-user convenience that could be recreated by specifying a scratch directory.

mscuthbert:
> I have another thought after putting this together: the status quo seems like a lot of overhead just to replicate what `NamedTemporaryFile` does by itself.

https://docs.python.org/2/library/tempfile.html

tempfile.NamedTemporaryFile([mode='w+b'[, bufsize=-1[, suffix=''[, prefix='tmp'[, dir=None[, delete=True]]]]]])
*New in version 2.3*

you've found something in the music21 code base that existed from back when minimum supported Python version was 2.1

:-D

will review more later, but likely this is the right approach.

mscuthbert:
> tempfile.NamedTemporaryFile([mode='w+b'[, bufsize=-1[, suffix=''[, prefix='tmp'[, dir=None[, delete=True]]]]]])
> _New in version 2.3_
> 
> you've found something in the music21 code base that existed from back when minimum supported Python version was 2.1

Actually, it's:

`New in version 2.6: The delete parameter.`

otherwise the file was deleted immediately after closing.  And Windows does not allow reading from the file while it's still open.  So it was sort of worthless on Windows until v2.6, and v2.5 of music21 was supported up until...2013?  So not that ancient history.

jacobtylerwalls:
Decided to remove the permission workarounds (the draft approach wasn't actually creating the directory that would be needed -- and I had no reason to believe that creating the new directory on the fly would or wouldn't have the same permission problems).

music21/environment.py
```diff
@@ -613,17 +611,43 @@ def getTempFile(self, suffix='', returnPathlib=True) -> Union[str, pathlib.Path]
...
         True
+
+        >>> oldScratchDir = e['directoryScratch']
+        >>> e['directoryScratch'] = None
+        >>> import getpass, os, stat
+        >>> os.chmod(e.getDefaultRootTempDir(), stat.S_IREAD)  # wipe out write, exec permissions
+        >>> 'music21-' + getpass.getuser() in e.getTempFile(returnPathlib=False)
+        True
+        >>> os.chmod(e.getDefaultRootTempDir(), stat.S_IREAD|stat.S_IWRITE|stat.S_IEXEC)  # restore
+        >>> e['directoryScratch'] = oldScratchDir  # restore
```
mscuthbert, line 622: I think that this is a little too complex for the doctest for this -- can we move to a unittest and put the restoration of directoryScratch into a "finally" clause of a try: so that if previous tests fail for some reason, we still restore the old test?

mscuthbert (commented):
Really like, but I think the Doctest is a bit too complex for doc-testing; can we move to unittest?

jacobtylerwalls:
Test runs on Catalina and High Sierra; is being skipped on Linux at Travis.

mscuthbert:

mscuthbert:
wonderful!  thanks!

PR #680, opened 2020-11-26
Create a fallback for environment.getDefaultRootTempDir()

tanchihpin0517:
Inspired by #679, handle #660 

Use tempfile.mkdtemp() to set and create default root temp directory.
This way is robust, cross-platform, and behave like the original routine.

~But it creates and deletes the temp directory for each python instance.~

### DIFF
--- a/music21/environment.py
+++ b/music21/environment.py
@@ -176,6 +176,7 @@ class _EnvironmentCore:
         # read will only right over values if set in field
         if forcePlatform is None:  # only read if not forcing platform
             self.read()  # load a stored file if available
+        self.defaultRootTempDir = None
 
     # SPECIAL METHODS #
 
@@ -498,19 +499,29 @@ class _EnvironmentCore:
         >>> e = environment.Environment()
         >>> e.getDefaultRootTempDir() == pathlib.Path(t) / 'music21'
         True
+
+        If failed to create the subdirectory (OSError is raised), this function
+        will return a temporary directory which is created by
+        tempfile.mkdtemp(prefix="music21-"), which is named with 'music21-'
+        plus 8 hashed codes. The location of this directory depends on OS.
+
+        Note the temporary dirctory will be different in each python session.
         '''
+        if self.defaultRootTempDir is not None and self.defaultRootTempDir.exists():
+            return self.defaultRootTempDir
+
         # this returns the root temp dir; this does not create a new dir
-        dstDir = pathlib.Path(tempfile.gettempdir()) / 'music21'
+        self.defaultRootTempDir = pathlib.Path(tempfile.gettempdir()) / 'music21'
         # if this path already exists, we have nothing more to do
-        if dstDir.exists():
-            return dstDir
+        if self.defaultRootTempDir.exists():
+            return self.defaultRootTempDir
         else:
             # make this directory as a temp directory
             try:
-                dstDir.mkdir()
-            except OSError:  # cannot make the directory
-                dstDir = dstDir.parent
-            return dstDir
+                self.defaultRootTempDir.mkdir()
+            except OSError:
+                self.defaultRootTempDir = pathlib.Path(tempfile.mkdtemp(prefix="music21-"))
+            return self.defaultRootTempDir
 
     def getKeysToPaths(self):
         '''


jacobtylerwalls:
I agree with the intent of using the standard Python library more when possible, however, a lot of places in the codebase currently expect the scratch directory to be a `pathlib.Path`:

```
>>> from music21 import *
>>> s = converter.parse('tinynotation: c1')
>>> s.write()
  File "/Users/jwalls/music21/music21/stream/__init__.py", line 253, in write
    return super().write(fmt=fmt, fp=fp, **keywords)
  File "/Users/jwalls/music21/music21/base.py", line 2554, in write
    return formatWriter.write(self,
  File "/Users/jwalls/music21/music21/converter/subConverters.py", line 974, in write
    xmlFp = self.writeDataStream(writeDataStreamFp, dataBytes)
  File "/Users/jwalls/music21/music21/converter/subConverters.py", line 936, in writeDataStream
    fp = self.getTemporaryFile()
  File "/Users/jwalls/music21/music21/converter/subConverters.py", line 234, in getTemporaryFile
    fp = environLocal.getTempFile(ext, returnPathlib=True)
  File "/Users/jwalls/music21/music21/environment.py", line 1028, in getTempFile
    filePath = envSingleton().getTempFile(suffix=suffix, returnPathlib=returnPathlib)
  File "/Users/jwalls/music21/music21/environment.py", line 608, in getTempFile
    fileDescriptor, filePath = tempfile.mkstemp(
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/tempfile.py", line 332, in mkstemp
    return _mkstemp_inner(dir, prefix, suffix, flags, output_type)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/tempfile.py", line 248, in _mkstemp_inner
    file = _os.path.join(dir, pre + name + suf)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/posixpath.py", line 76, in join
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not TemporaryDirectory
```

I'm warm to the idea of destroying temporary directories at the end of a Python session, but that is worth getting an opinion from @mscuthbert before going too much further.

tanchihpin0517:
> I agree with the intent of using the standard Python library more when possible, however, a lot of places in the codebase currently expect the scratch directory to be a `pathlib.Path`

Fixed :P. Thanks.

jacobtylerwalls:
Hi @tanchihpin0517 just trying to understand this proposal. Wouldn't we have a problem where:

1. User A creates a `TemporaryDirectory`, which sets it on the environment singleton `.defaultRootTempDir`
2. User B opens another python instance --> will they have a PermissionError trying to write to the shared `.defaultRootTempDir`?
3. User A ends the Python session or collects garbage
4. User B now might attempt to write files to a nonexistent directory.

Let me know if I'm understanding this right. Thanks.

tanchihpin0517:
> 2. User B opens another python instance --> will they have a PermissionError trying to write to the shared .defaultRootTempDir?

No, `TemporaryDirectory()` will create a unique directory for each call. It means User A and User B won't share the same temp directory. 

> 3. User A ends the Python session or collects garbage
> 4. User B now might attempt to write files to a nonexistent directory.

No, Python only delete A's temp directory. It won't do anything to B.

But if there is a scenario like this:
> User B want to read a temp file generated by User A, but User A has ended his python session.

or this:
> User A want to access a temp file generated from his previous python session.

Then it will fail because the first temp directory created by User A has been removed. That's what I worry about, although I think it shouldn't happen.

jacobtylerwalls:
Okay cool. I was assuming (wrongly) that these calls to `getDefaultRootTempDir()` somehow initialize the environment, but I realize now they don't. Thanks.

music21/environment.py
```diff
@@ -484,33 +485,20 @@ def toSettingsXML(self, ref=None):
...
-        else:
-            # make this directory as a temp directory
-            try:
-                dstDir.mkdir()
-            except OSError:  # cannot make the directory
-                dstDir = dstDir.parent
-            return dstDir
+        if self.defaultRootTempDir is None:
+            self.defaultRootTempDir = tempfile.TemporaryDirectory(prefix="music21-")
+        return pathlib.Path(self.defaultRootTempDir.name)
```
jacobtylerwalls, line 501: This is where it seems to me that User B would get the directory User A set for the entire environment.

jacobtylerwalls: I think I'm good now. The singleton is set when the module is imported. Every Python user is going to need to import music21. Sorry for the noise!

jacobtylerwalls:
Alright, sorry for the noise there while I walked through how the environment module works.

Taking a step back, I really like the idea of cleaning up these temp files. I think a new user is unlikely to go fishing around in `tmp/` for MuseScore files they generated and forgot to Save As, and an advanced user deploying an app is likely to wonder why music21 doesn't clean up its own temp files.

tanchihpin0517:
> Won't multi-user installs users share the same environment? Meaning the environment singleton's `.defaultRootTempDir` will be set to User A's TemporaryDirectory and then User B won't have a chance to create a TemporaryDirectory, will attempt to write to the other one and get a PermissionError or OSError? 

No,  the environment singleton only shared in one python session (or process from OS view)

> Or are you saying Linux users won't share the same environment singleton? 

Yes, more specifically saying, it does for all Python code, no matter which OS running on.

jacobtylerwalls:
My default reader is Finale, and I wrongly assumed MuseScore also opens unsaved copies of temp files. MuseScore just opens the temp file. So now I am slightly concerned that MuseScore users will wonder where their files went if they're not in the habit of saving as (and perhaps just use their recent files browser).

tanchihpin0517:
I didn't notice that. So should we use `tempfile.mkdtemp()` instead of `tempfile.TemporaryDirectory()`? `mkdtemp()` behave like `TemporaryDirectory()` but won't delete the temp directory automatically.

tanchihpin0517:
@jacobtylerwalls you are right. Deleting temp files automatically is harmful. I always assume that the temp files won't be opened after the end of python session. Thanks.
To avoid this problem, we should use `tempfile.mkdtemp()` to replace `tempfile.TemporaryDirectory()`.

mscuthbert:

mscuthbert:
Hi!  Previously the docs shows a fictitious but plausible directory name for this to write out.  Can we add that in with the #_DOCS_SHOW attribute?  I have no idea where the defaultRootTempDir() might be now.  Thanks!

tanchihpin0517:
Hi, I'm not familiar with how #_DOCS_SHOW works. Could you provide me some information about it? Like how to find the document.

jacobtylerwalls:
It's described here. Basically it's a way to have a line in the documentation that won't actually run as a test (because in our case here, checking for a certain temporary path would always fail):
https://web.mit.edu/music21/doc/developerReference/documenting.html#displaying-only-some-of-the-test-code-in-the-documentation

tanchihpin0517:
Hi, sorry for long time disappearing. I was struggling for many final projects recently (I'm a CS bachelor student haha). I have documented it. I'm not sure if it's right. please help me to check it. Thanks :)

mscuthbert:
Thanks @tanchihpin0517 -- and I appreciate how hard final projects are (as someone who assigns them to CS bachelor students.  :-)  )  -- with the revised changes, however, I'm not entirely sure how this helps the initial problem that this PR is supposed to address while also preserving existing temp dir solutions (I, for instance, have my music21 temp dir symlinked to my home directory on 4 different computers).  Can it check for the current solution and only fall back on yours if it is not writeable or does not exist?  THANKS!

tanchihpin0517:
Sure. But I find a perfect solution now. We just let the user who creates /tmp/music21 change the permission of that directory to writable for everyone, which solve any problem and won't change anything! :-D

tanchihpin0517:
Hm......I passed all tests except the last one, 'coverage/coveralls'.

>  coverage/coveralls — Coverage decreased (-0.002%) to 91.209% 

What does that mean? I couldn't figure out :(

tanchihpin0517:
I passed all tests except the last one, 'coverage/coveralls'.

> coverage/coveralls — Coverage decreased (-0.002%) to 91.209%

What does that mean? I couldn't figure out.

tanchihpin0517:
I passed all tests except the last one, coverage/coveralls.

> coverage/coveralls — Coverage decreased (-0.002%) to 91.209% 

What does that mean? I checked it but can't figure out where is the problem.

jacobtylerwalls:
I think programmatically changing permissions is too bold. It should be a last resort when catching an exception, if used at all.

Taking a step back, I believe when #679 was merged we covered the original issue you reported--that if User A creates the scratch directory `/tmp/music21`, user B won't have write access to it--because now we catch `PermissionError` and User B ends up writing to`/tmp/music21-userB` instead. I think if Linux users want to ensure that everybody shares the same directory, they should configure a scratch directory and not rely on the default. In other words, I think we now have sensible defaults and backups, and I'm wary of programmatically changing permissions every single time. (Someone will submit a ticket saying music21 is logging security events unnecessarily.)

The next path this PR took is I believe you wanted to use `tempfile.mkdtemp()` instead of `mkdir()` to create the original temp directories in `environment.getDefaultRootTempDir()`. That's probably still a useful optimization task, since the very last fallback for `tempfile.gettempdir()` is the current working directory, and I would think `mkdir()` could possibly fail there? Do you want to take this PR back in that direction?

PS -- I checked out v.6.3.0 and it appears even there the OS will clean up the temp directories and files on system restart. I was thinking #679 might have introduced that (good) behavior, but it appears we were already doing that.

jacobtylerwalls:
About Coveralls, every time you push, assuming the test suite passes, Coveralls updates the comment at the top of the thread with the missed lines report. Click the banner in the comment to jump to Coveralls. It shows your new contribution isn't covered by tests (probably because of the order in which the tests run -- your change covers the first time the temp directory is created). I would only worry about test coverage once we hear from @mscuthbert about the direction he wants to steer this in. Thanks!

tanchihpin0517:
> I think programmatically changing permissions is too bold. It should be a last resort when catching an exception, if used at all.
>
> Taking a step back, I believe when #679 was merged we covered the original issue you reported--that if User A creates the scratch directory `/tmp/music21`, user B won't have write access to it--because now we catch `PermissionError` and User B ends up writing to`/tmp/music21-userB` instead. I think if Linux users want to ensure that everybody shares the same directory, they should configure a scratch directory and not rely on the default. In other words, I think we now have sensible defaults and backups, and I'm wary of programmatically changing permissions every single time. (Someone will submit a ticket saying music21 is logging security events unnecessarily.)

I see. I didn't consider security problems.

> The next path this PR took is I believe you wanted to use `tempfile.mkdtemp()` instead of `mkdir()` to create the original temp directories in `environment.getDefaultRootTempDir()`. That's probably still a useful optimization task, since the very last fallback for `tempfile.gettempdir()` is the current working directory, and I would think `mkdir()` could possibly fail there? Do you want to take this PR back in that direction?

Because @mscuthbert  said he links his temp directory to different places, I think music21 is designed for allowing users to chase the temp files (or the directory). Then I think there is no need to use `tempfile.mkdtemp()` to replace the original one. Moreover, I think `mkdir` would never fail because `tempfile.gettempdir()` has made sure we can create the directory if it doesn't exists.

#679 has solved the current problem, so there is no problem now. I think this PR can be closed.

mscuthbert:
Okay, now that I know what this PR is doing (no permission changes, no change to default behavior, I still retain my massive speedup caches for already parsed files), I like it.  And I understand why we're storing `self.defaultRootTempDir` etc.  One last set of changes:

Merge the `if self.defaultRootTempDir is not None` and the `if self.defaultRootTempDir.exists()` into one line:  `if self.defaultRootTempDir is not None and self.defaultRootTempDir.exists():` in case the defaultRootTempDir was deleted during a music21 session (some people leave their interactive Jupyter notebook sessions on for days and an overzealous OS cleanup might have deleted the temp directory in the meantime).  Then after a `return` statement, there's no need for an `else:` which removes one layer of code indentation.

Update the docs to mention the fallback -- which has the `music21-` prefix instead of a separate directory.  No need for a doctest for that, but people who encounter that behavior should see it documented.

Thanks for this strong contribution to open source!

tanchihpin0517:
Done. Ready for review.

mscuthbert:
Thanks! Merging

PR #682, opened 2020-11-28
Sane Defaults for platform='nix'

wilsonify:
Changes needed to run unittests in music21/test 
out-of-the-box on Ubuntu 20.04

* set default environment when platform='nix'

* fix test case for testMultipartMeasures() per TODO

"this should show instruments, this is presently not showing instruments  probably b/c when appending to s Stream activeSite is set to that stream"

image


music21/test/testStream.py
```diff
@@ -167,24 +167,24 @@ def testMultipartStreams(self):
...
-        c = a[8][8:12]
-        d = a[8][12:16]
+        a = converter.parse(corpus.getWork(
+            workName='mozart/k155',
+            movementNumber='2',
+            fileExtensions='xml'
+        ))
+        b = a[1][0:8]
+        c = a[2][0:12]
+        d = a[3][0:16]
```
mscuthbert, line 182: this is an extremely old test that doesn't really follow the rules that it should -- what is a[1], a[2], a[3]?  Are they parts?  What is [0:8] etc. ?  If we can use a.parts[0], etc. this is probably much more what we'd like to have, and then call a slice from 0 to the index of measure(1) + 1, THANKS!

mscuthbert:

music21/test/testStream.py
jacobtylerwalls: I also found this recently and puzzled over it. The furthest I got was simplifying the parsing line to just:
`corpus.parse('mozart/k155', 2)`

wilsonify: I believe b,c,d are the different voices Violin I, Violin II, Viola
starting at 0 allows the names to be included in the resulting stream. 
I can introspect them and try to give them better names.

mscuthbert:
Nix users will thank you!

PR #683, opened 2020-11-28
Set bracket=False when bracket is undefined and show-number=None

jacobtylerwalls:
Before, on xml import, when brackets were undefined but show-number = None, this would result (beat 2):
implied-tuplet

Noticed this in the Grosse Fuge.


mscuthbert (approved):
Perfect!  Thanks!

Issue #684, opened 2020-11-29
Problems with parsing MIDI files with multiple tracks


GeorgeTouros:
**music21 version**

`6.3.0`

**Problem summary**
Parsing midi files fails to parse note duration and instrument information correctly

I try to parse and extract features from MIDI files that I have found on the Web. I am using the library found here: https://colinraffel.com/projects/lmd/
My goal is to use the feature extraction classes of music21 to build a classifier for my Thesis. 
When calling the converter.parse() method on the files, and then call the show() method, I get the following error:

```
meter: WARNING: Found a messed up beam pair <music21.beam.Beams <music21.beam.Beam 1/start>/<music21.beam.Beam 2/partial/right>>, <music21.beam.Beams <music21.beam.Beam 1/stop>/<music21.beam.Beam 2/stop>>, at index 5 of 
[None, <music21.beam.Beams <music21.beam.Beam 1/start>>, <music21.beam.Beams <music21.beam.Beam 1/stop>>, <music21.beam.Beams <music21.beam.Beam 1/start>/<music21.beam.Beam 2/partial/right>>, <music21.beam.Beams <music21.beam.Beam 1/stop>>, <music21.beam.Beams <music21.beam.Beam 1/start>/<music21.beam.Beam 2/partial/right>>, <music21.beam.Beams <music21.beam.Beam 1/stop>/<music21.beam.Beam 2/stop>>]
```

When the MuseScore application opens up to show the file, I get the following error:

![museScore3_error](https://user-images.githubusercontent.com/31730489/100544786-cd724780-3260-11eb-94b8-5e328a64b04c.png)

More importantly, the tempo and note duration is not parsed correctly, and neither is the track instrument information. 

**Steps to reproduce**
I've tested this using multiple files, but this one is the cleanest:

[0a2d626401e49eb46c59ec6dbd4f048fa305b001.zip](https://github.com/cuthbertLab/music21/files/5612297/0a2d626401e49eb46c59ec6dbd4f048fa305b001.zip)

```
>>> from music21 import *
>>> base_midi = converter.parse('0a2d626401e49eb46c59ec6dbd4f048fa305b001.mid', format='midi',
...                                                  forceSource=True, quantizePost=True)
>>> base_midi.show()
```

**Expected vs. actual behavior**
This is how the file is parsed when using MuseScore3 -> Open through the UI
[...]

PR #685, opened 2020-11-29
Don't assign durations to instruments during MIDI parsing

jacobtylerwalls:
Regression in #641. Was causing leading rests at the beginning of a part to be broken up into smaller chunks.
Noticed while poking at #684.


mscuthbert (approved):
Thanks!

PR #686, opened 2020-12-01
Set literal name parsed from SEQUENCE_TRACK_NAME or INSTRUMENT_NAME on instrument.partName or .instrumentName

jacobtylerwalls:
Refs #684 

**Before**: `instrument.partName` was not set on instruments created from parsing MIDI. If the instruments could successfully be normalized to m21 instruments, `.partName` on the `Part` would come out okay because of the way the property drills down to the instrument types. If a default instrument was created from a string like "Marcato" that doesn't normalize, the information was lost (there was no specific instrument for the property to drill down to).

**Now**: We store the text from SEQUENCE_TRACK_NAME or INSTRUMENT_NAME meta messages on `instrument.partName` so it will flow upstream to `Part.partName` and thus to an xml export, whether the text successfully normalizes to a specific instrument or not.

### DIFF
--- a/music21/midi/translate.py
+++ b/music21/midi/translate.py
@@ -689,6 +689,8 @@ def midiEventsToInstrument(eventList):
     '''
     Convert a single MIDI event into a music21 Instrument object.
     '''
+    from music21 import midi as midiModule
+
     if not common.isListLike(eventList):
         event = eventList
     else:  # get the second event; first is delta time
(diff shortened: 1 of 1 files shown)


mscuthbert:
will think about whether we want to override the PartName if an instrument can be detected from the MIDI program number -- a lot of MIDI files give good program numbers but just 'Instrument' or blank spaces for instrument name.  Perhaps checking for these two (putting a .strip() on decoded) and only setting part name if not matching either of these?  Otherwise looks great!

jacobtylerwalls:
I think this should come out okay with no add'l changes because .partName skips the generic Instrument with no details (track name) when returning the name of the first normalized instrument (program change). Will add a test👌🏻

jacobtylerwalls:
Oh now I see it, setting partName on that generic instrument (spaces etc) will introduce a problem -- you're right. Checking for spaces is simple. What about post processing to remove the generic instrument object if two are found at offset 0? (Thinking of this very test case with the instrument "Inst 1".)

jacobtylerwalls:
Now I'm thinking to handle each message separately, if TRACK_NAME set partName (even if spaces?); if INSTRUMENT_NAME set instrumentName, but only if not spaces. Still have the scenario where Inst 1 INSTRUMENT_NAME coincides with Guitar PROGRAM_CHANGE.

mscuthbert:
THANK YOU!  This is wonderful!

PR #687, opened 2020-12-01
Improve readability of parsing PITCH_BEND and POLYPHONIC_KEY_PRESSURE messages

jacobtylerwalls:
Noticed while debugging that the `else` case was capturing more than NOTE_ON and NOTE_OFF. So I looked up what was being captured and wrote cases for them. Needs a test file; none of the .midi files in the repo have PITCH_BEND messages, for instance.

### DIFF
--- a/music21/midi/__init__.py
+++ b/music21/midi/__init__.py
@@ -839,8 +839,8 @@ class MidiEvent:
             self.velocity = byte2
             return midiBytes[3:]
         elif self.type == ChannelVoiceMessages.POLYPHONIC_KEY_PRESSURE:
-            self.pitch = byte1
-            self.data = byte2
+            self.parameter1 = byte1  # pitch
+            self.parameter2 = byte2  # pressure
             return midiBytes[3:]
         raise TypeError(f'expected ChannelVoiceMessage, got {self.type}')  # pragma: no cover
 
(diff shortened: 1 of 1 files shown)


music21/midi/__init__.py
```diff
@@ -829,12 +829,20 @@ def parseChannelVoiceMessage(self, midiBytes: bytes) -> bytes:
...
+            self.parameter2 = byte2  # most significant byte
+            return midiBytes[3:]
+        elif self.type in (ChannelVoiceMessages.NOTE_ON, ChannelVoiceMessages.NOTE_OFF):
             # next two bytes:  pitch, velocity
             self.pitch = byte1
             self.velocity = byte2
             return midiBytes[3:]
+        elif self.type == ChannelVoiceMessages.POLYPHONIC_KEY_PRESSURE:
+            self.pitch = byte1
+            self.data = byte2
```
mscuthbert, line 843: Should this be parameter1, parameter2?

jacobtylerwalls: The values are pitch and pressure. Up to you.

jacobtylerwalls: The monophonic pressure message uses `self.data` for pressure, so that's what I was thinking.

mscuthbert:
It is possible to generate a MIDI file with PITCH_BEND messages just by writing from music21 a score with quarter tones or other microtones -- that would at least give some data for a test.  THANKS!

music21/midi/__init__.py
jacobtylerwalls: Looking more closely, .pitch and .data are both just synonyms for parameter 1 on setting; on getting, .pitch will only return the value if NOTE_ON or NOTE_OFF. so we definitely want parameter 1 and parameter2 because we'll lose byte2. Now that I see this I see that this patch is mostly just about readability of the code.

jacobtylerwalls:
Coveralls observed that the PITCH_BEND case was already tested by the doctest for MidiFile:
https://github.com/cuthbertLab/music21/blob/506b28c109360d3054604dff77468da184ddcb86/music21/midi/__init__.py#L1235-L1239

PR #688, opened 2020-12-02
Retain instruments outside of voices when making measures

jacobtylerwalls:
Refs #684  (Turned out to be easier than I thought!)

**Before:** elements outside of voices, such as Instruments, would vanish when running `.makeMeasures`.

### DIFF
--- a/music21/stream/__init__.py
+++ b/music21/stream/__init__.py
@@ -5756,6 +5756,9 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
             groups = []
             for i, v in enumerate(srcObj.voices):
                 groups.append((v.flat, i))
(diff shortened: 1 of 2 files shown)


mscuthbert:
LGTM pending final check (Travis is SLOW these days) -- OffsetMap is a concept that should be changed in v.7 -- no reason to use srcObj, and probably just using the tree-object and getting the voice from .sites would be simpler.

jacobtylerwalls:
Yeah, I read the blog posts at Travis, apparently they're being sucked dry by crypto miners. 🙄

mscuthbert:
Oh gee... I had no idea about the crypto miners.  Poor travis.

PR #697, opened 2020-12-05
Coverage

jacobtylerwalls:
Increase coverage

### DIFF
--- a/music21/common/numberTools.py
+++ b/music21/common/numberTools.py
@@ -707,7 +707,8 @@ def unitNormalizeProportion(values):
     #[0.20000000000000001, 0.59999999999999998, 0.20000000000000001]
 
 
-    Negative values should be shifted to positive region first
+    Negative values should be shifted to positive region first:
+
     >>> common.unitNormalizeProportion([0, -2, -8])
     Traceback (most recent call last):
     ValueError: value members must be positive
--- a/music21/common/numberTools.py
+++ b/music21/common/numberTools.py
@@ -442,7 +442,13 @@ def almostEquals(x, y=0.0, grain=1e-7):
     >>> common.almostEquals(1.001, 1, grain=0.1)
     True
 
-    For very small grains, just compare Fractions without converting...
+    OMIT_FROM_DOCS
+    For very small grains, just compare Fractions without converting:
+    >>> from fractions import Fraction
+    >>> x = Fraction(1e-10)
+    >>> y = Fraction(1e-10)
+    >>> common.almostEquals(x, y)
+    True
 
     :rtype: bool
     '''
@@ -701,9 +707,13 @@ def unitNormalizeProportion(values):
     #[0.20000000000000001, 0.59999999999999998, 0.20000000000000001]
 
 
+    Negative values should be shifted to positive region first
+    >>> common.unitNormalizeProportion([0, -2, -8])
+    Traceback (most recent call last):
+    ValueError: value members must be positive
+
     :rtype: list(float)
     '''
-    # note: negative values should be shifted to positive region first
     summation = 0
     for x in values:
         if x < 0:
@@ -1063,6 +1073,10 @@ def toRoman(num):
     Traceback (most recent call last):
     TypeError: expected integer, got <... 'str'>
 
+    >>> common.toRoman(0)
+    Traceback (most recent call last):
+    ValueError: Argument must be between 1 and 3999
+
     :rtype: str
     '''
     if not isinstance(num, int):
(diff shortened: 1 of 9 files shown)


music21/converter/__init__.py
```diff
@@ -632,11 +632,16 @@ def parseURL(self, url, format=None, number=None, **keywords):  # @ReservedAssig
         Note that this checks the user Environment
         `autoDownload` setting before downloading.
 
+        >>> us = environment.UserSettings()
+        >>> us['autoDownload'] = 'allow'
```
mscuthbert, line 636: We shouldn't be changing UserSettings during any test scripts.  in particular here, we're going to wipe out whatever the user's autoDownload already was and set it to deny afterwards.  I think that this should be set to what it was before and we can use environment.Environment() in a unit test instead.

music21/corpus/__init__.py
```diff
@@ -246,12 +246,11 @@ def getComposer(composerName, fileExtensions=None):
     )
 
 
-@property
```
mscuthbert, line 249: can you explain this change?  Thanks!

music21/common/numberTools.py
```diff
@@ -706,9 +707,13 @@ def unitNormalizeProportion(values):
     #[0.20000000000000001, 0.59999999999999998, 0.20000000000000001]
 
 
+    Negative values should be shifted to positive region first
+    >>> common.unitNormalizeProportion([0, -2, -8])
```
mscuthbert, line 711: Sphinx needs a blank line between docs and code examples to format properly.

music21/test/testStream.py
```diff
@@ -1418,11 +1418,9 @@ def testMeasureOffsetMap(self):
...
         self.assertEqual(sorted(list(mOffsetMap.keys())), [8.0])
 
-        # this should work but does not yet
-        # it seems that the flat score does not work as the flat part
-#         mOffsetMap = a.flat.measureOffsetMap('Note')
-#         self.assertEqual(sorted(mOffsetMap.keys()),
-#                [0.0, 4.0, 8.0, 12.0, 16.0, 20.0, 24.0, 28.0, 32.0]  )
+        mOffsetMap = a.flat.measureOffsetMap('Note')
+        self.assertEqual(sorted(mOffsetMap.keys()),
+                         [0.0, 4.0, 8.0, 12.0, 16.0, 20.0, 24.0, 34.0, 38.0])
```
mscuthbert, line 1423: Are these numbers definitely right?  the failing test was for 20, 24, 28, 32 -- I can't remember how this works, but want to make sure it's right

mscuthbert (changes requested):
THANK YOU!  This is really great!  The only thing that really needs to be changed is the download of the Jeannie file can't mess with UserSettings -- I think it's better as it was before, but then having a UnitTest that actually downloads a file.

(If we can, in the unittest, switch to something even smaller and not by Stephen Foster -- learning more about him, I'm trying to deemphasize him in the docs.  Not going to remove the piece but trying to use it in fewer examples.  THANKS!)

music21/corpus/__init__.py
jacobtylerwalls: Coveralls showed that the doctest wasn't running, and since all the doctest did was attempt to call the property, it suggested to me it wasn't behaving as intended, as here:
```
>>> corpus.noCorpus
<property object at 0x11e4dd450>
```
Is that the intended use?

music21/test/testStream.py
jacobtylerwalls: This chorale has a psalm tone, so these numbers looked correct when I opened the chorale.

jacobtylerwalls:
Agree. Now that I look, there already is a unit test for parseURL(), it just doesn't run on CI because it's in `TestExternal`. The difference between `Test` and `TestExternal` seems to be a difference of degree (as long as the file isn't too long...)

Do you suggest moving that test into the `Test` class? And then long-term trying to avoid loading of large external files in any event and eventually get rid of those extra test classes?

mscuthbert:
> Agree. Now that I look, there already is a unit test for parseURL(), it just doesn't run on CI because it's in `TestExternal`. The difference between `Test` and `TestExternal` seems to be a difference of degree (as long as the file isn't too long...)

TestExternal is allowed to do things like start playing music or pop up a graph on the screen or open MuseScore or even print to the console.  Test shouldn't have any effects other than sending errors or pass to unittest.

> Do you suggest moving that test into the `Test` class? And then long-term trying to avoid loading of large external files in any event and eventually get rid of those extra test classes?

It's probably worth moving the test to the Test class -- at some point Chris Ariza and I wanted Test to be able to run without an internet connection (he commuted up from Baltimore or at least NYC weekly and the WiFi wasn't so good on the Acela) but once we added the ComposerPopularity feature, we stopped carring about that.

jacobtylerwalls:
Decided to leave it where it was (it was long: 30 seconds on 1 core) and just write a new one for a 13 bar Chopin prelude.

mscuthbert (approved):
Looks great to me!

mscuthbert:
Thank you!

PR #703, opened 2020-12-09
MIDI translate -- ensure conductor track is first

mscuthbert:
* Document a number of under-documented midi.translate routines

* Add typing in MIDI

* Upgrade version since conductor track is breaking old pickles (for testing only)

* tempo.MetronomeMark -- FINALLY allow a number alone.

* Fix pyCharm linter errors

### DIFF
--- a/music21/_version.py
+++ b/music21/_version.py
@@ -42,7 +42,7 @@ When changing, update the single test case in base.py.
 Changing this number invalidates old pickles -- do it if the old pickles create a problem.
 '''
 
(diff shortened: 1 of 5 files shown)


mscuthbert:
@jacobtylerwalls  -- these were some of the docs I was thinking of.   Also found that conductor track was sometimes being put at the end of the file/stream, so adjusted priority so it is always the first part.

jacobtylerwalls:
Great, thanks for those docs changes and for the priority ordering. Wasn't familiar with the priority concept!

PR #705, opened 2020-12-11
Write individual files from opus.write() and return path to last

jacobtylerwalls:
Noticed while reviewing #651 . Every other `.write()` returns something, but not `opus.write()`. What do you think?

### DIFF
--- a/music21/stream/__init__.py
+++ b/music21/stream/__init__.py
@@ -23,6 +23,8 @@ this module.
 import collections
 import copy
 import itertools
+import math
+import pathlib
 import unittest
 import warnings
 import sys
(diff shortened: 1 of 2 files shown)


mscuthbert:
Looks good to me -- the only thing is that the 'fp=fp' doesn't make sense in Opus (pre-existing bug.  Maybe it should do something like fp-03.xml etc -- or just ignore fp for now.  But overwriting the same file over and over again really doesn't do anything.

Or perhaps, if we do this, Opus should return the single LAST filename, since it's would then be easy to reconstruct the other filepaths from it.  That way we're trying to follow the substitution equality trait where all .write() calls return a single file name?

jacobtylerwalls:
That's amusing, I didn't even notice the overwriting. I like just returning the last. Maybe print statements or warnings wouldn't hurt. archiveTools.compressXML spits out warnings when it updates files, which clutters the test suite output.

music21/stream/__init__.py
```diff
@@ -13381,21 +13382,37 @@ def mergeScores(self):
...
 
+        Individual files are written for each score; returns the last file written.
+
+        >>> sc1 = stream.Score()
+        >>> sc2 = stream.Score()
+        >>> o = stream.Opus()
+        >>> o.append([sc1, sc2])
+
+        #_DOCS_SHOW >>> o.write()
+        #_DOCS_SHOW PosixPath('/some/temp/path')
```
mscuthbert, line 13400: Should probably end in .xml or .musicxml

mscuthbert: And also end in -2.xml

music21/stream/__init__.py
```diff
@@ -13381,21 +13382,37 @@ def mergeScores(self):
...
         if fmt is not None and 'lily' in fmt:
             return Stream.write(self, fmt, fp, **keywords)
         elif common.runningUnderIPython():
             return Stream.write(self, fmt, fp, **keywords)
-        else:
-            for s in self.scores:
-                s.write(fmt=fmt, fp=fp, **keywords)
+        post = []
+        for i, s in enumerate(self.scores):
+            if fp is not None:
```
mscuthbert, line 13408: I think what we want to do her is factor the fp check out earlier and something like:

```
if fp is None:
    fp = ...whatever the function to get the fp from format is as a Path...
elif isinstance(fp, str):
    fp = pathlib.Path(fp)

fpParent = fp.parent
fpStem = fp.stem
fpSuffix = '.'.join(fp.suffixes)

numZeros = int(math.log10(len(self.scores))) + 1
formatString = f'0{numZeros}d'

for i, s in enumerate(self.scores):
     scoreName = fpStem + f'{i + 1:{formatString}}' + '.' + fpSuffix
     fpToUse = fpParent / scoreName
     ...
```

Something like that should assure that whatever is passed in as fp all the files get the right extension AND also have sequence numbers.  And if no fp is passed in, all the temporary files still have the same name with extensions.

mscuthbert: All this might be in a helper function that returns the entire list of post, so that .show() can do the right thing with it.

mscuthbert (changes requested):
Looks really good -- some improvements suggested.

music21/stream/__init__.py
jacobtylerwalls: Good stuff here. Not sure I follow what `show` would be able to do with it, since show will just write from scratch and then launch, and doesn't take an fp as an argument.

mscuthbert: Ah True!

mscuthbert:
Looks great!  Waiting for pylint, then merging!

PR #706, opened 2020-12-12
WIP -- handle out of measure bounds `<direction>` tags (and avoid creating gaps)

jacobtylerwalls:
Fixes an issue where `<direction>` tags with `<offset>` subelements yielding unique offsets in the measure would create gaps when exporting to xml:

```
>>> s = corpus.parse('weber')
>>> s.parts[0].measure(1).show('t')
...
{0.0} <music21.meter.TimeSignature 3/4>
{2.0} <music21.note.Rest rest>
{3.1064} <music21.tempo.MetronomeMark larghetto Quarter=60.0>
>>> s.show()
```

Output:
before

Now:
after

Source musicxml from the corpus file (preceding the initial rest; divisions=1024):
```
      <direction placement="above">
        <direction-type>
          <words default-y="40" font-size="8.4" relative-x="-74">Adagio ma non troppo</words>
        </direction-type>
        <sound tempo="72"/>
      </direction>
      <direction placement="above">
        <direction-type>
          <metronome default-y="42" font-family="Petrucci" font-size="10.8" halign="left">
            <beat-unit>quarter</beat-unit>
            <per-minute>60</per-minute>
          </metronome>
        </direction-type>
        <offset>3181</offset>
        <sound tempo="60"/>
      </direction>
```

~WIP for now; needs tests.~ done ( at least for MetronomeMark)

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -2782,6 +2782,18 @@ class MeasureExporter(XMLExporterBase):
 
         self.currentVoiceId = voiceId
 
+        # Stash fractional offsets from elements that will become <direction> tags
+        # and convert later to <offset> subelement
+        # to prevent unnecessary <forward> tags (i.e. rests)
+        for elem in m.getElementsByClass(
+            ('TempoIndication', 'Segno', 'Coda', 'Dynamic', 'RehearsalMark', 'TextExpression')
+        ):
+            integer = math.floor(elem.offset)
+            remainder = elem.offset - integer
+            if remainder > 0:
+                elem.editorial['offsetRemainder'] = remainder
+                elem.offset = integer
+
         # group all objects by offsets and then do a different order than normal sort.
         # that way chord symbols and other 0-width objects appear before notes as much as
         # possible.
@@ -4799,10 +4811,16 @@ class MeasureExporter(XMLExporterBase):
     def setOffsetOptional(self, m21Obj, mxObj=None, *, setSound=True):
         '''
         If this object has an offset different from self.offsetInMeasure,
+        after accounting for any stashed offset remainder,
         then create and return an offset Element.
 
         If mxObj is not None then the offset element will be appended to it.
         '''
+        try:
+            stashed = m21Obj.editorial['offsetRemainder']
+            m21Obj.offset = m21Obj.offset + stashed
+        except KeyError:
+            pass
         if m21Obj.offset == self.offsetInMeasure:
             return None
         offsetDifferenceInQl = m21Obj.offset - self.offsetInMeasure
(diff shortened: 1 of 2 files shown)


mscuthbert:
How did you KNOW that I had just filed in my own TODO yesterday to figure out what was breaking on the Weber corpus show???  Mind reader?  Or do you have a keyboard tracker planted on my computer?  WOW!

mscuthbert:
I want to think of this in a bit of depth.  It seems to me that encoding offset > end of the measure is actually non-sensical: if the result of a direction takes place after the measure ends, then it should be encoded in the following measure.  ("in 4/4, crescendo until you reach forte by beat 5" should actually be "crescendo until you reach forte by beat 1 of the next measure")  Cases where we want the direction to be logically part of one measure but displayed as part of the next measure, should be encoded with relative-x attributes.

So that I think solves the "how this should be encoded" problem.  What about the how it's actually encoded problem?  I can think of a couple of things to do:

(1) when parsing musicxml, ignore offset tags that are > the end of the measure and put them either at the end of the measure, or something like 1/4 beat before the end of the measure.  This solves the parsing of this example.  (The metronome mark should surely actually be at beat 1/offset 0, and has just been moved right to avoid collision with the tempo mark).

(2) when outputting direction tags, I don't think that looking at floor(offset) == offset is the solution.  In 6/8, for instance, a fractional offset of 1.5 is a very important beat position.  My idea is that if the offset position is > currentTimeSignature.barDuration.quarterLength then we want to put it in musicxml at the position defined by (1) above.  (This expensive search operation could be sped up by checking that offset > 0 first).  The MusicXML offset-divisions is what I believe should be stored in the editorial for the element: `direction.editorial['musicxmlOffsetPlacement'] = ...`  and then in generating the musicxml `<offset>` tag, we should look for the existence of that key first and only use the .offset of the element if it's absent.  This way we're not manipulating the offset of the actual element and putting it back (this is an approach I try to avoid, because it makes debugging really hard, since unless you happen to look at .offset at the right moment in your debugger, you can miss that it changed and then changed back)  

Thoughts on these possible directions?

jacobtylerwalls:
- Rest assured, no telepathy--only looking for piano music in the corpus to regression test the joinPartStaffs PR (that will probably become a v.7 breaking change at some point)

- Yes, I became unsatisfied with this approach shortly after writing the test and realizing I had no reason to believe that integer offsets were the most reasonable locations. Agree that changing offset on the fly is not great.

- If we think the beginning of the measure is realistically the logical location, when parsing musicxml, what about just putting these elements at the beginning of the measure and recalculating the offset placement on that basis before saving to `direction.editorial['musicxmlOffsetPlacement']`? Or were you suggesting using `relative-x`?

- I think this isn't just an end of measure scenario. Anywhere you have a calculated offset from the `<direction><offset>` that doesn't correspond with a musical offset, like 3.14159 in a 4/4 measure, I think we will risk having either gaps or voices created:
```
        s = converter.parse('tinynotation: 4/4 d2')
        mm = tempo.MetronomeMark(number=144)
        s.measure(1).insert(3.14159, mm)
        s.show()
```

midmeasure-gap

We could compare the keys in the offsetMap of a stream filtered on just these elements that map to directions, and find out if that key is not in stream.notesAndRests.offsetMap. If we get one, we need special handling to prevent creating forward tags or voices at those offsets. So for those directionElements, set the offset to the measure start and recalculate and store `direction.editorial['musicxmlOffsetPlacement']`. What do you think?

mscuthbert:
This is the part that I'm going to need to look more about Directions and Offsets in musicxml to determine.  There doesn't need to be a note in order to place a direction there, so we can't always snap to notes.  I'm less concerned about hidden-rests or gaps or voices created in the case you've presented above, where the measure isn't complete/doesn't agree with  the 4/4 time signature.  When changing to '4/4 d2 d2' there are no problems, though the metronom mark appears at the end of the measure rather than about 1 beat earlier.

jacobtylerwalls:
I had another think about this since dealing with positions for directions in #923.

> I'm less concerned about hidden-rests or gaps or voices created in the case you've presented above, where the measure isn't complete/doesn't agree with the 4/4 time signature.

Fair, incomplete measures aren't worth coding around, but I encounter incomplete voices in ensemble music where parts will be extracted, and cross-staff piano music is also valid but leaves gaps in m21. So I would think it's still worth avoiding creating gaps.

I wouldn't insist on this approach, of course, but at some point I can see about making a proof of concept for avoiding gaps. If it's a small diff then perhaps simpler than messing with the editorial object. Probably will revisit in the summer.

mscuthbert:
Incomplete voices are definitely fine and supported, as are incomplete measures in most cases (though paddingLeft especially and paddingRight are better to be encoded).  But voices that overfill the measure are also not something necessarily to code around.

jacobtylerwalls:
Will reopen if I return to this.

Issue #699, opened 2020-12-05
Performance bottleneck in music21.key.Key.getLeadingTone()


napulen:
**music21 version**

`6.3.0`   

**Problem summary**

While optimizing an algorithm, I noticed an (odd) performance bottleneck, on the `getLeadingTone()` function. This operation was taking several orders of magnitude longer than, for example, comparing two pitches, or creating an Interval.

**Steps` to reproduce**

Here is a short snippet to show this:

```python
In [1]: import music21

In [2]: music21.__version__
Out[2]: '6.3.0'

In [3]: import timeit

In [4]: setup = """
   ...: import music21
   ...: 
   ...: k = music21.key.Key('C')
   ...: """

In [5]: timeit.timeit('k.getLeadingTone()', setup=setup)
Out[5]: 33.03134616499983

In [6]: setup2 = """
   ...: import music21
   ...: 
   ...: c = music21.chord.Chord(("C4", "E4", "G4"))
   ...: """

In [7]: timeit.timeit('c.root()', setup=setup2)
Out[7]: 0.161935043000085
```

Not sure if its fair to use a `chord.Chord.root()` to compare, but you get the idea. This is consistent with what I found across many music21 operations. 33 seconds to get the LeadingTone 10,000 times, compared to 0.16 seconds for the root() of a chord.

I looked at the code of `getLeadingTone()` to see if I could make a PR myself. It seems to be doing a fairly straightforward operation: compute the interval between the tonic and the seventh degree, if it's different than a `M7`, compute a `M7` from the tonic and return that. Not sure why that could be taking so long. Since a leading tone is a constant, maybe it could be a `@cached_property`?

As an "advanced" user of music21, I feel kinda guilty submitting issues rather than PRs. If I get a chance later, I'll take a closer look and work on a PR.

jacobtylerwalls:
Identifying issues is useful work! :-)

First thing I see off the bat is that `self.pitchFromDegree(7)` is calculated twice.

What are the benchmarks for `.notesToInterval()` and `.pitchFromDegree()`?

napulen:
Heres `notesToInterval()`

```python
In [9]: setup = """
   ...: import music21
   ...: 
   ...: n1, n2 = music21.note.Note('C4'), music21.note.Note('B4')
   ...: """

In [10]: timeit.timeit("music21.interval.notesToInterval(n1,n2)", setup=setup)
Out[10]: 19.901850692000153
```
and `pitchFromDegree()`
```python
[...]

PR #707, opened 2020-12-13
Alternative (faster) implementation of DiatonicScale.getLeadingTone()

napulen:
Here is the new implementation of `DiatonicScale.getLeadingTone()` as discussed in #699.

Based on the last comment by @mscuthbert (each step appears exactly once in DiatonicScales), it should be safe to omit the `mod 12` in the condition.

Credit for faster check on 11 semitones (rather than `M7` Intervals) goes to @jacobtylerwalls.


mscuthbert:
Perfect.  Thanks!

PR #709, opened 2020-12-15
Prevent multiple periods in the filepath from opus.write()

jacobtylerwalls:
When writing a test for the line I hadn't covered yet, sure enough there was a bug. Multiple periods were being created in a couple places.

Also short-circuited on lack of `.scores` so that we don't raise an ugly `MathDomainError`.


mscuthbert:
Awesome.  Thanks!

PR #712, opened 2020-12-15
Retain midi channel

### LINKED ISSUE #465: Retain channel information from MIDI
We've got a somewhat related problem:
We're trying to read a MIDI file generated by a [Jammy guitar](https://playjammy.com/).
The Jammy uses 6 MIDI channels for the 6 strings of the guitar.
We would like to retain this information during MIDI => Music21 transformation in order to e.g. do analysis later on how a chord was fingered.
Is there some way to transform MIDI into a Music21 stream without losing the channel information?

Thanks!

(Note: This is a cross-post from here https://github.com/cuthbertLab/music21/issues/206, I wasn't sure if someone would see this in a closed ticket).

jacobtylerwalls:
Resolves #465

Store the channel parsed from a midi note on `.editorial['channel']` for note; `.editorial['channels']` for chord.

Update MIDI docs: `midi.base` is just `midi` now.

### DIFF
--- a/music21/midi/translate.py
+++ b/music21/midi/translate.py
@@ -353,6 +353,7 @@ def midiEventsToNote(eventList, ticksPerQuarter=None, inputM21=None):
     n.volume.velocity = eOn.velocity
     n.volume.velocityIsRelative = False  # not relative coming from MIDI
     # n._midiVelocity = eOn.velocity
+    n.editorial['channel'] = eOn.channel
     # here we are handling an issue that might arise with double-stemmed notes
     if (tOff - tOn) != 0:
         ticksToDuration(tOff - tOn, ticksPerQuarter, n.duration)
@@ -506,6 +507,7 @@ def midiEventsToChord(eventList, ticksPerQuarter=None, inputM21=None):
         c = chord.Chord()
     else:
         c = inputM21
+    c.editorial['channels'] = []
 
     if ticksPerQuarter is None:
         ticksPerQuarter = defaults.ticksPerQuarter
@@ -526,6 +528,7 @@ def midiEventsToChord(eventList, ticksPerQuarter=None, inputM21=None):
             p = pitch.Pitch()
             p.midi = eOn.pitch
             pitches.append(p)
+            c.editorial['channels'].append(eOn.channel)
             v = volume.Volume(velocity=eOn.velocity)
             v.velocityIsRelative = False  # velocity is absolute coming from
             volumes.append(v)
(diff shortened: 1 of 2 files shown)


mscuthbert:
When changing how files are parsed, be sure to update music21's most minor version number (the 1 or 6.5.1) and update the test in base.py since this will invalidate old caches of parsed files -- I've twice been bit by having old cached versions of files around when updating to master.  Thanks!

music21/midi/translate.py
```diff
@@ -515,6 +517,15 @@ def midiEventsToChord(eventList, ticksPerQuarter=None, inputM21=None):
     pitches = []
     volumes = []
 
+    def handleEvent(event, pitches, volumes, thisChord):
+        p = pitch.Pitch()
```
mscuthbert, line 521: Now that this is factored out, can it have docs and typings?  What are pitches?  pitch objects? etc.  I know it was just a way of reducing code duplication -- which is great! -- but not that it's moved out of context, it's harder to understand what is happening than before.  With docs and doctests and typing it's a big positive.

music21/midi/translate.py
```diff
@@ -515,6 +517,15 @@ def midiEventsToChord(eventList, ticksPerQuarter=None, inputM21=None):
...
+        p.midi = event.pitch
+        v = volume.Volume(velocity=event.velocity)
+        v.velocityIsRelative = False  # velocity is absolute coming from MIDI
+        pitches.append(p)
+        volumes.append(v)
+        thisChord.editorial['channels'].append(event.channel)
```
mscuthbert, line 527: since channels is MIDI specific and not respected for output, can we make it `importedFromMidiChannels`?  or something like that?  

Are there cases where notes/chords in a single part will be imported from multiple channels?  Should this be put on the Part object instead?

mscuthbert (commented):
It looks on the right track, but I'm a bit confused that we're putting this information on the Pitch object -- adding an editorial will weigh each pitch down a bit; why isn't it on the Stream?

music21/midi/translate.py
jacobtylerwalls: I assumed in the reporter's use case--with the Jammy guitar with different fingerings on each note represented by different channels--there would be a stream of notes in a single track all with different channels, and this is what was being preserved. I'll ask!

jacobtylerwalls: If this is too much overhead then maybe we don't want this in core, and then @nightscape can just fork and merge this branch.

mscuthbert: Pitches are quite lightweight -- they use slots, etc. to keep them really memory efficient.  The editorial was something I debated putting on but decided to do so because it only adds a single byte per pitch if it's None, but I think it will double the size of pitches if it's declared on every object, and then people parsing very large MIDI files will notice.

What really needs to happen is a complete refactor of the MIDI translate into two files toMidi and fromMidi with __init__.py going to objects.py and an object-oriented translator that can have keyword settings on what types of quantization, whether to add channel information, etc. but that's a MUCH larger project.

jacobtylerwalls:
Makes sense. Closing, but the commits are here if @nightscape wants it. I'll open a new PR with the docs cleanup.

Issue #523, opened 2020-04-07
from music21 import * has gotten slower


mscuthbert:
Where is the slowness coming? We had made this very efficient at some point and now it's nearing a second.  Probably something creating many extra objects that are not needed.

PR #716, opened 2020-12-17
Speed up startup by only importing joblib as needed

mscuthbert:
Moves init from 0.8sec to 0.4sec!  (47% speed increase!)

Also load chardet and PIL late -- down to 0.38 seconds.

Remove non-functioning testGraphImportStar


mscuthbert:
Fixes #523

PR #718, opened 2020-12-18
Remove deprecated Environment.launch()

jacobtylerwalls:
- [ ] someone should test on Windows! :)

### DIFF
--- a/music21/environment.py
+++ b/music21/environment.py
@@ -715,60 +715,8 @@ class _EnvironmentCore:
 
         Create a png, svg, etc. converter (or just a graphics converter) and call launch on it
         '''
-        # see common.fileExtensions for format names
-        filePath = common.cleanpath(filePath, returnPathlib=True)
-
-        m21Format, unused_ext = common.findFormat(fmt)
-        environmentKey = self.formatToKey(m21Format)
-        if environmentKey is None:
-            environmentKey = self.formatToKey(fmt)
-        if m21Format == 'vexflow':
-            try:
-                import webbrowser
-                filePath = filePath.as_uri()
-                webbrowser.open(filePath)
-                return
-            except ImportError:
-                print(f'Cannot open webbrowser, sorry. Go to file://{filePath}')
-        if app is not None:
-            # substitute app provided via argument
-            fpApp = app
-        elif environmentKey is not None:
-            fpApp = self._ref[environmentKey]
-        else:
-            fpApp = None
-
-        platform = common.getPlatform()
-        if fpApp is None:
-            if platform == 'win':
-                # no need to specify application here:
-                # windows starts the program based on the file extension
-                cmd = f'start {filePath}'
-            elif platform == 'darwin':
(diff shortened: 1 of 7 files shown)


music21/environment.py
```diff
@@ -714,60 +714,8 @@ def launch(self, fmt, filePath, options='', app=None):
...
-        elif platform == 'darwin':
-            cmd = f'open -a"{fpApp}" {options} "{filePath}"'
-        elif platform == 'nix':
-            cmd = f'{fpApp} {options} "{filePath}"'
-        else:
-            raise Exception(f'Unknown platform {platform}.')
-
-        os.system(cmd)
+        from music21.converter.subConverters import SubConverter
+        return SubConverter().launch(filePath, fmt=fmt, options=options, app=app)
```
mscuthbert, line 718: these are very similar routines, but not identical.  The SubConverter launch assumes that filePath has already been cleaned and that fmt has already been regularized.  Probably best to make SubConverter take over the safety checks that Enviornment once did.  Has this been tested for the tricky cases of graph.launch(), etc. listed in the docs?  (Also need docs updated to "undeprecate")  THANKS!

jacobtylerwalls: `SubConverter` subclasses take care of cleaning paths in `.write()`, so I don't think we should add any more checks to subclass `launch` methods. Obviously steering this is up to you, but here's why I think these changes are safe, and get closer to deprecating `Environment.launch` altogether someday:

Four places call `Environment.launch`:
  - `graph.write`
  - `lily.showPNG`
  - `lily.showSVG`
  - `scale.show`
  - technically also `vexflow.toMusic21j.TextExternal.testCuthbertLocal`, but I've never heard of that guy

***

`graph.show` passes a hard-coded `'png'` format.
It usually receives a temp `fp` from `stream.plot()`, but a power user might pass an unexpanded path to `graph.show()`, so yes, we could add a safety check, but I would do it in `graph.write` to be analogous to `SubConverter.writeDataStream`.

***
`lily.showPNG` passes a hard-coded `'png'` format.
The `fp` passed to `launch` is the fp returned by `lily.createPNG`, which is a `pathlib.Path`, so I don't think it needs to be cleaned.

***
`lily.showSVG()`: sim.
***
`scale.show` already runs `common.findFormat()` and will only `launch` if regularized fmt == 'scala'.
The `fp` it passes to `launch` is returned from `self.write('scala')`, so since it's already been used, it's too late to clean it.

jacobtylerwalls: I guess I could just finish the job and fully deprecate this!

mscuthbert:
Merging -- much easier for me to test on windows after merged.  :-)

jacobtylerwalls:
Re: the mailing list user having trouble on windows, @mscuthbert my guess is that this patch needed to preserve the "extra" set of quotes around the app name, which I was hoping python would handle for me. Do you have a way to test on Windows?

mscuthbert:
yes, I can try it.

PR #721, opened 2020-12-18
MuseScore package is once again just MuseScore 3

jacobtylerwalls:
Apparently this summer when I bumped the default MuseScore path to MuseScore 3.5.app, I caught the brief period when they weren't using just "MuseScore 3.app":

August 2020 (3.5.0)
(only 3.5.1 is accessible today?)
Screen Shot 2020-12-17 at 10 57 39 PM

October 2020 (3.5.2)
Screen Shot 2020-12-17 at 10 58 32 PM

- [ ] would be great to have someone test on Windows or at least unzip the MuseScore package and report back with what the .exe is called

### DIFF
--- a/music21/environment.py
+++ b/music21/environment.py
@@ -395,10 +395,10 @@ class _EnvironmentCore:
             for name, value in [
                 ('lilypondPath', 'lilypond'),
                 ('musicxmlPath',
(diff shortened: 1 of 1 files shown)


mscuthbert:
Thank you -- I think the next change we'll leave these all blank and have a system run a configuration search and set the first time show is called if not set directly.

PR #722, opened 2020-12-20
Test for rest creation during voice creation

jacobtylerwalls:
Making rests while making voices (which happens in MIDI parsing) depends on identifying gaps. Some site alchemy (?) was clearing the offsets, so `.findGaps()` was returning `None` (that is, unless your debugger is watching expressions, in which case observing the behavior changed the behavior).

```
handel = corpus.parse('handel')
out = handel.write('midi')
inn = converter.parse(out)
inn.parts[0].show('t', addEndTimes=True)

{0.0 - 240.0} <music21.stream.Voice 0x10dd1dac0>
    {0.0 - 1.5} <music21.note.Rest rest>
    {1.5 - 2.0} <music21.note.Note E>
    {2.0 - 2.5} <music21.note.Note F#>
    {2.5 - 3.0} <music21.note.Note E> 
    {3.5 - 3.8333} <music21.note.Note E>   <--- expected rest to precede
    {4.0 - 4.5} <music21.note.Note B>         <--- expected rest to precede
    {4.5 - 5.0} <music21.note.Note A>
```

There is probably a less expensive way to gather the offsets, but here I just casted the stream to a list to gather all the offsets. I tried clearing the cache for `GapStream` and that didn't seem to do it.

Is there possibly a different root cause: either an `.insert` or `.remove` call doing something it shouldn't with sites?

The attached test file consists of the last three measures of the Handel corpus example.


jacobtylerwalls:
The uncovered line in Coveralls looks like Good News -- it's a continue statement for an error condition in makeBeams, and we do frequently get reports of people processing midi and remarking on "found a messed up beam pair" etc

mscuthbert:
The solution works, but I think that there's a bigger problem in makeRests() if `list(v)` is able to fix it.  I'll try to get a look at it...

mscuthbert:
Indeed, this turned into a cleanup of all sorts of places where "coreInsert()" (or something like that) was being called, but "coreElementsChanged()" never was.  I can't figure out how to push onto this PR, so here's a patch file instead.  :-)
[diff_behavior.patch.txt](https://github.com/cuthbertLab/music21/files/5749180/diff_behavior.patch.txt)

jacobtylerwalls:
Sweet, when that is merged I can just strip this down to the regression test.

mscuthbert:
Thanks so much for finding this bug -- your work ended up leading to the fix of at least 6 other bugs in the system!

PR #724, opened 2020-12-20
Check CHORD_ALIASES when importing xml chord symbols

jacobtylerwalls:
The demo file in the corpus for chord symbols was exporting malformed xml, because on import, musicxml's `<kind>major-minor</kind>` was never mapped back to music21's `minor-major-seventh`, so it never matched the entry in `harmony.CHORD_ALIASES` and `minor-major` was written back out, which is not a valid musicxml chord kind.


mscuthbert:
Thank you Jacob!

PR #727, opened 2020-12-22
degree-alter 0 instead of empty

jacobtylerwalls:
Spec says `<degree-alter>` takes a decimal, so exporting `<degree-alter />` for "omit5" was causing validation errors.


mscuthbert:
Thanks!

PR #730, opened 2020-12-24
Start of a cleanup for ABC parsing

mscuthbert:
I wanted to add one tiny feature to ABC, but the tokenize method is way too complex as it is.  A start to cleanup methods by making current parsing positions be attributes on the ABCHandler class


mscuthbert:
Already merged

PR #589, opened 2020-08-22
Constructed m21 instrument from MIDI instrument name or sequence track name

jacobtylerwalls:
Fixed #588 

**Before**: `MetaEvent.INSTRUMENT_NAME` and `MetaEvent.SEQUENCE_TRACK_NAME` were ignored when parsing midi files into streams.

**Now**:  Instrument instances are constructed.

PR #732, opened 2020-12-26
Deduplicate instruments

### LINKED ISSUE #684: Problems with parsing MIDI files with multiple tracks

jacobtylerwalls:
Fixes #684
(In the sense that we identified no further TODOs, but for my comment about improving quantization someday, which would be a massive project.)

**Before:** MIDI parsing, after #589, constructs Instrument instances from either `SEQUENCE_TRACK_NAME` messages or `PROGRAM_CHANGE` messages.

This can lead to duplicates if both messages are present at the same offset, which can lead to downstream problems with `partitionByInstrument()`:
https://stackoverflow.com/questions/64413988/reading-midi-file-using-music21-with-partitionbyinstrument-to-get-notes-and-ch

**Now:** instrument instances at the same offset in a substream are consolidated IF the `partName` and `instrumentName` attributes agree, i.e. are identical or `None`.

Consolidation means:
- setting the attributes to agree
- removing additional instances of the same class if one class is found
- removing instances of the generic `Instrument` class if multiple classes are found

There is a change to when `stream/__init__.py` imports the instrument module to avoid a circular import (because I import `Stream` for typing).

### DIFF
--- a/music21/instrument.py
+++ b/music21/instrument.py
@@ -1788,7 +1788,7 @@ def deduplicate(s: Stream, inPlace: bool = False) -> Stream:
     if not returnObj.hasPartLikeStreams():
         substreams = [returnObj]
     else:
-        substreams = list(returnObj.getElementsByClass('Stream'))
+        substreams = returnObj.getElementsByClass('Stream')
 
     for sub in substreams:
         oTree = OffsetTree(sub.recurse().getElementsByClass('Instrument'))
--- a/music21/instrument.py
+++ b/music21/instrument.py
@@ -30,6 +30,8 @@ from music21 import base
 from music21 import common
 from music21 import interval
 from music21 import pitch
+from music21.stream import Stream  # for typing
+from music21.tree.trees import OffsetTree
 
 from music21.exceptions21 import InstrumentException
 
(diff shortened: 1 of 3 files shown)


mscuthbert:
Looks great!  Thanks!

PR #733, opened 2020-12-27
Fix musicxml validation error on lack of ending number

jacobtylerwalls:
Missed case from #598 -- musicxml doesn't allow '0' for an ending number; music21 does.

Noticed in corpus example (mozart/k458, 1)


mscuthbert:
Thanks!

PR #736, opened 2020-12-28
Cleanup MakeNotation

mscuthbert:
This was an attempt to fix makeMeasures trailing zero length, from April, but I can no longer remember exactly why I was working on it, so just merging the cleanup code.


jacobtylerwalls:
Sounds like you were jumping on #534, which we ended up fixing anyway.

mscuthbert:
Sounds like it!

PR #738, opened 2020-12-29
If options is blank, not the same as an extra space subprocess 

mscuthbert:
From recently merged change.

Just FYI @jacobtylerwalls to look out for next time we do a subprocess


jacobtylerwalls:
ah--thank you!

Issue #451, opened 2019-10-11
Better document behavior of chordify on grace notes.


giamic:
If one chordifies a score containing grace notes, these get a duration equal to the one of the notes they are referring to. This seems like a wrong behavior to me: grace notes should be kept as grace notes or, at most, suppressed.

A minimal working example is just:

`converter.parse('grace.mxl').chordify().write('musicxml', fp='grace_ch.mxl')`

where 'grace.mxl' is a file containing grace notes, for example like the one attached.
[grace.zip](https://github.com/cuthbertLab/music21/files/3717888/grace.zip)

mscuthbert:
This is currently the expected behavior, and I believe documented as such somewhere -- if not will document it.  

The reason for doing so is -- if you don't want this behavior, it is easy to remove all gracenotes from a stream:

```
notesToRemove = []
for n in s.recurse().notes:
     if n.duration.isGrace:
           notesToRemove.append(n)

for n in notesToRemove:
    n.activeSite.remove(n)
```

whereas if they were removed by default then it would be very hard to get the behavior that we currently have.

Will turn this into an issue of documenting better the behavior and the workaround.

MarkGotham:
Thank you both - this is helpful. Grace notes have had an undue influence on something I've been working on recently, so it's good to know what it (and isn't) planned for their future!

giamic:
Michael, would that be possible to add an option to choose the behavior? Something like `converter.parse('grace.mxl').chordify(keep_grace=False)`

mscuthbert:
Very unlikely -- it would add an additional thing to test and keep up to date with every combination of parameters.

wilsonify:
I submitted #742 to update 
documentation/source/usersGuide/usersGuide_27_graceNotes.ipynb

which I believe will effect
http://web.mit.edu/music21/doc/usersGuide/usersGuide_27_graceNotes.html

![image](https://user-images.githubusercontent.com/26659886/103380736-ce90d180-4ab7-11eb-9acc-488882ad9679.png)

PR #742, opened 2020-12-30
Grace Notes Documentation Update

wilsonify:
Update Grace Note Documents per #451

image


jacobtylerwalls:
howdy @wilsonify -- I would suggest moving this out of the unwritten appoggiature section, since grace notes are a superset of appoggiature. Thanks for picking this up!

mscuthbert:
Thanks -- I'll edit this a bit, because the docs aren't really about appogiature, they work on all grace notes, but thanks for the push!

mscuthbert:
Thanks Thomas for the prod.  I'm closing this in favor of #748  which adds a deeper demo and leaves the Appogiature section still to be written.

Issue #740, opened 2020-12-30
Performance bottleneck chord.Chord.isTriad()


napulen:
**music21 version**

`6.3.0`   

**Problem summary**

In my recent experiments optimizing an expensive algorithm, another performance bottleneck I noticed was this function `chord.Chord.isTriad()`. It is slower than `isMajorTriad()`, `isMinorTriad()`, `isDiminishedTriad()`, `isAugmentedTriad()`. It is also slower than `isSeventh()` although I haven't verified whether that was because I was calling `isSeventh()` afterward and there were some cached attributes already.

I spent some time looking at the implementation. This is the current implementation:
```python
        if self.pitchClassCardinality != 3:
            return False

        try:
            third = self.third
            fifth = self.fifth
        except ChordException:
            return False
        if third is None or fifth is None:
            return False
        for thisPitch in self.pitches:
            try:
                thisInterval = interval.notesToInterval(self.root(), thisPitch)
            except ChordException:
                return False
            if thisInterval.diatonic.generic.mod7 not in (1, 3, 5):
                return False
            if self.hasAnyRepeatedDiatonicNote():
                return False
        return True
```

I think this implementation can be improved. In fact, after looking at the `third` and `fifth` attributes, I wonder if this would be an equivalent (and much faster) check:

```python
        # if self.pitchClassCardinality != 3:
        #     return False
        # try:
        #     third = self.third
        #     fifth = self.fifth
        # except ChordException:
        #     return False
        # if third is None or fifth is None:
        #     return False
        # for thisPitch in self.pitches:
        #     try:
        #         thisInterval = interval.notesToInterval(self.root(), thisPitch)
        #     except ChordException:
        #         return False
        #     if thisInterval.diatonic.generic.mod7 not in (1, 3, 5):
        #         return False
        #     if self.hasAnyRepeatedDiatonicNote():
        #         return False
        if self.pitchClassCardinality == 3 and self.third and self.fifth:
            return True
        return False
```

[...]

PR #744, opened 2020-12-31
Remove bottlenecks on Chord.isTriad and Chord.isSeventh

mscuthbert:
Also speed up containsTriad and containsSeventh

Make helpers for isMajorTriad, isMinorTriad, isDim..., isAug...
Takes away a tiny bit of the speedup, for greater code clarity.

Might slightly reduce code coverage because of reduced length of covered code.


mscuthbert:
Fixes #740

PR #745, opened 2020-12-31
Adds roman.RomanNumeral.isNeapolitan() test ...

MarkGotham:
... to match chord.Chord.isAugmentedSixth()


mscuthbert:
I like the idea, but four layers of nesting breaks the code style.  A series of "if not..." "return False" works better for the paradigm of returning earlier.  can this be done without the nests?  

Can it have an optional flag to require first inversion?  (Generally better to sketch out in issues before giving a PR)

Neapolitan is good because there are two different ways of writing it (bII6 and N6) so can otherwise be hard to identify.  As opposed to say, "isSubdominant" which wouldn't be needed.  Can the doctests show roman.RomanNumeral('N6') passes?

PR #746, opened 2020-12-31
Bug in deduplicate() returned early

jacobtylerwalls:
Fixing an oversight in a recent feature.

### DIFF
--- a/music21/instrument.py
+++ b/music21/instrument.py
@@ -1770,15 +1770,20 @@ def deduplicate(s: Stream, inPlace: bool = False) -> Stream:
     >>> i3.partName = 'Piccolo'
     >>> i4 = instrument.Piccolo()
     >>> s2 = stream.Score()
-    >>> p = stream.Part()
-    >>> p.insert(0, i3)
-    >>> p.insert(0, i4)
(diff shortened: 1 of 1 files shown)


music21/instrument.py
```diff
@@ -1770,15 +1770,20 @@ def deduplicate(s: Stream, inPlace: bool = False) -> Stream:
...
-    >>> p = stream.Part()
-    >>> p.insert(0, i3)
-    >>> p.insert(0, i4)
-    >>> s2.append(p)
-    >>> s2.parts[0].getInstruments().elements
+    >>> p1 = stream.Part()
+    >>> p1.append([i3, i4])
+    >>> p2 = stream.Part()
+    >>> p2.append([instrument.Flute(), instrument.Flute()])
+    >>> s2.append([p1, p2])
```
mscuthbert, line 1777: This should be `s2.insert(0, p1); s2.insert(p2)` I believe.

mscuthbert (changes requested):
Looks great except for one doc error (that does not affect the output, but demonstrates an anti-paradigm in m21) -- I'll push a quick fix.

PR #751, opened 2021-01-01
roman.RomanNumeral.isMixture()

MarkGotham:
Note: I am the author of the code on which this is based, previously hosted at [MarkGotham/When-in-Rome](https://github.com/MarkGotham/When-in-Rome/blob/master/Code/anthology.py) and I agree to have this code re-released under music21's licenses, and the BSD license in particular.

### DIFF
--- a/music21/roman.py
+++ b/music21/roman.py
@@ -64,7 +64,10 @@ def _getKeyFromCache(keyStr: str) -> key.Key:
     create a new key and put it in the cache and return it.
     '''
     if keyStr in _keyCache:
-        # this will at least prevent small errors at a
+        # adding copy.copy will at least prevent small errors at a cost of only 3 nano-seconds
+        # of added time.  A deepcopy, unfortunately, take 2.8ms, which is longer than not
+        # caching at all.  And not caching at all really slows down things like RomanText.
+        # This at least will prevent what happens if `.key.mode` is changed
         keyObj = copy.copy(_keyCache[keyStr])
     else:
         keyObj = key.Key(keyStr)
@@ -3094,9 +3097,7 @@ class RomanNumeral(harmony.Harmony):
         >>> rn.isMixture()
         True
 
-        >>> k = key.Key('C')
-        >>> k.mode = 'Fake mode'
-        >>> rn = roman.RomanNumeral('iv', k)
+        >>> rn.key.mode = 'hypomixolydian'
         >>> rn.isMixture()
         False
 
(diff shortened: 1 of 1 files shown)


mscuthbert:
I'm trying to understand why coverage decreased in this.  Can you look at the report and see?

Be sure to build the documentation and look at it -- the docstring for this in RST/Sphinx/HTML will be a jumble.  

Reason why I think that "v" in major has to be mixture is because of usage beyond common-practice classical.  In non-common-practice classical V in minor is mixture(!) so that's the controversial side!

mscuthbert:
I'm trying to understand why coverage decreased in this.  Can you look at the report and see?

Be sure to build the documentation and look at it -- the docstring for this in RST/Sphinx/HTML will be a jumble.  

Reason why I think that "v" in major has to be mixture is because of usage beyond common-practice classical.  In non-common-practice classical V in minor is mixture(!) so that's the controversial side!

mscuthbert:
Check out https://coveralls.io/builds/36034249/source?filename=music21%2Froman.py#L3047 to see what's still not covered.

MarkGotham:
Don't be modest now ... I'm sure we're courting more controversy than that!

The coverage issue comes from various False cases like calling on a chord that isn't a triad or seventh, or a key that's not major or minor.

Other than controversy and coverage, that's all fine. Thanks @mscuthbert!

mscuthbert:
Note about Keys vs Scales:

```
In [2]: roman.RomanNumeral('I', scale.MajorScale('D')).isMixture()
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-2-76f44ea5b7a1> in <module>
----> 1 roman.RomanNumeral('I', scale.MajorScale('D')).isMixture()

~/git/music21base/music21/roman.py in isMixture(self, secondaryNumerals)
   3135             return False
   3136 
-> 3137         mode = self.key.mode
   3138         if mode not in ('major', 'minor'):
   3139             return False

AttributeError: 'MajorScale' object has no attribute 'mode'

```

mscuthbert:
Is this expected?

```
roman.RomanNumeral('IV', 'c').isMixture() is True
```

MarkGotham:
> Is this expected?
> 
> ```
> roman.RomanNumeral('IV', 'c').isMixture() is True
> ```
I would think so, in the same way that ii minor (not diminished) is, i.e.: because of borrowing the raised 6th degree. Speaking of which, with that rationale, then vio (A,C,Eb) should perhaps also be in too (albeit as a rather rare case).

I'm more concerned by #III+, which I think of as a pretty classic minor-key dominant.

Let's see if anyone else wants to weigh in on the discussion by taking it [here](https://twitter.com/MarkGothamMusic/status/1345746378555871232).

mscuthbert:
Thanks Mark!  My latest pushes also fix (or gets as close to fixing as efficiency can allow) a long-standing problem with cached keys in RomanNumeral objects.

PR #753, opened 2021-01-02
See if running PyLint multicore gives speedups

mscuthbert:


mscuthbert:
Indeed -- halves the time for Pylint and thus makes the coverage test the slowest test being run.

PR #757, opened 2021-01-04
Capture more instruments from MIDI programs

### LINKED ISSUE #397: Additional General MIDI Instrument Support
Relating to the discussion in #235, I encountered issues reading midi files which utilize guitar instruments. Because there were no instrument definitions for some of the general midi standards, the translated stream was loaded with piano tracks (which is the default instrument). @mscuthbert, you mentioned that not all instruments should necessarily have a definition. I'd to understand that a little better. When translating midi to stream, the instrument is loaded by ID. In the case that the ID is not found, it falls back on the piano. It seems in order to properly maintain the correct instrument when translating to/from midi/streams, it would need a definition. Since I'm still new to and learning music21, there may be some gap in my understanding, so please correct me if I am wrong. If an instrument definition is not the right approach, what else can be done in order to support proper handling of these files? I'm happy to implement whatever solution you'd suggest and submit the PR. Thanks.

jacobtylerwalls:
Fixes #397 , which described reverting to generic `Instrument` on encountering midi programs that didn't match the default midi programs on the existing classes, and provide a faster lookup in `instrumentFromMidiProgram()`.

The remaining TODOs would be to also set self.instrumentSound from a similar lookup.

`class Sampler` could be replaced with references to `KeyboardInstrument`, but that seems to decenter electronics and contemporary musical practices to me. An orchestra hit or a siren can't be played on a generic KeyboardInstrument--only a sampler. I'm having flashbacks to the few minutes before my 12pm section of Keyboard Skills where my students would unplug their headphones and jam on sirens and whistles and screams before class. It was pretty tight.

### DIFF
--- a/music21/instrument.py
+++ b/music21/instrument.py
@@ -2002,7 +2002,10 @@ def instrumentFromMidiProgram(number: int) -> Instrument:
     <music21.instrument.Accordion 'Accordion'>
     >>> instrument.instrumentFromMidiProgram(500)
     Traceback (most recent call last):
-    music21.exceptions21.InstrumentException: No instrument found with given midi program
+    music21.exceptions21.InstrumentException: No instrument found for MIDI program 500
+    >>> instrument.instrumentFromMidiProgram('43')
+    Traceback (most recent call last):
+    TypeError: Expected int, got <class 'str'>
     '''
 
     try:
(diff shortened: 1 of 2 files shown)


mscuthbert:
Thanks!

PR #763, opened 2021-01-09
WIP: Apply tuplet to multiple components to express 5/6 or 7/3 QL

### LINKED ISSUE #572: Enhancement: Find a complex type for some non-power of 2 tuplets
Found while running `stripTies` on Schoenberg op19/6, measure 8, but minimal example here:

```
eighth = note.Note(quarterLength=0.5)
tripletEighth = note.Note(quarterLength=(1/3))
eighth.tie = tie.Tie(type="start")
tripletEighth.tie = tie.Tie(type="stop")

s = stream.Stream()
s.append((eighth, tripletEighth))
s.makeMeasures(inPlace=True)
s.stripTies(inPlace=True)

singleNote = s.flat.notes[0]
(singleNote.quarterLength, singleNote.duration.type, singleNote.duration.tuplets[0])
```
```
(Fraction(5, 6), 'quarter', <music21.duration.Tuplet 6/5/32nd>)
```
Math checks out, but neither a musicxml export nor a roundtrip back using `makeTies` will do anything with the 6-in-the-space-of-5 32nds tuplet, which makes for difficult to read output (or lost information if the musicxml importer decides to drop the tuplet bracket on the note altogether).

I'm suggesting `quarterConversion` could try harder to come up with a complex type. In this instance it would involve storing five components of 16th notes, then a 3/2/16th tuplet applied to the whole thing. `m21ToXml` splits on components when exporting, so I would expect a clean export if we had a complex type. `makeTies` could also do something similar to ensure a roundtrip after `stripTies` came back the same.

I looked at how the duration algorithm works now, and for an input of 5/6 QL it finds 1/2 QL first, then looks at the remainder (1/3 QL) to see if it could be added as a second component, but by design does not allow tuplets. But we could use the facts that a valid component was found first and a suitably small-order triplet was found next as a cue to go get the LCM and make sure's it under an acceptable ceiling.

jacobtylerwalls:
Fixed #572 

**Before:** Durations such as 5/6 or 7/3 QL were expressed as non-power of 2 tuplets, which caused display problems.

**Now:** Upon extracting the largest type (such as 0.5 QL from 5/6 QL or 2.0 QL from 7/3 QL), if the remainder can be expressed as a single tuplet, express the largest type and the remainder in smaller components over which a tuplet can be applied.

<s>Added support to the musicxml exporter for creating the necessary visual `<tuplet>` elements from each split component.</s> Update: added a call to remake tuplet brackets inside `splitAtDurations()`

<s>Update: added the keyword `splitAtDurations` to `makeNotation()` and moved the logic in the xml exporter.</s> Update: calls new `splitAtDurations` method on Stream.

Update: adds attribute/slot `expressionIsInferred`

Update: extends tuplet brackets over tuplets that may not share the same type (3/2/eighth and 3/2/16th) but at least share the same ratio. This is to avoid unnecessarily proliferating start/stop brackets. Respected in MuseScore at least.


jacobtylerwalls:
musicxml needs a bit more work -- something to do with making the tuplet brackets close in the right place.
***

`stripTies` is one way we would get these values, as in the Schoenberg, but also MIDI parsing, esp. because the default quantization units are 0.25 and 0.33, so it is easy to end up with streams like this:

```
>>> s = stream.Stream()
>>> s.insert(0, note.Note(quarterLength=(1/3)))
>>> s.insert(0.75, note.Note(quarterLength=(1/4)))
>>> s.makeRests(fillGaps=True, inPlace=True)
>>> s.show()
```

**Before:** (not ugly, but wrong)

Screen Shot 2021-01-09 at 2 20 05 PM

**After:** (ugly, mostly correct, but still need to get the first tuplet to end in the right place):

Screen Shot 2021-01-09 at 2 21 01 PM



And here's Schoenberg, ties stripped (and then created again on xml export):
**Before:**
Screen Shot 2021-01-09 at 2 29 44 PM

**After:**
Screen Shot 2021-01-09 at 2 29 50 PM

**Source:**
Screen Shot 2021-01-09 at 2 31 37 PM


***
The current approach runs `makeTupletBrackets` on the components, but the components possibly needs to be extracted and inserted far earlier in the process so that `makeTupletBrackets` can be re-run on the whole stream to get correct stop values.

mscuthbert:
Ah!  All this comes as a necessary consequence of a refactor around music21 v.2 where components do not have their own tuplets.  In the past it would have been possible to express this:

```
    >>> duration.quarterConversion(5/6)
     QuarterLengthConversion(components=(DurationTuple(type='16th', dots=0, quarterLength=0.25),
                                         DurationTuple(type='16th', dots=0, quarterLength=0.25),
                                         DurationTuple(type='16th', dots=0, quarterLength=0.25),
                                         DurationTuple(type='16th', dots=0, quarterLength=0.25),
                                         DurationTuple(type='16th', dots=0, quarterLength=0.25)),
                             tuplet=<music21.duration.Tuplet 3/2/eighth>)
```

as:

```
    >>> duration.quarterConversion(5/6)
     QuarterLengthConversion(components=(
                                         DurationTuple(type='eighth', dots=0, quarterLength=0.5),
                                         DurationTuple(type='16th', dots=0, quarterLength=0.25, tuplet=<music21.duration.Tuplet 3/2/eighth>),
                                         DurationTuple(type='16th', dots=0, quarterLength=0.25, tuplet=<music21.duration.Tuplet 3/2/eighth>)),
     )
```

But the (huge speed increase and huge simplification) change to DurationTuples (from DurationComponent objects) made this impossible.

What would be very helpful to put in this PR is a new attribute on the Duration object (and its slots)

```
def __init__(self, ...):
      ...
      self.expressionIsInferred = False
```

this has an analogy to Pitch's `.spellingIsInferred` -- meaning that exporters and transformers should feel free to reorganize the components or tuplets at any point rather than needing to respect them.  And it should be set to True any time a Duration is set from QLs rather than type, dots, etc.

Then `.splitAtComponents()` might (unless a separate kwarg `strictSplit=False` is True) be able to eventually do something like:

```
    >>> n = note.Note(quarterLength=5/6)
    >>> n.duration.expressionIsInferred
    True
    >>> n.duration.components
    [ five 16th ]

    >>> [sub.duration for sub in n.splitAtComponents()] 
    [<duration ... normal eighth note with expressionIsInferred still True>,
      <duration ... 16th note triplet  with expressionIsInferred still True>,
      <duration ... 16th note triplet  with expressionIsInferred still True>]

    But 

    >>> [sub.duration for sub in n.splitAtComponents(strict=True)] 
    [<duration ... 16th note triplet  with expressionIsInferred still True>,
     <duration ... 16th note triplet  with expressionIsInferred still True>,
     <duration ... 16th note triplet  with expressionIsInferred still True>,
     <duration ... 16th note triplet  with expressionIsInferred still True>,
     <duration ... 16th note triplet  with expressionIsInferred still True>]

```

and be able to reconnect the first three 16ths under the tuplet into one eighth note.

later,  we would be able to do fancy things like reconnect the final two 16th note triplets together if there is a 16th note triplet following it in the stream, etc.

It would move closer to a very-long-term-goal of getting this to express properly

```
    >>> s = stream.Stream([note.Note(ql=1/3), note.Note(ql=1/2), note.Note(ql=1/6) ]
```

as containing a triplet of eighth, dotted eighth, sixteenth, rather than a triplet eighth - break tuplet - normal eighth - start tuplet again - triplet 16th.

None of that needs to go in this PR, but if `expressionIsInferred` can go in now, it'd make further work easier.  My sense is that ideally any manual setting of type, dots, tuplet, (or ideally, components) would set expressionIsInferred to False. But we don't have that in for Pitch's `spellingIsInferred` yet either because it would make it so hard to do.

jacobtylerwalls:
Done, although I may have described this incorrectly, since there is functionality to mutate components in `_updateComponents`, and if in that function I short circuit if `.expressionIsInferred` is False, I cause many failures.

jacobtylerwalls:
Experimented with a look-behind approach, but this got out of hand:

```diff --git a/music21/musicxml/m21ToXml.py b/music21/musicxml/m21ToXml.py
index 42ee5daad..ed77ed0d5 100644
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -2745,6 +2745,7 @@ class MeasureExporter(XMLExporterBase):
         # group all objects by offsets and then do a different order than normal sort.
         # that way chord symbols and other 0-width objects appear before notes as much as
         # possible.
+        lastNote = None
         for objGroup in OffsetIterator(m):
             groupOffset = m.elementOffset(objGroup[0])
             amountToMoveForward = int(round(divisions * (groupOffset
@@ -2767,7 +2768,7 @@ class MeasureExporter(XMLExporterBase):
                     self.parseOneElement(obj)
 
             for n in notesForLater:
-                self.parseOneElement(n)
+                lastNote = self.parseOneElement(n, lastNote=lastNote)
 
         if backupAfterwards:
             # return to the beginning of the measure.
@@ -2780,7 +2781,7 @@ class MeasureExporter(XMLExporterBase):
 
         self.currentVoiceId = None
 
-    def parseOneElement(self, obj):
+    def parseOneElement(self, obj, lastNote: Optional[Element] = None) -> Optional[Element]:
         '''
         parse one element completely and add it to xmlRoot, updating
         offsetInMeasure, etc.
@@ -2811,15 +2812,27 @@ class MeasureExporter(XMLExporterBase):
             for o in objList:
                 tempStream.append(o)
             stream.makeNotation.makeTupletBrackets(tempStream, inPlace=True)
+            # Cannot rely on earlier tuplet bracket definitions
+            if lastNote is not None:
+                mxTuplet = lastNote.findall('notations/tuplet')
+                if len(mxTuplet) == 1:
+                    if mxTuplet.get('type') == 'start':
+                        # need another mxTuplet for stop
+                        # which would be written by tupletToXmlTuplet()
+                        # but then I would also need previousNote.duration.tuplets[0]?
+                        # and then I'm passing two look-behind objects
+                        # and I also need to handle look-ahead
+                        pass
         else:
             objList = [obj]
 
+        returnElement = None
         parsedObject = False
         for className, methName in self.classesToMethods.items():
             if className in classes:
                 meth = getattr(self, methName)
                 for o in objList:
-                    meth(o)
+                    returnElement = meth(o)
                 parsedObject = True
                 break
 
@@ -2843,6 +2856,7 @@ class MeasureExporter(XMLExporterBase):
 
         for sp in postList:  # directions that follow the element
             root.append(sp)
+        return returnElement
 
     def prePostObjectSpanners(self, target):
         '''
@@ -3143,7 +3157,7 @@ class MeasureExporter(XMLExporterBase):
 
         return notations
 
-    def noteToXml(self, n: note.GeneralNote, noteIndexInChord=0, chordParent=None):
+    def noteToXml(self, n: note.GeneralNote, noteIndexInChord=0, chordParent=None) -> Element:
```

Don't know how to support closing the brackets correctly in musicxml unless we want to pre-walk the entire stream, run `splitAtDurations` in the same scenario we do now, but replace elements and re-make tuplet brackets BEFORE exporting. On that idea, it would be somewhat convenient for `stream.replace()` to take a list of elements to substitute.

jacobtylerwalls:
Since "pre-walking" is done via `makeNotation` anyway, and since tinkering with tuplets is probably better done there, here's a pass at using `splitAtDurations=True` in `makeNotation`.

Why the test is failing, I'm not sure. There are four `<tuplet>` tags in the output, so I'm either severely misunderstanding `findall` or there's a bug with it...!

jacobtylerwalls:
Well... closer....

Screen Shot 2021-01-10 at 10 29 31 PM

mscuthbert:
Btw -- I am still thinking about this -- sorry it's been so slow to get on it.  It's a big thought about whether something that was wrong but "looked good" is preferable to something that's right but looks really poor on output.

jacobtylerwalls:
Sounds good, no rush. Just tested a file against it and there's a bug -- corrupt XML output, so I can set this back to draft for now.

There have a been couple instances where I've thought .show() needs a makeNotation=False argument so that you can freeze what you have (for instance, if you've run makeNotation with splitAtDurations=False). That could be a way to accommodate both wrong-looking-good and right-looking-ugly scenarios.

mscuthbert:
show definitely needs a makeNotation=False.  But before we do that, I'm going to create a new issue to help get keywords from getting out of hand.

music21/stream/__init__.py
```diff
@@ -12260,6 +12301,9 @@ def makeNotation(self,
         >>> m.notes[1].pitch.accidental
         <music21.pitch.Accidental natural>
         '''
+
+        # TODO: splitAtDurations
+        # TODO: can this be unified with Stream.makeNotation?
```
jacobtylerwalls, line 12306: Inclined to see if this can be done with minimal effort?

music21/stream/__init__.py
```diff
@@ -6359,6 +6363,32 @@ def makeNotation(self,
...
+                    if noteObj.duration.type == 'complex':
+                        insertPoint = noteObj.offset
+                        objList = noteObj.splitAtDurations()
+                        container.replace(noteObj, objList[0])
+                        insertPoint += objList[0].quarterLength
+                        for subsequent in objList[1:]:
+                            container.insert(insertPoint, subsequent)
+                            insertPoint += subsequent.quarterLength
+
+                        notesWithSplitDurations += objList
```
jacobtylerwalls, line 6390: This should probably be factored out into `def splitAtDurations()`. My next thought is that that exists in `base.py` -- but lacking a way of dealing with stream containers. The only documented uses are on `GeneralNote`, but it accepts anything:

```
>>> s = stream.Stream()
>>> n = note.Note(quarterLength=5)
>>> s.insert(0, n)
>>> s.duration
<music21.duration.Duration 5.0>
>>> s.splitAtDurations()
  File "/Users/jwalls/music21/music21/base.py", line 3160, in splitAtDurations
    splitList = self.splitByQuarterLengths(quarterLengthList)
  File "/Users/jwalls/music21/music21/base.py", line 3041, in splitByQuarterLengths
    spannerList.extend(st.spannerList)
AttributeError: 'tuple' object has no attribute 'spannerList'
```

So maybe what is sketched out here (cribbed from the existing xml exporter) should move to `base.splitAtDurations`, wrapping what is already there?

jacobtylerwalls: Well, no, it should probably just be on stream.Stream so that it overrides base.splitAtDurations for Stream subclasses only.

jacobtylerwalls:
Here's what would suffice for a "don't make notation" argument (which doubles as "don't make a deepcopy")

```diff --git a/music21/converter/subConverters.py b/music21/converter/subConverters.py
index ecc445d38..9c6925e31 100644
--- a/music21/converter/subConverters.py
+++ b/music21/converter/subConverters.py
@@ -974,8 +974,15 @@ class ConverterMusicXML(SubConverter):
             defaults.title = ''
             defaults.author = ''
 
+        dataBytes: bytes = b''
         generalExporter = m21ToXml.GeneralObjectExporter(obj)
-        dataBytes: bytes = generalExporter.parse()
+        if ('Score' in obj.classes
+            and 'makeNotation' in keywords
+            and keywords['makeNotation'] is False):
+            # this keyword could have been wellFormed=True instead of makeNotation=False
+            dataBytes = generalExporter.parseWellFormedObject()
+        else:
+            dataBytes = generalExporter.parse()
 
         writeDataStreamFp = fp
         if fp is not None and subformats is not None:
```

Given that, do you want to wait until a larger keywords refactor is tackled before dropping this in?

jacobtylerwalls:
jotting a note for future self: above patch isn't enough, because `fixupNotationFlat` and measured also need to be skipped over

jacobtylerwalls:
Hey @napulen happy to hear of your interest here. This ended up touching a lot of parts of the system, so here's my recap of where I think things stand right now:

- [x] do math to avoid resorting to non-power of 2 tuplets to express 5/6 QL and 7/3 QL etc, which you can get from ties and gaps (#572) -- done and tested in this PR, but not mergeable until we have a way to make it less ugly on export (all those tiny rests in the second screenshot, at the top)
- [x] redraw tuplet brackets when splitting on components -- done and tested in this PR
- [x] join tuplet brackets when the ratio is the same but type is different -- done and tested in this PR, but may not end up needing it depending on what we do below, and it may not even be a great idea, given that MuseScore is conveniently lax about these mixed type brackets but Finale is inconveniently strict. There's some more roubst logic for this stuff in `TupletFixer` which I haven't looked at, and Myke should weigh in.
- [ ] `.show()` needs a way to skip making a copy and running makeNotation -- sketched out above, but not coded and tested yet
- [ ] do some magic to break into tuplets or consolidate from tuplets based on context (#904). needing to consolidate is why we haven't merged this PR yet, and needing to break into tuplets is your case with the Hensel pieces
- [x] new `expressionIsInferred` attribute on Durations to track whether it's okay to do magic -- done and tested in this PR, but should probably be merged separately

So what's left is the magic ✨ 

For a little proof of concept for 1/3 QL - 1/3 QL - 1 QL (rest) - 1/3 QL I would think it would go like:
- Do a scan through the notesAndRests (GeneralNote) of a measure or voice
- Is there an incomplete tuplet? Grab the `.next()` GeneralNote, but only if `.expressionIsInferred`
- Re-express that one if it helps, and set `.components`
- Then do a second pass from the beginning, and as long as `expressionIsInferred`, consolidate groups that are unnecessarily expressed as a tuplet (like the second screenshot at the top).

I don't want to take all the fun, so if you want to pair or split some of this up, send me a note.

jacobtylerwalls:
I would say May is the most likely point for me to dig in to the hard tasks here. I could probably get the expressionIsInferred bits and the don't makeNotation bits merged in smaller PRs first. So yeah, plenty of time to focus on Humdrum!

jacobtylerwalls:
Pieces of this were merged in other PRs, but the main idea -- applying a tuplet to multiple components -- can be more easily handled in a new PR once #904 is addressed; this one no longer applies cleanly.

PR #764, opened 2021-01-09
WIP: Humdrum parser bug when encountering manually-positioned rests

pkirlin:
Humdrum notation can (apparently!) contain manual-positioning information for rests, which unfortunately the current code detects as a note.

Details:

In humdrum, a normal eighth note rest might look like `8r`, but the position of this rest can be changed by adding the symbol for a pitch after the `r`; e.g., `8rGG` would mean an eighth note rest but positioned at the vertical space on the staff where the note `GG` is.  I cannot find any definite information on this in the humdrum documentation, unfortunately, but it appears in at least one file on the KernScores website: https://kern.humdrum.org/cgi-bin/ksdata?l=users/craig/ragtime/joplin&file=entertainer.krn&f=kern  (see measure 4).  

For the moment, this can be fixed by detecting rests first, then notes, in the `hdStringToNote` function, however I have no clue if this will break other things in the code, so I've marked this as a work in progress.  I'm hoping to get some clarification from @craigsapp.

### DIFF
--- a/music21/humdrum/spineParser.py
+++ b/music21/humdrum/spineParser.py
@@ -2193,12 +2193,17 @@ def hdStringToNote(contents):
 
     # http://www.lib.virginia.edu/artsandmedia/dmmc/Music/Humdrum/kern_hlp.html#kern
 
-    # 3.2.1 -- pitch
+    # 3.2.1 Pitches and 3.3 Rests
 
     matchedNote = re.search('([a-gA-G]+)', contents)
-
     thisObject = None
-    if matchedNote:
+
+    # Detect rests first, because rests can contain manual positioning information,
+    # which is also detected by the `matchedNote` variable above.
+    if contents.count('r'):
+        thisObject = note.Rest()
+
+    elif matchedNote:
         kernNoteName = matchedNote.group(1)
         step = kernNoteName[0].lower()
         if step == kernNoteName[0]:  # middle C or higher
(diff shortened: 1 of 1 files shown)


mscuthbert:
It looks like it'll work, and glad to merge if you think it is solid.  It wasn't in the original humdrum representation: https://web.archive.org/web/20040503210359/http://www.music-cog.ohio-state.edu/Humdrum/representations/kern.html#Rests

But as it is part of the current representation (https://www.humdrum.org/rep/kern#Rests) -- it's something not necessarily to support completely but to make sure it doesn't crash on, so this is a great easy solution to it.

craigsapp:
>  it's something not necessarily to support completely

If you do not support it, you will be getting a lot of junky data from Humdrum files, particularly the ones that I have proofread carefully.

pkirlin:
I'm running a few tests today with this change, and I'll see if it seems to work ok.  @craigsapp's comments on a different thread say that if a token has an `r` in it at all, it must be a rest, so it seems ok to me to check for those first, and then check for notes second, as this proposed change does.  I'll follow up in a bit.  Thanks to Craig and @mscuthbert for your comments and help!

mscuthbert:
Thanks @pkirlin !

pkirlin:
> Thanks @pkirlin !

You got it! I think it should work fine. I didn't press it in case people wanted more testing, but I think it's good to go!

PR #766, opened 2021-01-10
Check for elem.get('artic') is not None in articElement

### LINKED ISSUE #762: MEI import fails with 'AttributeError: 'NoneType' object has no attribute 'split''
I have looked at this problem more closely, and it happens in the following situation:
 - The MEI files contains something like ```<artic xml:id="artic-xxxx" />```.
In this case the function '_makeArticList (attrib)' is called, and fails on attrib.split, since attrib in this case is None.

 - In two of three situations where '_makeArticList' is called, it is checked whether the attrib is None before calling, but in articFromElement it is not. 

Beyond that it gets too complicated for me...

Of course the input file should not include non-sensical lines like the above, but sometimes this happens, and it would be nice if the import would not fail, especially since find the cause is not trivial. A better way would be just to ignore it. (IMHO the best way would be to ignore it, and write a log entry that the line was ignored and why.)

### LINKED ISSUE #765: MEI import fails when determining key signature from attributes (pname and mode)
In a MEI file from DME one finds ```<key mode="major" pname="d"/>```. The import tries to derive the key signature from it and fails, when combining tonic and accidental, since the accidental is set to None.

I think that this can be fixed by changing one line in _keySigFromAttrs:
From:
```return key.Key(tonic=elem.get('key.pname') + _accidentalFromAttr(elem.get('key.accid')), mode=mode)```
to:
```return key.Key(tonic=elem.get('key.pname') + _accidentalFromAttr(elem.get('key.accid',default='')),mode=mode)```

mscuthbert:
Thanks heinzr

Fixes #762

### DIFF
--- a/music21/mei/base.py
+++ b/music21/mei/base.py
@@ -1914,7 +1914,11 @@ def articFromElement(elem, slurBundle=None):  # pylint: disable=unused-argument
 
     **Contained Elements not Implemented:** none
     '''
-    return _makeArticList(elem.get('artic'))
+    articElement = elem.get('artic')
+    if articElement is not None:
+        return _makeArticList(articElement)
+    else:
+        return []
 
 
 def accidFromElement(elem, slurBundle=None):  # pylint: disable=unused-argument


jacobtylerwalls:
Looks like #762, not #765?

PR #767, opened 2021-01-10
Fix MEI _keySigFromAttrs None handling for accidental

### LINKED ISSUE #765: MEI import fails when determining key signature from attributes (pname and mode)

jacobtylerwalls:
(I anticipate that this) Fixes #765

### DIFF
--- a/music21/mei/base.py
+++ b/music21/mei/base.py
@@ -1039,8 +1039,13 @@ def _keySigFromAttrs(elem: Element) -> Union[key.Key, key.KeySignature]:
         # @key.accid, @key.mode, @key.pname
         # noinspection PyTypeChecker
         mode = elem.get('key.mode', '')
-        return key.Key(tonic=elem.get('key.pname') + _accidentalFromAttr(elem.get('key.accid')),
-                       mode=mode)
+        step = elem.get('key.pname')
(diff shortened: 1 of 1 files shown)


mscuthbert:
Thanks!

heinzr:
I have a question about this function: according to the function annotation, it is supposed to return key.key and key.keySignature. It does not seem to do this, though. If key.pname is not set, it determines ks, but does not return it.

jacobtylerwalls:
The annotation for the return type is one of the two given types, not both. You're right that the annotation could be more restrictive and just list `key.Key`. The intermediate ks is pretty throwaway; it doesn't even have a mode yet.

PR #769, opened 2021-01-11
Fix the instantiation of GeneralNote with lyrics

### LINKED ISSUE #768: Conversion of non-string lyric when instantiating music21.note.GeneralNote and childs
**music21 version**

`6.3.0`

**Problem summary**

When instantiating a `GeneralNote` object or any of its child classes with a `lyric` keyword argument equal to a non-string value, it gets converted to its string representation even if it's `None`.

**Steps to reproduce**

```python
>>> import music21
>>> note_a = music21.note.GeneralNote(lyric=None)  # Unexpected behavior.
>>> note_b = music21.note.GeneralNote(); note_b.lyric = None  # Desired behavior.
```

When passing `None` as the value for the `lyric` keyword argument, it's being processed through the [`addLyric`](https://github.com/cuthbertLab/music21/blob/4c25aeac5462de6c1d5c9513a9a6ed3ea19eb6ad/music21/note.py#L467-L538) method, which [converts](https://github.com/cuthbertLab/music21/blob/4c25aeac5462de6c1d5c9513a9a6ed3ea19eb6ad/music21/note.py#L522-L523) any non-string lyric to a string, disregarding the special type `None` that indicates the absence of lyrics.

```python
>>> type(note_a.lyric)
<class 'str'>
>>> note_a.lyric
'None'  # Unexpected behavior; converts the type to string.
```

```python
>>> type(note_b.lyric)
<class 'NoneType'>
>>> note_b.lyric
None  # Desired behavior, keeps the original type.
```

**More information**

In my opinion, it would be nice to replace [`addLyric`](https://github.com/cuthbertLab/music21/blob/4c25aeac5462de6c1d5c9513a9a6ed3ea19eb6ad/music21/note.py#L467-L538) by [`_setLyric`](https://github.com/cuthbertLab/music21/blob/4c25aeac5462de6c1d5c9513a9a6ed3ea19eb6ad/music21/note.py#L420-L430) [here](https://github.com/cuthbertLab/music21/blob/4c25aeac5462de6c1d5c9513a9a6ed3ea19eb6ad/music21/note.py#L379
), so the special `None` and `False` values are taken into account:

https://github.com/cuthbertLab/music21/blob/4c25aeac5462de6c1d5c9513a9a6ed3ea19eb6ad/music21/note.py#L379

0x2b3bfa0:
Previously, `GeneralNote(lyric=None)` would set a lyric with the literal `None` as a string. Fixes #768.

### DIFF
--- a/music21/note.py
+++ b/music21/note.py
@@ -375,7 +375,7 @@ class GeneralNote(base.Music21Object):
         self.expressions = []
         self.articulations = []
 
-        if 'lyric' in keywords:
+        if 'lyric' in keywords and keywords['lyric'] is not None:
             self.addLyric(keywords['lyric'])
 
         # note: Chords handle ties differently


jacobtylerwalls:
A more general fix would be to do something like this. This way any caller of `addLyric` gets the fix:

```diff --git a/music21/note.py b/music21/note.py
--- a/music21/note.py
+++ b/music21/note.py
@@ -519,6 +519,8 @@ class GeneralNote(base.Music21Object):
         >>> n1.lyrics[3].syllabic
         'single'
         '''
+        if text is None:
+            return
         if not isinstance(text, str):
             text = str(text)
         if lyricNumber is None:
```

jacobtylerwalls:
`insertLyric` has the same problem.

0x2b3bfa0:
My opinionated recommendation would be to just check if `isinstance(text, str)` elsewhere and raise `ValueError` otherwise, so it doesn't silently convert unexpected types to strings. The only special cases that should take an optional **None** value into account are:

   1. Instantiation of `GeneralNote` and childs with `lyric=None`, meaning that the object has no lyrics at all.
   2. Assignment of `None` to the `lyric` property in order to clear the lyrics of that object.

Any other conversion of different objects to string would need to be explictly performed by the caller:

```python
import music21

class MyLyric:  
    def __str__(self):
        text = "Always look on the bright side of life"
        return text[:2]

lyric = str(MyLyric())  # Note that we're performing the conversion on our own.
note = music21.note.GeneralNote(lyric=lyric)
```

Causing `insertLyric` and `addLyric` to implicitly ignore calls with **None** values could be as bad as causing them to blindly convert values to strings. _«[Explicit is better than implicit](https://www.python.org/dev/peps/pep-0020/)»_, and, ideally, these functions shouldn't need to deal with other types than `str`.

Users could have the choice of what to do when adding/inserting None lyrics:

```python
import music21

note = music21.note.GeneralNote()
lyric = None

# Either skip it:
if lyric is not None:
    note.addLyric(lyric)

# Or use an empty string instead:
note.addLyric(lyric or "")
```

mscuthbert:
On the issue page, I've given a workaround that I'm down with.  We're not getting rid of implicit coercion of lyrics; it's too ingrained in other people's usages, even with 0 (which evaluates to False) for things like:

```
for i, n in enumerate(s.flat.notes):
    n.lyric = i
```

yes, people should do `str(i)`, but it's in so much code already, it would be a backwards compatibility breakage for a benefit I'm not fully seeing.

mscuthbert:
Thanks -- and great first contribution 0x2b! -- will once all tests pass.

mscuthbert:
Coveralls failure not a fault of this PR -- noting it but merging.

0x2b3bfa0:
😊  Thank you very much, @mscuthbert! Glad to add a grain of sand to this excelent project.

> _On the issue page, I've given a workaround that I'm down with. We're not getting rid of implicit coercion of lyrics; it's too ingrained in other people's usages, even with 0 (which evaluates to False) for things like:_
> _yes, people should do str(i), but it's in so much code already, it would be a backwards compatibility breakage for a benefit I'm not fully seeing._

There was no real benefit other than sparing possibly unexpected results to future users, so leaving it as-is, as you suggest, is a wise choice. A breaking change like that would require a deprecation warning and ~65 additional changes inside this project, not to mention others.

PR #770, opened 2021-01-12
Extended duration symbols in ABC chords (marian-schultz)

marian-schulz:
Extension for the recognition of chords in ABCFormat according to http://abcnotation.com/wiki/abc:standard:v2.1#chords_and_unisons.

* For chords an outer length modifier is supported
Example:
[ceg]2 == [c2e2g2]
[c3e3g3]2 == [ceg]6 == [c6e6g6]

* The length of the chord is determined by its first note
Example:
[c2e4g8] == [c2e2g2]
* Empty chords are discarded
* Additional unittests for variations of abc Chords
* Fix: Missing chord symbols  for chords
* Additional unittests for chord symbols on note & chords

### DIFF
--- a/music21/abcFormat/__init__.py
+++ b/music21/abcFormat/__init__.py
@@ -1641,21 +1641,37 @@ class ABCChord(ABCNote):
 
     A subclass of ABCNote.
     '''
-    def __init__(self, src):
+
+    def __init__(self, src: str = ''):
         super().__init__(src)
         # store a list of component objects
         self.subTokens = []
 
     def parse(self, forceKeySignature=None, forceDefaultQuarterLength=None):
+        '''
+            Chord without length modifier: [ceg]
+            Chords with outer length modifier: [ceg]2, [ceg]/2
+            Cords with inner length modifier: [c2e2g2], [c2eg]
+            Cords with inner and outer length modifier: [c2e2g2]/2, [c/2e/2g/2]2
+        '''
+
         self.chordSymbols, nonChordSymStr = self._splitChordSymbols(self.src)
 
-        tokenStr = nonChordSymStr[1:-1]  # remove outer brackets
-        # environLocal.printDebug(['ABCChord:', nonChordSymStr, 'tokenStr', tokenStr])
+        # position of the closing bracket
+        pos = nonChordSymStr.index(']')
+        # Length modifier string behind der chord brackets
+        outerLengthModifierStr = nonChordSymStr[pos + 1:]
+        # String in the chord brackets
+        tokenStr = nonChordSymStr[1:pos]
 
-        self.quarterLength = self.getQuarterLength(
-            nonChordSymStr,
-            forceDefaultQuarterLength=forceDefaultQuarterLength
-        )
+        # environLocal.printDebug(['ABCChord:', nonChordSymStr, 'tokenStr', tokenStr, '
+        # outerLengthModifierStr', outerLengthModifierStr])
+
+        # Get the outer chord length modifier if present
+        outer_lengthModifier = 1.0
+        if outerLengthModifierStr:
+            outer_lengthModifier = self.getQuarterLength(outerLengthModifierStr,
+                                                         forceDefaultQuarterLength=1.0)
 
         if forceKeySignature is not None:
             activeKeySignature = forceKeySignature
@@ -1669,7 +1685,7 @@ class ABCChord(ABCNote):
         # may need to supply key?
         ah.tokenize(tokenStr)
 
-        chordDurationPost = None
+        inner_quarterLength = 0
         # tokens contained here are each ABCNote instances
         for t in ah.tokens:
             # environLocal.printDebug(['ABCChord: subTokens', t])
@@ -1678,14 +1694,22 @@ class ABCChord(ABCNote):
                 t.parse(
                     forceDefaultQuarterLength=self.activeDefaultQuarterLength,
                     forceKeySignature=activeKeySignature)
+
+                if t.isRest:
+                    continue
+
                 # get the quarter length from the sub-tokens
-                # note: assuming these are the same
-                chordDurationPost = t.quarterLength
-            if isinstance(t, ABCNote) and not t.isRest:
+                # All the notes within a chord should normally have the same length,
+                # but if not, the chord duration is that of the first note.
+                if not inner_quarterLength:
+                    inner_quarterLength = t.quarterLength
+
                 self.subTokens.append(t)
 
-        if chordDurationPost is not None:
-            self.quarterLength = chordDurationPost
+
+        # When both inside and outside the chord length modifiers are used,
+        # they should be multiplied. Example: [C2E2G2]3 has the same meaning as [CEG]6.
+        self.quarterLength = outer_lengthModifier * inner_quarterLength
 
 
 # ------------------------------------------------------------------------------
@@ -2193,9 +2217,19 @@ class ABCHandler:
             # get chords
             if c == '[':
                 j = self.pos + 1
+
+                # find closing chord bracket
                 while j < self.srcLen - 1 and self.strSrc[j] != ']':
                     j += 1
+
                 j += 1  # need character that caused break
+
+                # find outer chord length modifier
+                while j < self.srcLen and (self.strSrc[j].isdigit() or self.strSrc[j] in '/'):
+                    j += 1
+
+
+
                 # prepend chord symbol
                 if activeChordSymbol != '':
                     self.currentCollectStr = activeChordSymbol + self.strSrc[self.pos:j]
--- a/music21/abcFormat/__init__.py
+++ b/music21/abcFormat/__init__.py
@@ -1668,10 +1668,8 @@ class ABCChord(ABCNote):
         # outerLengthModifierStr', outerLengthModifierStr])
 
         # Get the outer chord length modifier if present
-        outer_lengthModifier = 1.0
-
         outer_lengthModifier = self.getQuarterLength(outerLengthModifierStr,
-                                                    forceDefaultQuarterLength=1.0)
+                                                     forceDefaultQuarterLength=1.0)
 
         if forceKeySignature is not None:
             activeKeySignature = forceKeySignature
--- a/music21/abcFormat/__init__.py
+++ b/music21/abcFormat/__init__.py
@@ -1651,15 +1651,15 @@ class ABCChord(ABCNote):
         '''
             Chord without length modifier: [ceg]
             Chords with outer length modifier: [ceg]2, [ceg]/2
-            Cords with inner length modifier: [c2e2g2], [c2eg]
-            Cords with inner and outer length modifier: [c2e2g2]/2, [c/2e/2g/2]2
+            Chords with inner length modifier: [c2e2g2], [c2eg]
+            Chords with inner and outer length modifier: [c2e2g2]/2, [c/2e/2g/2]2
         '''
 
         self.chordSymbols, nonChordSymStr = self._splitChordSymbols(self.src)
 
         # position of the closing bracket
         pos = nonChordSymStr.index(']')
-        # Length modifier string behind der chord brackets
+        # Length modifier string behind the chord brackets
         outerLengthModifierStr = nonChordSymStr[pos + 1:]
         # String in the chord brackets
         tokenStr = nonChordSymStr[1:pos]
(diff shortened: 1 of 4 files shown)


mscuthbert:
coveralls had a problem -- closing and reopening to get a fresh test.

mscuthbert:
Thanks -- look at the lint errors, etc. -- but it seems to be on the right track!

marian-schulz:
I have fixed the PEP8 violations and cleaned up my commit history.
The tests are all green now.

I apologize for not doing the code inspection before.

marian-schulz:
The chord symbols before a chord were missing. I have also fixed this and added an additional unittest.

music21/abcFormat/testFiles.py
```diff
@@ -816,6 +816,67 @@ def testAbc21DirectiveCarryNot(self):
...
+
+        af = abcFormat.ABCFile()
+        # default length of this test
+        abc_dl = 'L:1/8\n'
+
+        # Empty Chords should be skipped at all
+        for abc_chord in ['[]', '[z]']:
+            ah = af.readstr(abc_dl + '[]')
+            s = translate.abcToStreamScore(ah)
+            self.assertFalse(s.elements[1].getElementsByClass(chord.Chord),
```
mscuthbert, line 834: see below about elements[1]

music21/abcFormat/testFiles.py
```diff
@@ -816,6 +816,67 @@ def testAbc21DirectiveCarryNot(self):
...
+            ('[c/2e/2g/2]2', 0.5),
+            ('[c/2e/2g/2]/2', 0.125)
+        ]
+
+        for abc_chord, quarter_length in abc_chords:
+            ah = af.readstr(abc_dl + abc_chord)
+            s = translate.abcToStreamScore(ah)
+            self.assertEqual(s.duration.quarterLength, quarter_length,
+                             'invalid duration of chord "%s"' % abc_chord)
+            for e in s.elements[1].getElementsByClass(chord.Chord):
```
mscuthbert, line 857: can `.elements[1]` be changed to whatever it is we're actually getting, like `.parts[0]` etc. -- otherwise this test is fragile if other metadata elements or something else like a spanner is later a default part of abcToStreamScore.

music21/abcFormat/testFiles.py
```diff
@@ -816,6 +816,67 @@ def testAbc21DirectiveCarryNot(self):
...
+        # list of test abc chords and their quarter lengths at the default length of 1/8
+        # list[tuple(str, int)] = of abc chords and= [( abc_chord: str)]
+        abc_chords = [
+            ('[ceg]', 0.5),
+            ('[ceg]2', 1.0),
+            ('[c2e2g2]', 1.0),
+            ('[ce2g]', 0.5),
+            ('[ceg2]', 0.5),
+            ('[c2e2g2]/2', 0.5),
+            ('[c/2e/2g/2]', 0.25),
```
mscuthbert, line 846: Add some tests with sharps, accidentals, etc. to make sure that c#/2 still works.  Maybe a separate test so it doesn't break the pitch_name in 'CEG' test below.  Just always paranoid about that.

music21/abcFormat/testFiles.py
```diff
@@ -816,6 +816,67 @@ def testAbc21DirectiveCarryNot(self):
...
+            ah = af.readstr(abc_dl + abc_chord)
+            s = translate.abcToStreamScore(ah)
+            self.assertEqual(s.duration.quarterLength, quarter_length,
+                             'invalid duration of chord "%s"' % abc_chord)
+            for e in s.elements[1].getElementsByClass(chord.Chord):
+                for pitch_name in 'CEG':
+                    self.assertIn(pitch_name, e.pitchNames,
+                                  'Pitch not in Chord "%s"' % abc_chord)
+
+    def testAbc21ChordSymbol(self):
```
mscuthbert, line 862: great!

mscuthbert (changes requested):
Looks great!  The `.elements[1]` call is all that needs to be changed, to make sure that it's explicitly getting the type of element that it should be.

(btw -- `s.elements[1]` is almost always better as `e[1]` anyhow -- .elements[1] is slower and doesn't setActiveSite).

music21/abcFormat/testFiles.py
marian-schulz: Fixed this in all cases. I use the getElementbyClass of operator now.

music21/abcFormat/testFiles.py
marian-schulz: I agree.

marian-schulz:
Optimze the Testcases.
+ Use the ElementsByClass operator instead of an index access to .elements[x].
+ Add sharps, flats, natural and ocatve modificators to notes.

Broken rythm
+ Fixes broken rythm for Chords (i guess i broke that)
+ Add A Testcase to cover brocken rythm between chords and notes

music21/abcFormat/__init__.py
```diff
@@ -1641,21 +1641,37 @@ class ABCChord(ABCNote):
...
         super().__init__(src)
         # store a list of component objects
         self.subTokens = []
 
     def parse(self, forceKeySignature=None, forceDefaultQuarterLength=None):
+        '''
+            Chord without length modifier: [ceg]
+            Chords with outer length modifier: [ceg]2, [ceg]/2
+            Cords with inner length modifier: [c2e2g2], [c2eg]
+            Cords with inner and outer length modifier: [c2e2g2]/2, [c/2e/2g/2]2
```
jacobtylerwalls, line 1655: Chords

music21/abcFormat/testFiles.py
```diff
@@ -816,10 +816,127 @@ def testAbc21DirectiveCarryNot(self):
...
+            self.assertEqual(ist_left.duration.quarterLength, soll_left,
+                             f'Invalid left note/chord length of abc broken rythm: {abc}')
+            self.assertEqual(ist_right.duration.quarterLength, soll_right,
+                             f'Invalid right note/chord length of abc broken rythm: {abc}')
+
 
 if __name__ == '__main__':
     import music21
-    # music21.converter.parse(reelsABC21, format='abc').scores[1].show()
+    music21.converter.parse(reelsABC21, format='abc').scores[1].show()
```
jacobtylerwalls, line 941: Better to leave this commented so it doesn't launch a file.

jacobtylerwalls (commented):
Happened to stroll by and noticed a couple small typos or details.

music21/abcFormat/testFiles.py
```diff
@@ -831,7 +831,8 @@ def testAbc21Chords(self):
         for abc_chord in ['[]', '[z]']:
             ah = af.readstr(abc_dl + '[]')
             s = translate.abcToStreamScore(ah)
-            self.assertFalse(s.elements[1].getElementsByClass(chord.Chord),
+            part = s.getElementsByClass(stream.Part)
```
mscuthbert, line 834: this is returning a Stream iterator -- you probably want a specific part, like:

```
part = s.getElementsByClass(stream.Part)[0]
```

which is abbreviated as:

```
part = s.parts[0]
```

music21/abcFormat/testFiles.py
```diff
@@ -854,14 +855,15 @@ def testAbc21Chords(self):
             s = translate.abcToStreamScore(ah)
             self.assertEqual(s.duration.quarterLength, quarter_length,
                              'invalid duration of chord "%s"' % abc_chord)
-            for e in s.elements[1].getElementsByClass(chord.Chord):
+            part = s.getElementsByClass(stream.Part)
+            for e in part.getElementsByClass(chord.Chord):
```
mscuthbert: Similarly here.  I'll bet that `e` is never being set.  Temporarily put a print statement into the loop and you'll see it never runs.

music21/abcFormat/__init__.py
```diff
@@ -1669,9 +1669,9 @@ def parse(self, forceKeySignature=None, forceDefaultQuarterLength=None):
 
         # Get the outer chord length modifier if present
         outer_lengthModifier = 1.0
-        if outerLengthModifierStr:
-            outer_lengthModifier = self.getQuarterLength(outerLengthModifierStr,
-                                                         forceDefaultQuarterLength=1.0)
+
+        outer_lengthModifier = self.getQuarterLength(outerLengthModifierStr,
+                                                    forceDefaultQuarterLength=1.0)
```
mscuthbert, line 1674: misalignment of hanging indent (one space more)

music21/abcFormat/testFiles.py
marian-schulz: Thanks,
I fixed this and  notice an  additional Bug, i guess made by the translater. (I still have to research this)

for abc_chord in ['[]', '[z]']:
    ah = af.readstr(abc_dl + '[]')
    s = translate.abcToStreamScore(ah)
    chord = list(s.parts[0].notes)
    print(abc_chord, chord, ah.tokens)

[] [<music21.note.Note C>] [<music21.abcFormat.ABCMetadata 'L:1/8'>, <music21.abcFormat.ABCChord '[]'>]
[z] [<music21.note.Note C>] [<music21.abcFormat.ABCMetadata 'L:1/8'>, <music21.abcFormat.ABCChord '[]'>]

There is no Chord at all but a note C in the part. The C is not present in abc code nor token list at all.

music21/abcFormat/__init__.py
marian-schulz: add this extra space and removed the previous line.

music21/abcFormat/testFiles.py
marian-schulz: I wonder while the stream iterator implements .getElementsByClass at all ?
However .parts[0] looks better.
fixed in the last commit.

mscuthbert: It totally does.  StreamIterators are designed to be chained, so most explicit cases would be:

```
s.flat.iter.getElementsByOffset(10.0, 20.0).getElementsByClass('Note').getElementsByGroup('cue').addFilter(lambda n: n.pitch.alter == 1)
```

As of v.6., each of these "getElementsBy..." returns a new StreamIterator that adds a new filter on the notes.  So `s.getElementsByClass(stream.Part).getElementsByClass(chord.Chord)` would always return no elements, because first it has to be of class Part, and then of class Chord, and nothing will ever be like that.

It might be that adding a second filter of the same type should raise an exception, but there are a few cases where you might want this, such as:

`s.getElementsByClass(spanner.Spanner).getElementsByClass(expressions.Tremolo)`

which only gets Tremolos that are Spanners. 

Maybe that doesn't work, but there are a few cases like that.

mscuthbert:
looks really good -- one little change -- capital letter variable names are only used for classes.  Will merge if everything passes.  Thank you for the great contributions!

music21/abcFormat/testFiles.py
```diff
@@ -816,10 +816,130 @@ def testAbc21DirectiveCarryNot(self):
...
+        for abc_text in ('"C"C', '"C"[ceg]'):
+            ah = af.readstr(abc_dl + abc_text)
+            part = translate.abcToStreamScore(ah).parts[0]
+            chord_symbol = part.getElementsByClass(harmony.ChordSymbol)
+            self.assertTrue(chord_symbol, 'No ChordSymbol found in abc: "%s"' % abc_text)
+            for pitch_name in 'CEG':
+                self.assertIn(pitch_name, chord_symbol[0].pitchNames,
+                              'Pitch not in ChordSymbol of abc: "%s"' % abc_text)
+
+    def testAbc21BrokenRythm(self):
```
jacobtylerwalls, line 887: rhythm

music21/abcFormat/testFiles.py
```diff
@@ -816,10 +816,130 @@ def testAbc21DirectiveCarryNot(self):
...
+        # Test the chord symbol for note and chord
+        from music21 import abcFormat, note
+        from music21.abcFormat import translate
+
+        # default length of this test
+        abc_dl = 'L:1/4\n'
+
+        # test abc strings of broken rythm between 2 notes and/or chords and their
```
jacobtylerwalls, line 895: same

music21/abcFormat/testFiles.py
```diff
@@ -816,10 +816,130 @@ def testAbc21DirectiveCarryNot(self):
...
+            ah = af.readstr(abc_dl + abc)
+            part = translate.abcToStreamScore(ah).parts[0]
+            general_notes = part.getElementsByClass(note.GeneralNote)
+            self.assertEqual(len(general_notes), 2,
+                             f'Wrong numbers of Notes found in abc: {abc}!')
+            ist_left, ist_right = general_notes
+            self.assertEqual(ist_left.duration.quarterLength, soll_left,
+                             f'Invalid left note/chord length of abc broken rythm: {abc}')
+            self.assertEqual(ist_right.duration.quarterLength, soll_right,
+                             f'Invalid right note/chord length of abc broken rythm: {abc}')
```
jacobtylerwalls, line 939: here and above

jacobtylerwalls (commented):
Noticed a few instances where "rhythm" is misspelled.

mscuthbert:
Please don't force-push when making Pull Requests -- it makes it impossible to review what has changed.  Your open-source maintainers will appreciate it.

mscuthbert:
Congrats! Merged

PR #773, opened 2021-01-13
Filtering out null channel values

bearpelican:
Python3 specific bug:
track.getChannels throws an exception if one of the channels is None.
This is because the channels are being sorted, and python3 no longer allows comparison between int and None.

I don't have enough context, so feel free to close this PR and put up the correct fix instead.

Steps to REPRO:

1. Download midi file: 
`wget https://github.com/bearpelican/musicautobot/raw/master/data/midi/examples/A%20Thousand%20Miles%20-%20Vanessa%20Carlton%20-%20Verse-And-Pre-Chorus.mid`
2. Run Following
```python
import music21
file = 'A Thousand Miles - Vanessa Carlton - Verse-And-Pre-Chorus.mid'

mf = music21.midi.MidiFile()
mf.open(file)
mf.read()
mf.close()

[t.getChannels() for t in mf.tracks]
```

Screen Shot 2021-01-13 at 14 51 48

### DIFF
--- a/music21/midi/__init__.py
+++ b/music21/midi/__init__.py
@@ -1433,14 +1433,15 @@ class MidiTrack(prebase.ProtoM21Object):
 
     def getChannels(self):
         '''
-        Get all channels used in this Track (sorted)
+        Get all channels (excluding None) used in this Track (sorted)
 
         >>> mt = midi.MidiTrack(index=2)
         >>> noteOn = midi.MidiEvent(type=midi.ChannelVoiceMessages.NOTE_ON, channel=14)
         >>> noteOn2 = midi.MidiEvent(type=midi.ChannelVoiceMessages.NOTE_ON, channel=5)
         >>> noteOn3 = midi.MidiEvent(type=midi.ChannelVoiceMessages.NOTE_ON, channel=14)
+        >>> noteOn4 = midi.MidiEvent(type=midi.ChannelVoiceMessages.PROGRAM_CHANGE, channel=None)
 
-        >>> mt.events = [noteOn, noteOn2, noteOn3]
+        >>> mt.events = [noteOn, noteOn2, noteOn3, noteOn4]
 
         >>> mt.getChannels()
         [5, 14]
--- a/music21/midi/__init__.py
+++ b/music21/midi/__init__.py
@@ -1447,7 +1447,7 @@ class MidiTrack(prebase.ProtoM21Object):
         '''
         post = []
         for e in self.events:
-            if e.channel not in post:
+            if e.channel not in post and e.channel is not None:
                 post.append(e.channel)
         return sorted(post)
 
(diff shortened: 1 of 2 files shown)


jacobtylerwalls:
LGTM. The only call in music21 to this function is from the test suite, so the risk here is nil. I checked out v.5.7.2 and can reproduce the same issue, so predates some of the recent MIDI work. Thanks for identifying the issue.

jacobtylerwalls:
That said, the tests that currently expect `[None]` will need to be updated to `[]`.

mscuthbert:
Closing and reopening because tests were canceled.

music21/midi/__init__.py
```diff
@@ -1445,10 +1445,7 @@ def getChannels(self):
         >>> mt.getChannels()
         [5, 14]
         '''
-        post = []
-        for e in self.events:
-            if e.channel not in post:
-                post.append(e.channel)
+        post = { e.channel for e in self.events if e.channel is not None }
```
mscuthbert, line 1448: Can you rewrite using the original code just changing `if e.channel not in post and e.channel is not None` ?  Or define a set at the beginning and just put `if e.channel is None: \n continue` before a `post.add(e.channel)` line.  There are a lot of beginner programmers who read music21, and in teaching them, I find that explicit iteration is much easier for them to grasp than filtered set comprehensions.

Otherwise happy to take after tests pass.

bearpelican: Ah yes! I'll update the code to keep the same style

bearpelican:
> That said, the tests that currently expect `[None]` will need to be updated to `[]`.
> EDIT: in fact, that particular test I added for the conductor track feature in v.6. Since this function has been broken before that, I doubt there are applications relying on the behavior of inspecting a MIDI export in > v.6 and checking for `[None]`.

If expected behavior is [None, 1], we can also fix the sort error with:
```python
sorted([1, None], key=lambda x: x if x is not None else -1)
```

jacobtylerwalls:
Well, we don't have a test for this in the suite right now. I was referring to the test in the suite that fails with your change (i.e. you should adjust the expected behavior from `[None]` to `[]`).

Regarding whether the track that caused your exception should yield `[1]` or `[None, 1]`, my vote is for `[1]` as you have it. A word in the docstring to the effect that events with channel=None are ignored might be helpful. But Myke should  weigh in.

mscuthbert:
Yeah, that's perfect.  No channel = no note.  :-D  so fine to ignore.

bearpelican:
Ok updated the fix and the tests.

I also updated the docstring - 15f836b1c07ee09958aafaa12e902ae4e0308601.
Not sure how necessary it is, so I can revert the change as well.

mscuthbert:
Looks perfect!  Thanks for the contribution!

bearpelican:
Thanks for this amazing library! 😄

PR #777, opened 2021-01-14
Replace spanned elements when stripping ties

jacobtylerwalls:
Spanners would disappear if a terminus was stripped by `stripTies`:

```
sch = corpus.parse('schoenberg/opus19', 6)
sch.stripTies().measure(8).show()
```

Before:

stripped-no-spanner

After:

stripped-with-spanner

### DIFF
--- a/music21/common/objects.py
+++ b/music21/common/objects.py
@@ -306,6 +306,8 @@ class Timer:
 
     >>> t = common.Timer()
     >>> now = t()
+    >>> import time  #_DOCS_HIDE
+    >>> time.sleep(0.01)  #_DOCS_HIDE  -- some systems are extremely fast or have wide deltas
     >>> nowNow = t()
     >>> nowNow > now
     True
--- a/music21/stream/__init__.py
+++ b/music21/stream/__init__.py
@@ -6676,6 +6676,13 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
 
                 # set tie to None on first note
                 notes[posConnected[0]].tie = None
+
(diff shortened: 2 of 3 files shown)


jacobtylerwalls:
Now that I know more about `coreElementsChanged` I noticed that this could have been removed during my first patch for stripTies last summer because using `stream.remove()` instead of `pop()` obviates the need.

mscuthbert:
hmmm.... I'll have to think on this -- sometimes we call stripTies in order to do musical analysis but then want to show the original score (for instance, with all "complex" notes turned orange, or with a pitchClass analysis in the lyrics).  Will this change mean that the spanners disappear in the original score if it is shown?  It might be that we need to make a copy of the spanner and put that copy into the stripTies score.  Thoughts?

jacobtylerwalls:
This acts on the return object only, so if it's inPlace every transformation is destructive, and if not, then we're only acting on the copy. In your example are you thinking of an inPlace=False operation?

mscuthbert:
Ah! Of course -- you're right, we make a deepcopy so everything is new in the new output, and if we're doing inPlace=True then it's much less of a problem that the original notes no longer are part of spanners.

<s>It does make me think that there's a bigger (and pre-existing) bug on cleaning-up sites (isDead=True) in that notes that are removed still think they're in spanners.</s> Nope!  Checked the code -- replaceSpannedElements calls .replace which ensures that sites are cleaned up properly.  Okay -- looks good (and thanks for the timer test!) -- merging!

PR #781, opened 2021-01-15
Scale derived

marian-schulz:
I have added a new function to the key.Key class. 

Derive a fitting Key from the KeySigntature and the tonic of this Key.
Thew new key will have the same tonic as this key and the alternated pitches from
the key and the key signature. If alternated pitches of the same step exist in Key
and KeySignature, the alternated pitch is taken from the Keysignature.

Examples:
>>> k = key.Key('G', 'major')
>>> ks = key.KeySignature()
>>> ks.alteredPitches = ['Fn']
>>> k.deriveByKeySignature(ks)
[<music21.key.Key of G mixolydian>]

>>> k = key.Key('C', 'major')
>>> ks = key.KeySignature()
>>> ks.alteredPitches = ['F#']
>>> k.deriveByKeySignature(ks)
[<music21.key.Key of C lydian>]

Why this function ?
It is not only great, but also useful for handling the key in ABCFormat module. (I'm currently working on it).
In ABC a given key can be modified by additional 'accidentials' (alternated keys).

Example:
K:G =F

This means 'G major with an F natural' or also know as 'G mixolydian'. (Thx to Greg Chapman)
With this function it is possible to use a known key instead of a simple key signature when importing ABC to music21.

I am open for suggestions.

### DIFF
--- a/music21/key.py
+++ b/music21/key.py
@@ -999,9 +999,20 @@ class Key(KeySignature, scale.DiatonicScale):
     def deriveByKeySignature(self, KeySig: Union[KeySignature, str]) -> List['Key']:
         '''
         Derive a fitting Key from the KeySignature and the tonic of this Key.
-        Thew new key will have the same tonic as this key and the alternated pitches from
-        the key and the key signature. If alternated pitches of the same step exist in Key
-        and KeySignature, the alternated pitch is taken from the Keysignature.
+        Thew new key will have the same tonic as this key and the altered pitches from
+        the key and the key signature. If altered pitches of the same step exist in Key
+        and KeySignature, the altered pitch is taken from the Keysignature.
+        >>> k = key.Key('C', 'major')
+        >>> ks = key.KeySignature()
+        >>> ks.alteredPitches = []
+        >>> k.deriveByKeySignature(ks)
+        [<music21.key.Key of C major>]
+
+        >>> k = key.Key('G', 'major')
+        >>> ks = key.KeySignature()
+        >>> ks.alteredPitches = ['C#']
+        >>> k.deriveByKeySignature(ks)
+        [<music21.key.Key of G lydian>]
 
         >>> k = key.Key('G', 'major')
         >>> ks = key.KeySignature()
@@ -1013,7 +1024,13 @@ class Key(KeySignature, scale.DiatonicScale):
         >>> ks = key.KeySignature()
         >>> ks.alteredPitches = ['F#']
         >>> k.deriveByKeySignature(ks)
-        [<music21.key.Key of C lydian>, <music21.key.Key of C hypolocrian>]
+        [<music21.key.Key of C lydian>]
+
+        >>> k = key.Key('C#', 'major')
+        >>> ks = key.KeySignature()
+        >>> ks.alteredPitches = ['Cn']
+        >>> k.deriveByKeySignature(ks)
+        []
         '''
 
         def derived_scales(cls: Type):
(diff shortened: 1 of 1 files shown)


marian-schulz:
There is an Exception in class AbstractOctaveRepeatingScale(AbstractScale) -  i silence this by an 'except:'

This Exception is caused by the  __init__ default value 'intervalList' of type tuple of the subclass class OctaveRepeatingScale.

>> def __init__(self, tonic=None, intervalList=('m2',)):
There is a List in the name, but its a tuple ;)

The Exception is raised in AbstractOctaveRepeatingScale.buildNetwork(self, mode) when appending to the value of the previous mentioned tuple default value

>> AttributeError: 'tuple' object has no attribute 'append'
I wonder how that pass the linter.

music21/key.py
```diff
@@ -996,6 +996,56 @@ def tonicPitchNameWithCase(self):
...
+            # find closest-matching ConcreteScale based on the pitch collection
+            try:
+                # Derive a scale from the concrete scale using the keyPitches
+                sc = name(self.tonic).derive(other=keyPitches.values())
+                if sc.tonic == self.tonic:
+                    # create a key with the given tonic and  mode of the scale
+                    scales.append(Key(tonic=self.tonic, mode=sc.type))
+            except AttributeError:
+                # Something has failed within one of the scales
+                pass
```
jacobtylerwalls, line 1041: Thanks for finding the issue with intervalList! I have a separate patch for this in #785, so I suggest removing the workaround here.

music21/key.py
```diff
@@ -996,6 +996,56 @@ def tonicPitchNameWithCase(self):
             tonic = tonic.lower()
         return tonic
 
+    def deriveByKeySignature(self, KeySig: Union[KeySignature, str]) -> List['Key']:
+        '''
+        Derive a fitting Key from the KeySigntature and the tonic of this Key.
+        Thew new key will have the same tonic as this key and the elternated pitches from
+        the key and the key signature. Conflicting alternating keys are overriden by the
```
jacobtylerwalls, line 1003: "overridden". also--is "conflicting altered pitches" is more your meaning? I can't quite follow it as written.

music21/key.py
```diff
@@ -996,6 +996,56 @@ def tonicPitchNameWithCase(self):
             tonic = tonic.lower()
         return tonic
 
+    def deriveByKeySignature(self, KeySig: Union[KeySignature, str]) -> List['Key']:
+        '''
+        Derive a fitting Key from the KeySigntature and the tonic of this Key.
```
jacobtylerwalls, line 1001: "fitting" and "KeySignature"

music21/key.py
```diff
@@ -996,6 +996,56 @@ def tonicPitchNameWithCase(self):
...
+        Thew new key will have the same tonic as this key and the elternated pitches from
```
jacobtylerwalls, line 1002: "alternated", but perhaps "altered" is more your meaning?

jacobtylerwalls (commented):
Thanks for finding the issue with `intervalList`! Better to patch it than work around it, so I opened #785.

music21/key.py
marian-schulz: You're rigth, i did a rewording for this.

>>> If alternated pitches of the same step exist in Key and KeySignature, the alternated pitch is taken from the Keysignature.

Sound better ?

music21/key.py
marian-schulz: i did.

music21/key.py
marian-schulz: Fixed

marian-schulz:
I have made changes again.

1) I catch a ScaleException from 'DiatonicScale'.  The 'type' diatonic is not a valid mode.

3.) Duplicates in the results are sorted out.

2) The function derive from the scale module returns a 'closest-matching 'ConcreteScale'.  Therefore some scales are not exactly what I am looking for.
I now check if the pitches of the scale exactly match the given pitches.

**Now I get only one result in my tests. Is this to be expected ? Can there be two keys with the same tonic and key signature ? 
If I know for sure that there is only one key, I can simplify the code and increase the performance slightly.**

In case no key is found do you prefer an exception or a none/empty list as return value?

music21/key.py
marian-schulz: > Thanks for finding the issue with `intervalList`! Better to patch it than work around it, so I opened #785.

How do I rebase my commits so that the tests don't stumble over this issue ?

jacobtylerwalls:
Hi. Would you be able to pull from (or rebase on) master to apply the fix to the issue you identified earlier? Thanks!

marian-schulz:
i have found the time to rebase my commits to the recent master, and now the test are green.
excelent

mscuthbert:
I'm not sure that this is an attribute on Key or KeySignature objects -- anything that requires creating a key and another keySignature in order to get a list of keys seems like it's not placed in the right place.  Why is it returning a list of keys?  Is there an example where more than one key will be returned?

marian-schulz:
>> anything that requires creating a key and another keySignature in order to get a list of keys seems like it's not placed in the right place.

Was an interesting question.
It came up during my work on the ABC module. There a key can be modified by accidentals. I thought it would be nice to know if such a modified key signature corresponds to a known key and to set such a key instead of a key signature.

E.g. i diditn't know that:
K: G =f
is equal to:
K: G mix

>> Why is it returning a list of keys? Is there an example where more than one key will be returned?
No it isn't.
From a programmer's point of view, there is the possibility for multiple results. From the point of view of musicology, rather not. 

I will gladly change this if you think there is room for such functionality. 
Does music21 have some kind of toolbox for such cases ? (not worth (not worth for the core but useful) 

I'm sure i will keep it for myself anyway.

jacobtylerwalls:
I like the idea of solving for mode from sharps and tonic. Right now `asKey()` lets you solve for tonic from sharps and mode. There could be an additional argument `tonic=None` (and of course, impossible solutions would raise errors):

https://github.com/cuthbertLab/music21/blob/ef4bdd6ae787ffb189486a126e6f1a2ee0b556c7/music21/key.py#L383-L387

The specific problem you were having in translating from abc where you needed to be able to get from `K: G =f` to 0 sharps is I think an abc specific problem; perhaps better to just count the resulting sharps or flats from the original key and applying the alterations, and use __that__ to get a `Key` by solving for mode from sharps and tonic in the fashion I suggest above.

Thoughts?

jacobtylerwalls:
Hi @marian-schulz we're unlikely to take this as currently drafted, because although brute-force searching every scale for a match solved your use case, we would be more inclined to take a method implementing music theory to get to the answer, which is why I suggested enhancements to `asKey()`. Feel free to open a new pull request if you choose to go in that direction. Thanks.

PR #782, opened 2021-01-16
Use shell=True when executing "start" on Windows command line and assorted Windows fixes

### LINKED ISSUE #783: DirectoryInformation.findWorks() doesn't group movements in subdirectories correctly on Windows
**music21 version**

`6.6.0`

 
**Problem summary**
Doctest for `DirectoryInformation.findWorks()` is failing on Windows.

```
>>> di = corpus.work.DirectoryInformation('schoenberg',
...             corpusObject=corpus.corpora.CoreCorpus())
>>> di.findWorks()
```

**Result** (ignore my manual indentation of "Got:")

```
File "D:\a\music21\music21\music21\corpus\work.py", line 54, in corpus_work.DirectoryInformation.findWorks
Failed example:
    di.findWorks()
Expected:
    OrderedDict([(...'opus19', CorpusWork(title='Opus 19',
                                   files=[CorpusFile(path='schoenberg/opus19/movement2.mxl',
                                                     title='Movement 2',
                                                     filename='movement2.mxl',
                                                     format='musicxml',
                                                     ext='.mxl'),
                                          CorpusFile(path='schoenberg/opus19/movement6.mxl',
                                                     title='Movement 6',
                                                     filename='movement6.mxl',
                                                     format='musicxml',
                                                     ext='.mxl')],
                                    virtual=False))])
Got:
    OrderedDict(
        [('/opus19/movement2',
            CorpusWork(title='/Opus 19 /Movement 2',
                files=[
                    CorpusFile(
                    path='/opus19/movement2.mxl',
                    title='/Opus 19 /Movement 2',
                    filename='/opus19/movement2.mxl',
                    format='musicxml',
                    ext='.mxl')
                    ],
                virtual=False)),
        ('/opus19/movement6',
            CorpusWork(title='/Opus 19 /Movement 6',
                files=[
                    CorpusFile(
                    path='/opus19/movement6.mxl',
                    title='/Opus 19 /Movement 6',
                    filename='/opus19/movement6.mxl',
                    format='musicxml',
                    ext='.mxl')
                    ],
                virtual=False))])
 ```
**More information**
Windows on GitHub actions, see run at https://github.com/jacobtylerwalls/music21/runs/1714018939

### PR DESCRIPTION
**Pull request by jacobtylerwalls**
- Fixes issue reported on m21 list 1/11/21 by Brian M. regarding `.plot()` failing on Windows.

Compare the CI runs on my fork from https://github.com/jacobtylerwalls/music21/runs/1714077317 and https://github.com/jacobtylerwalls/music21/runs/1714098851 to see the lack of `FileNotFoundError` with this change.

- Allows replacing "PosixPath" with "WindowsPath" in doctests when running on Windows.
- Fixes another test depending on forward slash separators.
- Fixes many tests failing on Windows because strings over 260 characters were passed to os.path.exists
- Fixes extra new lines on Windows in output of tsvConverter
- Fixes #783: mixing `as_posix()` with `os.sep` in `DirectoryInformation.findWorks()`

### CODE REVIEW

**Comment by jacobtylerwalls** [comment]
@MarkGotham FYI I fixed a bug in the tsv writer, essentially, when passing file objects to `csv.writer`, the file object must always be opened with `newline=''`, otherwise newlines can multiply:
```
FAIL: testM21ToTsv (romanText_tsvConverter.Test)
  File "D:\a\music21\music21\music21\romanText\tsvConverter.py", line 750, in testM21ToTsv
    self.assertEqual(handler.tsvData[0][0], 'I')
AssertionError: '' != 'I'
```
python docs:
https://docs.python.org/3/library/csv.html#id3
passing test run:
https://github.com/jacobtylerwalls/music21/runs/1715618237

glad the test was there! :-)

File: music21/converter/__init__.py [hunk 1]
```diff
@@ -1163,10 +1163,15 @@ def parse(value: Union[bundles.MetadataEntry, bytes, str, pathlib.Path],
     # a midi string, must come before os.path.exists test
     elif not isinstance(value, bytes) and valueStr.startswith('MThd'):
         return parseData(value, number=number, format=m21Format, **keywords)
-    elif not isinstance(value, bytes) and os.path.exists(valueStr):
+    # windows systems can't test existence of filepath names > 260 chars
+    elif (not isinstance(value, bytes)
+          and len(valueStr) <= 260
```
mscuthbert, line 1168: this seems too restrictive both for Windows and other os's (why put a windows restriction onto other filesystems).  In windows, valueStr can be > 260 bytes -- up to 32768, but each directory of the valueStr must be smaller than 260 bytes, so "C:\My Windows Hard Drive\Users\Michael Scott Cuthbert's Home Directory\Myke's Documents and Settings\2020 Was a Very Hard Year\Music Files are Here\Keep Looking for Music Files\They are in here somewhere\Keep Digging\Here's my Directory\No not this one\Yes this one\Okay I think this is longer than 260 characters now\File.mxl" should be fully parseable.

We probably need a separate routine: `and osCanLoad(valueStr)` which returns True for non-windows, but checks the path length in Windows to have no parts > 260 and no total length > 32768.

music21/converter/__init__.py
```diff
@@ -1163,10 +1163,15 @@ def parse(value: Union[bundles.MetadataEntry, bytes, str, pathlib.Path],
...
+          and os.path.exists(valueStr)):
         return parseFile(valueStr, number=number, format=m21Format,
                          forceSource=forceSource, **keywords)
-    elif not isinstance(value, bytes) and os.path.exists(common.cleanpath(valueStr)):
+    elif (not isinstance(value, bytes)
+          and len(valueStr) < 240
```
mscuthbert, line 1173: similarly to above.

music21/environment.py
```diff
@@ -803,9 +803,8 @@ class Environment:
     the :class:`~music21.environment.UserSettings` object.
 
     >>> env = environment.Environment(forcePlatform='darwin')
-    >>> #_DOCS_SHOW env['musicxmlPath'] = '/Applications/Finale Reader.app'
-    >>> #_DOCS_SHOW env['musicxmlPath']
-    >>> print("PosixPath('/Applications/Finale Reader.app')") #_DOCS_HIDE
+    >>> env['musicxmlPath'] = '/Applications/Finale Reader.app'
+    >>> env['musicxmlPath']
     PosixPath('/Applications/Finale Reader.app')
```
mscuthbert, line 808: Won't this fail on windows?

music21/romanText/tsvConverter.py
```diff
@@ -299,16 +297,13 @@ def importTsv(self):
                     continue
                 values = line.strip().split('\t')
                 data.append([v.strip('\"') for v in values])
-        f.close()
```
mscuthbert, line 302: Thanks

music21/romanText/tsvConverter.py
```diff
@@ -299,16 +297,13 @@ def importTsv(self):
...
 
         return data
 
     def tsvToChords(self):
         '''
         Converts a list of lists (of the type imported by importTsv)
         into TabChords (i.e. a list of TabChords).
-
-        See docs at :class:`music.romanText.TsvHandler`.
```
mscuthbert, line 311: Thanks!

music21/environment.py
mscuthbert: AH! I get it now.  THANKS!

music21/test/testRunner.py
```diff
@@ -89,13 +90,16 @@ def fixDoctests(doctestSuite):
...
+    windows: bool = platform.system() == 'Windows'
     for dtc in doctestSuite:  # Suite to DocTestCase -- undocumented.
         if not hasattr(dtc, '_dt_test'):
             continue
 
         dt = dtc._dt_test  # DocTest
         for example in dt.examples:
             example.want = stripAddresses(example.want, '0x...')
+            if windows:
+                example.want = example.want.replace('PosixPath', 'WindowsPath')
```
mscuthbert, line 102: Very cool.  Thank you!

mscuthbert (changes requested):
Looks really good -- I'm afraid of limiting macOS file path lengths because of windows, so one requested change.

music21/converter/__init__.py
jacobtylerwalls: Agree that we should not apply this limit to other OSes. Like the idea of a new function.

I think the extended paths in your example require an opt-in admin setting, and only work on  >= Windows 10 v.1607 (2015). Python 3.6 can read this setting but will still raise `ValueError` if it's not set:
https://docs.python.org/3/using/windows.html#removing-the-max-path-limitation

GitHub actions happens to enable this setting on their Windows-latest CI server, but that was confusing some users who missed bugs on ordinary systems that way:
https://github.com/actions/virtual-environments/issues/1484

Is it safer to just try and catch `ValueError`? In your example we will still crash on earlier versions of Windows or where the setting is absent.

jacobtylerwalls: I only discovered this on GH actions because some of the musicxml strings are > 32768 chars:
https://github.com/jacobtylerwalls/music21/pull/2/checks?check_run_id=1753971325

mscuthbert:
Thanks!  Wonderful!

PR #383, opened 2019-04-08
WIP: Bugfixes in figure parsing for chord symbols

a-papadopoulos:
I fixed a couple of bugs I encountered while parsing a large dataset of lead sheets.

Basically, I want to ensure that given a ChordSymbol ch, music21 can always create a ChordSymbol given the figure ch.figure, and that the resulting chord symbol is always equivalent to the first.

I have included tests for the several cases I encountered, without any reference to the original dataset.

PR #792, opened 2021-01-24
Chord symbol fixes

jacobtylerwalls:
- Tests authored by @a-papadopoulos in #383
- Passing tests enabled, failing tests disabled
- Fix: set bass pitch when there is no chord; set default octave on bass pitches
- Fix: add bass note to chord's pitches when not in harmony
- Standardize `augmented-major seventh` to `augmented-major-seventh` like every other type
- Fix: treat "add" alterations on the seventh degree against a minor-seventh baseline, per musicXML spec
- Fix: parse inversions from musicXML when only `<inversion>` is present and `<bass>` is absent
- Fix: add chord step modifications before performing bass octave corrections

TODO: ~~one fix per commit to get the other tests to pass~~ What's left is just the "add", "omit", "subtract" business which is sticky and probably needs its own PR. I left those tests commented out with a link to #383.

### DIFF
--- a/music21/harmony.py
+++ b/music21/harmony.py
@@ -2806,7 +2806,7 @@ class Test(unittest.TestCase):
         self.assertEqual(pitches, ChordSymbol('Aminmaj7').pitches)
         self.assertEqual(pitches, ChordSymbol('Am#7').pitches)
 
-    def x_testRootBassParsing(self):
+    def testRootBassParsing(self):
         """
         This tests a bug where the root and bass were wrongly parsed,
         since the matched root and bass were globally removed from figure,
@@ -2826,7 +2826,7 @@ class Test(unittest.TestCase):
           </harmony>
         """
         figure = 'E7/E-'
-        pitches = ('E-3', 'E3', 'G#3', 'B3', 'D4')
+        pitches = ('E-2', 'E3', 'G#3', 'B3', 'D4')
         self.runTestOnChord(xmlString, figure, pitches)
 
 
--- a/music21/harmony.py
+++ b/music21/harmony.py
@@ -1945,7 +1945,7 @@ class ChordSymbol(Harmony):
         ['C3', 'E-3', 'G3']
 
         >>> [str(pi) for pi in CS(root='C', bass='B', kind='major-ninth').pitches]
-        ['B2', 'C3', 'D3', 'E3', 'G3']
+        ['B1', 'D2', 'C3', 'E3', 'G3']
 
         >>> [str(pi) for pi in CS(root='D', bass='F', kind='minor-seventh').pitches]
         ['F3', 'A3', 'C4', 'D4']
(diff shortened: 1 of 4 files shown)


jacobtylerwalls:
https://github.com/cuthbertLab/music21/blob/ed2049ef2897ffa69c73be45b329fc22fac6fec6/music21/chord/__init__.py#L929-L934

Shouldn't that be omitted root?

music21/harmony.py
```diff
@@ -59,7 +59,7 @@
     ('major-seventh', ['1,3,5,7', ['maj7', 'M7']]),  # Y
     ('minor-major-seventh', ['1,-3,5,7', ['mM7', 'm#7', 'minmaj7']]),  # Y: 'major-minor'
     ('minor-seventh', ['1,-3,5,-7', ['m7', 'min7']]),  # Y
-    ('augmented-major seventh', ['1,3,#5,7', ['+M7', 'augmaj7']]),  # N
+    ('augmented-major-seventh', ['1,3,#5,7', ['+M7', 'augmaj7']]),  # N
```
jacobtylerwalls, line 62: OK?

mscuthbert:
> https://github.com/cuthbertLab/music21/blob/ed2049ef2897ffa69c73be45b329fc22fac6fec6/music21/chord/__init__.py#L929-L934
> 
> Shouldn't that be omitted root?

Yeah, I just realized that when I was writing a comment.  You're right.  I've never heard of an omitted bass.

music21/harmony.py
```diff
@@ -1945,7 +1949,7 @@ def _updatePitches(self):
         ['C3', 'E-3', 'G3']
 
         >>> [str(pi) for pi in CS(root='C', bass='B', kind='major-ninth').pitches]
-        ['B2', 'C3', 'D3', 'E3', 'G3']
+        ['B1', 'D2', 'C3', 'E3', 'G3']
```
mscuthbert, line 1952: B1 is pretty low -- is this because we take C3 as a basis but why is D below C here.  I'm a little confused.

jacobtylerwalls: Culprit is this:

https://github.com/cuthbertLab/music21/blob/ed2049ef2897ffa69c73be45b329fc22fac6fec6/music21/harmony.py#L2033-L2038

Going into that block, the pitches are C3, E3, G3, B3, D4. The block bumps C, E, G to octave 5, then quits, instead of skipping the B and then acting on the D to bump the D to D6.

Then, the block that begins `while self._hasPitchAboveC4` brings everything down two octaves.

The change in this diff is that now bass notes are given a default octave of 3 just like roots. Before the root default octave was 3 and the bass octave was silent so we defaulted it from pitch.py.

That meant that this block of code acts on the D4, which it won't in the current version (because D4 > B4):

https://github.com/cuthbertLab/music21/blob/ed2049ef2897ffa69c73be45b329fc22fac6fec6/music21/harmony.py#L2043-L2050

This second block seems duplicative of the first block. I'm going to push another commit for that first block that de-fangs it a little bit.

music21/harmony.py
mscuthbert: yep!

PR #796, opened 2021-01-29
preferSecondaryDominants - initial implementation for V/

MarkGotham:


music21/roman.py
```diff
@@ -770,6 +770,33 @@ def romanNumeralFromChord(chordObj,
     ...     )
     <music21.roman.RomanNumeral #io6b3 in C major>
 
+    The preferSecondaryDominants option defaults to False, but if set to True,
+    then certain rare figures are swapped with their more common secondary equivalent.
+    This is currently limited to three robust cases, replacing:
+    #i with vii/ii,
+    #iv with vii/V,
+    and II with V/V.
```
mscuthbert, line 778: build sphinx to see layout issues here.

music21/roman.py
```diff
@@ -832,6 +859,11 @@ def romanNumeralFromChord(chordObj,
         'i64b3': 'Sw43',
         'io6b5b3': 'Ger65',
     }
+    secondaryDominantSwaps = {
+        ('#', 'i'): ('vii', 'ii'),
```
mscuthbert, line 863: I'd prefer not to have this one, since it's not a secondary  dominant -- would need some sort of "preferAppliedChords" instead of secondary dominant.

music21/roman.py
```diff
@@ -832,6 +859,11 @@ def romanNumeralFromChord(chordObj,
...
+        ('#', 'iv'): ('vii', 'V'),
```
mscuthbert, line 864: same as above.  Has this been tested with diminished chords? viio7 of V, etc.  I imagine it would work, but need to check

music21/roman.py
```diff
@@ -832,6 +859,11 @@ def romanNumeralFromChord(chordObj,
...
+        ('', 'II'): ('V', 'V'),
```
mscuthbert, line 865: what about the other common secondaries: V/vi? V/iii, and do these work in minor as well as major?

mscuthbert (changes requested):
Some requested changes for other common chords and tests in minor.

music21/roman.py
MarkGotham: Ok. I think we'll want that other functionality, but I'll leave it out for now. I'm also including a suggested rename of `preferSecondaryDominants` to `preferSecondaryV` to this effect.

music21/roman.py
MarkGotham: Now fully mapped out with major/minor and tests.

MarkGotham:
The Lint fail is due to `preferSecondaryDominants` calls within `analysis/reduceChordsOld.py` (despite the fact that this option has never previously been operational). Presumably the edict against changing names doesn't apply as strongly to ones that have never worked?!

In any case, I don't mind returning to `preferSecondaryDominants` but I agree with your idea of separating `V/x` from `vii/x` and I think it might be better to have the more specific names:

- `preferSecondaryV`
- `preferAppliedSeventh`

(and thereby hedge on the question of dominant function)

mscuthbert:
If we're going to rename it, it should probably be "preferAppliedChords" -- now that we're taking patches for v7 we can change names, but we should make it so that for one version preferSecondaryDominants is still supported, perhaps with a deprecation warning.

MarkGotham:
Thanks.

That's _version_ 7 right? Not 'v7' as in G, Bb, D, F in C? ;)

That's all fine, though perhaps not worth it for such a slight change. Reverting to the good old `preferSecondaryDominants` and kicking the applied chords can down the road.

MarkGotham:
Hey @mscuthbert - is this a good scope for this PR as is, or would you like preferAppliedChords in here too?

music21/roman.py
```diff
@@ -898,6 +948,12 @@ def romanNumeralFromChord(chordObj,
         elif rnString in ('Fr43', 'Sw43'):
             keyObj = _getKeyFromCache(chordObj.seventh.name.lower())
 
+    if preferSecondaryDominants and (ft.prefix == ''):  # Nothing on altered scale degrees
```
jacobtylerwalls, line 951: Ah, I see -- we're definitely seeing this two different ways.

In c minor, with `preferSecondaryDominants=True`, I would expect Eb G Bb to be left alone as III and expect En G# Bn to become V/vi.

Now that I see that out loud, I actually wouldn't expect a diminished chord on vi to be tonicized at all, so I would probably drop this swap entirely. Reactions?

jacobtylerwalls (commented):
Had some music theory questions here. Let me know if you have reactions or find my comments off base. Thanks, Mark!

music21/roman.py
```diff
@@ -770,6 +769,46 @@ def romanNumeralFromChord(chordObj,
...
+
+    The full list of supported swaps is:
+
+    * in both major and minor: II to V/V;
+
+    * for major key only: III to V/vi and VI to V/ii;
+
+    * in minor: I to V/iv, III to V/VI, and IV to V/VII (NB: no flat).
+
+    As always, those minor cases are debatable in relation to the minor type.
```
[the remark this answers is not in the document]
MarkGotham: Exactly so -- thanks.

music21/roman.py
```diff
@@ -832,6 +871,17 @@ def romanNumeralFromChord(chordObj,
...
+    secondaryDominantSwapsMajor = {
+        'II': 'V',  # II > V/V
+        'III': 'vi',  # III > V/vi
+        'VI': 'ii',  # VI > V/ii
+    }
+    secondaryDominantSwapsMinor = {
+        'I': 'iv',  # Debatable
+        'II': 'V',  # Only one shared with major
+        'III': 'VI',  # NB upper (VI)
+        'IV': 'VII',  # sic, no flat (bVII) by default
```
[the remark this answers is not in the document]
MarkGotham: Thanks. Yes this relates to the m21 convention for 6th / 7th degree handling by `quality` which (as you know) sets `VII` on the natural, rather than harmonic minor seventh (e.g. Bb rather than B in c). This is consistent within m21 (or at least it should be!) but is not with all other code libraries / schools of thought -- hence the reminder. Perhaps 'no flat needed' is clearer?

MarkGotham:
Thanks for the review @jacobtylerwalls !

Clearly the main issue here is our familiar friend -- how to handle 6th and 7th degrees in minor.

We might want to take a cautious view and try to hit all and only the very clear cut cases, perhaps limiting ourselves to:

- Major: II, III, VI
- Minor: II only?

Then again, there's no doubt that the default for `preferSecondary` will be `False`, so that potentially gives the argument a bit more creative freedom -- if a user actively choses to set it, then they should expect some secondaries!

Finally (probably one for a separate PR), do dominant sevenths change our view of what's clear cut? There's a good case to be made for any chord that `.isDominantSeventh` to return as V7/x, or at least any for which x would be diatonic to the original key (avoiding `/#x`) and also a major or minor triad (avoiding `/vio` as you mention). If so, we'd have to consider whether we're content to work with the fact that `.isDominantSeventh` (via `isSeventhOfType`) requires 'correct' spelling (via `isSeventh`).

music21/roman.py
MarkGotham: Indeed! As below. Thanks.

jacobtylerwalls:
> There's a good case to be made for any chord that .isDominantSeventh to return as V7/x, or at least any for which x would be diatonic to the original key (avoiding /#x) and also a major or minor triad (avoiding /vio as you mention).

Sounds good to me. That helps distinguish between VII and V7/III or III and V7/VI. If we had that I would throw in I7 (dominant seventh) -> V7/IV (avoidable by not using the keyword).

> There's a good case to be made for any chord that .isDominantSeventh to return as V7/x, or at least any for which x would be diatonic to the original key (avoiding /#x) and also a major or minor triad (avoiding /vio as you mention).

Seems consistent, no?

mscuthbert:
Sorry I let this drop so long.  Returning to stalled PRs this month.

MarkGotham:
Hi @mscuthbert , @jacobtylerwalls

Thanks for the notes on this.

Returning to this after a while, I think this PR is now
- much better ;)
- getting a good balance of more or fewer changes (see docs -- hard to do without having the context of chords before/after)
- free of dodgy bugs except the definitely separate matter reported in #1450. For present purposes this is very minor (excuse the pun ... ). Again, it's hard to say whether something "should be" `V/IV` or `V/iv` without the context of what follows. And this will pick up any improvements we make to that other issue automatically.

MarkGotham:
What do you think? Musical logic good now? Code could perhaps:
- re-factor to another (perhaps hidden) method like `_evaluateSecondary()`
- use that to de-indent the nesting (i.e., `return` at the first sign of irrelevance)

Thanks for your thoughts. This PR has gathered dust for nearly 2 years now so let's close soon one way or another. The functionality is a long promised by music21 and never previously implemented, but (as always), I'm happy to implement this stuff either for music21 or externally as you prefer so feel free to decline.

music21/roman.py
```diff
@@ -937,6 +936,110 @@ def romanNumeralFromChord(
...
+
+    In C major it does come through as a secondary
+
+    >>> rn = roman.romanNumeralFromChord(cd, 'C', preferSecondaryDominants=True)
+    >>> rn.figure
+    'V65/V'
+
+    "German Augmented sixth" chords are left intact, without change.
+    This is thanks to the constraints on
+    spelling and on the root of the secondry degree.
```
jacobtylerwalls, line 992:
```suggestion
    spelling and on the root of the secondary degree.
```

music21/roman.py
```diff
@@ -937,6 +936,110 @@ def romanNumeralFromChord(
...
+    Let's check that with a dominant seventh spelling and minor key context:
+
+    >>> cd = chord.Chord('Ab4 C5 Eb5 Gb5')
+    >>> rn = roman.romanNumeralFromChord(cd, 'c', preferSecondaryDominants=True)
+    >>> rn.figure
+    'bVIb753'
+
+    So that's a context in which the root is diatonic,
+    but the possible secondary root is not.
+    No let's do the opposite case with a root that is not diatonic
```
jacobtylerwalls, line 1008:
```suggestion
    Now let's do the opposite case with a root that is not diatonic
```

music21/roman.py
```diff
@@ -1111,6 +1214,19 @@ def romanNumeralFromChord(
         elif nationalityStart in ('Fr', 'Sw'):
             keyObj = _getKeyFromCache(chordObj.seventh.name.lower())
 
+    if preferSecondaryDominants:
+        if stepRoman != 'V':  # ignore if already "primary" dominant
+            if chordObj.isDominantSeventh() or chordObj.isMajorTriad():
```
jacobtylerwalls, line 1219: Just a cosmetic suggestion, but you don't have else branches, reduce indentation with:
```suggestion
    if (
        preferSecondaryDominants:
        and stepRoman != 'V':  # ignore if already "primary" dominant
        and (chordObj.isDominantSeventh() or chordObj.isMajorTriad())
    ):

File: music21/roman.py [hunk 11]
```diff
@@ -1111,6 +1214,19 @@ def romanNumeralFromChord(
...
+                possibleSecondaryTonic = chordObj.root().transpose('P4').name
+                degree = keyObj.getScaleDegreeFromPitch(possibleSecondaryTonic)
+                if degree:  # None if not in chord
+                    secondaryAsRoman = keyObj.romanNumeral(degree).romanNumeralAlone  # NB
+                    primaryFigure = romanNumeralFromChord(chordObj,
+                                                          key.Key(possibleSecondaryTonic),
+                                                          # Note preferSecondaryDominants False
```
jacobtylerwalls, line 1226: Make this more explicit by passing False rather than writing a comment.

jacobtylerwalls (changes requested):
I'm seeing this:
```python
c = chord.Chord("A C# E")
roman.romanNumeralFromChord(c, key.Key("C"), preferSecondaryDominants=True)
```
```
<music21.roman.RomanNumeral V6/II in C major>
```
But I would have expected lowercase ii.

music21/roman.py
MarkGotham: Thanks.

MarkGotham:
Thanks @jacobtylerwalls.

Wrt your test case leading to ... 
`<music21.roman.RomanNumeral V6/II in C major>`
... I agree with the issues and report it in #1450.

I've also just noticed a similar bug in the existing workings of `roman.romanNumeralFromChord` which I'll report on that same issue, or a new one.

For now, I've refactored a (verbose but reliable) workaround. See line 1035 for the relevant doc test.

I want to do more with this, adding questions of context and the parallel case of `vii/x` but it's clear that music21 is not really up for new functionality these days / at the moment so I'll implement that (along with a version of this) in When in Rome soon. All the same, (for the avoidance of doubt), music21 is still welcome to us and merge this limited PR if you so choose.

music21/roman.py
```diff
@@ -1111,6 +1214,28 @@ def romanNumeralFromChord(
...
+        degree = keyObj.getScaleDegreeFromPitch(possibleSecondaryTonic)
+        if degree:  # None if not in chord
+            # Note: This is super verbose but the alternatives are bug-ridden. See #1450
+            fake2ndChordDegrees = [degree, degree + 2 % 8, degree + 4 % 8]
+            fake2ndChordPitches = keyObj.pitchesFromScaleDegrees(fake2ndChordDegrees)
+            fake2ndChord = chord.Chord(fake2ndChordPitches)
+            secondaryAsRoman = romanNumeralFromChord(fake2ndChord,
+                                                     keyObj,
+                                                     preferSecondaryDominants=False
+                                                     ).romanNumeralAlone
```
jacobtylerwalls, line 1232: Thanks for hacking something up. I'm just wondering, if we merge #1481 soon, which I think can happen, can this just use `RomanNumeral()` rather than `romanNumeralFromChord()`? That way, we neither have to wait for #1480, nor add this workaround.

jacobtylerwalls: btw #1481 was merged.

MarkGotham: Yup, that now works. Thanks for the swift work.

MarkGotham:
Fail = some issue on scale in python 3.10 ... ?
Any idea @jacobtylerwalls ?

mscuthbert:
> Fail = some issue on scale in python 3.10 ... ?
> Any idea @jacobtylerwalls ?

You won (lost) the probability lottery.  It's something that tests that our random selection of non-deterministic scales is working properly and it's design to fail no less than 1 time in 10,000 or so.  You got a failure.  Closing and reopening to get a clear board, then will try to review in next couple of days.

MarkGotham:
> > Fail = some issue on scale in python 3.10 ... ?
> > Any idea @jacobtylerwalls ?
> 
> You won (lost) the probability lottery. It's something that tests that our random selection of non-deterministic scales is working properly and it's design to fail no less than 1 time in 10,000 or so. You got a failure. Closing and reopening to get a clear board, then will try to review in next couple of days.

Haha. This sounds familiar. I'd say that I've had that before ... but I've not put in 2x 10,000 PRs ... so maybe I'm just a lucky (unlucky) kind of guy ;)

mscuthbert:
Played around with this and it looks really good.  The only thing that I could find that someone could quibble about is whether F dom 7 in eb-minor is 'V7/v' or 'V7/V' -- or when the chord's root is d5 away from the tonic, whether it should be V7/vii or V7/viio -- but no matter this is still an improvement and other things can be fixed/patched later.

(I am going to change "spelt" to "spelled" consistent with US spelling, but not a criticism, just keeping consistency)

mscuthbert (approved):

mscuthbert:
Maybe it's closer to 1 in 1000?  But no -- I do at least 1000 commits a year and I've only seen it happen once.

It used to be easier to trigger but I changed the threshold at some point.  Probably better to make the threshold even higher for the future.  OTOH, it reminds me that this feature exists and that it can be used for some great automatic composition with state machines.

PR #798, opened 2021-01-30
Don't wipe chordKindStr value on already created NoChord object

### LINKED ISSUE #797: NoChord import sets chordKindStr with a value that prevents export
**music21 version**

`6.6.1`   

**Problem summary**
A harmony tag like `<harmony><kind text="">none</kind>` will correctly yield a `harmony.NoChord` object, but because the text attribute is `""`, `xmlToChordSymbol` will update the `chordKindStr` attribute with a value that can't be exported.

**Steps to reproduce**

From scratch instead of loading a file.
```
nc = harmony.NoChord()
nc.chordKindStr = ''
nc.write()
```

```
  File "/Users/jwalls/music21/music21/musicxml/m21ToXml.py", line 4584, in noChordToXml
    assert cs.chordKindStr not in (None, '')
AssertionError
```

**More information**
Refs #786

jacobtylerwalls:
Fixes #797


mscuthbert:
Thanks!  Good fix!

PR #799, opened 2021-01-30
Add part and measure context to MusicXML exceptions

jacobtylerwalls:
There were two existing efforts to do this--`measureParsingError()` which wasn't tested/functioning, and a try block on `xmlToSimpleNote()`--but it was simple enough to just abstract this to the exception classes.


mscuthbert:
This is brilliant -- yes!  What I wanted to happen always.  Thank you!

PR #803, opened 2021-02-01
improve safety of setElementOffset and various coreElementsChanged fixes

jacobtylerwalls:
**Motivation**
`setElementOffset()` appears to require `coreElementsChanged()` to be run afterward. This isn't mentioned in the docs, and it isn't observed everywhere in the codebase, possibly causing bugs.

<s>Most of those bugs are caught with prophylactic calls to `coreElementsChanged()` downstream. However, calling this unnecessarily is a performance problem that defeats the original purpose of the core methods.

This PR uses the existing attribute `streamStatus._dirty` to track the necessity of calling `coreElementsChanged` (and this was how I discovered that setElementOffset was dangerous).</s>

**Steps to reproduce**
The example that was here no longer applies, so I removed it.

### DIFF
--- a/music21/base.py
+++ b/music21/base.py
@@ -899,14 +899,14 @@ class Music21Object(prebase.ProtoM21Object):
 
         and
 
-            stream1.setElementOffset(n1, 20)
+            stream1.coreSetElementOffset(n1, 20)
 
         >>> import music21
         >>> aSite = stream.Stream()
         >>> aSite.id = 'aSite'
         >>> a = music21.Music21Object()
         >>> aSite.insert(0, a)
-        >>> aSite.setElementOffset(a, 20)
+        >>> aSite.coreSetElementOffset(a, 20)
         >>> a.setOffsetBySite(aSite, 30)
         >>> a.getOffsetBySite(aSite)
         30.0
@@ -923,7 +923,7 @@ class Music21Object(prebase.ProtoM21Object):
         0.0
         '''
         if site is not None:
-            site.setElementOffset(self, value)
+            site.coreSetElementOffset(self, value)
             if site is self.activeSite:
                 self._activeSiteStoredOffset = value  # update...
         else:
@@ -2177,7 +2177,7 @@ class Music21Object(prebase.ProtoM21Object):
             offset = value.quarterLength
 
         if self.activeSite is not None:
-            self.activeSite.setElementOffset(self, offset)
+            self.activeSite.coreSetElementOffset(self, offset)
         else:
             self._naiveOffset = offset
 
@@ -4002,7 +4002,7 @@ class Test(unittest.TestCase):
         # still have activeSite
         self.assertEqual(a.activeSite, b)
         # now the offset returns the value for the current activeSite
-        # b.setElementOffset(a, 40.0)
+        # b.coreSetElementOffset(a, 40.0)
         self.assertEqual(a.offset, 40.0)
 
         # assigning a activeSite to None
@@ -4403,7 +4403,7 @@ class Test(unittest.TestCase):
         b1 = bar.Barline()
         s.append(n1)
         self.assertEqual(s.highestTime, 30.0)
-        s.setElementOffset(b1, 'highestTime', addElement=True)
+        s.coreSetElementOffset(b1, 'highestTime', addElement=True)
 
         self.assertEqual(b1.getOffsetBySite(s), 30.0)
 
--- a/music21/base.py
+++ b/music21/base.py
@@ -2467,8 +2467,7 @@ class Music21Object(prebase.ProtoM21Object):
         for s in self.sites.get():
             if hasattr(s, 'coreElementsChanged'):
                 # noinspection PyCallingNonCallable
-                if s.streamStatus._dirty:
-                    s.coreElementsChanged(updateIsFlat=False, keepIndex=True)
+                s.coreElementsChanged(updateIsFlat=False, keepIndex=True)
 
     def _getPriority(self):
         return self._priority
@@ -3593,7 +3592,6 @@ class Music21Object(prebase.ProtoM21Object):
         self.duration = mm.secondsToDuration(value)
         for s in self.sites.get(excludeNone=True):
             if self in s.elements:
-                s.streamStatus._dirty = True
                 s.coreElementsChanged()  # highest time is changed.
 
     seconds = property(_getSeconds, _setSeconds, doc='''
--- a/music21/base.py
+++ b/music21/base.py
@@ -2467,7 +2467,8 @@ class Music21Object(prebase.ProtoM21Object):
         for s in self.sites.get():
             if hasattr(s, 'coreElementsChanged'):
                 # noinspection PyCallingNonCallable
-                s.coreElementsChanged(updateIsFlat=False, keepIndex=True)
+                if s.streamStatus._dirty:
+                    s.coreElementsChanged(updateIsFlat=False, keepIndex=True)
 
     def _getPriority(self):
         return self._priority
@@ -3592,6 +3593,7 @@ class Music21Object(prebase.ProtoM21Object):
         self.duration = mm.secondsToDuration(value)
         for s in self.sites.get(excludeNone=True):
             if self in s.elements:
+                s.streamStatus._dirty = True
                 s.coreElementsChanged()  # highest time is changed.
 
     seconds = property(_getSeconds, _setSeconds, doc='''
--- a/music21/base.py
+++ b/music21/base.py
@@ -899,14 +899,14 @@ class Music21Object(prebase.ProtoM21Object):
 
         and
 
-            stream1.coreSetElementOffset(n1, 20)
+            stream1.setElementOffset(n1, 20)
 
         >>> import music21
         >>> aSite = stream.Stream()
         >>> aSite.id = 'aSite'
         >>> a = music21.Music21Object()
         >>> aSite.insert(0, a)
-        >>> aSite.coreSetElementOffset(a, 20)
+        >>> aSite.setElementOffset(a, 20)
         >>> a.setOffsetBySite(aSite, 30)
         >>> a.getOffsetBySite(aSite)
         30.0
@@ -923,7 +923,7 @@ class Music21Object(prebase.ProtoM21Object):
         0.0
         '''
         if site is not None:
-            site.coreSetElementOffset(self, value)
+            site.setElementOffset(self, value)
             if site is self.activeSite:
                 self._activeSiteStoredOffset = value  # update...
         else:
@@ -2177,7 +2177,7 @@ class Music21Object(prebase.ProtoM21Object):
             offset = value.quarterLength
 
         if self.activeSite is not None:
-            self.activeSite.coreSetElementOffset(self, offset)
+            self.activeSite.setElementOffset(self, offset)
         else:
             self._naiveOffset = offset
 
@@ -4002,7 +4002,7 @@ class Test(unittest.TestCase):
         # still have activeSite
         self.assertEqual(a.activeSite, b)
         # now the offset returns the value for the current activeSite
-        # b.coreSetElementOffset(a, 40.0)
+        # b.setElementOffset(a, 40.0)
         self.assertEqual(a.offset, 40.0)
 
         # assigning a activeSite to None
@@ -4403,7 +4403,7 @@ class Test(unittest.TestCase):
         b1 = bar.Barline()
         s.append(n1)
         self.assertEqual(s.highestTime, 30.0)
-        s.coreSetElementOffset(b1, 'highestTime', addElement=True)
+        s.setElementOffset(b1, 'highestTime', addElement=True)
 
         self.assertEqual(b1.getOffsetBySite(s), 30.0)
 
--- a/music21/harmony.py
+++ b/music21/harmony.py
@@ -2028,7 +2028,7 @@ class ChordSymbol(Harmony):
         if inversionNum not in (0, None):
             for p in pitches[0:inversionNum]:
                 p.octave = p.octave + 1
-                # Repeat if 9th/11th/13th chord in 3rd inversion or greater
+                # Repeat if 9th/11th/13th chord in 4th inversion or greater
                 if inversionNum > 3:
                     p.octave = p.octave + 1
 
--- a/music21/musicxml/xmlToM21.py
+++ b/music21/musicxml/xmlToM21.py
@@ -925,8 +925,7 @@ class MusicXMLImporter(XMLParserBase):
         for sp in rm:
             self.spannerBundle.remove(sp)
 
-        if s.streamStatus._dirty:
-            s.coreElementsChanged()
+        s.coreElementsChanged()
         s.definesExplicitSystemBreaks = self.definesExplicitSystemBreaks
         s.definesExplicitPageBreaks = self.definesExplicitPageBreaks
         for p in s.parts:
@@ -1509,8 +1508,7 @@ class PartParser(XMLParserBase):
         for sp in rm:
             self.spannerBundle.remove(sp)
         # s is the score; adding the part to the score
-        if self.stream.streamStatus._dirty:
-            self.stream.coreElementsChanged()
+        self.stream.coreElementsChanged()
 
         if self.maxStaves > 1:
             self.separateOutPartStaves()
@@ -1639,8 +1637,7 @@ class PartParser(XMLParserBase):
             self.xmlMeasureToMeasure(mxMeasure)
 
         # self.removeEndForwardRest()
-        if part.streamStatus._dirty:
-            part.coreElementsChanged()
+        part.coreElementsChanged()
 
     #     def removeEndForwardRest(self):
     #         '''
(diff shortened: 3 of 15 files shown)


mscuthbert:
I'm not sure that this is the direction I'd like to go in -- music21 used to have flags like this (back in the pre-v2 system) and they just caused a lot of errors because every time a new method was added, it had to deal with the notion of whether the stream was `._dirty` (I'd probably call it more explicitly, `coreUpdateRequired` if we did go this route).  Sorting, offsets, etc. were all a mess, so this is something I'm very weary of from experience.

My feeling is that the only systems that could leave the system in a `dirty` state and the only systems that should check for `dirty` status should be those in stream.core and that are prefaced with `core` -- so I'm okay with a coreSetElementOffset() function that could be called by other systems.  

And though I'd prefer not to, I could be convinced to go with a flag `coreUpdateRequired` that other core systems could use to see if they can bypass calling `coreElementsChanged`.  But better for systems to understand themselves why they might need to call coreUpdate.  For instance, I can already see a bug in ABCTranslate that line 189 uses a coreAppend but 201 uses a coreInsert w/o a coreElementsChanged() -- probably append then update is okay, but update then append is not, but this is untested.

For instance, `informSites()` needs to always call `coreElementsChanged()` no matter if the stream is dirty or not, because it's the function that tells a stream -- "you thought you weren't dirty but you are!" since this can happen when an element's priority is changed which tells the sites that they need to re-sort the Stream.

I tried a lot of "justInTime" sorting/etc. but it was a mess.

What is the particular performance bottleneck that this code removes?  And how much of a performance improvement is it compared to just adding coreElementsChange() to setOffsetBySite?

jacobtylerwalls:
Sure, that makes sense. I didn't state the motivation very well--the primary thing wasn't performance, it was understanding which operations are dangerous, since I spotted some conflicting information to that effect--and once I had the flags in there to figure that out, I was ambivalent about leaving them in. Happy to remove them. What will be left, I think, are a handful of doc improvements or code fixes.

Re: your last suggestion, we can't add coreElementsChanged to setElementOffset, because coreInsert and coreAppend use setElementOffset, so that would defeat the whole point. I take it `setElementOffset` needs to become `coreSetElementOffset` and move to stream.core?

music21/stream/__init__.py
```diff
@@ -1611,6 +1606,9 @@ def _deepcopySubclassable(self, memo=None, ignoreAttributes=None, removeFromIgno
                 # noinspection PyArgumentList
                 new.coreStoreAtEnd(copy.deepcopy(e, memo))
 
+        # shouldn't this be required? (coreStoreAtEnd calls coreSetElementOffset)
+        # new.coreElementsChanged()
```
mscuthbert, line 1610: yes! almost certainly!

music21/stream/__init__.py
```diff
@@ -1668,8 +1666,19 @@ def setElementOffset(
...
         '''
+        DEPRECATED!
+        MOVED in v6.7 to :meth:`~music21.stream.core.coreSetElementOffset`
+        This backwards-compatible shim may be removed in the future.
+
         Sets the Offset for an element, very quickly.
+        Caller is responsible for calling :meth:`~music21.stream.core.coreElementsChanged`
+        afterward.
 
+        See docs on :meth:`~music21.stream.core.coreSetElementOffset`
```
mscuthbert, line 1677: Question that I really should know the answer -- is there any other non-core way to do something like this on a stream:

```
for i, el in enumerate([n1, n2, n3, n4]):
     s.setElementOffset(el, i)
```
I know we can do:

```
for i, el in enumerate([n1, n2, n3, n4]):
     el.setOffsetBySite(s, i)
```
but that is an expensive operation.

So, I don't think we should deprecate setElementOffset, only deprecate the addElement keyword (definitely) and probably setActiveSite keyword as well.  Then this method would simply be:

```
    def setElementOffset(
        self,
        element: base.Music21Object,
        offset: Union[int, float, Fraction, str],
        *,
        addElement=False,  # deprecated
        setActiveSite=True  # deprecated
     ):
     
         self.coreSetElementOffset(element,
                                   offset,
                                   addElement=addElement,  # to become always False in v7
                                   setActiveSite=setActiveSite)  # to become always True in v7
         self.coreElementsChanged(updateIsFlat=False)  # might change sorting, but not flatness.  Maybe other things can be False too.
```

When deprecating an entire function, there is a common.decorators.deprecated decorator that I put on the function saying when it will be deprecated and disappear.  It handles all the warnings, etc.  But just store that away because we won't deprecate.

music21/musicxml/xmlToM21.py
```diff
@@ -5661,6 +5651,7 @@ def updateVoiceInformation(self):
                 self.voicesById[v.id] = v
             self.useVoices = True
 
+            self.stream.coreElementsChanged()
```
mscuthbert, line 5654: thanks!

music21/base.py
```diff
@@ -899,14 +899,14 @@ def setOffsetBySite(self,
 
         and
 
-            stream1.setElementOffset(n1, 20)
+            stream1.coreSetElementOffset(n1, 20)
```
mscuthbert, line 902: see below for not deprecating setElementOffset, so we can revert.

music21/base.py
```diff
@@ -899,14 +899,14 @@ def setOffsetBySite(self,
...
 
         >>> import music21
         >>> aSite = stream.Stream()
         >>> aSite.id = 'aSite'
         >>> a = music21.Music21Object()
         >>> aSite.insert(0, a)
-        >>> aSite.setElementOffset(a, 20)
+        >>> aSite.coreSetElementOffset(a, 20)
```
mscuthbert, line 909: no need.

music21/base.py
```diff
@@ -923,7 +923,7 @@ def setOffsetBySite(self,
         0.0
         '''
         if site is not None:
-            site.setElementOffset(self, value)
+            site.coreSetElementOffset(self, value)
```
mscuthbert, line 926: this should definitely be the normal setElementOffset() -- so site is left intact

music21/base.py
```diff
@@ -2177,7 +2177,7 @@ def _setOffset(self, value):
             offset = value.quarterLength
 
         if self.activeSite is not None:
-            self.activeSite.setElementOffset(self, offset)
+            self.activeSite.coreSetElementOffset(self, offset)
```
mscuthbert, line 2180: revert -- use non-core version.

music21/base.py
```diff
@@ -4002,7 +4002,7 @@ def assignActiveSite(aa, bb):
         # still have activeSite
         self.assertEqual(a.activeSite, b)
         # now the offset returns the value for the current activeSite
-        # b.setElementOffset(a, 40.0)
+        # b.coreSetElementOffset(a, 40.0)
```
mscuthbert, line 4005: rev -- we can test core... in Stream

music21/base.py
```diff
@@ -4403,7 +4403,7 @@ def testHighestTime(self):
         b1 = bar.Barline()
         s.append(n1)
         self.assertEqual(s.highestTime, 30.0)
-        s.setElementOffset(b1, 'highestTime', addElement=True)
+        s.coreSetElementOffset(b1, 'highestTime', addElement=True)
```
mscuthbert, line 4406: rev.

music21/harmony.py
```diff
@@ -2028,7 +2028,7 @@ def _updatePitches(self):
         if inversionNum not in (0, None):
             for p in pitches[0:inversionNum]:
                 p.octave = p.octave + 1
-                # Repeat if 9th/11th/13th chord in 3rd inversion or greater
+                # Repeat if 9th/11th/13th chord in 4th inversion or greater
```
mscuthbert, line 2031: thanks!

music21/stream/__init__.py
```diff
@@ -2488,12 +2470,10 @@ def insertAndShift(self, offsetOrItemOrList, itemOrNone=None):
                     #                         'o+shiftPos', o+shiftPos])
 
                     # need original offset, shiftDur, plus the distance from the start
-                    self.setElementOffset(e, o + shiftPos)
+                    self.coreSetElementOffset(e, o + shiftPos)
```
mscuthbert, line 2473: outside of the for loop, this should have a coreElementsChanged(), before the insert() statement below, because insert() assumes (or is allowed to assume) that the stream is not dirty.

music21/stream/__init__.py
```diff
@@ -6691,7 +6670,6 @@ def updateEndMatch(nInner):
             # Recurse rather than depend on the containers being Measures
             # https://github.com/cuthbertLab/music21/issues/266
             returnObj.remove(nTarget, recurse=True)
-        returnObj.coreElementsChanged()
```
mscuthbert, line 6694: good.  -- though eventually this could get a speedup by using core commands to remove,...

...well, actually this should be rewritten eventually to use trees.

music21/stream/__init__.py
```diff
@@ -8335,7 +8313,6 @@ def scaleDurations(self, amountToScale, *, inPlace=False):
         for e in returnObj.recurse().getElementsNotOfClass('Stream'):
             e.duration = e.duration.augmentOrDiminish(amountToScale)
 
-        returnObj.coreElementsChanged()
```
mscuthbert, line 8338: Good.  Just putting a comment here where I can:  looking at the line 8322/8300 "We do not retain durations in any circumstance; if inPlace=True, two deepcopies are done" should be if "inPlace=False, two deepcopies of each duration are done" -- though actually deepcopying a duration is no longer slow, and I know an easy speedup since the v2.0 duration model.  (But no one uses this!)

music21/stream/__init__.py
```diff
@@ -8552,6 +8528,10 @@ def bestMatch(target, divisors):
                             and signedError != 0):
                         e.editorial.quarterLengthQuantizationError = signedError
 
+            # end for e in ._elements
+            # ran coreSetElementOffset
+            useStream.coreElementsChanged()
```
mscuthbert, line 8533: thanks!  can include `updateIsFlat=False` (but not isSorted, since the quantization might move elements.

music21/stream/__init__.py
```diff
@@ -8704,7 +8684,6 @@ def sliceByGreatestDivisor(self, *, addTies=True, inPlace=False):
             # call on component measures
             for m in returnObj.getElementsByClass('Measure'):
                 m.sliceByGreatestDivisor(addTies=addTies, inPlace=True)
-            returnObj.coreElementsChanged()
```
mscuthbert, line 8707: oh my...this routine should be moved off of stream to somewhere else in v7.  :-)  No one ever uses this.

music21/stream/__init__.py
```diff
@@ -10567,7 +10556,6 @@ def flattenUnnecessaryVoices(self, *, force=False, inPlace=False):
                     returnObj.coreInsert(shiftOffset + e.getOffsetBySite(v), e)
                 returnObj.remove(v)
 
-        returnObj.coreElementsChanged()
```
mscuthbert, line 10570: We need this, because we are doing a coreInsert above.  But it can move to the end of the `if len(flatten) == 1` line.

music21/stream/__init__.py
```diff
@@ -11024,8 +11012,6 @@ def activateVariants(self, group=None, *, matchBySpan=True, inPlace=False):
         # Now fix measure numbers given the saved information
         returnObj._fixMeasureNumbers(deletedMeasures, insertedMeasures)
 
-        # have to clear cached variants, as they are no longer the same
-        returnObj.coreElementsChanged()
```
mscuthbert, line 11028: If you don't mind, leave this, unless you're absolutely clear on the variant code -- it's old code and not super well documented or tested, so there might be some core-leaking there.

music21/stream/__init__.py
```diff
@@ -11882,7 +11871,7 @@ def makeVariantBlocks(self):
                     r.style.hideObjectOnPrint = True
                     for el in cV._stream:
                         oldOffset = el.getOffsetBySite(cV._stream)
-                        cV._stream.setElementOffset(el, oldOffset + shiftOffset)
+                        cV._stream.coreSetElementOffset(el, oldOffset + shiftOffset)
                     cV.insert(0.0, r)
```
mscuthbert, line 11875: needs a coreElementsChanged() before the cV.insert(0.0, r) line.

music21/stream/__init__.py
```diff
@@ -13349,10 +13338,6 @@ def makeNotation(self,
                                inPlace=True,
                                bestClef=bestClef,
                                **subroutineKeywords)
-            # note: while the local-streams have updated their caches, the
-            # containing score has an out-of-date cache of flat.
-            # this, must call elements changed
-            returnStream.coreElementsChanged()
         else:  # call the base method
```
mscuthbert, line 13341: Can we leave this back in?  I remember that this was a very subtle error that was hard to diagnose, and I'd rather have it than lose another day trying to figure out what happened.  Can add "Probably not needed."

music21/stream/core.py
```diff
@@ -191,9 +253,8 @@ def coreElementsChanged(
                 'coreElementsChanged should not be triggered on an immutable stream'
             )
 
-        if memo is None:
-            memo = []
-        memo.append(id(self))
+        if memo is not None:
+            memo.append(id(self))
```
mscuthbert, line 257: Actually, leave this... but should put after the `if` statement, `if id(self) in memo: return` and...

mscuthbert: in line `livingSite.coreElementsChanged()` we should be passing in the memo at least and probably eventually `updateIsFlat` and `clearIsSorted` -- this should be a significant speedup.

music21/stream/iterator.py
```diff
@@ -250,7 +250,20 @@ def __getattr__(self, attr):
...
+        srcStream is accessible, but not with "__getattr__", which joblib uses
+
+        >>> s.notes.srcStream is s
+        True
+        >>> s.notes.__getattr__('srcStream') is None
+        True
         '''
+        # Prevent infinite loop in feature extractor task serialization
+        if attr == 'srcStream':
+            return None
```
mscuthbert, line 265: good catch.  This might cause other bugs along the way, but it's a nice fix for now.

music21/stream/makeNotation.py
```diff
@@ -522,6 +522,7 @@ def makeMeasures(
             v = stream.Voice()
             v.id = voiceIndex  # id is voice index, starting at 0
             m.coreInsert(0, v)
+        m.coreElementsChanged()
```
mscuthbert, line 525: `if(voiceCount): m.coreElementsChanged()` to avoid calling in the normal case where there are no voices. (which is why I didn't discover this bug before)

music21/stream/makeNotation.py
```diff
@@ -619,8 +622,6 @@ def makeMeasures(
     for sp in spannerBundleAccum:
         post.append(sp)
 
-    post.coreElementsChanged()
```
mscuthbert, line 622: Ah -- we should find a way to try putting coreInsert above again now that you've likely fixed the bug.  This is a very slow operation that does not need to be.

mscuthbert (changes requested):
Great work!  I want to keep setElementOffset as a "public" version of coreSetElementOffset, so read the comments on that method first and I think it will make the other parts make more sense.

mscuthbert:
For some reason I decided to do a "core strength" routine in my Yoga today.

music21/stream/__init__.py
jacobtylerwalls: I had a question about this and another -- `stream.remove()` and `stream.insert()` both run `coreElementsChanged` at the end, so the idea in this particular instance is that it's duplicative.

However--looking at the call in `remove`, `clearIsSorted` is set to False -- so, I take it there may be a reason to call it after all, with the default `clearIsSorted=True`. Is that right?

music21/stream/makeNotation.py
jacobtylerwalls: We would have to call it on `m`, but since `m` is refreshed in every loop iteration, it would be impossible / trivial (take your pick).

music21/stream/core.py
jacobtylerwalls: Wasn't sure if "leave this" meant you wanted memo.append(id(self)) to still run even when memo is None. Currently we do nothing with the memo.

music21/stream/iterator.py
jacobtylerwalls: I didn't have much of a choice, because it was blowing up otherwise. I think it was the call to coreElementsChanged I removed from stripTies. I don't know how that exposed more functionality (or caused the serialization to happen differently).

mscuthbert:
Don't worry about coverage -- will merge if my latest commit passes.

jacobtylerwalls:
Got it, then in that case you probably need the original `if memo is None`

mscuthbert:
OH!  I just figured out why there can be an infinite loop now:

```
s = stream.Stream()
iter = s.iter
s2 = iter.srcStream
s2 is s
iter2 = s2.iter
iter2 is not iter
```

This might have come in around 6.1?  Where iterators are not reused anymore.

Should rename in v7 `iter` to `iter()` -- I had too much property love when I first switched to Python.

PR #810, opened 2021-02-04
Import and export multi-key piano staves

### LINKED ISSUE #808: Part staffs with multiple keys
**music21 version**

`6.7`   

**Problem summary**

41b and 41c in musicxml test are both failing -- check to see if it is music21 or MuseScore?

jacobtylerwalls:
Fixes #808 

LMK if you prefer that I not use `.keySignature` as a lookup, but it seemed more convenient than recursing and checking if not None, etc.

### DIFF
--- a/music21/musicxml/partStaffExporter.py
+++ b/music21/musicxml/partStaffExporter.py
@@ -441,17 +441,25 @@ class PartStaffExporterMixin:
         ...
         </measure>
         '''
+        # Is this source multi-key?
+        initialM21Key: Optional[KeySignature] = None
+        multiKey: bool = False
+        for ps in group:
+            if initialM21Key is None:
+                for ks in ps.recurse().getElementsByClass(KeySignature):
+                    initialM21Key = ks
+                    break
+            else:
+                firstKeySubsequentStaff = None
+                for ks in ps.recurse().getElementsByClass(KeySignature):
+                    firstKeySubsequentStaff = ks
+                    break
+                if firstKeySubsequentStaff != initialM21Key:
+                    multiKey = True
+                    break
+
         initialPartStaffRoot: Optional[Element] = None
         mxAttributes: Optional[Element] = None
-        initialM21Key: Optional[KeySignature] = group[0].flat.keySignature
-        multiKey: bool = False
-
-        for ps in group[1:]:
-            subsequentKey = ps.flat.keySignature
-            if subsequentKey != initialM21Key:
-                multiKey = True
-                break
-
         for i, ps in enumerate(group):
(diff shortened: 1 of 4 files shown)


music21/stream/iterator.py
```diff
@@ -261,6 +261,7 @@ def __getattr__(self, attr):
         True
         '''
         # Prevent infinite loop in feature extractor task serialization
+        # TODO: investigate if this can be removed once iter becomes iter()
```
mscuthbert, line 264: We'll only be able to do that after .iter is no longer an alias for .iter()

music21/musicxml/partStaffExporter.py
```diff
@@ -438,6 +443,15 @@ def setEarliestAttributesAndClefsPartStaff(self, group: StaffGroup):
         '''
         initialPartStaffRoot: Optional[Element] = None
         mxAttributes: Optional[Element] = None
+        initialM21Key: Optional[KeySignature] = group[0].flat.keySignature
```
mscuthbert, line 446: Check that group is not empty.  A StaffGroup, like any spanner, can have zero items in it.  Especially if someone has been deleting parts.

mscuthbert: And yes, we should be using `.recurse().getElementsByClass(key.KeySignature)` not because of keySignature, but because `.flat` is really expensive in terms of time and memory.

Btw -- I'm going to add to StreamIterator, a `.first()` method, which returns the first item or None if there are no matching items.  This will help in a lot of places.

music21/musicxml/partStaffExporter.py
```diff
@@ -438,6 +443,15 @@ def setEarliestAttributesAndClefsPartStaff(self, group: StaffGroup):
...
+        multiKey: bool = False
+
+        for ps in group[1:]:
+            subsequentKey = ps.flat.keySignature
```
mscuthbert, line 450: same here.

mscuthbert (changes requested):
Thank you!  And, yes, unfortunately, we're going to need to use the full `getElementsByClass()` call on recurse for speed/memory/good-practice.

music21/musicxml/partStaffExporter.py
jacobtylerwalls: Yeah, safer to check the group since the function is public (I was lazily depending on empty groups being filtered out by joinableGroups() but better to be safe.)

mscuthbert:
Unfortunately there are some problems with the same Key object appearing in multiple parts.  Investigating.

mscuthbert:
Never mind -- it's from an earlier PR

PR #812, opened 2021-02-04
Make chord/voice boundary agree with quantization units

### LINKED ISSUE #809: Conform MIDI chord-vs.-voice boundary to the quantization units being used
**music21 version**

`6.7`

**Problem summary**
MIDI import is too quick to create voices and too reluctant to create chords because the "chord subbing" tolerance is different from the quantization units. The "chord subbing" is not configurable--a 64th note, regardless of the quantization units, meaning any onsets after a 64th note window jump to another voice instead of being collected onto a chord. Then when quantization runs later, it's too late--you've got voices when chords would have been appropriate, given the quantization scheme.

music21's MIDI parsing is very literal, which is great, but I think there are a couple of low-effort batteries-included improvements we could do in this area. Eventually I'll write up the other idea I have on quantization.

**Steps to reproduce**
File on #684
```
s = converter.parse('example-from-684'.mid)
s.parts[4].voices[0].show('t')
```

Before:  {5.0} <music21.note.Note E4>
After this proposal : {5.0} <music21.chord.Chord E4 C#4>

**Suggested fix**
```diff --git a/music21/midi/translate.py b/music21/midi/translate.py
index b2c8606da..e4eab20d4 100644
--- a/music21/midi/translate.py
+++ b/music21/midi/translate.py
@@ -1755,6 +1755,12 @@ def midiTrackToStream(
     i = 0
     iGathered = []  # store a list of indexes of gathered values put into chords
     voicesRequired = False
+
+    if 'quarterLengthDivisors' in keywords:
+        quarterLengthDivisors = keywords['quarterLengthDivisors']
+    else:
+        quarterLengthDivisors = defaults.quantizationQuarterLengthDivisors
+
     if len(notes) > 1:
         # environLocal.printDebug(['\n', 'midiTrackToStream(): notes', notes])
         while i < len(notes):
@@ -1779,10 +1785,15 @@ def midiTrackToStream(
                 tSub, unused_eSub = onSub
                 tOffSub, unused_eOffSub = offSub
 
-                # can set a tolerance for chordSubbing; here at 1/16th
-                # of a quarter
-                chunkTolerance = ticksPerQuarter / 16
-                if abs(tSub - t) <= chunkTolerance:
+                # let tolerance for chord subbing follow the quantization
+                if quantizePost:
+                    divisor = max(quarterLengthDivisors)
+                # fallback: 1/16 of a quarter (64th)
+                else:
+                    divisor = 16
+                chunkTolerance = ticksPerQuarter / divisor
+                # must be strictly less than the quantization unit
+                if abs(tSub - t) < chunkTolerance:
                     # isolate case where end time is not w/n tolerance
                     if abs(tOffSub - tOff) > chunkTolerance:
                         # need to store this as requiring movement to a diff
@@ -1839,10 +1850,6 @@ def midiTrackToStream(
     s.coreElementsChanged()
     # quantize to nearest 16th
     if quantizePost:
-        if 'quarterLengthDivisors' in keywords:
-            quarterLengthDivisors = keywords['quarterLengthDivisors']
-        else:
-            quarterLengthDivisors = None
         s.quantize(quarterLengthDivisors=quarterLengthDivisors,
                    processOffsets=True,
                    processDurations=True,
```

jacobtylerwalls:
Fixes #809

The test that had to be adjusted demonstrates the improvement.

```
s = converter.parse('/Users/jwalls/music21/music21/midi/testPrimitive/test09.mid', forceSource=True)
made = s.parts[0].makeMeasures()
made.measure(54).show()
```

**Before**

voice-chord-before

**After**

voice-chord-after

**Update:** fixed an existing bug that would have increased in frequency -- when setting duration on MIDI chords, only the first chord member's note off is used. Now we use the last, which I would expect to be more likely to be correct.

### DIFF
--- a/music21/midi/translate.py
+++ b/music21/midi/translate.py
@@ -512,7 +512,8 @@ def midiEventsToChord(eventList, ticksPerQuarter=None, inputM21=None):
 
     Changed in v.7 -- Uses the last DeltaTime in the list to get the end time.
     '''
-    tOn = 0
+    tOn: int = 0  # ticks
+    tOff: int = 0  # ticks
 
     if inputM21 is None:
         c = chord.Chord()
--- a/music21/midi/translate.py
+++ b/music21/midi/translate.py
@@ -501,6 +501,15 @@ def midiEventsToChord(eventList, ticksPerQuarter=None, inputM21=None):
     >>> c.duration.quarterLength
     2.0
 
+    Providing fewer than four events won't work.
+
+    >>> c = midi.translate.midiEventsToChord([dt1, me1, me2])
+    Traceback (most recent call last):
+    music21.midi.translate.TranslateException: fewer than 4 events provided to midiEventsToChord:
+    [<MidiEvent DeltaTime, t=0, track=1, channel=None>,
+        <MidiEvent NOTE_ON, t=0, track=1, channel=None, pitch=45, velocity=94>,
+        <MidiEvent NOTE_ON, t=0, track=1, channel=None, pitch=46, velocity=94>]
+
     Changed in v.7 -- Uses the last DeltaTime in the list to get the end time.
     '''
     tOn = 0
@@ -521,7 +530,7 @@ def midiEventsToChord(eventList, ticksPerQuarter=None, inputM21=None):
     # this is a format provided by the Stream conversion of
     # midi events; it pre groups events for a chord together in nested pairs
     # of abs start time and the event object
-    if isinstance(eventList, list) and isinstance(eventList[0], tuple):
+    if isinstance(eventList, list) and eventList and isinstance(eventList[0], tuple):
         # pairs of pairs
         for onPair, offPair in eventList:
             tOn, eOn = onPair
@@ -533,7 +542,7 @@ def midiEventsToChord(eventList, ticksPerQuarter=None, inputM21=None):
             v.velocityIsRelative = False  # velocity is absolute coming from
             volumes.append(v)
     # assume it is a flat list
-    else:
+    elif len(eventList) > 3:
         onEvents = eventList[:(len(eventList) // 2)]
         offEvents = eventList[(len(eventList) // 2):]
         # first is always delta time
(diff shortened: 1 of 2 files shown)


music21/midi/translate.py
```diff
@@ -535,7 +538,9 @@ def midiEventsToChord(eventList, ticksPerQuarter=None, inputM21=None):
         offEvents = eventList[(len(eventList) // 2):]
         # first is always delta time
         tOn = onEvents[0].time
-        tOff = offEvents[0].time
+        # use the off time of the last chord member
+        # -1 is the event, -2 is the delta time for the event
+        tOff = offEvents[-2].time
```
mscuthbert, line 543: Are we absolutely sure that this will always be the case?  Just want to be totally sure.  I think so given the offEvents definition.

music21/midi/translate.py
```diff
@@ -551,8 +556,8 @@ def midiEventsToChord(eventList, ticksPerQuarter=None, inputM21=None):
     if (tOff - tOn) != 0:
         ticksToDuration(tOff - tOn, ticksPerQuarter, c.duration)
     else:
-        # environLocal.printDebug(['cannot translate found midi event with zero duration:',
-        #                         eventList, c])
+        environLocal.warn(['midi chord with zero duration will be treated as grace',
+                            eventList, c])
```
mscuthbert, line 560: Good.

music21/midi/translate.py
```diff
@@ -520,9 +523,9 @@ def midiEventsToChord(eventList, ticksPerQuarter=None, inputM21=None):
     # of abs start time and the event object
     if isinstance(eventList, list) and isinstance(eventList[0], tuple):
```
mscuthbert, line 524: if isinstance(eventList, list) **and eventList** and isinstance(eventList[0], tuple):

mscuthbert: (this was broken before for empty MIDI files, but good to fix now

music21/midi/translate.py
```diff
@@ -520,9 +523,9 @@ def midiEventsToChord(eventList, ticksPerQuarter=None, inputM21=None):
...
         # pairs of pairs
-        tOff = eventList[0][1][0]
-        for onPair, unused_offPair in eventList:
+        for onPair, offPair in eventList:
             tOn, eOn = onPair
+            tOff, unused_eOff = offPair
```
mscuthbert, line 528: tOff should be initialized to 0 when tOn is because we could have an undefined variable

music21/midi/translate.py
mscuthbert: Should check that there are indeed at least 4 events in eventList.

mscuthbert (commented):
Looks really good -- I have a few Qs about empty eventLists (most predating this contribution) before merging.

PR #813, opened 2021-02-04
Don't raise CalledProcessError in .launch()

### LINKED ISSUE #804: Catch non-zero exits in SubConverter.launch?
https://github.com/cuthbertLab/music21/blob/2a3ba8d6ba759897bbd93879b70cae23ca111fd9/music21/converter/subConverters.py#L196

`Finale.exe` at least fails to return 0 when called from a python notebook, which bubbles up an annoying `CalledProcessError` even though everything worked fine. Would it be worth wrapping this in a `try:` / `except subprocess.CalledProcessError` to deal with this use case? If handling it silently seems bad, maybe we could just print a one-liner warning instead of forcing the developer to deal with the full exception?

jacobtylerwalls:
Fixes #804

### DIFF
--- a/music21/converter/subConverters.py
+++ b/music21/converter/subConverters.py
@@ -193,7 +193,7 @@ class SubConverter:
                 cmd = ('open', '-a', str(app), str(filePath))
         else:
             raise SubConverterException(f'Cannot launch files on {platform}')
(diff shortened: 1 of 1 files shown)


Meekohi:
Thank you!

Issue #816, opened 2021-02-04
Lyric Search -- overzealous on non-RE


mscuthbert:
Need to treat everything as a literal character for plain-text search.

mscuthbert:
Regression in the change in 6.7

PR #817, opened 2021-02-04
Fix some bugs that crept into lyrics

mscuthbert:
Composite lyrics with assembleText

re.escape on plaintext searches.


mscuthbert:
Fixes #816

PR #586, opened 2020-08-02
Fixes #502: Strip ties from chords only when all members have a stop tie

jacobtylerwalls:
Fixes #502 fixes #430 (duplicates)

**Before**: `stripTies()` would accept at face value a tie type of stop from a `Chord` object, even though not all of its members might participate in a tie. This would leave downstream effects in `chordify()` where verticalities would go missing.

**Now**: `stripTies()` only treats a `Chord` as the end of a tie if every member has a tie type of stop. That is, unless `matchByPitch` is used.

--

(PS -- LMK if you think linking back to the issue tracker is overkill -- seems useful in the event there is either a discussion there or evidence of something happening in the wild, but would certainly defer on this style point!)

PR #818, opened 2021-02-04
V7 stripTies signature

jacobtylerwalls:
- remove retainContainers
- default matchByPitch = True

I cribbed the doctest example for `matchByPitch=True` from your https://github.com/cuthbertLab/music21/pull/586#issuecomment-679428578


mscuthbert:
Looks good -- and I agree that matchByPitch default should be True.  Do we have a test of matchByPitch=False?

jacobtylerwalls:
Very simple one, but yes, at the end of the doc test I added--shows stripTies isn't executed despite the pitches matching.

mscuthbert:
Great!  Merging!

PR #822, opened 2021-02-05
Refactor separateOutPartStaves

mscuthbert:
* Discard the original PartStaff, and create all new ones from template.  Then assign elements to staff if it is not excluded, but the second time an element is encountered, make a copy of it -- so except for measures, etc., we make n-1 copies of duplicate elements still.

* Add tests that duplicate elements are not appearing in staff.

* template uses coreInsert methods

* fix original template code to ensure that endElements remain endElements.

* add recurse() to addGroupForElements, but do not yet turn on in parsing.


jacobtylerwalls:
Nice, at a glance this looks like a good way to avoid so many insert and remove operations in the old code.

PR #823, opened 2021-02-05
Import and export polymetric grand staffs

jacobtylerwalls:
Spotted this while we were working on templates and multi-keys.

Polymetric grand staffs were imported before templates, but then templates sort of broke it (just like keys) by letting the template copy the top meter into the bottom staff, leaving the bottom staff with two meters.

Then, export wasn't ever supported, but now it is.

### DIFF
--- a/music21/musicxml/partStaffExporter.py
+++ b/music21/musicxml/partStaffExporter.py
@@ -20,6 +20,7 @@ from xml.etree.ElementTree import Element, SubElement
 
 from music21.key import KeySignature  # for typing
 from music21.layout import StaffGroup
+from music21.meter import TimeSignature  # for typing
 from music21 import stream  # for typing
 from music21.musicxml import helpers
 from music21.musicxml.xmlObjects import MusicXMLExportException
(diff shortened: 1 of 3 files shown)


music21/musicxml/partStaffExporter.py
```diff
@@ -440,23 +443,68 @@ def setEarliestAttributesAndClefsPartStaff(self, group: StaffGroup):
...
-            else:
-                firstKeySubsequentStaff = None
-                for ks in ps.recurse().getElementsByClass(KeySignature):
-                    firstKeySubsequentStaff = ks
-                    break
-                if firstKeySubsequentStaff != initialM21Key:
-                    multiKey = True
-                    break
+
+        def isMultiAttribute(m21Class, comparison: str = '__eq__') -> bool:
```
mscuthbert, line 483: this is beautiful.

mscuthbert (approved):
Wonderful!  

I'm about to get in a PR for iterator.first() [and last()] so I'll merge this and then immediately refactor this to make the finding of initial instances etc. simpler.

I'm not going to back-port this to v6.7 since it's more of an enhancement than a bug-fix in existing promised functionality.

PR #826, opened 2021-02-06
measures() returns only relevant spanners

jacobtylerwalls:
https://github.com/cuthbertLab/music21/blob/4f67e0ddb4e5d9fce459d7aa0802b278de1c9d06/music21/stream/__init__.py#L3840-L3841

done!

https://github.com/cuthbertLab/music21/blob/4f67e0ddb4e5d9fce459d7aa0802b278de1c9d06/music21/stream/__init__.py#L3876

easy, because that was done in #734:

https://github.com/cuthbertLab/music21/blob/4f67e0ddb4e5d9fce459d7aa0802b278de1c9d06/music21/stream/__init__.py#L6303-L6304

### DIFF
--- a/music21/braille/segment.py
+++ b/music21/braille/segment.py
@@ -1670,8 +1670,12 @@ def prepareSlurredNotes(music21Part, **keywords):
     for slur in music21Part.spannerBundle.getByClass(spanner.Slur):
         firstNote = slur[0]
         lastNote = slur[1]
-        beginIndex = allNotes.index(firstNote)
-        endIndex = allNotes.index(lastNote)
+
+        try:
+            beginIndex = allNotes.index(firstNote)
+            endIndex = allNotes.index(lastNote)
+        except exceptions21.StreamException:
+            continue
 
         delta = abs(endIndex - beginIndex) + 1
 
--- a/music21/braille/segment.py
+++ b/music21/braille/segment.py
@@ -1670,12 +1670,8 @@ def prepareSlurredNotes(music21Part, **keywords):
     for slur in music21Part.spannerBundle.getByClass(spanner.Slur):
         firstNote = slur[0]
         lastNote = slur[1]
-
-        try:
-            beginIndex = allNotes.index(firstNote)
-            endIndex = allNotes.index(lastNote)
-        except exceptions21.StreamException:
-            continue
+        beginIndex = allNotes.index(firstNote)
+        endIndex = allNotes.index(lastNote)
 
         delta = abs(endIndex - beginIndex) + 1
 
--- a/music21/stream/__init__.py
+++ b/music21/stream/__init__.py
@@ -3832,13 +3832,26 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
         <music21.stream.Measure 8 offset=0.0>
         <music21.stream.Measure 9 offset=4.0>
 
-        if gatherSpanners is True then all spanners in
-        the score are gathered and
-        included. (this behavior may change in the future)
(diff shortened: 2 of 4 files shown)


music21/braille/segment.py
```diff
@@ -1670,12 +1670,8 @@ def prepareSlurredNotes(music21Part, **keywords):
     for slur in music21Part.spannerBundle.getByClass(spanner.Slur):
         firstNote = slur[0]
         lastNote = slur[1]
-
-        try:
-            beginIndex = allNotes.index(firstNote)
-            endIndex = allNotes.index(lastNote)
-        except exceptions21.StreamException:
-            continue
```
mscuthbert, line 1678: what if the note that a Spanner refers to has been removed from the Stream?  Like:

```
>>> n = note.Note('C')
>>> sp = spanner.Slur([n])
>>> s = stream.Stream()
>>> s.append([sp, n])
>>> s.remove(n)
>>> { call the braille segment routine }
```

music21/stream/__init__.py
```diff
@@ -3972,18 +3973,12 @@ def hasMeasureNumberInformation(measureIterator):
...
-                # returnObj.insert(sp.getOffsetBySite(mStreamSpanners), sp)
-                returnObj.coreInsert(sf.elementOffset(sp), sp)
-
-                # environLocal.printDebug(['Stream.measures: copying spanners:', sp])
-
         # used coreInsert
         returnObj.coreElementsChanged()
+
+        if gatherSpanners:
+            returnObj.coreGatherMissingSpanners()
```
mscuthbert, line 3980: Oh, hey!  This was already here.  How nice!  :-D  wow.

I wonder if the default should be not to `requireAllPresent` so that we can extract measure 3-4 and see "are there any slurs beginning or ending in mm. 3-4?"  It'd be a lighter transition from v6 if at least any possibly relevant spanner was there.  We could make "gatherSpanners" be an enum (another PR adds a common.enums file that could go in.) of True, False, or CompleteOnly  -- feel free to leave that for another PR.

I'll go and update coreGatherMissingSpanners() to be more efficient.

music21/braille/segment.py
jacobtylerwalls: Yep, I'll revert. A warning might be nice.

mscuthbert (changes requested):
For now, just want the braille check to be restored until we can be sure it doesn't break something else.  Will do some followup work with coreGatherMissingSpanners to not run coreElementsChanged() if there are no spanners, and to make it possible to not add spanners that were not present in the previous stream.  As it is, you could do this:

```
>>> n1, n2, n3 = [ three whole Notes ]
>>> sl = spanner.Slur([n1, n2])
>>> s.append([timeSignature, sl, n1, n2, n3])
>>> s.makeMeasures(inPlace=True)
>>> sf = s.flat
>>> sf.remove(sl)  # slur is gone
>>> s2 = sf.makeMeasures()
>>> excerpt = s2.measures(1, 2)
>>> sl in excerpt  # True!
```

I'll make an argument to coreGatherMissingSpanners to pass in another spannerBundle that missing spanners must be a part of.  But will get that post this PR.

music21/braille/segment.py
jacobtylerwalls: But in that case indexing into the slur will fail, and that's not wrapped. We should probably be using `.getFirst()` and wrapping in a try.

music21/stream/__init__.py
jacobtylerwalls: Nice, agreed that that would be a lighter transition.

PR #827, opened 2021-02-06
Change special offsets to an Enum

mscuthbert:


documentation/nbvalNotebook.py
```diff
@@ -46,6 +45,8 @@ def runAll():
             return None
 
 
+# pylint is not realizing the finally clause is important.
+# pylint: disable=try-except-raise
```
jacobtylerwalls, line 49: I think pylint is just saying you can use try/finally instead of try/except/finally.

mscuthbert: Ah!  I did not realize you could use finally without except!

PR #828, opened 2021-02-06
makeMeasures no longer flattens part-like substreams

jacobtylerwalls:
`makeMeasures` no longer automatically flattens streams with part-like substreams. Forces the user to `.flat` first if they're OK with that.

It will continue to <s>flatten</s> account for substreams inside measures, such as voices.

### DIFF
--- a/music21/analysis/windowed.py
+++ b/music21/analysis/windowed.py
@@ -58,6 +58,8 @@ class WindowedAnalysis:
         # environLocal.printDebug(self.processor)
         if 'Stream' not in streamObj.classes:
             raise WindowedAnalysisException('non-stream provided as argument')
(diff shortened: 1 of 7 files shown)


mscuthbert:
Looks great -- at first I was afraid that all anaysis tools would fail unless `.flat` was called, which would be a NO-GO.  But so long as it's just windowed analysis, I'm fine with that.

mscuthbert:
This is good.  Note -- I'm planning on getting rid of "partLikeStreams" in v7 -- if you want Stream/Stream/Measure it should be Score/Part/Measure.  Anything else is a mistake.

jacobtylerwalls:
music21 -- A toolkit for computational music study, but only if you call `.flat`

PR #830, opened 2021-02-07
Fix braille line length rendering

jacobtylerwalls:
Two fixes:
- parse maxLineLength keyword
- fix a constraint on trailing whitespace that was demanding impossible solutions in short line length scenarios, better to leave some whitespace than to give up (just so happens that your original test, with the dotted quarter, hit precisely at the right location on the line!)


mscuthbert:
Wonderful!  Thank you!

PR #832, opened 2021-02-07
Remove functions marked as deprecated before/gone in v7

mscuthbert:
Whoops, thought I did this first thing.


mscuthbert:
In fact I did...

PR #836, opened 2021-02-08
Improve braille segment divisions

jacobtylerwalls:
Fixes:
 - IndexError when a BrailleSegmentDivision is found at a location where the segment is already empty (such as the beginning of a stream, or another segment break from a double barline)
 - Segment breaks weren't being created from double barlines after 48 elements (element count variable kept being cleared every measure)
 - Measure element group numbers weren't being incremented after segment breaks
 
Tests that had to be adjusted:
- `test_drill10_2` now shows a segment break being created correctly from a double barline after 48 elements
- `test_example11_2` now shows element group numbers being incremented

See the test changes by looking at the diff on the original English at b4625f56698858ce9b64ccf4fb509fe53bb7a395


music21/braille/segment.py
```diff
@@ -1843,50 +1844,81 @@ def getRawSegments(music21Part, setHand=None):
...
     '''
     allSegments = []
 
     currentSegment = BrailleSegment()
 
+    elementsInCurrentSegment: int = 0
+
+    startANewSegment: bool = False
+
+    # TODO: why is this skipping the measure layer and getting voices?
```
mscuthbert, line 1875: probably shouldn't!  :-)

mscuthbert (approved):
Looks great to me!

Btw -- the braille directory is one place where it's better to start switching variable names to underscore_case, at least internal variables, but possibly even method names, etc.  My friends in the blind community report that these variables are much easier for their screen readers to read for them, and it's the part of the system that has a large blind developer and user community.

PR #837, opened 2021-02-08
Metronom: fixed fontStyle attribute

marian-schulz:
A small fix to an error I stumbled across. 
The value 'bold' does not belong to fontStyle but to fontWeight.

MusicXml don't accept font-style="bold"

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -5050,7 +5050,7 @@ class MeasureExporter(XMLExporterBase):
         >>> MEX.dump(MEX.xmlRoot.findall('direction')[1])
         <direction>
           <direction-type>
-            <words default-y="45" enclosure="none" font-style="bold"
+            <words default-y="45" enclosure="none" font-weight="bold"
                 justify="left">slow</words>
           </direction-type>
         </direction>
--- a/music21/style.py
+++ b/music21/style.py
@@ -252,18 +252,30 @@ class TextStyle(Style):
         if value is None:
             self._fontStyle = None
         else:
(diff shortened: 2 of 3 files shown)


marian-schulz:
I took a look at the style module and realised that something was not implemented quite right there either. 

Bold is definitely a fontWeight and if that is set as a style, there is trouble with musicXML. 

But the style module says that it allows bold as a style and not only that it even knows 'bolditalic' a hybrid of fontWeight and fontStyle.

I did a second commit here to adress this issue.

I have created a second commit that I think solves the problem elegantly by setting 'bold' as style internally fontWeight and for 'bolditalic' the fontStyle 'italic' and the fontWeight 'bold'.

The alternative would be to adjust the music21ToXML and set the values there as musicXML likes it or to remove 'bold' as style and do without 'bolditalic' altogether.

I think my solution creates the least friction.
Agree ? 


Translated with www.DeepL.com/Translator (free version)

marian-schulz:
In a 3th commit i adapted the doc test in music21ToXml for the MetronomeMark.

marian-schulz:
sorry, did forget to update my master to music21/master

mscuthbert:
Music21 calls both bold and italic fontStyle not font-weight.  In this way it fits more with software like MS Word / Google Docs than like CSS/MusicXML.

mscuthbert (changes requested):
Music21 calls both bold and italic fontStyle not font-weight.  In this way it fits more with software like MS Word / Google Docs than like CSS/MusicXML.

Will accept a change to fix bug in output to MusicXML, not a change to music21's internal model.

marian-schulz:
I close this pull request.
I will did a new one, fixing the problem in the xml export

PR #838, opened 2021-02-08
fix the mxml export for unspecified font-style values 'bold' and 'bolditalic'

marian-schulz:
I have fixed an incompatibility between the fontStyle property of music21 and the corresponding font-style property in mxml.
Unlike music21, mxml does not know  font-styles 'bold' and 'bolditalic'. 

When exporting, the font-style values are changed to compatible values and properties in the mxml output as follows
a.) 
For fontStyle='bold' in music21, font-weight="bold" is set in mxml.  (The font-style property is removed in mxml)

b.)
For fontStyle='bolditalic' in music21 font-weight='bold' and font-style='italic' is set in mxml.

I added 2 Testcases for this issue, and the bug in my personal usecase is solved too ;)


mscuthbert:
Thanks!  Great fix!

PR #841, opened 2021-02-09
V7 stripTies(inPlace=True) returns None

jacobtylerwalls:
Now that we don't have to dance with `retainContainers` to get a return value, we can make this banana peel go away (in this example, the call was already commented out but you get the idea):

https://github.com/cuthbertLab/music21/blob/c79932c43beb528dba428641d0e133a836770f5c/music21/features/base.py#L369-L375


mscuthbert:
LGTM thanks!

PR #842, opened 2021-02-09
Try reducing the number of pylint messages

mscuthbert:
Make a commit that has a pylint error and see if it just gets one message


mscuthbert:
Nope -- not working.  Not worth it.  Will wait until Pylint 2.7.0 comes out and just deal with the multiple messages until then.

PR #845, opened 2021-02-10
Add stream.Stream.splitAtDurations

jacobtylerwalls:
Previously, `splitAtDurations()` when called on Streams would either return something unhelpful like a 1-tuple `(<music21.stream.Part at 0x...>,)` or raise an error, see https://github.com/cuthbertLab/music21/pull/763#discussion_r573203124.

Now it replaces complex notes in their contained streams with the replacements from `base.splitAtDurations` before returning that 1-tuple.


jacobtylerwalls:
There would be alternatives, of course, to returning a dict, and in fact a list of substreams is probably all that is needed. The alternative to returning anything would be a new StreamStatus for needing tuplet brackets made (which would help with performance ... currently `haveTupletBracketsBeenMade()` is an O(n) scan through every note in a stream. But that could be another PR down the road.

music21/stream/__init__.py
```diff
@@ -2709,6 +2709,105 @@ def replaceDerived(startSite=self):
...
+        False
+        >>> s3.spanners.first().getFirst().duration
+        <music21.duration.Duration 4.0>
+        >>> s3.spanners.first().getLast().duration
+        <music21.duration.Duration 1.0>
+        '''
+
+        def processContainer(container: Stream):
+            replacements: List[note.GeneralNote] = []
+            for noteObj in container.getElementsByClass('GeneralNote'):
```
mscuthbert, line 2775: I don't know that we want to limit to just GeneralNotes.  can be `getElementsNotOfClass(['Stream', 'Variant', 'Spanner'])` -- there can be other things that are not note-related that have complex durations.

music21/stream/__init__.py
```diff
@@ -2709,6 +2709,105 @@ def replaceDerived(startSite=self):
...
+                if noteObj.duration.type == 'complex':
```
mscuthbert, line 2776: I much prefer, `if element.duration.type != 'complex':  continue` to reduce the indentation.

music21/stream/__init__.py
```diff
@@ -2709,6 +2709,105 @@ def replaceDerived(startSite=self):
 
         replaceDerived()
 
+    def splitAtDurations(self, *, recurse=False) -> Optional[Dict]:
```
mscuthbert, line 2712: I'm trying to keep the return signatures and the non-keyword arguments of overridden methods as close to the original as possible.  So my preference would be that the return signature is the same as the original -- even though it's a bit ridiculous.  For one: Streams are not really hashable (I'm surprised that it allowed it as the key for a dict) -- even if we keep it, it should return an empty dict if nothing is found.  There's something that can bite you about:

```
# simplify durations:
for el in s.recurse():
     didSplit = el.splitAtDurations()
     if didSplit:
         print(didSplit[0])  # Bang! dies on a nested stream.
```

IDEs and typecheckers won't know why el.splitAtDurations() returned a Dict this time instead of a List.

Maybe we could have a `returnChangedElements=False` keyword where if someone specifically opts in to this, it returns something more useful.  But that should be List[Tuple[Stream, List[Music21Object]]] to avoid the hashing on Streams.  So then you can do:

```
for changedStream, changedElements in s.splitAtDurations(recurse=True, returnChangedElements=True):
     print(f'{changedStream}: {changedElements}')
```

mscuthbert (changes requested):
Love the approach, but not so keen on the function signature.  See notes below.

music21/stream/__init__.py
jacobtylerwalls: Agree. Thanks for sketching out the opt-in keyword, but I don't think we need the complexity. I like just returning the sort tuple. There would be other ways to track containers that need to have tuplets re-drawn, like setting a stream status flag in the subroutine while it's being edited.

My question is -- do you still want the stream itself to have its durations split? So a 5.0QL stream would be split into a 4.0QL stream and a 1.0QL stream? I'm thinking that's unnecessary.

jacobtylerwalls: Ah, right, it's moot because it wasn't working (AttributeError: 'tuple' object has no attribute 'spannerList').

mscuthbert:
> There would be alternatives, of course, to returning a dict, and in fact a list of substreams is probably all that is needed. The alternative to returning anything would be a new StreamStatus for needing tuplet brackets made (which would help with performance ... currently `haveTupletBracketsBeenMade()` is an O(n) scan through every note in a stream. But that could be another PR down the road.

This is a good thing to have!

mscuthbert:
Thanks!  This is a good improvement.  No one considers a Stream to be a complex element needing splitting.  Merging.  (And thanks for working through the mid-PR major stream refactor)

PR #848, opened 2021-02-12
Refactor stream.__init__ to stream.base, sim meter

mscuthbert:
Move testStream into stream.tests, make sure runs during multiprocessTest


mscuthbert:
Lowers TimeSignature creation time from 700 microseconds to 500 microseconds

mscuthbert:
Still way worse than Note which we've gotten down to 17microseconds.  But there are more notes.

PR #852, opened 2021-02-13
Add doctest and use .first() in getInstruments

jacobtylerwalls:


mscuthbert:
This is great -- I wonder if you'd mind setting the keywords to keyword-only?  It's a change that I'm implementing throughout and feel free to do it whenever there's no obvious order to keywords.

I am still thinking of eventually getting rid of these for a more general solution, but not in v7 in any case.

PR #856, opened 2021-02-14
MIDI import makes measures

### LINKED ISSUE #824: MIDI could make measures
**music21 version**

7

**Problem summary**
MIDI doesn't make measures. It makes:
- Score
-- Part
---- Voice

The lack of a measure layer leads to user questions (hence the tip in #754), but we have the knowledge of time signatures, so I'm not sure what's holding us back from just making the measures. A user will probably import midi, `.show()` it to find out how it turned out, see measures, but in reality they only have measures by virtue of `.show()` making notation.

The part-direct-to-voice hierarchy also leads to funny things, like making notation on `.show()` attempting to make voices a second time, because there aren't measures:

https://github.com/cuthbertLab/music21/blob/4f67e0ddb4e5d9fce459d7aa0802b278de1c9d06/music21/stream/__init__.py#L6306-L6316

That second-time-through `makeVoices` turns out okay because it doesn't recurse, so there are no notes in the part container, and it quits. But that should be made safer? Print a warning? If somebody `.insert()`'s notes into a part created from MIDI (with voices), then this second-time-around makeVoices is going to do nothing, unless the added flat notes have overlaps, in which case you'll get 4 voices instead of 2.

Opinions?

jacobtylerwalls:
Fixes #824 

Now, measures are made when MIDI files are imported.

### DIFF
--- a/music21/_version.py
+++ b/music21/_version.py
@@ -42,7 +42,7 @@ When changing, update the single test case in base.py.
 Changing this number invalidates old pickles -- do it if the old pickles create a problem.
 '''
 
-__version_info__ = (7, 0, 0)  # can be 4-tuple: (7, 0, 5, 'a2')
+__version_info__ = (7, 0, 1)  # can be 4-tuple: (7, 0, 5, 'a2')
 
 v = '.'.join(str(x) for x in __version_info__[0:3])
 if len(__version_info__) > 3 and __version_info__[3]:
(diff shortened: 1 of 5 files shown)


mscuthbert (approved):
Super great!  Merging. 

(note that the version was also incremented in another PR, so your cache may need to be manually cleared)

PR #857, opened 2021-02-14
Inform StreamStatus when accidentals or tuplets made

jacobtylerwalls:
Optimization in that if we've already made accidentals or tuplets once, we can retrieve the cached value instead of doing another scan through the notes. which will be more expensive in the case of extremely diatonic music or streams with no tuplets, etc.


mscuthbert:
This looks good to me.  Feel free to merge if after resolving conflicts and merging master everything still looks good.

PR #861, opened 2021-02-17
Clear figure when transposing ChordSymbol

### LINKED ISSUE #486: wrong figure for transposed chords
I've found a little issue with the transpose method for `ChordSymbol`s. Basically, the `transpose()` method isn't being redefined in `ChordSymbol`, and as a result, the figure is not updated.
```
c = ChordSymbol('Am')
c = c.tranpose(5)
c.figure
```
still returns 'Am', even if `c.findFigure()` correctly returns 'Dm'. The only way I've found around this is to manually set `c._figure = None` to force `c.figure` to recompute the figure, but that's not very satisfying, plus I might be missing other stuff.
Am I doing something wrong?

jacobtylerwalls:
Fixes #486

### DIFF
--- a/music21/harmony.py
+++ b/music21/harmony.py
@@ -2265,13 +2265,6 @@ class ChordSymbol(Harmony):
         <music21.harmony.ChordSymbol Dm>
         >>> cs.figure
         'Dm'
-        
-        >>> nc = harmony.NoChord()
-        >>> nc.figure
-        'N.C.'
-        >>> nc.transpose(8, inPlace=True)
-        >>> nc.figure
-        'N.C.'
         '''
         post = super().transpose(value, inPlace=inPlace)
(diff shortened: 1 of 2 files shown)


music21/harmony.py
```diff
@@ -2355,6 +2380,22 @@ def writeAsChord(self):
...
+        >>> nc.figure
+        'N.C.'
+        >>> nc.transpose(8, inPlace=True)
+        >>> nc.figure
+        'N.C.'
+        '''
+        if not inPlace:
+            return copy.deepcopy(self)
+        else:
+            return self
```
jacobtylerwalls, line 2397: in general, are we returning None when inPlace?

mscuthbert: whoops -- meant that.  Thank you!

PR #863, opened 2021-02-18
Fix crash in export of flattened piano music

jacobtylerwalls:
Fixes a failure to show flattened piano scores. Could be worth a backport to v.6?


mscuthbert:
Agreed -- worth backporting too.  Can you prepare a PR against the v6 branch with just this commit "cherry-pick" 'd into it?  THANKS!  I meant to release a 6.7.1 anyhow since there are two other major bug fixes that I back-ported to that branch already.

PR #866, opened 2021-02-19
Avoid reiterating accidentals on diatonic or tied pitches in excerpts with key signatures

### LINKED ISSUE #865: Accidental reiterated on tied chromatic pitch if key signature present
**music21 version**

7

**Problem summary**
A tied pitch has its accidental unnecessarily reiterated when running `makeNotation` if <s>the key is no sharps or flats</s> there is a key object somewhere.

**Steps to reproduce**

```
from music21 import converter
p = converter.parse('tinynotation: 4/4 f#1~ f#1')
p.measure(1).insert(0, key.KeySignature(sharps=0))
# makeNotation performs some checks before calling makeAccidentals
p.makeNotation(inPlace=True)
p.show()
```

image

**More information**

Noticed this in the User's Guide, the last example in Ch. 9 shows an image of this working correctly (based on the series of calls to chordify, etc.), but it no longer works.

jacobtylerwalls:
Fixes #865


jacobtylerwalls:
Test failure here is an interesting case -- what is the expected behavior for cautionary accidentals when a key change is involved?

```
>>> from music21 import *
>>> bm = converter.parse('''tinynotation: 4/4 d4 f# a f# d8 e f# g a2
... e-4 g b- g e-8 f g a- b-2
... e4 g# b g# e8 f# g# a b2 f4 a c' a f8 g a b- c'2
... g-4 b- d'- b- g-8 a- b- c'- d'-2 d'4 b g b d'8 c' b a g2''', makeNotation=False)
>>> bm.insert(0.0, key.KeySignature(2))
>>> bm.insert(8.0, key.KeySignature(-3))
>>> bm.insert(16.0, key.KeySignature(4))
>>> bm.insert(24.0, key.KeySignature(-1))
>>> bm.insert(32.0, key.KeySignature(-6))
>>> bm.insert(40.0, key.KeySignature(0))
>>> bm.makeNotation(inPlace=True, cautionaryNotImmediateRepeat=False)
>>> bm.measure(11).notes[0].pitch.accidental.displayStatus
True
```

Finale ignores the `<accidental>` tags written by music21 with this PR, but MuseScore observes them:

musescore-cautionary


What should music21 do?

jacobtylerwalls:
The change in the PR brings this branch of `stream.makeNotation` into agreement with `Part.makeAccidentals` by just using `if i > 0`:

https://github.com/cuthbertLab/music21/blob/ed021ae6804fd6af3e9723b502094170be7773ab/music21/stream/base.py#L12990-L12999

mscuthbert:
> Test failure here is an interesting case -- what is the expected behavior for cautionary accidentals when a key change is involved?
> 
> musescore-cautionary
> 
> What should music21 do?

I think that if the accidental has been cancelled by the key signature, the performer knows not to play it.    So here, we shouldn't put the accidentals in.

I wish we could put the accidental (natural in this case) in the first time the note is encountered if it's encountered between 10 and 20 notes after the new key change.  :-)  that's when I always miss it.  I notice the key change for the first few measures, but if I haven't yet needed to play a note not affected by the new key, that's when I screw up.

jacobtylerwalls:
Now this PR just avoids getting the previous measure's pitches if we have a key change. I take it this example (failed test) is OK? If so I can update the test:


Screen Shot 2021-02-23 at 10 41 01 AM

jacobtylerwalls:
This is the other failed test, same scenario:

Screen Shot 2021-02-23 at 10 50 07 AM

And it looks like the cautionary accidentals have been fiddled with in both tests before:

https://github.com/cuthbertLab/music21/blob/e7b102a4539cb68741ff4eabc4b9dee581c44e1e/music21/braille/test.py#L3170-L3178

music21/stream/tests.py
```diff
@@ -2501,6 +2501,34 @@ def testMakeAccidentalsTies(self):
...
+        p.makeNotation(inPlace=True)
+        self.assertEqual(p.measure(2).notes.first().pitch.accidental.displayStatus, False)
+
+    def testMakeNotationTiesKeyChange(self):
+        from music21 import converter
+        p = converter.parse('tinynotation: 4/4 f#1~ f#1')
+        # Insert key change where held-over note is chromatic
+        p.measure(2).insert(0, key.KeySignature(sharps=-1))
+        pMade1 = p.makeNotation()
+        self.assertEqual(pMade1.measure(2).notes.first().pitch.accidental.displayStatus, False)
```
mscuthbert, line 2519: I couldn't find anything about this case directly in the Gould, _Behind Bars_ book, but the closest recommendation is if there's a system break mid-tie to repeat the accidental afterwards (she says at one point, better without parentheses, but then demonstrates it with parentheses more often.  We don't have a displayParentheses status unfortunately).

music21/stream/tests.py
```diff
@@ -2501,6 +2501,34 @@ def testMakeAccidentalsTies(self):
...
+
+        p.measure(1).insert(0, key.KeySignature(sharps=-1))
+        # This is no longer a key "change", should still work based on the tie
+        pMade2 = p.makeNotation()
+        self.assertEqual(pMade2.measure(2).notes.first().pitch.accidental.displayStatus, False)
```
mscuthbert, line 2524: this is correct.

mscuthbert (commented):
This is an improvement for most cases, but I think in the case where there is a natural next to the note after a key change (such as if the f# is moved up an octave in the tests, and the key moved to is C) then the lack of an accidental display becomes especially confusing.  I don't know if it's an easy fix.

music21/stream/tests.py
jacobtylerwalls: Ah, but after the tie note concludes, yes? Not on the terminus of the tie itself? I've never seen an accidental interrupt a tie: https://github.com/cuthbertLab/music21/issues/865#issuecomment-782503317

music21/stream/base.py
```diff
@@ -6552,7 +6552,7 @@ def makeNotation(self,
                 if m.keySignature is not None:
                     ksLast = m.keySignature
 
-                if i > 0 and ksLast is None:
+                if i > 0:
```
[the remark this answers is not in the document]
jacobtylerwalls: ksLast is not None if there is any key anywhere to the left, so yeah, it was firing all over the place, including merely having D# in e minor tied over to another D#.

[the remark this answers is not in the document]
jacobtylerwalls: Sorry, "firing" meaning "reiterating accidentals on tie termini" because this branch wasn't executing

music21/stream/tests.py
jacobtylerwalls: But I will zip over to IMSLP and pull down some Reger to find out ...

jacobtylerwalls: This edition of a Reger intermezzo validates Gould's point about system breaks employing a parenthetical accidental on a chromatic end-tie:

Screen Shot 2021-02-23 at 2 18 53 PM
Screen Shot 2021-02-23 at 2 18 58 PM

jacobtylerwalls: Ugh, putting the key change AFTER the tie ends is just cheating, robbing us of a model:
Screen Shot 2021-02-23 at 2 23 07 PM

jacobtylerwalls: Too cumbersome to post a photo of a hard copy, but now I've got an example 12 measures from the end of Messiaen, op.1/vi, "Cloches d'angoisse et larmes d'adieu" where the accidentals are reiterated on the end-tie across a key change (and for one of the three piano staves, a page break also) on a note chromatic in the new key.

B major chord being tied over during a key change from 5 sharps to 2 sharps. The D# was explicit on the start tie because of an earlier D-natural. Then the D# is reiterated after the barline and key change (the key change was printed with cancelled naturals after the two sharps)

jacobtylerwalls: I arranged that prelude in college, and I had a copy sticking out of my bag when an acquaintance spotted it, and assuming it was an original composition, expressed alarm at what seemed to be a pretentious title, translating on the fly, "Bells of ...sadness and tears of ... goodbye?!" as if to ask, "are you doing alright, fella?"

jacobtylerwalls: `tiePitchSet` should have chromatic notes (in the new key) filtered out if `m.keySignature` is not None, then, so they get reiterated, maybe.

jacobtylerwalls:
Needs some tests for cautionary accidentals without ties and maybe also have a look to see if the DRY issues can be refactored.

music21/stream/makeNotation.py
```diff
@@ -1486,6 +1488,74 @@ def getTiePitchSet(prior):
...
+        'pitchPastMeasure',
+        'useKeySignature',
+        'searchKeySignatureByContext',
+        'tiePitchSet',
+        'inPlace',
+    ):
+        try:
+            del(srkCopy[illegalKeyword])
+        except KeyError:
+            pass
```
jacobtylerwalls, line 1512: Next step is just to get the four or so legal keywords left and just put them in the signature.

mscuthbert:
Thanks for all the work -- I'll do a deep review of PRs over the weekend (probably Sunday).  I need to get out of the "PRs first thing in the morning [Hawaii time]" routine -- getting me a bit behind on other work, even though I am LOVING the activity!

mscuthbert:
sorry I've been slow on this.  Added some comments to the previous page.

jacobtylerwalls:
np. just reiterating in case you don't see the other page first that I think the tests on this PR demo the desired behavior; LMK if the comments can be clearer.

music21/stream/tests.py
```diff
@@ -2601,6 +2601,48 @@ def testMakeAccidentalsTies(self):
...
                                 (i, allNotes[i].pitch.accidental.displayStatus, ds[i]))
 
+    def testMakeNotationTiesKeyless(self):
+        from music21 import converter
+        p = converter.parse('tinynotation: 4/4 f#1~ f#1')
+        # Key of no sharps/flats
+        p.measure(1).insert(0, key.KeySignature(sharps=0))
+        # calls makeAccidentalsInMeasureStream()
+        p.makeNotation(inPlace=True)
+        self.assertIs(p.measure(2).notes.first().pitch.accidental.displayStatus, False)
```
jacobtylerwalls, line 2611: Of the added assertions, this is one of the two that fails on master (first screenshot on the issue). The others just add coverage.

music21/stream/tests.py
```diff
@@ -2601,6 +2601,48 @@ def testMakeAccidentalsTies(self):
...
+        # Same, but with a natural
+        p.measure(2).insert(0, key.KeySignature(sharps=-1))
+        made = p.makeNotation()
+        self.assertIs(made.measure(2).notes.first().pitch.accidental.displayStatus, True)
+
+        p = converter.parse('tinynotation: 4/4 f#1~ f#1')
+        p.measure(1).insert(0, key.KeySignature(sharps=-1))
+        # This is no longer a key "change", should still work based on the tie
+        made = p.makeNotation()
+        self.assertIs(made.measure(2).notes.first().pitch.accidental.displayStatus, False)
```
jacobtylerwalls, line 2638: The second one that fails on master (third screenshot on the ticket).

music21/stream/makeNotation.py
```diff
@@ -1612,6 +1615,93 @@ def getTiePitchSet(prior):
...
+                # If there is any key signature object to the left,
+                # just get the chromatic pitches from previous measure
+                # G-naturals in C major following G-flats in F major need cautionary
+                # G-naturals in C major following G-flats in Db major don't
+                ksLastDiatonic = [p.name for p in ksLast.getScale().pitches]
+                pitchPastMeasure = [p for p in s[i - 1].pitches
+                    if p.name not in ksLastDiatonic]
+            # Get tiePitchSet from previous measure
+            try:
+                previousNoteOrChord = s[i - 1][-1]
```
mscuthbert, line 1678: should this be `s[i - 1][note.NotRest][-1]`?

mscuthbert: (in case there is a repeat-mark at the end or barline or something)

music21/stream/tests.py
```diff
@@ -2601,6 +2601,48 @@ def testMakeAccidentalsTies(self):
...
+
+        # Wipe out the tie; accidental should be reiterated
+        for n in p.flat.notes:
+            n.tie = None
+        made_no_ties = p.makeNotation()
+        self.assertIs(made_no_ties.measure(2).notes.first().pitch.accidental.displayStatus, True)
```
mscuthbert, line 2644: Great!

mscuthbert (approved):
All great -- I'm going to make one little change to help make sure we're getting ties from the right previous note in the (decently common) case that there's a right-barline or repeat sign or something preceding a key change.

music21/stream/makeNotation.py
jacobtylerwalls: yeah, looks like I was porting the old logic over without checking it. great catch!

mscuthbert:
Looking this over in regards to #1345 I'm wondering if makeAccidentalsInMeasureStream should instead be called from Stream.makeAccidentals() since we can easily make one method do the work of both and avoid redundancy?

jacobtylerwalls:
Yep, now that we've removed the limitation on having other objects in the measure stream, a generic stream could be the "stream having measures". And then it would be terrific to see `Part.makeAccidentals` go away completely (which we could do, right?)

mscuthbert:
I think so.  I'll do this in some small quantities...  I had an immediate research-project-fix that depended on this working.

PR #868, opened 2021-02-19
v7 KeySignature sharps defaults 0 instead of None

### LINKED ISSUE #867: Default KeySignature crashes on string representation
**music21 version**

7

**Steps to reproduce**

```
>>> ks = key.KeySignature()
>>> ks
  File "/Users/jwalls/music21/music21/prebase.py", line 236, in __repr__
    strRepr = self._reprInternal()
  File "/Users/jwalls/music21/music21/key.py", line 378, in _reprInternal
    return 'of ' + self._strDescription()
  File "/Users/jwalls/music21/music21/key.py", line 352, in _strDescription
    output = 'pitches: [' + ', '.join([str(p) for p in self.alteredPitches]) + ']'
  File "/Users/jwalls/music21/music21/common/decorators.py", line 171, in inner
    instance._cache[funcName] = method(instance, *args, **kwargs)
  File "/Users/jwalls/music21/music21/key.py", line 443, in alteredPitches
    if self.sharps > 0:
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

Default arg for `sharps` in the constructor is None, to support altered pitches. I suggest making the default in v7 `sharps: int = 0`, and then updating the documentation for nontraditional ks's to advise sending alteredPitches to the constructor. 

Then the repr will have to include a safety check for this as well (sharps is None and alteredPitches == [])

jacobtylerwalls:
Fixes #867 

Any interest in also changing the default arg from `sharps=None` to `sharps: int = 0`?

### DIFF
--- a/music21/key.py
+++ b/music21/key.py
@@ -281,10 +281,10 @@ class KeySignature(base.Music21Object):
     >>> legal
     <music21.key.Key of c# minor>
 
-    To set a non-traditional Key Signature, create a KeySignature object and then
-    set the alteredPitches list:
+    To set a non-traditional Key Signature, create a KeySignature object
+    with `sharps=None`, and then set the `alteredPitches` list:
 
-    >>> unusual = key.KeySignature()
+    >>> unusual = key.KeySignature(sharps=None)
     >>> unusual.alteredPitches = ['E-', 'G#']
     >>> unusual
     <music21.key.KeySignature of pitches: [E-, G#]>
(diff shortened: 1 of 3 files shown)


mscuthbert:
Yeah. let's change default value to 0 -- having special cases for None sharps seems to be a mistake.

There are a lot of things in music21 where originally I wanted to distinguish between None and an empty value.  Like None for Accidental vs. Natural (which has been annoying but at least there's a conceptual difference between a note whose accidental is unknown vs. one whose accidental is known to be natural -- e.g. in chant), but I can't think of a case where we want a None KS that is different from no flats or sharps, since it's easy enough to just not put a KS into the stream.

music21/key.py
```diff
@@ -315,7 +318,7 @@ class KeySignature(base.Music21Object):
 
     classSortOrder = 2
 
-    def __init__(self, sharps=None):
+    def __init__(self, sharps: int = 0):
```
mscuthbert, line 321: this should be typed as Optional[int]

mscuthbert (changes requested):
One change to typing, which should help typing to remember that `.sharps` might not be a number.

PR #871, opened 2021-02-21
Add cases to isWellFormedNotation

jacobtylerwalls:
Previously, Opus objects would always fail `isWellFormedNotation()`.

Update: now, we fail streams that have "loose" notes and rests and aren't a Measure or Voice.

My hunch is that if we sew up the gaps in `isWellFormedNotation` then we can more quickly answer user questions by just directing them to this method.

### DIFF
--- a/music21/stream/base.py
+++ b/music21/stream/base.py
@@ -9281,13 +9281,23 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
         >>> o2 = stream.Opus([s2])
         >>> o2.isWellFormedNotation()
         False
+
+        Only Measures and Voices are allowed to contain notes and rests directly:
+        >>> m.isWellFormedNotation()
+        True
+        >>> s2.append(note.Rest())
+        >>> s2.isWellFormedNotation()
+        False
         '''
         def allSubstreamsHaveMeasures(testStream):
             return all(s.hasMeasures() for s in testStream.getElementsByClass('Stream'))
 
-        # if a measure, we assume we are well-formed
-        if 'Measure' in self.classes:
+        # if a measure or voice, we assume we are well-formed
+        if 'Measure' in self.classes or 'Voice' in self.classes:
             return True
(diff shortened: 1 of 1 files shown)


music21/stream/base.py
```diff
@@ -9269,29 +9269,32 @@ def isWellFormedNotation(self):
...
+            return all(s.hasMeasures() for s in testStream.getElementsByClass('Stream'))
+
         # if a measure, we assume we are well-formed
         if 'Measure' in self.classes:
             return True
         elif 'Part' in self.classes:
             if self.hasMeasures():
                 return True
+        elif 'Opus' in self.classes:
+            return all(allSubstreamsHaveMeasures(s) for s in self.scores)
```
mscuthbert, line 9295: need a test on `stream.Opus().isWellFormedNotation()`  should return False (no scores), but given that `all([])` is True, it will return True.

mscuthbert (changes requested):
Are Empty Score and Parts Well-formed?  I suspect No.  It's a bit to ask, but I think that one of the things that well-formed needs to be is no notes/rests/etc. outside of measures, and non-empty -- maybe a change, but a good time to implement it.  OR an empty Opus/Score/Part can be well-formed IF there are no GeneralNotes (or spanners?) in them but otherwise not?  THanks!

jacobtylerwalls:
I'm inclined to leave aside empty streams for now, including Opus, but checking for "loose" notes is easy.

jacobtylerwalls:
Spanners usually show up at high levels of the hierarchy, so I wouldn't check for those.

mscuthbert:
Merging -- I think that I'll want to put in a keyword flag, "mustBeNonEmpty=False" later to catch elements without required subentities, but probably correct to not do that now.

In layout.py there is a second definition of well-formed, where Streams go: Page -> System -> Staff -> Measure [-> Voice] -> Notes ... eventually should check on that type of hierarchy too.  But I've only used layout.py's hierarchy when I'm parsing OMR scores.

PR #872, opened 2021-02-21
Bring keywords into method signatures in Braille module and observe maxLineLength for keyboard music

jacobtylerwalls:
Brings keywords into the method signatures in the Braille module.

This exercise revealed/verified:
- as pointed out on the list last fall, `segmentBreaks` keyword is inoperative, <s>so it's commented out now</s>
- keyboard segments still weren't observing the `maxLineLength` parameter, fixed
- old documentation misstated some default values, fixed

Sorry for having so many patches open, but once I got to know the Braille module I was in agreement with you that this needed to be cleaned up, and so opening this now is more of an FYI so that we don't duplicate effort.


jacobtylerwalls:
Extraction of music from voices doesn't appear to work, but I think I know why -- as is common, some confusion over the nesting of measures and voices. That can be a future PR.

mscuthbert (approved):
This is a great set of changes!  It does show the acute need for a BrailleTranslator object to keep these keywords in one place.  But that can be a second refactor.  THANKS!

My only Q is about whether to leave segmentBreaks in the keywords and have a warning displayed that it does nothing rather than have it break existing code, since the translator is being used by a number of people who might not know how to repair it.  BUT if we do a larger refactor in this v. then the Q becomes moot.  Leaving open for 24 hrs to think more, but I think it's a "Merge"

jacobtylerwalls:
Not a heavy lift to put it back in with a warning, happy to wait a few days and let you mull.

music21/braille/translate.py
```diff
@@ -55,6 +55,10 @@
     that should be present in a line of braille.
 * **segmentBreaks** ([]): A list consisting of (measure number, offset start)
     tuples indicating where the music should be broken into segments.
```
jacobtylerwalls, line 57: That wouldn't be difficult to implement -- I think we're talking about a for-loop and then
`obj.measure(x).insert(y, BrailleSegmentDivision())` so pretty trivial.

Given that, I'm inclined to just comment the keywords back in and let a subsequent PR implement that feature and any tests it needs, etc

jacobtylerwalls:
`[]` isn't dangerous if it's only being read from, but I see why I originally changed this to None.

music21/braille/segment.py
```diff
@@ -1661,20 +1650,8 @@ def prepareSlurredNotes(music21Part, **keywords):
...
-        showLongSlursAndTiesTogether = keywords['showLongSlursAndTiesTogether']
-    elif slurLongPhraseWithBrackets:
-        showLongSlursAndTiesTogether = True
-    else:
-        showLongSlursAndTiesTogether = SEGMENT_SHOWLONGSLURSANDTIESTOGETHER
-
-    if slurLongPhraseWithBrackets is False:
-        pass
+    if not slurLongPhraseWithBrackets:
+        showLongSlursAndTiesTogether = False
```
jacobtylerwalls, line 1654: FYI -- I'd like another pair of eyes on this. This was deeply confusing, and not apparent how it could be recreated using method signatures.

The issue is the two keywords `showLong...` and `slurLong...`

Previously, if `showLong...` was silent, we just allied `slurLong` and `showLong` to the same value. `slurLong` defaulted True, so we got True and True. This is why I made the change to the documentation for the keyword `showLong`--it said the default was False, but that's not what was happening.

However, if you explicitly set `showLong`, you get that value, even if it conflicts with `slurLong`. I don't think this is tested anywhere.

To get the tests to pass, I had to keep the current behavior for showLong=silent and slurLong=False, but since there is no way to represent silence (without using Optional[bool]=None), I just did `if not slurLong: showLong=False`, which means there is no way to affirmatively provide `slurLong=True` and `showLong=False`, which was theoretically possible but not tested.

Acceptable? Or do you want to insert some sort of Optional[bool]=None for showLong? Given the pretzel here and the fact the documentation was misleading and no tests are failing, this seemed fair.

jacobtylerwalls: ♻️  'ing back to this, I think `Optional[bool]=None` is the most conservative thing we can do to ensure we don't break people's workflows. I'll spruce this up.

mscuthbert (approved):
Explicit is better than implicit.  We can work on simplifying this later in next version

Issue #869, opened 2021-02-19
Enharmonic simplifications for romanNumeralFromChord


gulnazaki:
**music21 version**

`6.7.0`   

**Currently there is no way to do enharmonic simplifications for roman numeral analysis**

**Steps to reproduce**

```
>>> from music21 import *
>>> k = key.Key('C#')

>>> ch = chord.Chord('E-4 G-4 B-4')
>>> roman.romanNumeralFromChord(ch, k)
<music21.roman.RomanNumeral bbiii in C# major>

>>> ch = chord.Chord('F4 A-4 C-5')
>>> roman.romanNumeralFromChord(ch, k)
<music21.roman.RomanNumeral bivo5b3 in C# major>
```

**Expected vs. actual behavior**
It can be argued that the above are `ii`  and `iii` chords respectively if the enharmonics of the pitches above (`D#4 F#4 A#4`) and (`E#4 G#4 B#4`) where chosen instead.

I am not able to find any way to simplify enharmonics in order to avoid this and get simpler roman numerals, since the `simplifyEnharmonics` chord method is key independent. I can go forward with a PR adding a `simplifyEnharmonics` option to the `romanNumeralFromChord` if you find this feature useful.

jacobtylerwalls:
Would `simplifyMultipleEnharmonics` help your use case?

```
>>> ch
<music21.chord.Chord E-4 G-4 B-4>
>>> pitch.simplifyMultipleEnharmonics([n.pitch for n in ch], keyContext=key.Key('C#'))
[<music21.pitch.Pitch D#4>, <music21.pitch.Pitch F#4>, <music21.pitch.Pitch A#4>]
```

gulnazaki:
Thanks a lot! 

I wasn't aware of this method, it definitely does the trick.

jacobtylerwalls:
Maybe there's an opportunity for syntactic sugar here so that the user doesn't have to collect the pitch objects from the chord and call a method on the pitch module; maybe there could be a `chord.respell(key, inPlace=False)` method taking a key or key sig object.

gulnazaki:
Actually, there could be an optional `keyContext` argument to the `chord.simplifyEnharmonics` defaulting to `None`, that would change the simplification goal if provided. 
It would be more intuitive and there is no need to add another method I believe.

I could do a PR about it, if you prefer.

jacobtylerwalls:
[...]

PR #873, opened 2021-02-21
Added keyContext option in chord.simplifyEnharmonics

gulnazaki:
As discussed in #869, this provides the ability to simplify enharmonics based on a supllied key, without having to call `pitch.simplifyMultipleEnharmonics` for the pitch objects.

I also believe that a simpler criterion could be added as an option to the `pitch.simplifyMultipleEnharmonics`, for the key context case. Namely, for every pitch choosing between its enharmonics if they belongs to the key pitches (in a greedy manner). 

If you find this useful, I could add it to this PR or create a new one.

### DIFF
--- a/music21/chord/__init__.py
+++ b/music21/chord/__init__.py
@@ -3796,7 +3796,7 @@ class Chord(note.NotRest):
 
         >>> c.simplifyEnharmonics(inPlace=True, keyContext=key.Key('A-'))
         >>> c.pitches
-        (<music21.pitch.Pitch D->, <music21.pitch.Pitch F>, <music21.pitch.Pitch A>)
+        (<music21.pitch.Pitch D->, <music21.pitch.Pitch F>, <music21.pitch.Pitch A->)
         '''
         if inPlace:
             returnObj = self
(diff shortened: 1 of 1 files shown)


mscuthbert:
Merging -- but a Q -- if keyContext is None, should it try to get a key from `getContextByClass('Key')`?  that might make it even simpler to run.  If we think so, I'm happy to have a PR on that.

gulnazaki:
@mscuthbert Yes it is a good idea, but perhaps it limits the usage of the function (if someone wanted to do only a chord based simplification it is not possible if there is a key context, while now they can set the key explicitely if they want). It is more of a design choice, so I leave it up to you. 

Let me know if you find the option to use the extra criterion I mentioned above useful (and correct?), since it would speed up the process.

mscuthbert:
I think that the actual best thing is to have the signature be:

`Chord.simplifyEnharmonics(*, inKey: Optional[key.KeySignature] = None, keyContextAware: bool = False)`

so that by default it runs the (fast-ish) pitch.simplifyMultipleEnharmonics without a key context.  with `ch.simplifyEnharmonics(inKey=key.Key('D'))` is uses that key (slower), and with `ch.simplifyEnharmonics(keyContextAware=True)` (slowest) it calls `getContextByClass("KeySignature")` before calling pitch.simplifyMultipleEnharmonics.

I'm not sure if we currently treat KeySignature differently from Key in simplifyMultipleEnharmonics.  Ideally we would allow KeySignatures but do things better with Keys (especially in minor).

PR #874, opened 2021-02-22
Pylint -- remove unnecessary comprehensions

jacobtylerwalls:
Fixes code failing on the new Pylint check for unnecessary comprehensions when generators will do.


mscuthbert:
AH!  I was just looking at that.  Thanks!

PR #877, opened 2021-02-23
[v.6] Fix crash in export of flattened piano music

jacobtylerwalls:


mscuthbert:
lint issues are from a newer version of PyLint than we are using on `m21_6`.  Merging

PR #882, opened 2021-02-27
Improve ChordSymbol construction from keywords

### LINKED ISSUE #793: xmlToChordSymbol() calls _updatePitches()
Find a way to avoid calling `ChordSymbol._updatePitches()` from `xmlToChordSymbol()`.

Decide whether users are supposed to be manipulating `figure`, or whether there needs to be a public version of `_updatePitches()`.

jacobtylerwalls:
Fixes #793 

- Improve handling of keyword arguments to `ChordSymbol` constructor, prerequisite in order to ...
- Refactor MeasureParser.xmlToChordSymbol so as to avoid calling `_updatePitches()`
- Bugfix: Allows calling `ChordSymbol('Cm', inversion=1)` or `cs.inversion(1)`


music21/chord/__init__.py
```diff
@@ -1914,6 +1917,8 @@ def inversion(self, newInversion=None, *, find=True, testRoot=None, transposeOnS
 
                 if 'inversion' in self._overrides:
                     del self._overrides['inversion']
+                if 'bass' in self._overrides:
+                    del self._overrides['bass']
```
mscuthbert, line 1921: ah! I thought we were getting rid of bass overrides?

mscuthbert (commented):
I'll have to look more closely at this but it looks good in a first view!

music21/chord/__init__.py
jacobtylerwalls: Still being set here: https://github.com/cuthbertLab/music21/blob/d3770d0907e3af00d45153da25b8c064bf2bbec4/music21/harmony.py#L210-L214

Was there an effort to get rid of this before?

mscuthbert: I just remember something about, "no one overrides the bass note!  just the roots.  But yes, if it's still being set somewhere then we should keep it going.

jacobtylerwalls: AH, got it, when we were trying to paraphrase Rameau better.

mscuthbert:
I keep reading this, but it'll still take more time for me to refamiliarize myself with the Harmony module (these GitHub diffs that don't show code that is merely moved differently isn't helpful  :-)  )  -- sorry this is taking time.

jacobtylerwalls:
You know what, I can leave some review comments to make this easier to review (given that I don't have incremental commits).

music21/musicxml/xmlToM21.py
```diff
@@ -4498,30 +4498,41 @@ def xmlToChordSymbol(self, mxHarmony):
...
+            mxBassAlter = mxBass.find('bass-alter')
+            if mxBassAlter is not None:
+                # can provide integer or float to create accidental on pitch
+                b.accidental = pitch.Accidental(float(mxBassAlter.text))
 
-        self.setEditorial(mxHarmony, cs)
-        self.setPrintStyle(mxHarmony, cs)
-        self.setPrintObject(mxHarmony, cs)
-        # TODO: attr: print-frame
-        # TODO: attrGroup: placement
```
jacobtylerwalls, line 4522: moved below to 4579

music21/musicxml/xmlToM21.py
```diff
@@ -4536,59 +4547,25 @@ def xmlToChordSymbol(self, mxHarmony):
             if mxRootAlter is not None:
                 # can provide integer or float to create accidental on pitch
                 r.accidental = pitch.Accidental(float(mxRootAlter.text))
-            # set Pitch object on Harmony
-            cs.root(r)
```
jacobtylerwalls, line 4540: handled by kwarg now

music21/musicxml/xmlToM21.py
```diff
@@ -4536,59 +4547,25 @@ def xmlToChordSymbol(self, mxHarmony):
...
 
-        if kindText is not None:  # two ways of doing it...
-            cs.chordKind = mxKind.text.strip()
-            # Get m21 chord kind from dict of musicxml aliases ("dominant" -> "dominant-seventh")
-            if cs.chordKind in harmony.CHORD_ALIASES:
-                cs.chordKind = harmony.CHORD_ALIASES[cs.chordKind]
-            mxKindText = mxKind.get('text')  # attribute
-            if mxKindText is not None:
-                if not ('NoChord' in cs.classes and mxKindText == ''):
-                    cs.chordKindStr = mxKindText
```
jacobtylerwalls, line 4553: moved above to 4528, but the NoChord condition had to be adjusted because we don't have an object yet

music21/musicxml/xmlToM21.py
```diff
@@ -4536,59 +4547,25 @@ def xmlToChordSymbol(self, mxHarmony):
...
-        # TODO: attr: use-symbols
-        # TODO: attr: stack-degrees
-        # TODO: attr: parentheses-degrees
-        # TODO: attr: bracket-degrees
-        # TODO: attrGroup: print-style
-        # TODO: attrGroup: halign
-        # TODO: attrGroup: valign
```
jacobtylerwalls, line 4560: moved to 4585

music21/musicxml/xmlToM21.py
```diff
@@ -4536,59 +4547,25 @@ def xmlToChordSymbol(self, mxHarmony):
...
-
-        mxInversion = mxHarmony.find('inversion')
-        if textStripValid(mxInversion):
-            try:
-                # must be an int
-                cs.inversion(int(mxInversion.text.strip()), transposeOnSet=False)
-            except ValueError:
-                pass
```
jacobtylerwalls, line 4568: moved to 4523 above, and the casting to int was removed because it's handled in `_updateBasedOnXMLInput`, but if you prefer the belt-and-suspenders, we can cast here also

music21/musicxml/xmlToM21.py
```diff
@@ -4536,59 +4547,25 @@ def xmlToChordSymbol(self, mxHarmony):
...
-        clearBassOverride: bool = False
-        mxBass = mxHarmony.find('bass')
-        if mxBass is not None:
-            # required
-            b = pitch.Pitch(mxBass.find('bass-step').text)
-            # optional
-            mxBassAlter = mxBass.find('bass-alter')
-            if mxBassAlter is not None:
-                # can provide integer or float to create accidental on pitch
-                b.accidental = pitch.Accidental(float(mxBassAlter.text))
```
jacobtylerwalls, line 4580: moved above to 4513

music21/musicxml/xmlToM21.py
```diff
@@ -4536,59 +4547,25 @@ def xmlToChordSymbol(self, mxHarmony):
...
-            # set Pitch object on Harmony
-            cs.bass(b)
-        else:
-            cs.bass(r)  # set the bass to the root if root is none
-            # only for the purpose of getting _updatePitches() to run
-            # then, we will need to get rid of this untruthful override
-            # since there may be an inversion!
-            clearBassOverride = True
-
```
jacobtylerwalls, line 4589: this is one of the 3 hacky blocks being jettisoned

music21/musicxml/xmlToM21.py
```diff
@@ -4597,19 +4574,21 @@ def xmlToChordSymbol(self, mxHarmony):
             # TODO: - should allow float, but meaningless to allow microtones in this context.
             seta(hd, mxDegree, 'degree-alter', 'interval', transform=int)
             seta(hd, mxDegree, 'degree-type', 'modType')
-            cs.addChordStepModification(hd, updatePitches=False)
```
jacobtylerwalls, line 4600: moved to 4577, but with updatePitches=True, which we added in #667

music21/musicxml/xmlToM21.py
```diff
@@ -4597,19 +4574,21 @@ def xmlToChordSymbol(self, mxHarmony):
...
-
-        if cs.chordKind != 'none':
-            cs._updatePitches()
-            # environLocal.printDebug(['xmlToChordSymbol(): Harmony object', h])
-            if cs.root().name != r.name:
-                cs.root(r)
```
jacobtylerwalls, line 4606: hacky block 2/3 to remove

music21/musicxml/xmlToM21.py
```diff
@@ -4597,19 +4574,21 @@ def xmlToChordSymbol(self, mxHarmony):
...
-
-        if clearBassOverride and cs.inversion() is not None:
-            # 0 inversion = step 1; 2nd inversion = step 5; 4th inversion = step 9, etc.
-            wantedStep = (cs.inversion() * 2) + 1
-            cs.bass(cs.getChordStep(wantedStep))
-            cs._updatePitches()
```
jacobtylerwalls, line 4612: hacky block 3/3 to remove (this one's mine!)

music21/musicxml/xmlToM21.py
```diff
@@ -4498,30 +4498,41 @@ def xmlToChordSymbol(self, mxHarmony):
...
+
         mxFrame = mxHarmony.find('frame')
 
-        # environLocal.printDebug(['mxToChordSymbol():', mxHarmony])
-        if mxFrame is not None:
-            cs = tablature.ChordWithFretBoard()
-        elif kindText == 'none':
-            cs = harmony.NoChord()
-        else:
-            cs = harmony.ChordSymbol()
```
jacobtylerwalls, line 4516: moved below to 4551

music21/musicxml/xmlToM21.py
```diff
@@ -4498,30 +4498,41 @@ def xmlToChordSymbol(self, mxHarmony):
...
+        r: Optional[pitch.Pitch] = None
+        inversion: str = ''
+        chordKind: str = ''
+        chordKindStr: str = ''
 
         mxKind = mxHarmony.find('kind')
         if textStripValid(mxKind):
-            kindText = mxKind.text.strip()
-        else:
-            kindText = None
```
jacobtylerwalls, line 4507: `None` and `''` are both Falsy when evaluating kwargs in `_updateBasedOnXMLInput` so I just simplified this, but if you want to preserve `Optional[str]` we could do that. NoChord objects need an affirmative `'none'` string.

jacobtylerwalls (commented):
maybe this will help! no rush, of course, merely strolling by during my not-very-precisely-measured pomodoro breaks.

music21/chord/__init__.py
mscuthbert: Ah -- bass can also be overridden to a different octave.

music21/harmony.py
```diff
@@ -245,28 +246,30 @@ def _updatePitches(self):
...
+    def _updateBasedOnXMLInput(self, keywords, *, parseInversion=True):
         '''
         This method must be called twice, once before the pitches
         are rendered, and once after. This is because after the pitches
         are rendered, the root() and bass() becomes reset by the chord class
         but we want the objects to retain their initial root, bass, and inversion
         '''
         for kw in keywords:
+            if not keywords[kw]:
+                continue
```
mscuthbert, line 258: I think that all the keywords should be explicit if parseInversion is going to be added.  It seems weird to have one dict[keywords] and one lone actual keyword.

mscuthbert (commented):
tried to read this again, but didn't get through it.

music21/harmony.py
jacobtylerwalls: True, that could be confusing. Easier to just remove `parseInversion`!

jacobtylerwalls:
> tried to read this again

If you want me to split this up so that each major block being moved is in its own commit, let me know, that shouldn't take too long.

jacobtylerwalls:
Noticed this branch fixes this bug:

```python
>>> cs = harmony.ChordSymbol("C")
>>> cs.inversion(1)
  File "/Users/jwalls/music21/music21/chord/__init__.py", line 2017, in inversion
    raise ChordException('Could not invert chord...inversion may not exist')
music21.chord.ChordException: Could not invert chord...inversion may not exist
```

mscuthbert (changes requested):
I've finally had a chance to read all this and the overall concept is approved; there are some changes requested though that I want to make sure work.

Going to immediately do a push with docs updates that I made while reviewing this.

music21/harmony.py
```diff
@@ -253,20 +256,22 @@ def _updateBasedOnXMLInput(self, keywords):
         but we want the objects to retain their initial root, bass, and inversion
         '''
         for kw in keywords:
+            if not keywords[kw]:
```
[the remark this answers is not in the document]
jacobtylerwalls: Yes. I think it can stay private, though.

music21/chord/__init__.py
```diff
@@ -1040,23 +1074,30 @@ def bass(self,
 
             if not foundBassInChord:  # it's not there, needs to be added
                 self.pitches = (newbass, *(p for p in self.pitches))
+            # to be changed in v8 -- do not add this bass to the chord itself.
```
jacobtylerwalls, line 1077: is the idea to raise an Exception?

mscuthbert: That's probably the best solution.  In any case it'd be nice if bass() worked very similarly to root().  There's a good reason to want to have a root that is not present in the chord or is not the normal one that is there.  To add a new bass note, all you need to do is append a note to the chord -- and if you're not sure of the octave you can use transposeBelowPitch (or whatever the name of that function is).

PR #885, opened 2021-03-04
Combinatoriality: DRY compression

MarkGotham:


music21/serial.py
```diff
@@ -1059,27 +1059,12 @@ def areCombinatorial(self, transType1, index1, transType2, index2, convention):
...
+            trans1 = getattr(self, convention + 'CenteredTransformation')(transType1, index1)
             pitches1 = trans1.pitchClasses()
-            trans2 = self.originalCenteredTransformation(transType2, index2)
+            trans2 = getattr(self, convention + 'CenteredTransformation')(transType2, index2)
             pitches2 = trans2.pitchClasses()
-            for i in range(6):
-                testRow.append(pitches1[i])
-            for i in range(6):
-                testRow.append(pitches2[6 + i])
+            testRow = pitches1[:6] + pitches2[:6]
```
jacobtylerwalls, line 1067: Hi Mark -- is there a music-theoretical reason this is equivalent to `pitches2[6:]` above for `original`? It's been a while ...

jacobtylerwalls: Indeed, the computation is different now. Print the `testRow` and run the second example in the doctest:

master:
[5, 4, 1, 11, 10, 0, 11, 9, 10, 8, 2, 3]
PR:
[5, 4, 1, 11, 10, 0, 0, 1, 4, 6, 7, 5]

music21/serial.py
```diff
@@ -1059,27 +1059,12 @@ def areCombinatorial(self, transType1, index1, transType2, index2, convention):
...
-                testRow.append(pitches2[i])
-            return pcToToneRow(testRow).isTwelveToneRow()
-        elif convention == 'original':
-            testRow = []
-            trans1 = self.originalCenteredTransformation(transType1, index1)
+        if convention in ['zero', 'original']:
+            trans1 = getattr(self, convention + 'CenteredTransformation')(transType1, index1)
             pitches1 = trans1.pitchClasses()
-            trans2 = self.originalCenteredTransformation(transType2, index2)
+            trans2 = getattr(self, convention + 'CenteredTransformation')(transType2, index2)
```
jacobtylerwalls, line 1065: Seems to me this is brittle against the possibility of future renaming.

mscuthbert:
It's hard to find getattr that are constructed from strings -- there needs to be a strong reason for setting them.  Does this speed anything up?  It looks like it may just be harder to maintain.  Need a justification for this PR.

MarkGotham:
Thanks for the review @mscuthbert and @jacobtylerwalls. I aimed for a small PR here, but there are already several considerations to unpick, including both the style and the substance:

**1. getattr:** Ok, I see why we wouldn't use getattr for compressions in this way. Noted, retracted, and no problem.

**2. Other compression:** All the same, there's still room for compression here (implemented in the latest version), not least because ... 

**3. Logic.** In trying to do a simple tidy up, we've chanced upon some incorrect logic. The combinatoriality comparison has to be between the first hexachords of the two rows whether using `zeroCenteredTransformation` or `originalCenteredTransformation`. Previously the comparison ran differently for those two cases and the test / demo simply asserted `False` so didn't pick up on the issue. This PR fixes that and implements more demos / docs.

**4. A bigger change** (not implemented for now but worth considering): As long as this method returns only True/False, there's no reason to offer the option of which convention to use, or to implement the alternatives. If it returned pitches or the like then it would make sense, but as it stands we could remove the whole convention option without loss of functionality.

music21/serial.py
MarkGotham: Thanks for flagging this up. I've responded in the summary below.

music21/serial.py
MarkGotham: Noted. Thanks

music21/serial.py
```diff
@@ -1048,42 +1048,58 @@ def areCombinatorial(self, transType1, index1, transType2, index2, convention):
...
 
+        First, let's take a row we know to have a combinatoriality pair:
+
         >>> moses = serial.getHistoricalRowByName('SchoenbergMosesAron')
         >>> moses.pitchClasses()
         [9, 10, 4, 2, 3, 1, 7, 5, 6, 8, 11, 0]
-        >>> moses.areCombinatorial('P', 1, 'I', 4, 'zero')
+
+        Combinatoriality holds here between P0 and I3
+        >>> moses.areCombinatorial('P', 0, 'I', 3, 'zero')
```
jacobtylerwalls, line 1058: Need a blank line preceding every code example (see png):

no-blank-line

music21/serial.py
```diff
@@ -1048,42 +1048,58 @@ def areCombinatorial(self, transType1, index1, transType2, index2, convention):
...
+        True
+
+        By definition what's true for :meth:`~music21.serial.zeroCenteredTransformation`
+        is also true for :meth:`~music21.serial.originalCenteredTransformation`:
```
jacobtylerwalls, line 1062: Pre-existing bug, but these links don't render, because these are methods on `ToneRow`. Adding `.ToneRow.` fixes it -- are you game to replace the two sets here and at the top of the docstring, and the third (set) of links on the `findZero...` and `findOriginal...` methods?

jacobtylerwalls (changes requested):
Thanks for finding and explaining the logical error. Noted some small documentation bugs. 

Interested to hear if Myke thinks now is the time to go ahead and remove (or just disregard) the convention parameter.

music21/serial.py
MarkGotham: Sure thing - thanks for catching this.

MarkGotham:
Oops that commit should be `fix 6 links`

mscuthbert:
Oh, I get it -- since we're using the same convention for both transformations, then it doesn't really matter which one you choose.  In that case, let's get rid of it.  It's v.7. so we can break little things like this.  Maybe better though to leave in _unused_convention=None and note that it will be removed in v.8.

mscuthbert:
(btw -- these "I", "RI", "R" + one of "T" or "P" should become Literal types.

jacobtylerwalls:
> Maybe better though to leave in _unused_convention=None and note that it will be removed in v.8.

I think we should just break it. 🔨  
We've been breaking signatures all over the package in order to make things keyword-only, and this method has five non-self positional arguments. v7 is going to be a possibly hard, but super awesome upgrade.

@MarkGotham are you game to just drop the fifth argument? Once we have that I think this is mergeable even if you're not ready to sign up for converting the characters to literal types just yet.

jacobtylerwalls:
> (btw -- these "I", "RI", "R" + one of "T" or "P" should become Literal types.

My understanding is that we can't do this until v. 8 when we drop support for Py 3.7:

https://github.com/cuthbertLab/music21/blob/19cbe27a55b7ab8e79686bf9a9ca88b7a883defb/music21/roman.py#L2655-L2657

MarkGotham:
> > (btw -- these "I", "RI", "R" + one of "T" or "P" should become Literal types.
> 
> My understanding is that we can't do this until v. 8 when we drop support for Py 3.7:
> 
> https://github.com/cuthbertLab/music21/blob/19cbe27a55b7ab8e79686bf9a9ca88b7a883defb/music21/roman.py#L2655-L2657

Thanks. So then the typing for now should be:
`transType1: int['P', 'T', 'R', 'I', 'RI'],`
... with a planned move to `Literal` later?

In summary, do typing for the first 4 arguments, remove the 5th and we're done here?

jacobtylerwalls:
> So then the typing for now should be:
transType1: str['P', 'T', 'R', 'I', 'RI'],
... with a planned move to Literal later?

Honestly I would leave those four arguments alone until the move to `Literal` in v.8. I haven't seen strings typed that way in our package nor in the typing module docs, so there would be a learning curve / cost to building and checking Sphinx for something we're changing in one version from now. Unless Myke wants to involve himself here and offer guidance, but for such a small point I think this is mergeable if we just rip out that fifth parameter and leave a "changed in v7 -- " notice.

A ValueError might be nice if someone passes a garbage string, but I wouldn't insist on it to merge this PR.

mscuthbert:
str is not a Generic type and can't be typed with literals in mypy

jacobtylerwalls (approved):

jacobtylerwalls:
Mark, would you like to take a look at the changes I just pushed and see if they look sensible to you? If so, then I think we can merge this next week. Thanks for getting this patch in!

MarkGotham:
Looks good, thanks!

PR #886, opened 2021-03-04
Provide more examples of setting .id on Voice objects

jacobtylerwalls:
<s>Figured it would be a small win to at least print .id rather than the memory location, but was this considered and rejected before?</s> <s>The MusicXML parser sets random numbers for voice IDs--is there a desire to hide this from users?</s>

EDIT: I see this is done after all -- just a matter of sprucing up some examples to model that this works!

### DIFF
--- a/documentation/source/usersGuide/usersGuide_58_sitesContext.ipynb
+++ b/documentation/source/usersGuide/usersGuide_58_sitesContext.ipynb
@@ -217,8 +217,7 @@
     }
    ],
    "source": [
-    "v = stream.Voice()\n",
-    "v.id = 1\n",
+    "v = stream.Voice(id=1)\n",
     "v.append(es)\n",
     "es.sites.getSiteCount()"
    ]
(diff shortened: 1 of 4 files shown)


jacobtylerwalls:
Alright, now that I know what's going on here, my suggestion is for `makeVoices()` to just set the voice numbers, just like `makeNotation` and the MusicXML parser so that MIDI, Humdrum, etc., can get nice defaults. I could do that on this PR (should have opened an issue, sorry...)

`makeNotation` sets the voice .id with an int, and the musicXML parser sets a string. Good time to define this behavior? New attribute `.number` or `.voiceNumber`?

https://github.com/cuthbertLab/music21/blob/ef4bdd6ae787ffb189486a126e6f1a2ee0b556c7/music21/musicxml/xmlToM21.py#L5616

mscuthbert:
This is a great idea -- let me think about whether we should explicitly define `voice.number` separately from ID -- the Q is whether it's universal across the score, within a PartStaff or across parts.  having it across PartStaffs would help our friend trying to track the tenor line -- IF it was set that way in musicxml.  But what happens if it ends and then begins again?

PR #887, opened 2021-03-05
Flatten unnecessary voices in makeTies

jacobtylerwalls:
Noticed `makeTies` had a comment about leaving unnecessary voices behind, and figured that was easy to clean up with a call to `flattenUnnecessaryVoices()`.

Then promoted the doctest where that was demo'd (along with some version changed annotations) above the OMIT_FROM_DOCS line.

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -6269,8 +6269,6 @@ class Test(unittest.TestCase):
         n = note.Note()
         v1 = stream.Voice([n])
         m = stream.Measure([v1])
-        xmlOut = self.getXml(m)
-        self.assertIn('<voice>1</voice>', xmlOut)
         n2 = note.Note()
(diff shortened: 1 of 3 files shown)


music21/musicxml/m21ToXml.py
```diff
@@ -6269,8 +6269,6 @@ def testLowVoiceNumbers(self):
         n = note.Note()
         v1 = stream.Voice([n])
         m = stream.Measure([v1])
-        xmlOut = self.getXml(m)
-        self.assertIn('<voice>1</voice>', xmlOut)
```
mscuthbert, line 6273: Maybe better to change to `self.assertNotIn()` since we want to test that unnecessary voices get flattened.

mscuthbert (changes requested):
Super great -- can we restore one test but change it to assertNotIn instead?

mscuthbert:
Thanks!

PR #892, opened 2021-03-06
Faster implementation of hyphenToCamelCase

jacobtylerwalls:
cProfile shows me this will speed up musicxml parsing by about 8%, since compiling a regex every time a note is parsed is expensive.

master
```
>>> from timeit import timeit
>>> timeit('music21.common.hyphenToCamelCase("relative-x")', setup='import music21')
4.062028805999999
```

PR
```
>>> from timeit import timeit
>>> timeit('music21.common.hyphenToCamelCase("relative-x")', setup='import music21')
1.016820898000006
```

The only difference in behavior is this, but I see this as acceptable (we don't have XML attributes named with numbers), and I documented it:

master
```
>>> music21.common.hyphenToCamelCase('music-21')
'music-21'
```

PR
```
>>> music21.common.hyphenToCamelCase('music-21')
'music21'
```

I could add a check for this to get the old behavior, but I don't think we need the old behavior.

### DIFF
--- a/music21/common/stringTools.py
+++ b/music21/common/stringTools.py
@@ -87,33 +87,34 @@ def getNumFromStr(usrStr: str, numbers: str = '0123456789') -> Tuple[str, str]:
 
 def hyphenToCamelCase(usrStr: str, replacement: str = '-') -> str:
     '''
-    given a hyphen-connected-string, change it to
+    Given a hyphen-connected-string, change it to
     a camelCaseConnectedString.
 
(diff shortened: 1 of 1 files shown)


mscuthbert:
Thanks!  :-)  Speedups are good!

(we probably could have gotten a very good speedup just by compiling the standard case of replacement = '-' and using that if hyphen was called.  But this is just as good, and music-21 should go to music21 anyhow!)

PR #893, opened 2021-03-06
Remove redundant calls to setColor() and setPosition()

jacobtylerwalls:
Import and export versions of `setPrintStyle()` call `setColor()`, so additional calls to `setColor()` are redundant.
https://github.com/cuthbertLab/music21/blob/1d917269a9cc64db4783cb67ee972a55b4a7fb40/music21/musicxml/xmlToM21.py#L460-L469
https://github.com/cuthbertLab/music21/blob/ef4bdd6ae787ffb189486a126e6f1a2ee0b556c7/music21/musicxml/m21ToXml.py#L824-L833

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -3274,7 +3274,7 @@ class MeasureExporter(XMLExporterBase):
             chordOrN = n
         else:
             chordOrN = chordParent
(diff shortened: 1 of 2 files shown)


mscuthbert:
Thank you!  Looks great to me!  Color was done early on in the translation process, which is probably why it was not caught.

PR #895, opened 2021-03-06
Speed up isIterable()

jacobtylerwalls:


mscuthbert:
single quotes around `'__iter__'` otherwise THANK YOU great!

mscuthbert:
Thanks!  I wonder if we can get rid of "bytes" and consider that iterable without consequence.  having it there is a major slowdown for checking if a list is iterable.

PR #896, opened 2021-03-06
Let Style objects inherit from ProtoM21Object

jacobtylerwalls:
```
>>> from music21 import style
>>> s = style.Style()
>>> if 'Style' in s.classes:
...    print('make toast')
... 
AttributeError: 'Style' object has no attribute 'classes'
```


mscuthbert:
Thanks!

PR #897, opened 2021-03-06
Restore two coreElementsChanged() calls

jacobtylerwalls:
Restore two calls removed in #803
- flattenUnnecessaryVoices calls remove(), which calls coreElementsChanged, but remove() might need correct offsets when doing its work
- And activateVariants() just looks like a review comment was missed.


mscuthbert:
Thank you!

PR #898, opened 2021-03-07
Speed up elementOffset()

jacobtylerwalls:
Slightly slower for 'highestTime' but faster for everything else. musicXML parses 20% faster.

master
```
>>> timeit('s.elementOffset(n)',
setup='from music21 import note, stream;n = note.Note();s = stream.Stream([n])')
3.110492326
```

PR
```
>>> timeit('s.elementOffset(n)',
setup='from music21 import note, stream;n = note.Note();s = stream.Stream([n])')
0.410327549999991
```

### DIFF
--- a/music21/stream/base.py
+++ b/music21/stream/base.py
@@ -1849,7 +1849,8 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
                 raise base.SitesException(
                     f'an entry for this object 0x{id(element):x} is not stored in stream {self}')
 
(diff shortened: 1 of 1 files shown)


mscuthbert:
THANK YOU!

probably can go even faster with `if type(x) is str` -- no one should subclass str.

jacobtylerwalls:
True. I tested real quickly and saw marginal improvements for floats and marginal degradation for strings, at least running Python 3.9.2. So not a clear win for now, but maybe someday

PR #903, opened 2021-03-08
voicesToParts() now preserves Instruments

### LINKED ISSUE #902: Attribute 'Instrument' of a two-voices part is lost when applying Stream.voicesToParts() method
Version of music21:  6.5.0
Good morning! Here martín from DIDONE project. First of all, thanks a lot for creating this library and keeping the support and improvements. I just installed the new 6.7 version but i doesn't seem to have solved my problem, so here I open an issue in case is useful for anyone:

**Problem summary**
Attribute 'Instrument' of a two-voices part is lost when applying Stream.voicesToParts() method. When having a Stream that defines two voices (in this case two wind insturments), I wanna use this method to split them in two andoperate with them separatedly. 
The point is that BEFORE the method I can run part.getInstrument() and get something like '<music21.instrument.Instrument 'P1: Oboi: SmartMusic SoftSynth 1'>' . But, after the method is called, this information is lost and the new parts that correspond to the previous don't have this attrubute, so if you call part.getInstrument() you just get an empty string. Therefore, afterwards I have no way to identify the type of the instrument.

**Steps to reproduce**

My code to separate the voices goes like follows:

            if has_voices:
                parts_splited = part.voicesToParts().elements
                num_measure = 0
                for measure in part.elements: #add missing information to both parts (dynamics, text annotations, etc are missing)
                    if isinstance(measure, stream.Measure) and any(not isinstance(e, stream.Voice) for e in measure.elements):
                        not_voices_elements = [e for e in measure.elements if not isinstance(e, stream.Voice)] #elements such as clefs, dynamics, text annotations...
                        for p in parts_splited:
                            p.elements[num_measure].elements += tuple(e for e in not_voices_elements if e not in p.elements[num_measure].elements)
              

And it works just fine. I just would like to call later parts_splited[0].getInstrument() to identify it.
I have come with 'kind of a ' workaround in the source code, but I think I needs expert hands on it. In the definition of the method (file __init__.py of stream class), I just added this line to get all the information of the original part to the subparts:

        for i in range(partCount):
            p = Part()
            # p.elements = self.elements #
            s.insert(0, p)

in lines 10402-10404. It does preserve the info, however some aspects of the object now chenged (looking at my code, the first elements of p.elements is not longer a measure but a instrument, so the iteration just doesen't work) and therefore I didnt wanted to go on with it yet not to rewriter all the code without knowing if its agood idea to make this change in the source code or not.

In any case, I just wanted to let you know about this issue in the method and hope it works to help to improve the library.

Hope to hear from you soon. 

Kind regards.

jacobtylerwalls:
Fixes #902 

Depends on the Instrument being at the level containing the Measure (such as Part), which is the standard.

Continues to use `mergeElements` like the rest of the routine -- which will create shallow references, so the same `Instrument` will be in both parts in the resulting score. But it appears that is how the rest of the routine works today.

### DIFF
--- a/documentation/source/usersGuide/usersGuide_12_music21object.ipynb
+++ b/documentation/source/usersGuide/usersGuide_12_music21object.ipynb
@@ -850,7 +850,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "A `Music21Object` that is inside one or more Streams should be able to get its most recently stream via its `.activeSite` attribute.  We've put `n` in `s`, which is called (now incorrectly) `'empty stream'`, so n's `.activeSite` should be `s`."
+    "A `Music21Object` that is inside one or more Streams should be able to get its most recently referenced stream via its `.activeSite` attribute.  We've put `n` in `s`, which is called (now incorrectly) `'empty stream'`, so n's `.activeSite` should be `s`."
    ]
   },
   {
--- a/music21/stream/base.py
+++ b/music21/stream/base.py
@@ -10608,6 +10608,7 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
         >>> ce = c.voicesToParts()
         >>> ce.show('t')
         {0.0} <music21.stream.Part Piano-v0>
+            {0.0} <music21.instrument.Instrument 'P1: Piano: '>
             {0.0} <music21.stream.Measure 1 offset=0.0>
                 {0.0} <music21.clef.TrebleClef>
                 {0.0} <music21.key.Key of D major>
@@ -10623,6 +10624,7 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
                 {0.0} <music21.note.Rest rest>
                 {4.0} <music21.bar.Barline type=final>
         {0.0} <music21.stream.Part Piano-v1>
+            {0.0} <music21.instrument.Instrument 'P1: Piano: '>
             {0.0} <music21.stream.Measure 1 offset=0.0>
                 {0.0} <music21.clef.BassClef>
                 {0.0} <music21.key.Key of D major>
@@ -10756,6 +10758,10 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
                         continue
                     pInner = partDict[voiceIdInner]
                     pInner.insert(self.elementOffset(mInner), copy.deepcopy(mActive))
+
+        # Place references to any instruments from the original part into the new parts
+        for p in s.parts:
+            p.mergeElements(self, classFilterList=('Instrument',))
 
         if self.hasMeasures():
             for m in self.iter.getElementsByClass('Measure'):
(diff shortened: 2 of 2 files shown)


mscuthbert:
won't this end up with two references to the same Instrument object within the same hierarchy making a flat return impossible?

mscuthbert:
(This is why `.flat` needs to become `.flatten()` so we can configure such things to allow for automatically copying or removing second instances to same object).

jacobtylerwalls:
voicesToParts() has always inserted references for everything, as far as I can tell. The test I wrote last summer verifies this, where the same `k` or `ts` is asserted to be in every part:

https://github.com/cuthbertLab/music21/blob/669fbd4bb94e3b7bc4b6f78241014c9c7d6789df/music21/stream/tests.py#L5606-L5632

```
>>> vp = s.voicesToParts()  # stream from above test
>>> for ks in vp.recurse().getElementsByClass('Key'):
...   print(ks, id(ks))
... 
E major 4504293136
E major 4504293136
```

I take it this is not acceptable given https://github.com/cuthbertLab/music21/pull/734#issuecomment-773727019

But it's not .flat or caching that seems to be a problem. What's going wrong in the other projects?

on main

```
>>> vp = score from the above test.voicesToParts()
>>> vpf = vp.flat
>>> for e in vpf:
...   print(e, id(e))
... 
<music21.bar.Barline type=regular> 4873756624  **
<music21.bar.Barline type=regular> 4873756624  **
<music21.clef.TrebleClef> 4873867472
<music21.clef.TrebleClef> 4873870688
E major 4427029280  **
E major 4427029280  **
<music21.meter.TimeSignature 2/4> 4873153312  **
<music21.meter.TimeSignature 2/4> 4873153312  **
<music21.note.Note C> 4873508848
<music21.note.Note C> 4873756192
<music21.note.Note C> 4873756480
<music21.bar.Barline type=final> 4873867424
<music21.bar.Barline type=final> 4873867328
```

jacobtylerwalls:
I see -- it's the sheer fact that it's not helpful from a stream manipulation standpoint:

PR
```
>>> s = stream.Stream()
>>> s.repeatAppend(stream.Voice(), 2)
>>> s.insert(0, instrument.Violin())
>>> vp = s.voicesToParts()
>>> vp.first().first()
<music21.instrument.Violin 'Violin'>
>>> vp.first().first().partName = 'Outside'
>>> vp.show('t')
{0.0} <music21.stream.Part 4402836672-v0>
    {0.0} <music21.instrument.Violin 'Outside: Violin'>
{0.0} <music21.stream.Part 4402836672-v1>
    {0.0} <music21.instrument.Violin 'Outside: Violin'>
```

So this routine just shouldn't call `mergeElements()` anywhere.

mscuthbert:
I think that we need more guards on putting the same object in the same hierarchy twice.  

OR, finally bite the bullet and do what I wanted from the beginning -- allow the same object to appear in the same hierarchy more than once.  It means a major refactor of the Core storage to allow for lists of offsets for an object rather than a single offset.  I think that all normal ways of approaching offsets ( .getOffsetBySite(), etc.) would still return a single number (the first one in the list), but we would need specialized ways of getting all the offsets.

The latter would be major work.

PR #901, opened 2021-03-08
getSpannedElementIds() checks _elements

jacobtylerwalls:
Dramatic speed improvement for musicxml export.

<s>95%</s> 76% of the time in musicxml export after `makeNotation()` for this corpus example --i.e. inside `parseWellFormedObject()`--is spent in `getSpannedElementIds()`:

EDIT: updated to use `ScoreExporter` to ensure I wasn't counting calls during makeNotation:

master
```
>>> s = corpus.parse('cpebach')
>>> sx = musicxml.m21ToXml.ScoreExporter(s)
>>> cProfile.run('sx.parse()', sort='cumtime')
         3441542 function calls (3429186 primitive calls) in 2.838 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    2.839    2.839 {built-in method builtins.exec}
        1    0.000    0.000    2.839    2.839 <string>:1(<module>)
        1    0.000    0.000    2.839    2.839 m21ToXml.py:1361(parse)
        1    0.000    0.000    2.672    2.672 m21ToXml.py:1543(parsePartlikeScore)
        2    0.001    0.000    2.658    1.329 m21ToXml.py:2349(parse)
       64    0.001    0.000    2.635    0.041 m21ToXml.py:2699(parse)
       64    0.000    0.000    2.425    0.038 m21ToXml.py:2719(mainElementsParse)
       78    0.006    0.000    2.422    0.031 m21ToXml.py:2744(parseFlatElements)
      942    0.012    0.000    2.384    0.003 m21ToXml.py:2813(parseOneElement)
     1006    0.078    0.000    2.257    0.002 spanner.py:746(getBySpannedElement)
   137822    0.165    0.000    2.178    0.000 spanner.py:396(getSpannedElementIds)
   137822    0.183    0.000    1.311    0.000 spanner.py:400(<listcomp>)
   415643    0.779    0.000    1.148    0.000 iterator.py:156(__next__)
   138463    0.109    0.000    0.707    0.000 base.py:291(__iter__)
   139634    0.376    0.000    0.609    0.000 iterator.py:92(__init__)
   419959    0.298    0.000    0.298    0.000 iterator.py:525(updateActiveInformation)

```

PR
```
>>> cProfile.run('sx.parse()', sort='cumtime')
         1236511 function calls (1224153 primitive calls) in 0.770 seconds
```

master
```
[...]

PR #905, opened 2021-03-09
Further spanner storage optimizations

jacobtylerwalls:
Sorry for the specific form of torture of getting multiple PRs a day, but it occurred to me (and didn't take long) to check the git blame for the spanner change we just merged in #901.

It was in b9929ef where there were changes system-wide to identity/contains, and I spotted a few other changes from that commit that should be reverted relating to spanner storage.

Update: added another speed improvement -- instead of getting the entire list of spannedElementIds and then checking if `in`, we can just check if the element is in the spanner, which uses the now-optimized `__contains__`, so that we can short-circuit when we find the element.

### DIFF
--- a/music21/spanner.py
+++ b/music21/spanner.py
@@ -346,7 +346,7 @@ class Spanner(base.Music21Object):
         return iter(self.spannerStorage)
 
     def __len__(self):
-        return len(self.spannerStorage)
+        return len(self.spannerStorage._elements)
 
     def getSpannedElements(self):
         '''
@@ -370,10 +370,7 @@ class Spanner(base.Music21Object):
         >>> sl.getSpannedElements() == [n1, n2, c1]  # make sure that not sorting
         True
         '''
-        post = []
-        for c in self.spannerStorage.elements:
-            post.append(c)
-        return post
+        return list(self.spannerStorage._elements)
 
     def getSpannedElementsByClass(self, classFilterList):
         '''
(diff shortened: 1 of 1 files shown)


mscuthbert:
I think that this is great, but I want to think a bit about, if I do `slur = score.recurse().getElementByClass('Slur')` and then later I do `for n in score.flat.stripTies().notes: if n in slur: do something` -- do I care that the answer has changed.  I think I don't: `n` is not in slur, n's `.derivation.origin` is in slur, not `n`.  So I think that this is an improvement, but just want to think for one more day about it.

jacobtylerwalls:
Sure, happy to think this through. Where is the bit you are thinking will cause a change in functionality? There shouldn't be any different results (tests fail if I fiddle with it in other ways), since previously `return spannedElement in self.spannerStorage` would run the generic Stream `__contains__()`, which also checks identity:

https://github.com/cuthbertLab/music21/blob/e07d7ecb77f4dc0f9390709c058449156107810b/music21/stream/base.py#L538-L541

So it's the same pattern, just reimplemented to avoid the StreamIterator overhead of checking `.elements` which hits the cache and possibly sorts etc

And then replacing `if idTarget in sp.getSpannedElementIds():` should give the same result since it was also already checking identity: `id()`.

mscuthbert:
> since previously return spannedElement in self.spannerStorage would run the generic Stream __contains__(), which also checks identity

Ah yes, You're right!  Merging, thanks.

PR #908, opened 2021-03-10
Add attribute expressionIsInferred to Durations

jacobtylerwalls:
Cherry-picked from #763 so that we can attack this in manageable pieces.


mscuthbert (approved):
really great.  A good manageable chunk

PR #911, opened 2021-03-11
Miscellany

### LINKED ISSUE #831: braille.segment.prepareSlurredNotes indexes into spanners unsafely
use slur.getFirst() and wrap in try block instead of expecting slur[0] to work.

_Originally posted by @jacobtylerwalls in https://github.com/cuthbertLab/music21/pull/826#discussion_r571676014_

jacobtylerwalls:
- Fix doc for humdrum `parseData()`
- Remove check for not hasattr(st, 'lineWidths')

- Model calling list() instead of  .elements in doctest

- Fixes #831 -- use `getFirst()` to index into slurs

- Documentation cleanup in partStaffExporter (Sphinx problem with python reserved word)

- Bugfix in `previous()` -- put className in a tuple before passing as classList

- Speed up `Note.__eq__()`

- Prefer tuples to lists in `setStyleAttributes` (hottest function in parser)

- Slight speed up to `Iterator.__next__()`
    Avoid doing self.index - 1

- Optimize `addSpannedElements()`

- Speed up `_setAbsoluteY()` for values 'above' and 'below'
    `numToIntOrFloat()` was being called.

- More uses of `.first()`

### DIFF
--- a/music21/tree/fromStream.py
+++ b/music21/tree/fromStream.py
@@ -182,7 +182,7 @@ def asTree(inputStream, flatten=False, classList=None, useTimespans=False, group
     >>> etFlat.getPositionAfter(0.5)
     SortTuple(atEnd=0, offset=1.0, priority=0, classSortOrder=20, isNotGrace=1, insertIndex=...)
 
-    >>> etFlatNotes = tree.fromStream.asTree(score, flatten=True, classList=[note.Note])
+    >>> etFlatNotes = tree.fromStream.asTree(score, flatten=True, classList=(note.Note,))
     >>> etFlatNotes
     <ElementTree {12} (0.0 <0.20...> to 8.0) <music21.stream.Score exampleScore>>
 
--- a/music21/base.py
+++ b/music21/base.py
@@ -2055,6 +2055,8 @@ class Music21Object(prebase.ProtoM21Object):
             activeS = self.activeSite  # might be None...
             if activeS is None:
                 return None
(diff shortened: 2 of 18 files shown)


music21/tree/fromStream.py
```diff
@@ -182,7 +182,7 @@ def asTree(inputStream, flatten=False, classList=None, useTimespans=False, group
     >>> etFlat.getPositionAfter(0.5)
     SortTuple(atEnd=0, offset=1.0, priority=0, classSortOrder=20, isNotGrace=1, insertIndex=...)
 
-    >>> etFlatNotes = tree.fromStream.asTree(score, flatten=True, classList=[note.Note])
+    >>> etFlatNotes = tree.fromStream.asTree(score, flatten=True, classList=(note.Note,))
```
mscuthbert, line 185: In docs, I tend to pass lists because tuples with trailing commas confuse new programmers.  Here it doesn't matter because beginners shouldn't be working with `tree`, but just FYI.

mscuthbert (approved):
Looks good to me -- one comment on the differences between docs and programming below.

In docs, I probably won't be using `first()` and `last()` until v.8, very much, because otherwise as soon as v7 comes out we'll get a lot of Qs about "I tried .first() and it said AttributeError" so if there's an older way of doing something that's still valid I usually wait one full version before pushing forward on it.  Just experience doing music21 support.

PR #913, opened 2021-03-12
Full measure rest fixes

jacobtylerwalls:
- Avoid splitting complex full measure rests on export (e.g. 4.5QL rest in 9/8 was split into 4.0 QL and 0.5QL)
- Set correct attribute `.fullMeasure` on import


mscuthbert:
Thanks!

PR #916, opened 2021-03-13
Search derivations and id when finding joinableGroups

jacobtylerwalls:
The patch in #863 to fix showing flat scores was too restrictive. Noticed today when trying to run `s.measure(1).show()`. 

@mscuthbert contributed logic for this in dcb32fc946, so my proposal is to just run it up front when filtering the joinable groups.


mscuthbert:
Looks good to me -- thanks for finding it!

PR #917, opened 2021-03-13
Update a comment that references python-Levenshtein

dhgutteridge:
python-Levenshtein (the name it goes by, rather than "pyLevenshtein") is now
available on pypi.org (https://pypi.org/project/python-Levenshtein/) and can
be installed via pip. Update the reference to this newer version (which has
fixed bugs with potential security issues that were present in the older
version). While here, also mention the forceDifflib parameter.


mscuthbert:
Thank you so much for this AMAZING library -- and for updating it to be pip installable.

PR #919, opened 2021-03-14
Barline doesn't need to inherit from StyleMixin

### LINKED ISSUE #918: Barlines in thawed streams raise AttributeErrors when accessing style
**music21 version**

`7.0.1`

**Problem summary**
Unserializing a stream with a barline corrupts the `_style` attribute. Probably something to do with multiple inheritance and properties doing fancy footwork.

**Steps to reproduce**

```
>>> s = stream.Stream([bar.Barline()])
>>> data = converter.freezeStr(s, fmt='pickle')
>>> s2 = converter.thawStr(data)
>>> thawedBarline = s2.first()
>>> thawedBarline.hasStyleInformation
  File "/Users/jwalls/music21/music21/base.py", line 675, in hasStyleInformation
    return not (self._style is None)
AttributeError: _style
>>> '_style' in dir(thawedBarline)
True
>>> hasattr(thawedBarline, '_style')
False
```

**More information**
Encountered this loading a cached file from the corpus (writing a unit test for an upcoming unrelated patch).

I believe this is the only class to inherit from these two classes, each of which have slightly different implementations of `hasStyleInformation`.
https://github.com/cuthbertLab/music21/blob/747316f350319f0832d704c5d06a0c0bc2f07cbe/music21/bar.py#L96

- We could consolidate to just have one implementation of that method on `base.Music21Object` which catches AttributeError (which you would think would be unnecessary), which seemed to make my problems go away, but that could be masking something deeper. OK for now, I guess?

jacobtylerwalls:
Fixes #918

### DIFF
--- a/music21/_version.py
+++ b/music21/_version.py
@@ -42,7 +42,7 @@ When changing, update the single test case in base.py.
 Changing this number invalidates old pickles -- do it if the old pickles create a problem.
 '''
 
-__version_info__ = (7, 0, 1)  # can be 4-tuple: (7, 0, 5, 'a2')
+__version_info__ = (7, 0, 2)  # can be 4-tuple: (7, 0, 5, 'a2')
 
 v = '.'.join(str(x) for x in __version_info__[0:3])
 if len(__version_info__) > 3 and __version_info__[3]:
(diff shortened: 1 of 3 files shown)


mscuthbert (approved):
Thanks!

PR #920, opened 2021-03-16
returns downward directions for stems in tabs

### LINKED ISSUE #912: Stem problem when exporting a tab stream in MusicXML
**music21 version**

`7.0.1`

**Problem summary**
When writing a tab stream in MusicXML, it looks like m21 tries to make stems and that makes an error as a TabClef can't return an appropriate direction for stems (function `getStemDirectionForPitches` in `clef.py`) :

```
music21/music21/clef.py", line 262, in getStemDirectionForPitches
    midLine = self.lowestLine + 4
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'
```

Apparently, stems are rather rare in tablatures, but when they appear, they are all in the same direction, generally downwards (see [this link](http://www.xx-europe.de/tablature-breizh-guitare-mandocello-bouzouki-cistre.html) for example).

Would it be appropriate when the clef is TabClef to either disable stems either force them to be downwards ?

PS : curiously the error does not appear when omitting the time signature, but I was not able to see the relation here.

**Steps to reproduce**

```
from music21 import *

test_tab = stream.Stream()
p = stream.Part()
m = stream.Measure()
c = clef.TabClef()

n1 = note.Note(64,quarterLength=0.25)
s1 = articulations.StringIndication()
s1.number = 2
f1 = articulations.FretIndication()
f1.number = 5
n1.articulations = [s1,f1]

n2 = note.Note(67,quarterLength=0.25)
s2 = articulations.StringIndication()
s2.number = 2
f2 = articulations.FretIndication()
f2.number = 8
n2.articulations = [s2,f2]

m.append(clef.TabClef())
m.append(meter.TimeSignature('4/4'))
m.append(n1)
m.append(n2)

p.append(m)

test_tab.append(p)
test_tab.show('text')
test_tab.write('musicxml','output')
```

louisbigo:
Fixes #912

### DIFF
--- a/music21/clef.py
+++ b/music21/clef.py
@@ -365,10 +365,6 @@ class TabClef(PitchClef):
     >>> a = clef.TabClef()
     >>> a.sign
     'TAB'
-
-    As tab stems are assumed to be mostly downwards (at least for guitar tabs),
-    the method `getStemDirectionForPitches` is overriden from `Clef` to
-    systematically return "down".
     '''
 
     def __init__(self):
@@ -383,7 +379,10 @@ class TabClef(PitchClef):
         firstLastOnly: bool = True,
         extremePitchOnly: bool = False,
     ) -> str:
-        return "down"
+        '''
+        Overridden to simply return 'down' for guitar tabs.
+        '''
+        return 'down'
 
 # ------------------------------------------------------------------------------
 
--- a/music21/clef.py
+++ b/music21/clef.py
@@ -1113,7 +1113,7 @@ class Test(unittest.TestCase):
         m.append(meter.TimeSignature('4/4'))
         m.append(n1)
         m.append(n2)
-        m = m.makeBeams()
+        m.makeBeams(inPlace=True)
 
         self.assertEqual(m.notes[0].stemDirection, 'down')
 
--- a/music21/clef.py
+++ b/music21/clef.py
@@ -365,12 +365,25 @@ class TabClef(PitchClef):
     >>> a = clef.TabClef()
     >>> a.sign
     'TAB'
+    
+    As tab stems are assumed to be mostly downwards (at least for guitar tabs),
+    the method `getStemDirectionForPitches` is overriden from `Clef` to
+    systematically return "down".
     '''
 
     def __init__(self):
         super().__init__()
         self.sign = 'TAB'
         self.line = 5
+
+    def getStemDirectionForPitches(
+        self,
+        pitchList: Union[pitch.Pitch, Iterable[pitch.Pitch]],
+        *,
+        firstLastOnly: bool = True,
+        extremePitchOnly: bool = False,
+    ) -> str:
+        return "down"
 
 # ------------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


music21/clef.py
```diff
@@ -365,13 +365,26 @@ class TabClef(PitchClef):
...
         self.line = 5
 
+    def getStemDirectionForPitches(
+        self,
+        pitchList: Union[pitch.Pitch, Iterable[pitch.Pitch]],
+        *,
+        firstLastOnly: bool = True,
+        extremePitchOnly: bool = False,
+    ) -> str:
+        return "down"
```
jacobtylerwalls, line 386: music21 style is single quotes

music21/clef.py
```diff
@@ -365,13 +365,26 @@ class TabClef(PitchClef):
     >>> a = clef.TabClef()
     >>> a.sign
     'TAB'
+    
+    As tab stems are assumed to be mostly downwards (at least for guitar tabs),
+    the method `getStemDirectionForPitches` is overriden from `Clef` to
```
jacobtylerwalls, line 370: overridden has two "d"s (gets me all the time).

Also, this documentation is already on `getStemDirectionForPitches`, so I don't think you need to refer to the name of the function, but mentioning the polymorphism is good! Something simple like "Overridden to simply return 'down' for guitar tabs." would do, IMO.

jacobtylerwalls (commented):
Thanks for this, and for including a test! Left some pointers to help you spruce this up before Michael takes a look. If you're not familiar with GitHub Actions, you can access all the Pylint violations by clicking the red X. And to see the violations locally before publishing your work, you can `pip install pylint` and then run `pylint music21/file/you/were/editing.py`. Pardon me if you know this already.

Oh, and if in the body of the PR you use a magic GitHub keyword like "Fixed" or "Fixes" before the issue number, it will give a little dashed underline to indicate they're actually linked so that when this is merged, the issue is closed (i.e. Michael doesn't have to go hunt for it and remember to close it). 🎉

louisbigo:
I didn't know about flake checks. Thanks for all the useful tips, I'm learning a lot here.

louisbigo:
So I guess everything is fine now, and I just have to wait for the branch to be merged (I know the PR list is long !) - correct me if I'm wrong and there's something more I should do. Thanks for the help so far.

music21/clef.py
```diff
@@ -1085,6 +1097,26 @@ def testContexts(self):
...
+        m = stream.Measure()
+
+        n1 = note.Note(64, quarterLength=0.25)
+        n2 = note.Note(67, quarterLength=0.25)
+
+        m.append(clef.TabClef())
+        m.append(meter.TimeSignature('4/4'))
+        m.append(n1)
+        m.append(n2)
+        m = m.makeBeams()
```
jacobtylerwalls, line 1116: Very minor detail, but I would just use `inPlace=True` rather than overwriting `m`.

louisbigo: Done. Didn't notice this `inPlace` parameter, thanks.

mscuthbert:
What a perfect first PR  -- thanks Louis and thanks Jacob for helping get this to simple review and approve status!

mscuthbert:
btw -- there are tabs (renaissance, etc.) that are always up.  So I could see in the future TabClef having a configurable "self.overriddenStemDirection = 'down'"  which would let it be more easily changed.  But this gives a sense for how someone would do it anyhow if they were so keen to do.

It also gives a sense of how to do it for percussion clefs that always want up stems.

PR #921, opened 2021-03-17
v7 makeRests defaults inPlace=False

jacobtylerwalls:
All but a handful of calls in the test suite set the `inPlace` keyword already, so this was an easy change.

(side note -- the "obvious" problems... have to do with placing rests inside measures, in the last example?)
Update: handled in #922, so removed the TODO line here to avoid a forseeable merge conflict.


mscuthbert:
LGTM!

A v7 cleanup would also be to make some of these `.makeX` calls all keyword only.

PR #923, opened 2021-03-19
Set position on dynamics, wedges, and tempos on import

jacobtylerwalls:
Set position on `<direction>` elements on import, which wasn't handled.

- Set position on dynamics, spanners, and metronome marks
-- Fixes an issue with vocal dynamics not appearing above the staff, e.g. in the Beach corpus example. (placement='above' was not sufficient; also needed the y-values)
- Also set relative positions on coda/segno elements (only default positions were handled)
- Add tests for position import for all 5 direction types (TextExpressions were already working, but the other 4 had changes; just test all 5)

The file parsing shouldn't add much time to the test suite; the Beach file is parsed elsewhere, so the cached version loads here (in fact, this was how I found the bug with barlines in cached streams -- by trying to use `if el.hasStyleInformation` in the list comprehension below).

### DIFF
--- a/documentation/source/about/faq.rst
+++ b/documentation/source/about/faq.rst
@@ -95,7 +95,7 @@ Can I synthesize new sounds with `music21`?
 
     Yes, and no.  `Music21` is primarily built for manipulating symbolic 
     musical data not sound waves.  There are lots of great programs for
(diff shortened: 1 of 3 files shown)


mscuthbert:
Thanks!

PR #924, opened 2021-03-20
v7 SpannerBundle.replaceSpannedElement() only takes elements, not ids

jacobtylerwalls:
One last PR for spanners.

After the last few PRs there remains only one use of `getSpannedElementIds()`, which inefficiently builds the entire list before providing it to the caller to search, in `SpannerBundle.replaceSpannedElement()`, which alternatively takes an element, and which could be tested for spanner membership with the faster `Spanner.__contains__()` pattern we just optimized in 839c7e5.

- commit 1: sends elements, not ids, to `SpannerBundle.replaceSpannedElement()`, since we just optimized that, and raises an exception if you try to do it the slow way by passing an id.
- <s>commit 2: instead of raising an exception, provide backwards compatibility by checking if the input is `int`, so as to treat as an id and pass to new function `replaceSpannedElementSlow` with a deprecation warning. </s> dropped from the branch
- commit 3: unrelated optimization -- make a similar change to `RecursiveIterator.__next__()` for speed as in 113929c for consistency.

If you think we don't need to bother with backwards compatibility for this power usage -- (wouldn't a power user want to know they were doing it the slow way?) -- I can drop commit 2 and force-push.  I don't think we really need to go to the effort of having the shim, but it was fast to just code it up and push it. Your call.

### DIFF
--- a/music21/repeat.py
+++ b/music21/repeat.py
@@ -788,7 +788,7 @@ class Expander:
         for m in srcStream:
             # processes uses the spanner bundle stored on this Stream
             self._repeatBrackets.spannerBundle.replaceSpannedElement(
-                id(m.derivation.origin), m)
+                m.derivation.origin, m)
 
         # srcStream = self._srcMeasureStream
         # post = copy.deepcopy(self._srcMeasureStream)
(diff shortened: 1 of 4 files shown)


mscuthbert:
Let's do it the fast way.  If we're talking about Python ids and not music21 `.id` then it's a poweruser only thing and we don't need to give a backwards compatible way of working.  So we can simplify.

mscuthbert (approved):
Looks good -- we can definitely remove getSpannedElementById() in v8 and maybe fully deprecate it here.

PR #927, opened 2021-03-21
MIDI -- insert conductor events into parts before making measures

jacobtylerwalls:
Missed case from #856 . Need to insert conductor events (time signatures in particular) before making measures, else the whole thing is for naught.


mscuthbert (approved):
Looks great to me!

PR #930, opened 2021-03-23
Preserve value of spellingIsInferred in simplifyMultipleEnharmonics()

### LINKED ISSUE #929: spellingIsInferred should remain True when Chord created from MIDI numbers is respelled
**music21 version**
reproduced on current and on 5.1.0; expect it goes back to the introduction of the attribute

**Problem summary**
`.spellingIsInferred` should still be True on chord members created from MIDI numbers that happened to have been simplified, as I understand it.

**Steps to reproduce**

```
>>> b_maj = chord.Chord([59, 63, 66])
```

```
>>> [p.spellingIsInferred for p in b_maj.pitches]  # expect [True, True, True]
[True, False, True]
```

**More information**
Noticed in discussing #928

jacobtylerwalls:
Fixes #929

Also replace some bare asserts in the Chord module tests.

### DIFF
--- a/music21/chord/__init__.py
+++ b/music21/chord/__init__.py
@@ -5876,6 +5876,14 @@ class Test(unittest.TestCase):
         self.assertEqual(unscrambledChord3.pitches[3].name, 'F')
         self.assertEqual(unscrambledChord3.pitches[4].name, 'A-')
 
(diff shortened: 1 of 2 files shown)


mscuthbert (approved):
Thank you!  Looks great!

PR #935, opened 2021-03-28
Use "music21 Part" as the default part name

jacobtylerwalls:
These look like reasonable results.

Finale

Screen Shot 2021-03-28 at 1 55 27 PM

MuseScore

Screen Shot 2021-03-28 at 1 55 51 PM

OSMD (with @supersational 's feature!)

Screen Shot 2021-03-28 at 1 58 32 PM

### DIFF
--- a/music21/_version.py
+++ b/music21/_version.py
@@ -42,7 +42,7 @@ When changing, update the single test case in base.py.
 Changing this number invalidates old pickles -- do it if the old pickles create a problem.
 '''
 
-__version_info__ = (7, 0, 3)  # can be 4-tuple: (7, 0, 5, 'a2')
+__version_info__ = (7, 0, 4)  # can be 4-tuple: (7, 0, 5, 'a2')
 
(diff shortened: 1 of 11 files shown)


mscuthbert:
I'm not sure I like this change -- I actually think that we should be moving in the opposite direction, leaving off "music21 Fragment" and author name instead of giving defaults there.  If the particular problem is in osmd output not allowing empty part names, I'm okay with osmd's converter manipulating the stream before sending it to XML rendering.

jacobtylerwalls:
Yeah, I get that. Just tested and setting `.partName` on the stream directly should suffice without inserting a `Piano`, so if the change is that soft, rather than inserting Pianos -- seems okay to do inPlace. I'll close this and leave a comment on #326.

jacobtylerwalls:
Had a quick look at the MusicXML spec -- `<part-name>` is required and must be a string. Not sure if that forbids having an empty tag `<part-name/>`, which is what we're doing.

Strikes me as a low ROI area of the converter to be reworking at this point though.

PR #941, opened 2021-03-31
Check literal duration from MusicXML against notated type and create unlinked durations

### LINKED ISSUE #779: Superfluous rest created when when Music21 interprets a bad MusicXML
**music21 version**

`6.1 through 6.5, most likely other versions `   

**Problem summary**

**TLDR:** 
When a MusicXML measure contains elements with more duration than should be specified by the time signature, rests are incorrectly created and appended to the beginning of the measures, with duration equal to the aforementioned duration discrepancy. Because of this, Music21 content no longer becomes equal to MusicXML content. More issues arise when the durations are shorter, but they are less important issues.

**Steps to reproduce**

I'm adding external files because the issue stems from bad XML files themselves, not anything that can easily be reproduced by creating a stream manually.
[Rachmaninoff_Prelude_Op._23_No._5_G_Minor.mxl.zip](https://github.com/cuthbertLab/music21/files/5821344/Rachmaninoff_Prelude_Op._23_No._5_G_Minor.mxl.zip)

```
import music21
m21_stream = music21.converter.parse('/path/to/MusicXML/Rachmaninoff_Prelude_Op._23_No._5_G_Minor.mxl')
m21_stream[3][190].show('t') ###Measure 6, Part P1-Staff1
m21_stream[4][190].show('t') ###Measure 6, Part P1-Staff2
### To view MusicXML content, my suggestion is to use MuseScore or Finale. OSMD was okay, but didn't seem as clear or accurate as MuseScore or Finale.
```

**Expected**:
When Music21 reads in a bad MusicXML where elements have duration longer than the duration specified by the XML time signature, Music21 converts these measures and deals with the MusicXML errors however it chooses to, but does not add any incorrect or superfluous content to the stream measures when this occurs.

**Actual**:
When Music21 reads in a bad MusicXML where elements have duration longer than the duration specified by the XML time signature, Music21 converts these measures and creates the elements as written with overextended duration, but also creates extra incorrect rest elements that were not in the XML, appended to the beginning of measures and voices.

**More information**

**Long read, and my thought process:**
This shouldn't have to be a problem, but it's one I've unfortunately run into.

In a MusicXML, when the duration of a part's measure's elements extended beyond the expected duration of the measure, a funny phenomenon occurs in the XML to M21 conversion. If an instrument has multiple staves or voices, for every staff/voice that is "next" after an overextended measure, a rest is added to the beginning of the next staff/voice *equal to the duration by which the last staff/voice had been overextended. Here's a concrete example:

In an included piece by Rachmaninoff, measure 6, there are two staves and multiple voices per part. 
**In the MusicXML:**
- In measure 6, staff 1, voice 1, there are 4 beats worth of elements, which is fine.
- In staff 1, voice 2, there are 4.5 beats worth of elements, which will later cause a cascade of issues.
- Voices 5 and 6 are not present in staff 1, and are written after the duration error in the MusicXML.
- In measure 6, staff 2, voice 5, there are 4 beats worth of elements, but a duration issue in this part did precede.
- In staff 1, voice 6, there are 4.5 beats worth of elements, on top of the preceding duration error.
- Voices 1 and 2 are not present in staff 2, and are written before or during the duration error in the MusicXML.

**In the M21 stream:**
- In measure 6, staff 1, voice 1, the elements in this staff have the right duration, so we're okay. 
- In staff 1, voice 2, the note D has a duration of 1.5, and the rest has a duration of 3, making 4.5 beats total, 0.5 beats over. The handling of this measure I believe is okay. However, all future voices/staves will start with a rest of 0.5 duration.
- In staff 1, voice 5, this voice was previously absent in the staff, but because it exists in the MusicXML part, and I believe Music21 added it to this staff for the purpose of adding the extra 0.5 duration rest.
- Staff 1, voice 6 is the same as above. Note that neither voice 5 nor 6 extend beyond the time signature duration, so they don't affect future staves/voices.
- In measure 6, staff 2, voice 1, because MusicXML's denote two staves of a piano as the *same* part, Music21 will have the same issue of offsetting staff 2 elements by 0.5 because of what happened in staff 1, but now they're just separated into different staves. Staff 2, voice 5 now has an extra rest in the front. Because the MusicXML itself had 4 beats of elements, the offset of the next voice is not affected beyond the same 0.5-beat offset and 0.5-duration rest added in the beginning.
- In measure 6, staff 2, voice 6, the same 0.5-beat offset occurred, resulting in the same 0.5-duration rest added in the beginning.
- Voices 1 and 2 don't show up in the second staff because by the time the duration error occurred, they were passed by in the MusicXML.
- Let's say, you had a 7th voice after voice 6. After testing, I have seen that, in staff 2 voice 7, there would be a 1.0-beat rest added to compensate for the 0.5-beat duration error in staff 1 voice 2 and the 0.5-beat duration error in staff 2 voice 6 accumulating.

```
###Time signature 4/4
(Pdb) m21_stream[3][190].show('t') ### Measure 6, Part P1-Staff1
{0.0} <music21.stream.Voice 1>
    {0.5} <music21.chord.Chord D4 F#4 D5>
    {0.75} <music21.chord.Chord D4 F#4 D5>
    {1.0} <music21.chord.Chord D4 F#4 D5>
    {1.5} <music21.note.Note F#>
    {1.75} <music21.note.Note A>
    {2.0} <music21.note.Note B->
    {2.5} <music21.chord.Chord G3 D4 E4>
    {3.0} <music21.note.Note G>
    {3.5} <music21.chord.Chord B-3 D4 E4>
{0.0} <music21.stream.Voice 2>
    {0.0} <music21.note.Note D>
    {1.5} <music21.note.Rest rest> ###duration of 3 beats, total voice duration of 4.5 beats
{0.0} <music21.stream.Voice 5>
    {0.0} <music21.note.Rest rest> ### duration of 0.5 = 4.5 - 4.0 beats
{0.0} <music21.stream.Voice 6>
    {0.0} <music21.note.Rest rest> ### duration of 0.5 = 4.5 - 4.0 beats
{1.5} <music21.clef.BassClef>
{2.5} <music21.clef.TrebleClef>

(Pdb) m21_stream[4][190].show('t') ### Measure 6, Part P1-Staff2
{0.0} <music21.stream.Voice 5>
    {0.0} <music21.note.Rest rest> ### duration of 0.5 = 4.5 - 4.0 beats
    {0.5} <music21.note.Rest rest>
    {1.0} <music21.chord.Chord A2 F#3 A3>
    {1.25} <music21.chord.Chord A2 F#3 A3>
    {1.5} <music21.chord.Chord A2 F#3 A3>
    {2.0} <music21.note.Note F#>
    {2.25} <music21.note.Note A>
    {2.5} <music21.note.Note B->
    {3.0} <music21.chord.Chord D3 E3>
    {3.5} <music21.chord.Chord G1 G2>
    {4.0} <music21.chord.Chord D3 E3> ###duration of 0.5 beats, total voice duration of 4.5 beats
{0.0} <music21.stream.Voice 6>
    {0.0} <music21.note.Rest rest> ### duration of 0.5 = 4.5 - 4.0 beats
    {0.5} <music21.chord.Chord D1 D2>
    {2.0} <music21.note.Rest rest>  ###duration of 3 beats, total voice duration of 5 beats
```

I didn't finish diagnosing the problem, but it occurs in music21/musicxml/xmltom21.py, around line 2320-2340 in measureParser.parse(), probably at the v.makeRests() function. Before this, the stream contains elements only from the MusicXML (as expected), but all elements (at least GeneralNotes) were offset by the amount of beats the last staff/voice went over by, e.g. a voice object looking like this.

```
###Time signature 4/4
{0.0} <music21.stream.Voice 5>
    {0.5} <music21.note.Rest rest>
    {1.0} <music21.chord.Chord A2 F#3 A3>
    {1.25} <music21.chord.Chord A2 F#3 A3>
    {1.5} <music21.chord.Chord A2 F#3 A3>
    {2.0} <music21.note.Note F#>
    {2.25} <music21.note.Note A>
    {2.5} <music21.note.Note B->
    {3.0} <music21.chord.Chord D3 E3>
    {3.5} <music21.chord.Chord G1 G2>
    {4.0} <music21.chord.Chord D3 E3>
```
My guess is that the coreInsert() uses some kind of info written in the XML to decide how far to backtrack the offset when creating streams, and in this case, it didn't back up far enough because the last measure's elements went over the expected duration. Then 4.5 current offset of the last voice's 

After the v.makeRests() call, the voice would look like this:

```
###Time signature 4/4
{0.0} <music21.stream.Voice 5>
    {0.5} <music21.note.Rest rest>
    {1.0} <music21.chord.Chord A2 F#3 A3>
    {1.25} <music21.chord.Chord A2 F#3 A3>
    {1.5} <music21.chord.Chord A2 F#3 A3>
    {2.0} <music21.note.Note F#>
    {2.25} <music21.note.Note A>
    {2.5} <music21.note.Note B->
    {3.0} <music21.chord.Chord D3 E3>
    {3.5} <music21.chord.Chord G1 G2>
    {4.0} <music21.chord.Chord D3 E3>
```
and suddenly the stream has rest elements that don't match the content in the MusicXML.

I was able to write a work-around routine to deal with this. I use a regular expression to strip the staff label from the part ID's, so part ID's P1-Staff1 and P1-Staff2 both become P1, P1 and match each other. After this, I go sequentially through the measures/voices until I find at least one duration overshoot. After this, I delete the first element of every staff/voice remaining to that part, because I assume the element will be a superfluous rest.

Although this routine was easy enough to write, I write this report so that other people are aware of the issue and so maybe it can be rectified in the source code. I'd also love to explore this more and see if I can fix in source, if you happen to be busy. I've already spent a couple days with this problem.

jacobtylerwalls:
... and create unlinked durations if discrepancies.

Fixes #779 

<s>I say I'm not going to jump on these and then it turns out to be just one more line to add. Needs tests before it's ready. </s> (Added one test.) And verify this is the direction we want to go.

### DIFF
--- a/music21/musicxml/xmlToM21.py
+++ b/music21/musicxml/xmlToM21.py
@@ -3199,8 +3199,7 @@ class MeasureParser(XMLParserBase):
 
         If the `<duration>` doesn't match the `<type>` and `<dots>`,
         an unlinked duration is created so that `.quarterLength` agrees with
-        `<duration>` but the notated types can still be represented by
-        `.components`.
+        `<duration>` but the notated types can still be represented.
 
         Create a second dot on `mxNote` and parse again, observing the identical
         `quarterLength`:
--- a/music21/musicxml/xmlToM21.py
+++ b/music21/musicxml/xmlToM21.py
@@ -6515,7 +6515,7 @@ class Test(unittest.TestCase):
           <step>D</step>
           <octave>5</octave>
         </pitch>
-        <duration>{defaults.divisionsPerQuarter * 0.5}</duration>
+        <duration>{defaults.divisionsPerQuarter * 0.5 * (2/3) * (2/3)}</duration>
         <voice>1</voice>
         <type>eighth</type>
         <time-modification>
--- a/music21/musicxml/xmlToM21.py
+++ b/music21/musicxml/xmlToM21.py
@@ -3259,8 +3259,9 @@ class MeasureParser(XMLParserBase):
         # can't do this unless we have a type, so if not forceRaw
         if not forceRaw:  # a cooked version builds up from pieces
             dt = duration.durationTupleFromTypeDots(durationType, numDots)
-            if dt.quarterLength == qLen:
+            if dt.quarterLength == qLen and not tuplets:
                 # raw == cooked, so we're done
+                # but this comparison gives false positives if tuplets are involved
                 return d if inputM21 is None else None
             if d is not None:
                 d.clear()
@@ -3271,6 +3272,7 @@ class MeasureParser(XMLParserBase):
 
             for tup in tuplets:
                 d.appendTuplet(tup)
+                qLen = common.opFrac(qLen * tup.tupletMultiplier())
 
             # Second check against qLen (raw), now with tuplets
             # if != , create unlinked Duration and set raw qLen
@@ -6506,13 +6508,13 @@ class Test(unittest.TestCase):
         '''
         test that a note with nested tuplets gets converted properly.
         '''
-        mxN = '''
+        mxN = f'''
         <note default-x="347">
         <pitch>
           <step>D</step>
           <octave>5</octave>
         </pitch>
-        <duration>4</duration>
+        <duration>{defaults.divisionsPerQuarter * 0.5}</duration>
         <voice>1</voice>
         <type>eighth</type>
         <time-modification>
--- a/music21/musicxml/xmlToM21.py
+++ b/music21/musicxml/xmlToM21.py
@@ -3235,42 +3235,48 @@ class MeasureParser(XMLParserBase):
             # TODO: empty-placement
 
         # two ways to create durations, raw (from qLen) and cooked (from type, time-mod, dots)
-        if forceRaw:
-            if d is not None:
-                # environLocal.printDebug(['forced to use raw duration', durRaw])
-                durRaw = duration.Duration()  # raw just uses qLen
-                # the qLen set here may not be computable, but is not immediately
-                # computed until setting components
-                durRaw.quarterLength = qLen
-                try:
-                    d.components = durRaw.components
-                except duration.DurationException:  # TODO: Test
-                    if qLen:
-                        qLenRounded = 2.0 ** round(math.log2(qLen))
-                    else:
-                        qLenRounded = 0.0
-                    environLocal.printDebug(
-                        ['mxToDuration',
-                         f'rounding duration to {qLenRounded} as type is not'
-                         + 'defined and raw quarterLength '
-                         + f'({qLen}) is not a computable duration'])
-                    # environLocal.printDebug(['mxToDuration', 'raw qLen', qLen, durationType,
-                    #                         'mxNote:',
-                    #                         ET.tostring(mxNote, encoding='unicode'),
-                    #                         'last mxDivisions:', divisions])
-                    durRaw.quarterLength = qLenRounded
-            else:
-                d = duration.Duration(quarterLength=qLen)
-        else:  # a cooked version builds up from pieces
+        if d is not None:
+            # environLocal.printDebug(['forced to use raw duration', durRaw])
+            durRaw = duration.Duration()  # raw just uses qLen
+            # the qLen set here may not be computable, but is not immediately
+            # computed until setting components
+            durRaw.quarterLength = qLen
+            try:
+                d.components = durRaw.components
+            except duration.DurationException:  # TODO: Test
+                if qLen:
+                    qLenRounded = 2.0 ** round(math.log2(qLen))
+                else:
+                    qLenRounded = 0.0
+                environLocal.printDebug(
+                    ['mxToDuration',
+                        f'rounding duration to {qLenRounded} as type is not'
+                        + 'defined and raw quarterLength '
+                        + f'({qLen}) is not a computable duration'])
+                durRaw.quarterLength = qLen = qLenRounded
+        else:
+            d = duration.Duration(quarterLength=qLen)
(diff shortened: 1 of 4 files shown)


jacobtylerwalls:
Speed still looks a wash! -- if not possibly a teeny improvement.


```
>>> timeit("converter.parse('music21/corpus/beach/prayer_of_a_tired_child.musicxml', forceSource=True, storePickle=False)", setup='from music21 import converter', number=100)
```

main
36.482573290999994

PR
36.418761322

music21/musicxml/xmlToM21.py
```diff
@@ -3235,43 +3256,50 @@ def xmlToDuration(self, mxNote, inputM21=None):
...
+                else:
+                    qLenRounded = 0.0
+                environLocal.printDebug(
+                    ['mxToDuration',
+                        f'rounding duration to {qLenRounded} as type is not'
+                        + 'defined and raw quarterLength '
+                        + f'({qLen}) is not a computable duration'])
+                durRaw.quarterLength = qLen = qLenRounded
+        else:
+            d = duration.Duration(quarterLength=qLen)
```
mscuthbert, line 3278: Doesn't this need the same guard as above?

mscuthbert (approved):
Incredible work.  One Q, but if I misunderstood, feel free to merge.

music21/musicxml/xmlToM21.py
jacobtylerwalls: All I did with this pair of if/else branches was to unindent. Separating out the next comment so I can make an issue.

jacobtylerwalls:
I'll leave this open until I can add a test case for grace notes.

PR #942, opened 2021-03-31
v7 Write `.musicxml` files instead of `.xml` if `fp` not given

jacobtylerwalls:
There was a [comment](https://github.com/cuthbertLab/music21/issues/456#issuecomment-544558681) a while ago about doing this for v6, so when I saw `.endswith('-2.xml')` in the temp files cleanup patch, I thought "let's cross this off now".

(Let me know if I've severely misunderstood the point of the comment!)

### DIFF
--- a/music21/common/formats.py
+++ b/music21/common/formats.py
@@ -94,11 +94,11 @@ def findFormat(fmt):
     Note that .mxl and .mx are only considered MusicXML input formats.
 
     >>> common.findFormat('mx')
-    ('musicxml', '.xml')
+    ('musicxml', '.musicxml')
     >>> common.findFormat('.mxl')
-    ('musicxml', '.xml')
+    ('musicxml', '.musicxml')
     >>> common.findFormat('musicxml')
-    ('musicxml', '.xml')
+    ('musicxml', '.musicxml')
     >>> common.findFormat('lily')
     ('lilypond', '.ly')
     >>> common.findFormat('lily.png')
@@ -121,9 +121,6 @@ def findFormat(fmt):
     ('vexflow', '.html')
     >>> common.findFormat('capx')
     ('capella', '.capx')
-
-    >>> common.findFormat('mx')
-    ('musicxml', '.xml')
 
 
     Works the same whether you have a leading dot or not:
--- a/music21/converter/subConverters.py
+++ b/music21/converter/subConverters.py
@@ -838,7 +838,7 @@ class ConverterMusicXML(SubConverter):
     '''
     registerFormats = ('musicxml', 'xml')
     registerInputExtensions = ('xml', 'mxl', 'mx', 'musicxml')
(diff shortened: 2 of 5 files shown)


music21/common/formats.py
```diff
@@ -122,9 +122,6 @@ def findFormat(fmt):
     >>> common.findFormat('capx')
     ('capella', '.capx')
 
-    >>> common.findFormat('mx')
-    ('musicxml', '.xml')
-
```
jacobtylerwalls, line 127: duplicated above, so removed it.

jacobtylerwalls:
> I love this! .xml was never a good file extension for musicXML

I believe .musicxml is the new sanctioned extension for uncompressed files as of MusicXML 3.1

mscuthbert:
Yes.  We should make sure the .xml files are still fully supported as input and output formats (e.g., if someone gives an `fp="file1.xml"` we should respect that), also inside of .mxl archives -- but that should have already been supported.  It's just what the default is for if fp is not given that should change.  Will review now.

jacobtylerwalls:
> (e.g., if someone gives an `fp="file1.xml"` we should respect that)

I think this needs some work still. I'll signify when it's ready.

jacobtylerwalls:
Apparently we were coercing any `fp` to end in .xml, even if it had no suffix or an incorrect one. We execute this branch because subformats is `[]`, not `None`, and then replace .xml with .musicxml.

https://github.com/cuthbertLab/music21/blob/b46a9533ed848049f1d8f4b0ab4a387904876636/music21/converter/subConverters.py#L984-L988

I pushed a commit skipping this branch and adding a test. Do we want to continue coercing the suffix on `fp`? The other formats don't do this:
```
>>> from music21 import *
>>> s = stream.Stream([note.Note()])
>>> s.write(fmt='midi', fp='garbage')
'garbage'
```

jacobtylerwalls:
> Will review now.

absolutely no rush!

music21/converter/subConverters.py
```diff
@@ -838,27 +838,32 @@ class ConverterMusicXML(SubConverter):
     '''
     registerFormats = ('musicxml', 'xml')
     registerInputExtensions = ('xml', 'mxl', 'mx', 'musicxml')
```
mscuthbert, line 840: we can get rid of "mx" here too -- it was from the days before musicxml had its own extension besides .xml and was only a music21 supported method.  No one has used it for years.

mscuthbert (approved):
Looks great to me.  Bye-bye 99% of my xml files!

mscuthbert:
Need a way of generating compressed XML files next!

jacobtylerwalls:
> Need a way of generating compressed XML files next!

Perhaps #596? 📬

mscuthbert:
oh, yeah.  :-D

PR #945, opened 2021-04-02
Fix duplicate bass note in dominant seventh ChordSymbol

jacobtylerwalls:
There was a small bug with duplicate bass pitches on dominant seventh chords that was just an alias/name bug.

(At first, I thought #882 would take care of it, but it's unrelated.)


mscuthbert:
Thanks!  The musicxml vs music21 naming of dominant-seventh chords strikes again!

PR #946, opened 2021-04-03
v7 Stream constructor improvements

jacobtylerwalls:
Inspired by mailing list 4/2/21

This would be the error if someone tried to follow my advice to `.show('text')` on a stream where iterators were inserted (user error!). Easy improvement.

**Update: ** also add `coreGuard...` check to Stream constructor.

```
>>> src = stream.Stream()
>>> otherStream.append(src.iter)
>>> otherStream.show('t')
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/music21/stream/base.py", line 1844, in elementOffset
    o = self._offsetDict[id(element)][0]
KeyError: 4469622288

During handling of the above exception, another exception occurred:

  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/music21/base.py", line 894, in getOffsetBySite
    raise e
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/music21/base.py", line 880, in getOffsetBySite
    a = site.elementOffset(tryOrigin, returnSpecial=returnSpecial)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/music21/stream/base.py", line 1855, in elementOffset
    raise base.SitesException(
music21.sites.SitesException: an entry for this object 0x10a690610 is not stored in stream <music21.stream.Stream 0x109ffcd90>

The above exception was the direct cause of the following exception:

  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/music21/stream/base.py", line 258, in show
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/music21/base.py", line 2647, in show
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/music21/converter/subConverters.py", line 552, in show
    print(obj._reprText(**keywords))
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/music21/stream/base.py", line 2985, in _reprText
    return self.recurseRepr(addEndTimes=addEndTimes,
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/music21/stream/base.py", line 2962, in recurseRepr
    msg.append(singleElement(element, indent))
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/music21/stream/base.py", line 2937, in singleElement
    offGet = in_element.getOffsetBySite(self)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/music21/base.py", line 901, in getOffsetBySite
    raise SitesException(
music21.sites.SitesException: an entry for this object <music21.stream.Score 0x10a2cb7c0> is not stored in stream <music21.stream.Stream 0x109ffcd90>
```

### DIFF
--- a/music21/stream/base.py
+++ b/music21/stream/base.py
@@ -177,20 +177,22 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
 
     A list of measures will let each be appended:
 
-    >>> s3 = stream.Part([stream.Measure(n1), stream.Measure(note.Rest())])
+    >>> m1 = stream.Measure(n1, number=1)
+    >>> m2 = stream.Measure(note.Rest(), number=2)
+    >>> s3 = stream.Part([m1, m2])
     >>> s3.show('text')
-    {0.0} <music21.stream.Measure 0 offset=0.0>
+    {0.0} <music21.stream.Measure 1 offset=0.0>
         {1.0} <music21.note.Note E->
-    {1.5} <music21.stream.Measure 0 offset=1.5>
+    {1.5} <music21.stream.Measure 2 offset=1.5>
         {0.0} <music21.note.Rest rest>
 
-    Here, every element is a Stream that's not a Measure, so we insert:
+    Here, every element is a Stream that's not a Measure, so we instead insert:
 
-    >>> s4 = stream.Score([stream.Part(n1), stream.PartStaff(note.Rest())])
+    >>> s4 = stream.Score([stream.PartStaff(n1), stream.PartStaff(note.Rest())])
     >>> s4.show('text')
-    {0.0} <music21.stream.Part 0x102fba940>
+    {0.0} <music21.stream.PartStaff 0x...>
         {1.0} <music21.note.Note E->
-    {0.0} <music21.stream.PartStaff 0x102fba6d0>
+    {0.0} <music21.stream.PartStaff 0x...>
         {0.0} <music21.note.Rest rest>
     '''
     # this static attributes offer a performance boost over other
@@ -271,23 +273,23 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
         if not common.isIterable(givenElements):
             givenElements = [givenElements]
 
-        for e in givenElements:
-            # Raises StreamException if not a Music21Object
-            self.coreGuardBeforeAddElement(e)
-
         # Append rather than insert if every offset is 0.0
         # but not if every element is a stream subclass other than a Measure
         # (i.e. Opus, Score, Part, or Voice)
-        append: bool = all(e.offset == 0.0 for e in givenElements)
+        append: bool = False
+        try:
+            append = all(e.offset == 0.0 for e in givenElements)
+        except AttributeError:
+            pass  # appropriate failure will be raised by coreGuardBeforeAddElement()
         if append and all(e.isStream for e in givenElements):
             if all(not e.isMeasure for e in givenElements):
                 append = False
 
-        if append:
-            for e in givenElements:
+        for e in givenElements:
+            self.coreGuardBeforeAddElement(e)
+            if append:
                 self.coreAppend(e)
-        else:
-            for e in givenElements:
+            else:
                 self.coreInsert(e.offset, e)
 
         self.coreElementsChanged()
--- a/music21/stream/base.py
+++ b/music21/stream/base.py
@@ -115,9 +115,9 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
     The duration, however, can be "unlinked" and explicitly
     set independent of the Stream's contents.
 
-    The first element passed to the Stream is an optional list,
-    tuple, or other Stream of music21 objects which is used to
-    populate the Stream by inserting each object at
+    The first element passed to the Stream is an optional single
+    Music21Object or a list, tuple, or other Stream of Music21Objects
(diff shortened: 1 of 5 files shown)


music21/stream/base.py
```diff
@@ -231,6 +231,7 @@ def __init__(self, givenElements=None, *args, **keywords):
             # TODO: perhaps convert a single element into a list?
             for e in givenElements:
                 try:
+                    self.coreGuardBeforeAddElement(e)
```
mscuthbert, line 234: Thanks -- I think though this is a good chance to change the behavior of the Stream constructor when given a list of elements.  My sense is that we check first if all elements have offset 0 then we use coreAppend instead of coreInsert.  The though originally was that we would do things like:

```
>>> n1 = note.Note('C4')
>>> n1.offset = 0.0
>>> n2 = note.Note('D4')
>>> n2.offset = 1.0
>>> s = stream.Stream([n1, n2])
```

but since about v1 of music21, we've gotten away from setting the offsets of notes etc. before adding to Streams.  A lot of our functions though still assume we'll do that.  Instead more often people want to do this:

```
>>> s = stream.Stream([note.Note('C4'), note.Note('D4')])
```

and they're shocked that both notes are at offset 0.

So a test that if all offsets are 0 we append, otherwise we insert. -- [but maybe a check that it's not a list of parts??? since those SHOULD all be at zero]

This is also definitely the time to change the behavior of line 230, to check that givenElements is not a single Stream object.  My students always make this error, which should work:

```
>>> n = note.Note()
>>> m = stream.Measure(n)
>>> p = stream.Part(m)
>>> sc = stream.Score(p)
```

and they're shocked that `sc` just contains a single note.  This is a paradigm that we'd like to have work

music21/stream/core.py
```diff
@@ -384,18 +384,26 @@ def coreGuardBeforeAddElement(self, element, *, checkRedundancy=True):
...
+        Traceback (most recent call last):
+        music21.exceptions21.StreamException: cannot insert StreamIterator into a Stream
+        Iterate over it instead (Users Guide chs. 6 and 26)
         '''
         # using id() here b/c we do not want to get __eq__ comparisons
         if element is self:  # cannot add this Stream into itself
             raise StreamException('this Stream cannot be contained within itself')
+        if 'StreamIterator' in element.classes:
+            raise StreamException('cannot insert StreamIterator into a Stream\n'
+            'Iterate over it instead (Users Guide chs. 6 and 26)')
```
mscuthbert, line 406: can we have a `try` except around this, because this is also the place presumably to check that the object IS a music21 object. Right here, we get two-for-one, because non-music21 objects will probably not have a `.classes` element.

mscuthbert (approved):
Looks very good to me.  Please read notes about the constructor changes.  And we can either fix it here (though that sounds like scope creep) or make an issue and I or you can work on it.

music21/stream/base.py
jacobtylerwalls: I can see the value in that. My impulse is to extend the all-are-parts rule to be all-are-streams-not-measures to get Voice, Score, Opus, Stream, etc. Only Measure is the supported positioned stream in a stream.

What about clefs, meters, etc. (corner case) -- okay with those being appended if they're listed in the wrong order?

jacobtylerwalls: well, voice is a supported positioned stream in a stream also, but you can put two voices at 0.0 and we shouldn't mess with that.

music21/variant.py
```diff
@@ -947,12 +947,7 @@ def addVariant(
             tempVariant.append(sVariant)
         else:  # sVariant is not a measure
             sVariantMeasures = sVariant.getElementsByClass('Measure')
-            # apparently expression cannot be simplified. -- this is a mistake
-            # since sVariantMeasures will never == [] even if there are no measures.
-            # yet switching this to `if not sVariantMeasures` breaks things.
-            # TODO(msc) -- figure this out and fix it.
-            # noinspection PySimplifyBooleanCheck
-            if sVariantMeasures == []:  # If there are no measures, work element-wise
+            if not sVariantMeasures:  # If there are no measures, work element-wise
```
jacobtylerwalls, line 950: Obviously we can back this out if you don't want to review this right now (should have been on a separate commit), but just noticed while fixing the variant test.

mscuthbert:
Voices can definitely have multiple at zero or multiple following each other.  But the default here should probably be that they both all go at the beginning, like Parts.

I think -- but don't fully remember -- that Scores within Opus elements can either be positioned to all be at time zero -- in which case they represent multiple scores within an anthology [choose one to play at a time], or be positioned to follow each other as appending, as in multiple movements.  BUT it's just a behavior I've had in my head.  And MAYBE they should alway follow each other, since Chordifying an Opus should not be having more than one score's chords merged with each other.

music21/variant.py
mscuthbert: It looks good -- nothing we have in the tests seems to be bitter by it.  I think it's time to bit the bullet and remove it, and see if bugs arise from it.

mscuthbert:
Looks great to me!  I'll fix up the changes in my comment as a separate issue

PR #948, opened 2021-04-04
Added tests and made tinyNotation.Converter more strict

### LINKED ISSUE #366: TinyNotation validates some bad inputs
The TinyNotation parser currently throws away anything that it doesn't tokenize, which means it permits some inputs that it probably shouldn't.

e.g., `g1 a g; g1 a g` is fine (n.b. the semicolon), since `;` is never captured by a regex.

It probably should check the input string after tokenizing to see if there's anything leftover and complain.

creyes17:
Fixes #366 

Hi friends,

I saw https://github.com/cuthbertLab/music21/issues/366 and thought I'd contribute a suggested fix. Specifically, we formerly could add semicolons (or any other extra, irrelevant characters) to any note, rest, or time signature and it would still validate the input. With `raiseExceptions=True`, we would expect it to let us know there was something wrong.

Because we were erroneously allowing the extra characters, it didn't think something was wrong and didn't raise the exception. In this pull request, I tighten up the regular expressions to exclude these extra characters.

WARNING: If any current users are inadvertently depending on the old, overly-generous syntax, they may see an unexpected change in behaviour. (For example, formerly `g1 a g; g1 a g` would have been accepted input and the `;` would have been ignored. Now, the `g;` token is considered invalid and would be silently skipped if `raiseExceptions` is not set to `True`.)

It took me a minute to wrap my head around the nuances of the tinyNotation syntax. It helped me to document the grammar, so I added that to the documentation as well. (If that was overly verbose, I'm happy to remove it or trim it down).

[Extended Backus-Naur form information from Wikipedia](https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form)
```
(* Grammar (in Extended Backus-Naur form) *)
(* https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form *)
(* Items in parentheses are grouped *)
(* Items in curly braces appear zero or more times *)
(* Items in square brackets may appear exactly zero or one time *)
(* Items in double quotes are literal strings *)
(* Each rule is ended by a semicolon *)

TINY-NOTATION = TOKEN, { WHITESPACE, TOKEN } ;
TOKEN = ( TIME-SIGNATURE | TUPLET | REST | NOTE );
TIME-SIGNATURE = INTEGER, "/", INTEGER ;
INTEGER = { "0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9" } ;
WHITESPACE = " ", { " " } ;
TUPLET = ( "trip" | "quad" ), "{",
    [ WHITESPACE ],
    ( REST | NOTE ),
    { WHITESPACE, ( REST | NOTE ) },
    [ WHITESPACE ],
"}" ;
REST = "r", [ DURATION ], [ MODIFIER ] ;
DURATION = ( EVEN-NUMBER, { "." } | { "." }, EVEN-NUMBER | ".", { "." } ) ;
EVEN-NUMBER = { INTEGER }, ( "0" | "2" | "4" | "6" | "8" ) ;
NOTE = PITCH, [ DURATION ], [ TIE ], { MODIFIER } ;
PITCH = (
    ( LOW-A | LOW-B | LOW-C | LOW-D | LOW-E | LOW-F | LOW-G ), [ ACCIDENTAL ] |
    ( HIGH-A | HIGH-B | HIGH-C | HIGH-D | HIGH-E | HIGH-F | HIGH-G ), [ ACCIDENTAL ], { "'" } |
    ( HIGH-A | HIGH-B | HIGH-C | HIGH-D | HIGH-E | HIGH-F | HIGH-G ), { "'" }, [ ACCIDENTAL ]
) ;
LOW-A = "A", { "A" } ;
LOW-B = "B", { "B" } ;
LOW-C = "C", { "C" } ;
LOW-D = "D", { "D" } ;
LOW-E = "E", { "E" } ;
LOW-F = "F", { "F" } ;
LOW-G = "G", { "G" } ;
(* I'm pretty sure high notes should only have a single letter, but there *)
(* are examples in music21.key.tonalCertainty where the lower-cased *)
(* letter is repeated. *)
HIGH-A = "a", { "a" } ;
HIGH-B = "b", { "b" } ;
HIGH-C = "c", { "c" } ;
HIGH-D = "d", { "d" } ;
HIGH-E = "e", { "e" } ;
HIGH-F = "f", { "f" } ;
HIGH-G = "g", { "g" } ;
ACCIDENTAL = ( EDITORIAL | SHARPS | FLATS | NATURAL ) ;
EDITORIAL = "(", ( SHARPS | FLATS | NATURAL ), ")" ;
SHARPS = "#", { "#" } ;
FLATS = "-", { "-" } ;
NATURAL = "n" ;
TIE = "~" ;
MODIFIER = (
    EQUALS-MODIFIER |
    UNDERSCORE-MODIFIER |
    SQUARE-MODIFIER |
    ANGLE-MODIFIER |
    PARENS-MODIFIER |
    STAR-MODIFIER
) ;
EQUALS-MODIFIER = "=", DATA ;
UNDERSCORE-MODIFIER = "_", DATA ;
SQUARE-MODIFIER = "[", DATA, "]" ;
ANGLE-MODIFIER = "<", DATA, ">" ;
PARENS-MODIFIER = "(", DATA, ")" ;
STAR-MODIFIER = "*", DATA, "*" ;
DATA = ALPHANUMERIC, { ALPHANUMERIC } ;
(* The following is just shorthand. I think this is more *)
(* clear than listing out all 62 alphanumeric characters. *)
ALPHANUMERIC = ? Any alphanumeric character. So "a-z", "A-Z", or "0-9" ?
```


creyes17:
Hmm... I see the tests failed on python 3.7, but I'm not able to replicate locally on python version 3.7.10.

creyes17:
Hmm... this looks like a network issue as opposed to an actual failing test. I'll try again tomorrow to get the tests passing

creyes17:
Everything passes locally for me using `python -c 'from music21.test.testSingleCoreAll import ciMain as ci; ci()'` on both python 3.9.2 and python 3.7.10. It also worked using `python -m music21.test.multiprocessTest` as well on python 3.9.2.

jacobtylerwalls:
Hi, and thanks for this patch. I'll leave the substantive review to others (LGTM at first glance!), but for the small suggestion to make the grammar function public so that it appears in the Sphinx-compiled docs. But perhaps not worth making that change until Myke weighs in on the approach. (Also, if you could please use a GitHub keyword like "Closes" or "Fixes #366" in the body of the PR description, it will automagically close on merge.)

I opened another PR in part to find out whether I would hit the same race condition. I skated through scot-free--maybe give yours another push? I'm sure this would be very difficult to reproduce locally; you'd need to wire up different tempdirs and/or containers given that `parseURL()` just reads the cached version the second time through.

creyes17:
Thanks for the feedback Jacob!

I've updated the PR description and that's a good idea about the Sphinx docs. I agree that we should wait for Myke's input there though.

music21/tinyNotation.py
```diff
@@ -1256,6 +1375,56 @@ def testOne(self):
...
+                pass
+            except AssertionError:
+                # The test failed. Reraise the original test failure
+                raise
+            except Exception as e:
+                raise AssertionError(
+                    "Should have raised a TinyNotationException for input "
+                    f'"{error_state["string"]}" because {error_state["reason"]} '
+                    f"but actually raised {e.__class__.__name__}."
+                ) from e
```
jacobtylerwalls, line 1425: Can we just use `with self.assertRaises(TinyNotationException):` here? Shorter, consistent with the rest of the suite, and avoids missed lines in the Coveralls report.

music21/tinyNotation.py
```diff
@@ -942,11 +1065,7 @@ def __init__(self, stringRep='', **keywords):
         self.generalBracketStateRe = re.compile(r'(\w+){')
         self.tieStateRe = re.compile(r'~')
 
-        self.tokenMap = [
-            (r'(\d+\/\d+)', TimeSignatureToken),
-            (r'r(\S*)', RestToken),
-            (r'([a-gA-G]\S*)', NoteToken),  # last
-        ]
+        self.tokenMap = _getDefaultTokenMap()
```
jacobtylerwalls, line 1068: Nice -- factoring this out makes it easier to override, it seems to me.

music21/tinyNotation.py
creyes17: Yeah, we can do that. The only reason I went with this approach is that I wanted to leave a more descriptive error message if the test fails.

creyes17: I'll just reraise with additional information. That should be simpler

creyes17: Oh wait! Today I learned that they've updated `assertRaises` to accept a `msg` parameter =D
https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

creyes17:
Should I update the patch version in `music21/_version.py` as well to `(7, 0, 4)`? I don't know what the convention is for this project

jacobtylerwalls:
> Should I update the patch version in `music21/_version.py` as well to `(7, 0, 4)`? I don't know what the convention is for this project

We've been doing that when making changes to file parsing in formats that are cached (because cached files are only used if they match the current version number). TinyNotation streams aren't serialized (to my knowledge), so I think we're good here. Thanks for asking.

mscuthbert:
This is super cool — I really like seeing the grammar here.  Sorry for the poor timing of my break (see music21list) but will review when I get back.  :-)  Thanks Chris.

Some things I see in the comment (haven’t looked at the code) 

The “trip{}”  “quad{}” are the only two pre-defined examples of something called “state” which generally can hold TINY-NOTATION.  So other states can be defined with ALPHANUMERIC “{“  Certain states like TRIP and QUAD can restrict what is inside the state.

UNDERSCORE-MODIFIER can be followed by any character except whitespace or “=“.   This allows for writing “C2_a&nbsp;low&nbsp;C” for output in Vexflow, etc.  

EQUALS-MODIFIER in default use (assigning .id) should only be DATA, but overridden can be followed by any character except whitespace or “_”.   

I know of one project that overrides the modifiers in order to present style attributes such as =color(red)

Any place where high letters are repeated is technically a bug, but WOW, I wish I had chosen to go with that syntax instead of single quotes which have given me no end of trouble.  Maybe I can make that an alternative syntax.

I’m going to need to port any changes to music21j afterwards so some things might need to be modified slightly so that the code there stays in sync with here.   Music21j has a non-standard token “partBreak” which changes the tinyNotation from a Part to a Score.  It’s something that I’ve meant to port over here. 

WHITESPACE can also include carriage returns (at least in m21j version).  Very helpful for long examples.

THANKS!  — Myke

creyes17:
Thanks Myke, I hope you had (are having?) a good time off! I'll make those updates this weekend.

music21/tinyNotation.py
```diff
@@ -1256,6 +1375,90 @@ def testOne(self):
...
+                    "only the same upper-cased letter may be repeated to "
+                    "indicate lower octaves"
+                ),
+            },
+        ]
+
+        for error_state in error_states:
+            with self.assertRaises(TinyNotationException, msg=(
+                    "Should have raised a TinyNotationException for input "
+                    f'"{error_state["string"]}" because {error_state["reason"]}.'
```
mscuthbert, line 1408: Prefer single quotes for all code unless they also include single quotes (as in some tiny notation).

Please do not use implicit string concatenation.  It's a little faster than explicit concat, but tends to lead to bugs later.

creyes17:
Not yet, but I've put time on my calendar this week for it.

mscuthbert:
I think that the full grammar is too much for the User's Guide -- imagine seeing that right the first time you're encountering TinyNotation which is supposed to be a super simple way of getting notes in quickly.  Maybe we can move it to a separate reference page?  Or, I was totally fine with having it in the tinyNotation.py file -- that's where people who really want to know about TinyNotation will go.

OH!  There's an Extending TinyNotation page.  We can move it there.  That's an advanced topic.  BUT probably even better as an Appendix to the User's Guide, which is linked from the Extending TinyNotation page.

creyes17:
That sounds good to me, although I can't find the "Extending TinyNotation" page. I see a reference to it in the [Table of Contents](https://web.mit.edu/music21/doc/usersGuide/usersGuide_99_Table_of_Contents.html?highlight=expanding%20subclassing) but it's not actually a link.

For now, I've moved the documentation back into `tinynotation.py`.

As far as creating an Appendix, I might need some more explicit instructions. Should I create a new Jupyter Notebook file and link to it from the table of contents?

mscuthbert (approved):
Looks great to me -- a better place to get started in going forward!  THANKS Chris!  :-)

mscuthbert:
Don't worry about the Appendix, etc. -- I'll keep working on User's Guide

PR #949, opened 2021-04-04
Let floating KeyAnalyzer reach final measure in streams without pickups

jacobtylerwalls:
Measure numbers were assumed to be 0-indexed (`for i in range(self.numMeasures):`), but accessed with `.measure()`, which gets the _actual_ measure numbers, so this only worked exactly right for pieces with pickup measures, like the Bach chorale in the doctest. That is, for pieces without pickup measures, the last measure was never reached.

### DIFF
--- a/music21/analysis/floatingKey.py
+++ b/music21/analysis/floatingKey.py
@@ -59,6 +59,22 @@ class KeyAnalyzer:
      <music21.key.Key of f# minor>, <music21.key.Key of f# minor>, <music21.key.Key of f# minor>,
      <music21.key.Key of f# minor>, <music21.key.Key of f# minor>,
      <music21.key.Key of f# minor>, <music21.key.Key of f# minor>]
+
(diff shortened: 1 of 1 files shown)


music21/analysis/floatingKey.py
```diff
@@ -70,38 +86,55 @@ def __init__(self, s=None):
 
         self.weightAlgorithm = divide
         if s.hasPartLikeStreams():
-            p = s.parts.first()
+            p = s.iter.parts.first()
```
jacobtylerwalls, line 89: This is just adding safety in case `s` is a generic `stream.Stream` (only Score has this property)

jacobtylerwalls:
Absolutely no need to reply at the moment; but I just stumbled across `indicesNotNumbers=True` which is probably a smaller and easier way to fix this. So I'll have a look and resubmit when ready.

mscuthbert:
Thanks -- yes, indicesNotNumbers is what programmers should use.  Measure numbers are a convienence for beginners.  :-)

PR #950, opened 2021-04-04
Add `forceSource` to `Converter.parseURL()`

jacobtylerwalls:
The module level function had this keyword, but it was dead, because it didn't exist on `Converter.parseURL()`.

Noticed when looking at the CI failure on #948 .

### DIFF
--- a/music21/converter/__init__.py
+++ b/music21/converter/__init__.py
@@ -631,7 +631,8 @@ class Converter:
         self.subConverter.keywords = keywords
         self.subConverter.parseData(dataStr, number=number)
 
-    def parseURL(self, url, format=None, number=None, **keywords):  # @ReservedAssignment
+    def parseURL(self, url, *, format=None, number=None,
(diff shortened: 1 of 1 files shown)


music21/converter/__init__.py
```diff
@@ -1041,14 +1046,17 @@ def parseData(dataStr, number=None, format=None, **keywords):  # @ReservedAssign
 
 # pylint: disable=redefined-builtin
 # noinspection PyShadowingBuiltins
-def parseURL(url, number=None, format=None, forceSource=False, **keywords):  # @ReservedAssignment
+def parseURL(url, *, format=None, number=None,
+             forceSource=False, **keywords):  # @ReservedAssignment
```
jacobtylerwalls, line 1050: keyword-only should make this moot, but noticed that format and number were not in the same order as the other methods. I get why we're making everything keyword-only 🎉

mscuthbert:
Great test!

Issue #775, opened 2021-01-13
Ties in Chords Missing After Write Stream to MIDI and Read Back


myersjm:
**music21 version**

`9.8.7`   

**Problem summary**

I am working on a large project involving constructing streams and then writing them to file as MIDI. I am constructing chords with notes where some of the notes are tied and others aren't. In my larger code, notes that are not tied are disappearing when I read back the MIDI from file. I tried to replicate this in a simpler context, but could not reproduce the exact behavior. But the simpler code is not doing what I expect, either. The ties present before writing to file are correct, but when I read back the file, the stream's chords do not have the ties anymore. This could be either a user error or an actual bug. Any help would be much appreciated!

**Steps to reproduce**

```
n0 = note.Note("C4")
n0.tie = tie.Tie('start')

n1 = note.Note("C4")
n2 = note.Note("F4")
n1.tie = tie.Tie('continue')

n3 = note.Note("C4")
n4 = note.Note("F4")
n3.tie = tie.Tie('stop')

note_arr0 = [n0]
note_arr = [n1, n2]
note_arr2 = [n3, n4]
c0 = chord.Chord(note_arr0)
c1 = chord.Chord(note_arr)
c2 = chord.Chord(note_arr2)

s = stream.Stream()
s.append(c0)
s.append(c1)
s.append(c2)
s.show()

# Then I read back the stream, and the ties are now gone:

fp = s.write('midi', fp='./Data/why.mid')
path = 'C:/Users/myersjm/Documents/11 Senior Year/Senior Thesis/Data/why.mid'

score = converter.parse(path)
score.show()

```

**Expected vs. actual behavior**

Before writing, the ties are correct and present in the stream:
![Before_Write](https://user-images.githubusercontent.com/31910744/104517669-746a3300-55c4-11eb-8df4-f1040043b28d.PNG)
After I write to file and read it back, the stream does not have any ties anymore, which is an issue, as it should be identical to before:
![After_Write](https://user-images.githubusercontent.com/31910744/104517667-73d19c80-55c4-11eb-8916-0a4f106efa0d.PNG)

**More information**

My main question is, am I using the ties correctly? If so, are there some underlying things that happen when you write a stream to MIDI and read it back where some of the tie information, or note information, is lost? My original issue was that some notes are disappearing in chords after I write them to file, but I am thinking this could be tied to an issue with how I am using ties. Thanks for any help! I am using Windows 10, if that is useful information, by the way.

[...]

Issue #620, opened 2020-09-14
Showing MIDI of a chordified stream does not follow chord ties


RayHuck:
**music21 version**

`6.1.0`   

**Problem summary**

- When calling `chordify()` and then `.show('midi')` on a stream, the ties are being ignored on the notes within the chords

**Steps to reproduce**
- Have corpus parse some piece and chordify and show it both as musicxml and midi
- Observe that it produces the correct musicxml representation
- Observe that it produces a midi representation that is different than what is written as it ignores ties between chords
```
    # original
    corpus.parse('bach/bwv108.6.xml').show('midi')

    # correct musicxml chordified version
    corpus.parse('bach/bwv108.6.xml').chordify().show()    

    # incorrect choridified midi version
    corpus.parse('bach/bwv108.6.xml').chordify().show('midi')
    
```

jacobtylerwalls:
Thanks for the report! 👍

Exporting MIDI runs `stripTies()` in an effort to represent ties by prolonging durations, but after #586, `stripTies()` doesn't do anything with partially-tied chords, since putting the result back on a single chord would necessitate mixing durations among chord members! If you look at the behavior in `v.5.7.2`, your example outputs about half the chordify slices (see #502), so I consider this behavior better: better to at least get all the notes in there before worrying about representing all the ties.

Now, in terms of representing the ties, maybe there could be an argument `makeVoices` to `stripTies()` that, in single voice scenarios, like `chordify()` output, would unpack all the chord members, split ties, create rhythmic overlaps, and then run `stream.makeVoices()` to separate out the overlapping notes, and then run `stripTies()` on each of the resulting voices. Something like that.

jacobtylerwalls:
It occurs to me that what I just described would be a way to unchordify something--to make chords into polyphony. Maybe it should be its own routine, and then `midi.translate` could call `unchordify()` before `stripTies()`.

mscuthbert:
Hi @jacobtylerwalls -- there is a routine that is intended to do this, called ".explode()" which currently is an exact synonym for ".voicesToParts()" but I made it separate in order to in some future date make it smarter.  I can imagine having a keyword attribute to this:  chordNotesToParts=ChordNotesToParts.NONE (an enum -- I'm tired of string configurations), which might also have ChordNotesToParts.ALL and ChordNotesToParts.COMPLEX_TIES -- where the latter would put everything in a new part when there's a chord where not all note components have the same tie status.

[...]

PR #956, opened 2021-04-11
Prevent false positive tie-merging by `stripTies()` when `matchByPitch=True`

jacobtylerwalls:
**Motivation**
`matchByPitch` defaults True in v7. So it's worth doing some improvements before the release. I noticed a few things that seemed off when retesting the example in #775, which has a set of commands to build this stream:

continue

If we do nothing, in v7, `matchByPitch=True` will match all of these notes and replace with a dotted half. In v6, it performs no match.

**Problems**

1. `matchByPitch=True` doesn't reject comparisons between Chords and Notes. So the second onset is connected to the first onset, and then we have a valid match all the way across.  Update: now, matchByPitch=False won't merge these either even if the ties are fixed up. In other words, we will check for equal cardinality in both cases.
2. `matchByPitch=False` does not and should not connect these either, thanks to #586 which made it so that the final chord must have all "stop" ties, not mixed-ties. But I think there is a similar problem with "continue" tie types -- if I adjust the example to put a stop tie on the last F4, we find a valid connection all the way through and replace with a dotted half. I think we should be strict about "continue" also. Essentially, we're giving the user the choice to be strict about tie types (matchByPitch=False), or permissive about tie types, preferring to matchByPitch instead.

**Now**
- `matchByPitch=True` rejects comparisons between Chords and Notes
- `matchByPitch=False` (strict mode) -- a tie type of "continue" on a chord is not taken at face value to be a valid connection, since it might be a mixed-tie-status chord.


mscuthbert:
Give me a minute to think through the last implication -- I'd really like never to have pitches removed during StripTies.  Seems dangerous.  Not sure if the fix should be here, or in another PR.

jacobtylerwalls:
Paraphrasing to see if I follow. The idea would be when matching by tie-type (not by pitch), to collect the set of all pitches across the matched span and create them on the replacement rather than merely prolong the first note, whatever its pitches were?

Example cases: stripped version of m.1 would become what you see in m.2?
Screen Shot 2021-04-19 at 3 11 01 PM

Screen Shot 2021-04-19 at 3 08 37 PM

mscuthbert:
> Paraphrasing to see if I follow. The idea would be when matching by tie-type (not by pitch), to collect the set of all pitches across the matched span and create them on the replacement rather than merely prolong the first note, whatever its pitches were?

I think that if any notes change we shouldn't merge the ties at all.  So in both cases, m1 remains the same after stripTies.  The point of it is to figure out where it's safe to merge two notes/chords together and both of these cases, information is lost.  (Well, not 100% sure about case 2, but definitely in case 1)

jacobtylerwalls:
Ah, so this is starting to sound like removing the `matchByPitch` keyword entirely because we no longer want to allow `matchByPitch=False` (which was the default until #818 !)

> The point of it is to figure out where it's safe to merge two notes/chords together and both of these cases, information is lost.

I would possibly gently push back: there could be some use cases for a destructive stripTies operation where it's more important to guarantee that there are no ties than to guarantee that all the pitches were preserved. Maybe I don't care about the pitches at all, maybe it's a drum set part, but I DEFINITELY care that every `NotRest` I iterate over is an onset of a sound, not a continuation, so that my rhythm and beat counting turn out right.

`inPlace` defaults False and `matchByPitch` defaults True, both of which have to be flipped to get irrecoverable pitch destruction, so it's opt-in.

What I remember from working on stripTies before is that it's used in three big places:
 - MIDI output
 - chordify
 - feature extractors

With matchByPitch=True I think those three places are all set. The question being now if we need to take matchByPitch=False away from anybody else wanting a way to guarantee no ties. Well, it doesn't ABSOLUTELY guarantee no ties, that would be more like `force=True`, and we don't have that. We have:
 - `matchByPitch=True` : permissive about tie-types, strict about pitch
 - `matchByPitch=False`: permissive about pitch, strict about tie types

Maybe this needs an enum where the match choices are pitch, type, or force.

In any case, just wanted to clarify I'm not introducing more tie stripping here, this PR as it stands reduces tie stripping by getting stricter in two scenarios; the scenario you were focusing on was status quo :-)

jacobtylerwalls:
> I want to enable the use case (drum, things, etc.) that you discuss (though I'd think that a gong-hit with a resonance tie that adds a cymbal crash on beat 2 does sound like two not one onset).

Right, and this use case is starting to sound like the complex-ties-to-voices, i.e. explode() scenario that is a someday-enhancement to get MIDI output to be correct (see #620 and #775).

And thanks, with that bit of history now I see how matchByPitch=False has outlived its usefulness. ☑️

jacobtylerwalls:
>matchByPitch=False has outlived its usefulness. 

Sorry -- should have said in the case of chords. I see why it was there (and still works) for single notes.

jacobtylerwalls:
Opted for a minimal change for now -- like `matchByPitch=True`, have `matchByPitch=False` ALSO check for equal cardinality of notes/chords.

How's this?

jacobtylerwalls:
You were on to something that we should be stricter and not just merge things. The failing test in the feature extractors had to do with bridging voices, which is currently mostly undefined in `stripTies`-- we only isolate and processs voices if they are contained in the Part directly (should no longer happen after the MIDI changes in #856). As you know, we don't have a voice-bridging routine at present, so stripTies just takes a .flat of the Part.

8a4bc98 ensured that cardinality was checked everywhere, including a case we hadn't discussed -- matchByPitch=True (default) on the STOP of a tie-span. We took the STOP at face value and checked neither cardinality nor pitch when matchByPitch=True.

In the Schoenberg example, with matchByPitch=True we were merging the D# in left hand measure with the trichord in the righthand measure:

**Source**
Screen Shot 2021-04-26 at 7 41 23 PM


**Stripped with match by pitch -- before bug fixes in 8a4bc98 and 17cda1c**
Screen Shot 2021-04-26 at 7 43 33 PM


So based on that, I added a test for this and will now push a commit changing the feature extractor results which I expect are more correct now.

jacobtylerwalls:
> we don't have a voice-bridging routine at present

That's not stated right--we do have `voicesToParts()`. Maybe that could be explored down the road for use here.

mscuthbert:
OMG -- voices make this EVEN more complicated.  Will read more on Tuesday and hope to get back to you then on this approach.

jacobtylerwalls:
Up to you, but this seems like a candidate for prioritizing for v7 since we've defaulted `matchByPitch=True`, which will generate more traffic into these existing bugs.

If you're short on time, I could cherry-pick the first bullet point under "Now:" to a new PR and at least merge that.

> Now:
> - matchByPitch=True rejects comparisons between Chords and Notes
> - matchByPitch=False (strict mode) -- a tie type of "continue" on a chord is not taken at face value to be a valid connection, since it might be a mixed-tie-status chord.

mscuthbert:
> > we don't have a voice-bridging routine at present
> 
> That's not stated right--we do have `voicesToParts()`. Maybe that could be explored down the road for use here.

 Not exactly the same thing as here.

mscuthbert:
You're right that we need to get this in before v7, but I'd like the default to be as cautious about removing notes as possible.  If that's where we are, we can merge now and then get other parts in later.  I'd rather have a few too many ties than have any notes removed that shouldn't be.

jacobtylerwalls:
> I'd like the default to be as cautious about removing notes as possible. If that's where we are, we can merge now and then get other parts in later. 

Yes, both of the bugfixes on this branch involve making _fewer_ merges/removals. And I have nothing else in mind, so I think we can just merge. Updated the test comments to remark where the failures would occur on current version.

mscuthbert:
LGTM -- feel free to merge if all tests pass

PR #958, opened 2021-04-12
Add safety to `Stream._getDuration()`

### LINKED ISSUE #957: stream._cache can disappear at the most inopportune moment
**music21 version**

`6.3.0`   

**Problem summary**

The cache can get cleared between adding _cache_['Duration'] and setting _cache_['Duration'].quarterLength.

**Steps to reproduce**
I set duration.tuplets on a note that was already in a stream. Some time later, someone called stream.duration, and the following code in stream/__init__.py (_getDuration) crashed:

            # environLocal.printDebug(['creating new duration based on highest time'])
            self._cache['Duration'] = duration.Duration()
            self._cache['Duration'].quarterLength = self.highestTime <--- BOOM, KeyError, which seems impossible, right?
            return self._cache['Duration']

Some investigation showed that self.highestTime cleared the cache in coreElementsChanged().  To be more clear, self.highestTime called _getQuarterLength on my modified duration, so updateQuarterLength gets called, which ultimately calls informClient/informSites/coreElementsChanged, which sets stream._cache_ to {}, making:
            self._cache['Duration'].quarterLength = self.highestTime
raise KeyError while trying to evaluate self._cache['Duration'] on an empty cache.

Proposed fix:
    In place of:
            self._cache['Duration'] = duration.Duration()
            self._cache['Duration'].quarterLength = self.highestTime
    do:
            highestTime = self.highestTime # might clear self._cache!
            self._cache['Duration'] = duration.Duration()
            self._cache['Duration'].quarterLength = highestTime

Workaround:
In my new Humdrum converter's tuplet handling code, when I set duration.tuplets on notes that are already in a stream, I now immediately get the quarterLength, to avoid this situation.

```

I'm afraid I don't have a minimal example.

```

**More information**

jacobtylerwalls:
Fixes #957 

(few days, few minutes, what is a Duration anyway?)

Thanks again to @gregchapman-dev for tracking down the interaction with `coreElementsChanged()` so that we could get a reproducer.

### DIFF
--- a/music21/duration.py
+++ b/music21/duration.py
@@ -2845,7 +2845,10 @@ class Duration(prebase.ProtoM21Object, SlottedObjectMixin):
     @property
     def tuplets(self) -> Tuple[Tuplet, ...]:
         '''
-        return a tuple of Tuplet objects
+        Return a tuple of Tuplet objects.
+        May leave a stream containing objects having this duration
+        in an unusable state, requiring :meth:`~music21.stream.core.coreElementsChanged`
+        to be called. For this reason, prefer using :meth:`appendTuplet` to add tuplets.
         '''
         if self._componentsNeedUpdating:
             self._updateComponents()
--- a/music21/stream/base.py
+++ b/music21/stream/base.py
@@ -7882,8 +7882,7 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
             return self._cache['Duration']
         else:
             # environLocal.printDebug(['creating new duration based on highest time'])
-            self._cache['Duration'] = duration.Duration()
-            self._cache['Duration'].quarterLength = self.highestTime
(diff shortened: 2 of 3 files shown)


gregchapman-dev (approved):
Looks good to me.  I like your fix better than my proposed fix. 👍

jacobtylerwalls:
Decided to document what I learned about the necessity of calling coreElementsChanged.

Should spruce that up someday--there's not a good workaround if you're trying to *remove* tuplets, since `appendTuplet` only appends, and it breaks our principle that only core methods should require calling `coreElementsChanged`.

mscuthbert:
In the setter on tuplets you can call `.informClient()` which passes the buck to the Note object, etc. to inform the stream that elements have changed on it.

PR #960, opened 2021-04-12
Inform client when setting tuplets

### LINKED ISSUE #959: Tuplet setter doesn't inform client
Decided to document what I learned about the necessity of calling `coreElementsChanged()` when setting `.tuplets` directly.

Should spruce that up someday--there's not a good workaround if you're trying to *remove* tuplets, since `appendTuplet` only appends, and it breaks our principle that only core methods should require calling `coreElementsChanged()`.

_Originally posted by @jacobtylerwalls in https://github.com/cuthbertLab/music21/issues/958#issuecomment-818046530_

***

I see a few approaches:

- tuplet setter could inform the client instead of `appendTuplet` having to do it (I tried this and it broke a test, but the test may have been measuring the wrong thing, so it may unwravel some shoelaces that need to be unwraveled anyway)
- new method `removeTuplet` to address the gap in functionality
- clear tuplets in `Duration.clear()`

Not going to jump on this, doesn't seem urgent for v7.

jacobtylerwalls:
Fixes #959 

The one existing test that needed changing looked appropriate.

### DIFF
--- a/music21/duration.py
+++ b/music21/duration.py
@@ -1821,8 +1821,7 @@ class Duration(prebase.ProtoM21Object, SlottedObjectMixin):
         True
         '''
         newTuplet.frozen = True
-        self.tuplets = self.tuplets + (newTuplet,)
-        self.informClient()
+        self.tuplets = self._tuplets + (newTuplet,)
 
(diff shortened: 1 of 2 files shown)


mscuthbert (approved):
Looks great!

Issue #398, opened 2019-05-16
Implementing percussion instrument support


bzier:
I encountered issues reading midi files which utilize a drum track (channel 10 - midi percussion). Issue #235 lists percussion as one of the line items for version 6. I am opening this issue to begin the discussion for the solution/implementation design for supporting these types of tracks. @mscuthbert, your insight is much appreciated. Thanks.

bzier:
@mscuthbert ping

mscuthbert:
We've just started work on v6 today

jain-sah:
Can we know an ETA by when the percussion support will be there? We are working on a school project which needs to be finished by summer. So we will wait for this if it gets completed early.

Also, what features exactly are we targetting for v6?

It would be a great help to know what's coming.

mscuthbert:
sorry -- I really can't commit to a timeline on anything; I'll try when I can.  If you absolutely need a feature implemented, there is a paid consulting option: https://web.mit.edu/music21/doc/about/faq.html .

nblumoe:
Would it already be possible to export musicxml or MIDI files from music21, where Musescore would recognise them as drum tracks?

So far I failed to do that. I tried using a `BassDrum` instrument, setting MIDI channel to 10 (and using appropriate pitches, e.g. 35 for the bass drum), adding a percussion clef and even a new class `DrumSet` (the latter allowed me to have Musescore pick up the instrument name at least, but that isn't sufficient).

However, when the musicxml gets imported to Musescore, it doesn't recognize it as a GM Drumset, but will use a Piano sound instead. Once I change the staff properties there to select the `Unpitched Percussion > Drumset` instrument, it works correctly. For the pitched instruments (e.g. Bagpipes, Woodblock) Musescore picks up the instrument correctly

Is there a way to export midi or musicxml from music21 which would be recognised correctly as a drum track, allowing me to play back the track correctly without the manual fix on the instrument in musescore?
I am wondering if there is any attribute I need to specify for this to work or if it cannot be done that easily with `music21` at this stage.

nblumoe:
I figured out that the midiChannel from instruments won't be used in the export, because `channelInstrumentData` from `midi.translate` will just assign new channels from the `acceptableChannelList` ignoring whatever was set on existing instruments.

I was able to hack in support for MIDI channel 10 with a monkey patch. Clearly that hack isn't the right way forward, but just let me know if anyone is interested in it.

jacobtylerwalls:
[...]

PR #961, opened 2021-04-16
Import/Export MIDI channel and import MIDI program

jacobtylerwalls:
**Feature**:
Read from an `Instrument` instance's `.midiChannel` attribute when assigning MIDI channels on export. Update: also set program and channel on import.

- Give unpitched percussion a default midiChannel of 10 (9 in music21, which uses 0-indexing on Instrument instances).
- Fix 0-index vs. 1-index confusion in `instrument.autoAssignMidiChannel()`
- Improve docs
- Set .midiChannel and .midiProgram on Instrument instances on import.

Refs #398 
Also refs #465 but doesn't address note-by-note channel changes, which, if we continue on the current model, would need Instrument instances created each time.

<s>**Future work for another PR**:

- Ensure instrument identification from MIDI files captures most drum programs. As discussed in #398 it's possible we will need a DrumSet class. </s> Not evident to me (yet) that there's more work needed, so striking this comment.

### DIFF
--- a/music21/midi/translate.py
+++ b/music21/midi/translate.py
@@ -729,7 +729,7 @@ def midiEventsToInstrument(eventList):
 
     >>> me = midi.MidiEvent()
     >>> me.type = midi.ChannelVoiceMessages.PROGRAM_CHANGE
-    >>> me.data = 54  # Voice Oohs
+    >>> me.data = 53  # MIDI program 54: Voice Oohs
     >>> midi.translate.midiEventsToInstrument(me)
     <music21.instrument.Vocalist 'Voice'>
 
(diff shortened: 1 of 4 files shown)


music21/midi/translate.py
```diff
@@ -2390,7 +2480,9 @@ def midiTracksToStreams(
 
 def streamToMidiFile(
     inputM21: stream.Stream,
+    *,
     addStartDelay: bool = False,
+    acceptableChannelList: List[int] = None,
```
mscuthbert, line 2485: This needs to be Optional[List[int]]

mscuthbert (changes requested):
Looks really good -- I wonder if we should however get totally out of the 0-indexed midiChannel world except in writing/reading from MIDI? -- both 0 and 1-indexed midiPrograms are used in many places, but I don't know if users ever think of 1-indexed midiChannels.

One needed change to typing.  Feel free to merge afterwards.

PR #962, opened 2021-04-16
Import musicxml measures containing only ChordSymbols

jacobtylerwalls:
A sequence of two such measures previously raised `ZeroDivisionError`.

**Example source:**
```
    <measure number="4" width="289.46">
      <harmony print-frame="no">
        <root>
          <root-step>G</root-step>
          </root>
        <kind text="m">minor</kind>
        </harmony>
      <harmony print-frame="no">
        <root>
          <root-step>E</root-step>
          <root-alter>-1</root-alter>
          </root>
        <kind>major</kind>
        <offset>24</offset>
        </harmony>
      <forward>
        <duration>48</duration>
        </forward>
      </measure>
    <measure number="5" width="357.54">
      <harmony print-frame="no">
        <root>
          <root-step>D</root-step>
          </root>
        <kind>major</kind>
        </harmony>
      <forward>
        <duration>48</duration>
        </forward>
      </measure>
```

Because m.4 was treated as "short", in the next measure we executed the last-measure-was-short branch and then crashed on a `ZeroDivisionError`:

```
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/music21/musicxml/xmlToM21.py", line 2028, in adjustTimeAttributesFromMeasure
    barDurationProportion = m.barDurationProportion()
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/music21/stream/base.py", line 12538, in barDurationProportion
    return opFrac(self.highestTime / barDuration.quarterLength)
ZeroDivisionError: float division by zero
```

### DIFF
--- a/music21/harmony.py
+++ b/music21/harmony.py
@@ -1387,8 +1387,8 @@ class ChordSymbol(Harmony):
     chords, by default appear as chord symbols in a score and have duration of
     0.
 
-    To obtain the chord representation of the in the score, change the
-    :attr:`music21.harmony.ChordSymbol.writeAsChord` to True. Unless otherwise
+    To obtain the chord representation of the `ChordSymbol` in the score, change
+    :attr:`~music21.harmony.ChordSymbol.writeAsChord` to True. Unless otherwise
     specified, the duration of this chord object will become 1.0. If you have a
     leadsheet, run :meth:`music21.harmony.realizeChordSymbolDurations` on the
     stream to assign the correct (according to offsets) duration to each
--- a/music21/harmony.py
+++ b/music21/harmony.py
@@ -1388,7 +1388,7 @@ class ChordSymbol(Harmony):
     0.
 
     To obtain the chord representation of the `ChordSymbol` in the score, change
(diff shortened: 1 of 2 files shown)


music21/harmony.py
```diff
@@ -1387,8 +1387,8 @@ class ChordSymbol(Harmony):
     chords, by default appear as chord symbols in a score and have duration of
     0.
 
-    To obtain the chord representation of the in the score, change the
-    :attr:`music21.harmony.ChordSymbol.writeAsChord` to True. Unless otherwise
+    To obtain the chord representation of the `ChordSymbol` in the score, change
+    :attr:`~music21.harmony.Harmony.writeAsChord` to True. Unless otherwise
     specified, the duration of this chord object will become 1.0. If you have a
     leadsheet, run :meth:`music21.harmony.realizeChordSymbolDurations` on the
```
mscuthbert, line 1393: this should be :func:

mscuthbert (approved):
Looks good -- I'll fix the existing doc problem separately.

PR #965, opened 2021-04-18
Remove Ambitus.getPitchRanges()

jacobtylerwalls:
`s.analyze('range')` is now a synonym for `s.analyze('span')`, which is a much faster algorithm. That means in the rare cases where there were no instances of a repeated note in a stream, the previous nonzero minPitch is now zero.

***

Prevents segfaults I was seeing when trying to cache metadata.

**More information**
This algorithm is O(n^2), so any extra values should be filtered out of `psFound` before running an expensive computation. We don't need more than two instances of any `ps`, because they'll just produce the same interval.

https://github.com/cuthbertLab/music21/blob/ebc47a1fbc9d65981bc52928593b4b564ae96cd5/music21/analysis/discrete.py#L1079-L1091

While debuggging I printed `psRange` being computed for one stream to a file, and it consumed > 500 MB before I interrupted the interpreter.

### DIFF
--- a/music21/analysis/discrete.py
+++ b/music21/analysis/discrete.py
@@ -1064,6 +1064,15 @@ class Ambitus(DiscreteAnalysis):
         >>> s = stream.Stream()
         >>> ambitusAnalyzer.getPitchRanges(s)
         (0, 0)
+
+        OMIT_FROM_DOCS
+
+        >>> s = stream.Stream()
+        >>> for i in range(12, 61):
+        ...    n = note.Note(i)
+        ...    s.repeatAppend(n, 2)
+        >>> ambitusAnalyzer.getPitchRanges(s)
+        (0, 48)
         '''
         ssfn = subStream.flat.notes
 
@@ -1075,15 +1084,19 @@ class Ambitus(DiscreteAnalysis):
             elif 'Note' in n.classes:
                 pitches = [n.pitch]
             for p in pitches:
-                psFound.append(p.ps)
+                # third insertion would be irrelevant
+                # we will sort next, and having two identical
+                # values is enough to produce an interval of 0.
+                if psFound.count(p.ps) < 2:
+                    psFound.append(p.ps)
         psFound.sort()
-        psRange = []
+        psRange = set()
         for i in range(len(psFound) - 1):
             p1 = psFound[i]
             for j in range(i + 1, len(psFound)):
                 p2 = psFound[j]
                 # p2 should always be equal or greater than p1
-                psRange.append(p2 - p1)
+                psRange.add(p2 - p1)
 
         if not psRange:
             return (0, 0)
--- a/music21/analysis/discrete.py
+++ b/music21/analysis/discrete.py
@@ -1024,7 +1024,6 @@ class Ambitus(DiscreteAnalysis):
 
         >>> s = stream.Stream(note.Rest())
         >>> p.getPitchSpan(s)
-        None
 
         OMIT_FROM_DOCS
 
@@ -1032,7 +1031,7 @@ class Ambitus(DiscreteAnalysis):
 
         >>> s.insert(4, harmony.ChordSymbol('C6'))
         >>> p.getPitchSpan(s)
-        None
+
         '''
         if subStream is self._referenceStream and self.minPitchObj and self.maxPitchObj:
             return self.minPitchObj, self.maxPitchObj
--- a/music21/analysis/discrete.py
+++ b/music21/analysis/discrete.py
@@ -997,10 +997,10 @@ class Ambitus(DiscreteAnalysis):
 
         # environLocal.printDebug([self._pitchSpanColors])
 
-    def getPitchSpan(self, subStream):
+    def getPitchSpan(self, subStream) -> Tuple[Optional[pitch.Pitch], Optional[pitch.Pitch]]:
         '''
-        For a given subStream, return the pitch with the minimum and
-        maximum pitch space value found.
+        For a given subStream, return a tuple consisting of the two pitches
+        with the minimum and maximum pitch space value.
 
         This public method may be used by other classes.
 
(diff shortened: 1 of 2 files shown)


mscuthbert:
Curious -- are we ever using the min range anywhere in the system?  That seems like the main culprit for speed and I can't imagine what it might be useful for.  (In any real situation, it's going to be m2 or M2 I imagine.  Oh, maybe it shows if there are quarter tones?  But there's a better way to figure that out too).

We should be using `srcStream.pitches` -- much faster also.   (And Note also has a .pitches as well)

jacobtylerwalls:
Having two instances of the same pitch in a Stream is sufficient to get a 0.0 min. So--no, not very useful! (Quarter tones shouldn't affect our decision, since `ps` is a float. So the same two quarter tones would yield 0.0.)

This is the only place it's used, to get colors for intervals (I suppose an interval of 0.0 needs a color?)

https://github.com/cuthbertLab/music21/blob/ebc47a1fbc9d65981bc52928593b4b564ae96cd5/music21/analysis/discrete.py#L970-L977

Mozart K155:
2455 notes
len(psRange) would be 2455
len(psFound) would be 2455 ** 2 = 6,027,025

PR:
len(psRange) 93
inner loop iterations would would be 93 ** 2 = 8,649

If you want it faster, yes, we could drop the second instance of the `ps` and just use a set

len(psRange) 47
inner loops 47 ** 2 = 2,209

And then change `_generateColors()` to always assume a min of 0.

> We should be using srcStream.pitches -- much faster also. (And Note also has a .pitches as well)

Nice, I can address this after you let me know your preference on the other point.

mscuthbert:
WOW.  Your work DOES do 99.9% of the savings.  But yes, I think it's a time to remove getPitchRanges and rename to getAmbitusPs or something like that and in _generateColors assume a min of 0.

I think it can be removed without any deprecation cycle.  It's pretty hidden.

jacobtylerwalls:
I think we can just remove. If we don't care about the minimum value any more, same result can be had getting the ambitus from the other method (`getPitchSpan()`) and taking the difference of the two values. No reason to do a second search through the score. (Future self: someday distill a reproducer for these segfaults not involving external files to report to cPython.)

music21/analysis/discrete.py
```diff
@@ -1348,6 +1299,11 @@ def analyzeStream(streamObj, *args, **keywords):
     if args:
         method = args[0]
 
+    if method == 'range':
+        # getPitchRanges() was removed in v7
+        # this synonym is being added for compatibility
+        method = 'span'
+
```
jacobtylerwalls, line 1306: I guess this is not 100% great in that someone who extended the functionality with another analysis method would have to override analyzeStream() also in order to not get this synonym.

jacobtylerwalls: Ohhhh but if I get rid of the synonym then I'm signing up for rewriting the user's guide examples. I'll give Myke time to weigh in.

jacobtylerwalls:
No rush, but when we come back to this, it's perhaps also a time to address the following: this branch calls getPitchSpan() three times:

- Ambitus() constructor (which calls _generateColors())
- getPitchSpan() on the following line
- .getSolution() calls process()

https://github.com/cuthbertLab/music21/blob/06b2122b81ba86c777dae90fe707a4c0a3c3cba6/music21/metadata/__init__.py#L1268-L1275

Not a showstopper, but we could do fewer searches if we decide where to cache this information.

music21/analysis/discrete.py
```diff
@@ -970,8 +975,8 @@ def _generateColors(self, numColors=None):
         if numColors is None:
             if self._referenceStream is not None:
                 # get total range for entire piece
-                minPitchObj, maxPitchObj = self.getPitchSpan(self._referenceStream)
-                difference = int(maxPitchObj.ps - minPitchObj.ps)
+                self.minPitchObj, self.maxPitchObj = self.getPitchSpan(self._referenceStream)
+                difference = int(self.maxPitchObj.ps - self.minPitchObj.ps)
                 minPitch, maxPitch = 0, difference
             else:
                 minPitch, maxPitch = 0, 130  # a large default
```
mscuthbert: I eventually want to get rid of this paradigm of setting two variables by creating two tuples that reference each other.

```
minPitch = 0
maxPitch = difference
```

seems so much clearer.  And in fact, it uncovers that `minPitch = 0` should be outside the `if` statement.

music21/analysis/discrete.py
```diff
@@ -1039,7 +1047,12 @@ def getPitchSpan(self, subStream):
         minPitchIndex = psFound.index(min(psFound))
         maxPitchIndex = psFound.index(max(psFound))
 
-        return pitchesFound[minPitchIndex], pitchesFound[maxPitchIndex]
+        minPitchObj, maxPitchObj = pitchesFound[minPitchIndex], pitchesFound[maxPitchIndex]
+
+        if subStream is self._referenceStream:
+            self.minPitchObj, self.maxPitchObj = minPitchObj, maxPitchObj
```
mscuthbert, line 1066: as above.  Two lines is clearer than one.

music21/analysis/discrete.py
```diff
@@ -1024,15 +1024,14 @@ def getPitchSpan(self, subStream) -> Optional[Tuple[pitch.Pitch, pitch.Pitch]]:
 
         >>> s = stream.Stream(note.Rest())
         >>> p.getPitchSpan(s)
-        None
```
mscuthbert: generally I try to write either

```
>>> print(p.getPitchSpan(s))
None
```

or

```
>>> p.getPitchSpan(s) is None
True
```

It is the difference between doing a statement for side effects alone

```
>>> p.hello = 'hi'
```

and doing something and testing that it returns None.  Of course this does that also, but it's less clear for readers that that's what we're doing.

mscuthbert (commented):
Looks good -- a few comments that might be worth fixing now, or getting to later.

music21/analysis/discrete.py
jacobtylerwalls: > I guess this is not 100% great in that someone who extended the functionality with another analysis method would have to override analyzeStream() also in order to not get this synonym.

On second thought, this isn't so bad. The name was previously reserved because m21 was using it. And someone can override this OR override analysisClassFromMethodName. I think it's okay. Better to support backwards compatibility for now.

PR #966, opened 2021-04-19
Improve Trills and Make sure "methods" are methods

mscuthbert:
A lot of early documentation did not make a strong distinction between functions and methods.  Went through some places and clarified.

Noticed during review of #962


mscuthbert:
Added bug fixes and improvements for trills.

PR #969, opened 2021-04-20
Create inexpressible durations from tiny durations instead of raising

jacobtylerwalls:
I had this lying around (not pulling an all-nighter!)

**Motivation**:
Some tiny durations raise exceptions when accessing their types. My real-world case for this was reducing an orchestral score using `chordify()` and ending up with a 1/72 QL from the difference between one violin part with a 9/8/32nd tuplet run and another part with 8 32nd notes. This is how we handle large types, so here I extend it to small types.

MusicXML export should still fail if the type is inexpressible (and in one case they were getting all the way through and exporting invalid musicxml), but better to fail there (seems to me) than when just merely creating the duration and accessing the type.

```
>>> from music21 import *
>>> d = duration.Duration(1/72)
>>> d.type
  File "/Users/jwalls/music21/music21/duration.py", line 2913, in type
    elif len(self.components) == 1:
  File "/Users/jwalls/music21/music21/duration.py", line 2440, in components
    self._updateComponents()
  File "/Users/jwalls/music21/music21/duration.py", line 1732, in _updateComponents
    qlc = quarterConversion(self._qtrLength)
  File "/Users/jwalls/music21/music21/duration.py", line 700, in quarterConversion
    tuplet, component = quarterLengthToNonPowerOf2Tuplet(qLen)
  File "/Users/jwalls/music21/music21/duration.py", line 412, in quarterLengthToNonPowerOf2Tuplet
    closestSmallerType, unused_match = quarterLengthToClosestType(qLen / qFrac.denominator)
  File "/Users/jwalls/music21/music21/duration.py", line 310, in quarterLengthToClosestType
    raise DurationException('Cannot return types smaller than 2048th; '
music21.duration.DurationException: Cannot return types smaller than 2048th; qLen was: 1/576
```

**Now**
- Create inexpressible durations when doing m21 transformations (such as chordify)
- Raise when attempting to export inexpressible durations to musicxml

### DIFF
--- a/music21/duration.py
+++ b/music21/duration.py
@@ -653,11 +653,11 @@ def quarterConversion(qLen: OffsetQLIn) -> QuarterLengthConversion:
     # Tuplets...
     qLen = opFrac(qLen)
     # try match to type, get next lowest for next part...
-    closestSmallerType, unused_match = quarterLengthToClosestType(qLen)
     try:
+        closestSmallerType, unused_match = quarterLengthToClosestType(qLen)
         nextLargerType(closestSmallerType)
     except DurationException:
-        # too big...
+        # too big or too small
         return QuarterLengthConversion((DurationTuple(type='inexpressible',
                                                       dots=0,
                                                       quarterLength=qLen),), None)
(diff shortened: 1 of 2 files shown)


mscuthbert:
Perfect!  Sounds great.  Though 1/72 isn't so small as to be unreasonable at all -- especially from Chordifying 9:8 -- so that is a separate bug that should be fixed at some point.

PR #970, opened 2021-04-20
Improve `Stream.__getitem__()`

### LINKED ISSUE #947: Decide whether to support class strings as keys for `Stream.__getitem__`
https://github.com/cuthbertLab/music21/blob/ebc47a1fbc9d65981bc52928593b4b564ae96cd5/music21/stream/base.py#L322-L324

Doesn't work as described; only names work, not strings:

```
>>> h = corpus.parse('handel')
>>> h[stream.Part]
<music21.stream.Score 0x110207ee0>
>>> h['stream.Part']
  File "/Users/jwalls/music21/music21/stream/base.py", line 451, in __getitem__
    raise KeyError(f'provided key ({k}) does not match any id or group')
KeyError: 'provided key (stream.Part) does not match any id or group'
>>> h['Part']
  File "/Users/jwalls/music21/music21/stream/base.py", line 451, in __getitem__
    raise KeyError(f'provided key ({k}) does not match any id or group')
KeyError: 'provided key (Part) does not match any id or group'
```

**More information**
Shall we correct the docs or add the feature?

Here's adding the feature:
```diff
diff --git a/music21/stream/base.py b/music21/stream/base.py
index 4d8f92316..3e792a754 100644
--- a/music21/stream/base.py
+++ b/music21/stream/base.py
@@ -398,7 +398,8 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
         KeyError: "provided class (<class 'music21.layout.StaffLayout'>) does
             not match any contained Objects"
 
-        Changed in v7. -- out of range indexes now raise an IndexError, not StreamException
+        Changed in v7. -- out of range indexes now raise an IndexError, not StreamException.
+        Unsupported types now raise TypeError. Strings now supported for class names.
         '''
         # need to sort if not sorted, as this call may rely on index positions
         if not self.isSorted and self.autoSort:
@@ -447,15 +448,20 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
                 groupStream = self.iter.getElementsByGroup(k)
                 if groupStream:
                     return groupStream.stream()
-                else:
-                    raise KeyError(f'provided key ({k}) does not match any id or group')
-        elif isinstance(k, type(type)):
+                # else:
+                    # raise KeyError(f'provided key ({k}) does not match any id or group')
+
+        if isinstance(k, (str, type)):
             # assume it is a class name
             classStream = self.iter.getElementsByClass(k)
             if classStream:
                 return classStream.stream()
-            else:
+            elif isinstance(k, type):
                 raise KeyError(f'provided class ({k}) does not match any contained Objects')
+            else:
+                raise KeyError(f'provided key ({k}) does not match any id, group, or class name')
+
+        raise TypeError(f'Get items by int, slice, id, or class; got {type(k)}')
```

jacobtylerwalls:
Fixes #947 

Changed in v7:
  - [earlier PR:] out of range indexes now raise an IndexError, not StreamException
  - class searches now recursive
  - negative slice indices now work
  -  Unsupported types now raise TypeError
  - Class and string searches now return a `StreamIterator` rather than a `Stream`
  - slice searches return a list, not a `Stream`

### DIFF
--- a/music21/stream/iterator.py
+++ b/music21/stream/iterator.py
@@ -89,6 +89,16 @@ class StreamIterator(prebase.ProtoM21Object):
     * For `activeInformation` see above.
 
     Changed in v.5.2 -- all arguments except srcStream are keyword only.
+
+    OMIT_FROM_DOCS
+
+    Informative exception for user error:
+
+    >>> s = stream.Stream()
+    >>> sIter = stream.iterator.StreamIterator(s, filterList=[note.Note])
+    Traceback (most recent call last):
+    TypeError: filterList expects Filters or callables,
+    not types themselves; got <class 'music21.note.Note'>
     '''
     def __init__(self,
                  srcStream: 'music21.stream.Stream',
@@ -122,6 +132,10 @@ class StreamIterator(prebase.ProtoM21Object):
             filterList = [filterList]
         elif isinstance(filterList, (set, tuple)):
             filterList = list(filterList)  # mutable....
+        for x in filterList:
+            if isinstance(x, type):
+                raise TypeError(
+                    f'filterList expects Filters or callables, not types themselves; got {x}')
         # self.filters is a list of expressions that
         # return True or False for an element for
         # whether it should be yielded.
--- a/music21/stream/iterator.py
+++ b/music21/stream/iterator.py
@@ -957,19 +957,24 @@ class StreamIterator(prebase.ProtoM21Object):
         >>> s = converter.parse('tinyNotation: 4/4 GG4 AA4 BB4 r4 C4 D4 E4 F4 r1')
         >>> s[note.Note].last().id = 'last'
         >>> for n in s[note.Note]:
-        ...     if n.octave == 4:
-        ...         n.groups.append('high')
+        ...     if n.octave == 3:
+        ...         n.groups.append('tenor')
 
-        >>> s.recurse().getElementsByQuerySelector('#last').first()
-        <music21.note.Note F>
-        >>> list(s.recurse().getElementsByQuerySelector('.high'))
+        >>> list(s.recurse().getElementsByQuerySelector('.tenor'))
         [<music21.note.Note C>,
          <music21.note.Note D>,
          <music21.note.Note E>,
          <music21.note.Note F>]
+
         >>> list(s.recurse().getElementsByQuerySelector('Rest'))
-        [<music21.note.Rest quarter>,
-         <music21.note.Rest whole>]
+        [<music21.note.Rest rest>,
+         <music21.note.Rest rest>]
+
+        Note that unlike with stream slices, the querySelector does not do anything special
+        for id searches.  `.first()` will need to be called to find the element (if any)
+
+        >>> s.recurse().getElementsByQuerySelector('#last').first()
+        <music21.note.Note F>
 
         New in v.7
         '''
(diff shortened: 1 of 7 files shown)


jacobtylerwalls:
Starting the implementation before we settle on whether to do a recursive search.

mscuthbert:
Let's do a substantial change in v.7 -- maybe we can say this:

* int = index -- return one Element(and change the docs to deemphasize non-sorted streams)
* slice e.g.,`4:7` -- return list of Elements in the slice
* class -- return recursive stream iterator returning elements by type
* string -- return stream recursive iterator with .getElementsByQuerySelector()

`QuerySelector()` ?????  -- it's something I've been wanting to add since, OH, v.1 -- setting up something like an HTML query selector:

* s['#hello'].first()  -- matches element matching ID 'hello' -- same as DOM querySelectorAll('#hello')
* s['.high'] --  matches element matching group 'high' -- same as DOM class selector
* s['Note'] -- matches elements of class 'Note' -- same as DOM tag/type selector

this would be the basic query selector and all we would implement now.  BUT it opens the avenue for setting up more sophisticated selectors such as https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors   The most important would be something like:

s['Note[duration.type="quarter"]'] 

(the issue that would come up though is that 'note.Note' cannot be considered to be a "note"-class followed by a "Note" group.)

jacobtylerwalls:
Cool -- at first I was slightly skeptical of getting too fancy/unpythonic, but I read this over a couple times and it makes sense to me, given the character of our Stream objects.

Made class searches recursive. Might have found an issue with #944 , see alternative line for `classIter` (or wait for me to report it :-D)

Noticed negative slice indices weren't working, also, so fixed it.

Next step (at some point, not immediately) is to implement the basic query selector. And some more tests.

music21/stream/iterator.py
```diff
@@ -337,6 +351,9 @@ def __getitem__(self, k):
 
         (nothing is printed)
         '''
+        if isinstance(k, type):
+            self.addFilter(k)
+            return self
```
jacobtylerwalls, line 356: This is here to support chained lookups like s[x][y]... needs docs and test if we want it.

mscuthbert: Ah -- I don't think that this will exactly work.  maybe remove it for now and then we can put it back in during a refactor of Stream.getItem, because otherwise we'll end up doing too much duplicated code to get something like: Stream.parts['Note'][4] to work, and it'd be weird because `.parts` is a normal, non recursive iterator, but we're now training people to think of `Stream[Note]` as making a recursive iterator.

music21/stream/iterator.py
```diff
@@ -921,6 +938,53 @@ def getElementsByClass(self, classFilterList, *, returnClone=True):
...
+
+        >>> list(s.recurse().getElementsByQuerySelector('.tenor'))
+        [<music21.note.Note C>,
+         <music21.note.Note D>,
+         <music21.note.Note E>,
+         <music21.note.Note F>]
+
+        >>> list(s.recurse().getElementsByQuerySelector('Rest'))
+        [<music21.note.Rest rest>,
+         <music21.note.Rest rest>]
```
mscuthbert, line 971: This and one test in Stream.base will immediately fail when the rest Repr is put in.  I had thought it was already put in, and was annoyed when I didn't get 'quarter', 'whole' here; hence that quick PR

PR #973, opened 2021-04-21
Raised TimeSignatureException when instantiating with a string representing floats

### LINKED ISSUE #964: meter.TimeSignature objects with `.0` in string result in incorrect output
**music21 version**

`6.7.1`   

**Problem summary**
If a `meter.TimeSignature` object is created with a float in the `value` string, it has an incorrect numerator and denominator.

**Steps to reproduce**

```
>>> x = TimeSignature("5.0/4.0")
```

I'm expecting the following:
```
>>> x
<music21.meter.TimeSignature 5/4>
```
but am instead getting
```
>>> x 
<music21.meter.TimeSignature 50/40>
```
**More information**
I haven't tried to debug it, but I think the issue is in the regex line in `meter.core.tools.slashToTuple` (see line 61). 
https://github.com/cuthbertLab/music21/blob/450859475497dd70510e5ebde6184536bdac15a8/music21/meter/tools.py#L50-L64
I'm not sure if it's better to assert that the two `value` elements are of type `int` when loaded, or allow them if they're still digits.

Luke-Poeppel:
…ing. This fixes #964.

### DIFF
--- a/music21/meter/tools.py
+++ b/music21/meter/tools.py
@@ -123,6 +123,10 @@ def slashMixedToFraction(valueSrc: str) -> Tuple[NumDenomTuple, bool]:
     Traceback (most recent call last):
     music21.exceptions21.MeterException: cannot match denominator to numerator in: 3+2+5/8+3/4+2+1+4
 
+    >>> meter.tools.slashMixedToFraction('3.0/4.0')
+    Traceback (most recent call last):
+    music21.exceptions21.TimeSignatureException: Cannot create time signature from "3.0/4.0"
+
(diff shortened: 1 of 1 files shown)


jacobtylerwalls:
Thanks--Looks great. Would you be able to add a test to cover this case? Could just be appended to the end of the doctest on slashMixedToFraction.

jacobtylerwalls:
Thanks for the report and patch!

mscuthbert:
Thanks @Luke-Poeppel !  :-)

PR #975, opened 2021-04-21
Prevent intent boxes from turning into task list

jacobtylerwalls:
FYI @mscuthbert I don't like how the intent boxes on the feature request template show up as task lists when viewing the rows of the issue board. We don't have to use actual Markdown for this. What do you think?


mscuthbert:
Good plan

PR #979, opened 2021-04-22
Remove SpacerRests, normal rests with hideObjectOnPrint

mscuthbert:
After doing this, however, maybe spacer rests ARE different from just hidden rests?  Now after wanting to remove them for years, I'm conflicted.  @jacobtylerwalls -- want to weigh in?


jacobtylerwalls:
I don't see a reason to have two ways to represent a deliberate gap. Undeliberate gaps are just absences, but we should have one way to represent deliberate gaps, and we already use hidden rests in musicxml I/O and with `makeRests(hideRests=True)` so I would reduce complexity here.

Issue #223, opened 2017-06-03
StreamException: the object * is already found in this stream


euneirophrenia:
Hi! Nice library to begin with and sorry if I'm posting an issue if the issue turns out to actually be my brain.

I'm trying to extract some features from a midi file.

Feature: jSymbolic.AcousticGuitarFractionFeature
File      : http://midkar.com/jazz/500_miles_high-Chick-Corea_ee.mid
Code: 
```python
import music21.features as feat
s=music21.converter.parse("./Test/500_miles_high-Chick-Corea_ee.mid")
ext=feat.jSymbolic.AcousticGuitarFractionFeature
print (ext(s).extract().vector)
```

Problem: 
```python
Traceback (most recent call last):
File "main.py", line 16, in <module>
    print(ext(s).extract().vector)
File "C:\Python34\lib\site-packages\music21\features\base.py", line 209, in extract
    self._process() # will set Feature object to _feature
File "C:\Python34\lib\site-packages\music21\features\jSymbolic.py", line 3462, in _process
    s = self.data['partitionByInstrument']`
File "C:\Python34\lib\site-packages\music21\features\base.py", line 627, in __getitem__
    return self._forms[key]
File "C:\Python34\lib\site-packages\music21\features\base.py", line 341, in __getitem__
    x = instrument.partitionByInstrument(self._base)
File "C:\Python34\lib\site-packages\music21\instrument.py", line 1730, in partitionByInstrument
    p.insert(sub.elementOffset(e), e)
File "C:\Python34\lib\site-packages\music21\stream\__init__.py", line 1677, in insert
    self._addElementPreProcess(element)
File "C:\Python34\lib\site-packages\music21\stream\core.py", line 290, in _addElementPreProcess
    (element, id(element), self, id(self)))
music21.exceptions21.StreamException: the object (<music21.tempo.MetronomeMark larghetto Quarter=60.0>, id()=116292632) is already found in this Stream (<music21.stream.Part Flute>, id()=306686944)
```

(i'm so sorry for the poor formatting, hope it's still somewhat readable EDIT: fixed it a little).
I've been going through the same problems as in this issue 
https://github.com/cuthbertLab/music21/issues/150
(even the opening one, by the way, I had to change 2 lines at jsymbolic.py 3130 and 3216, adding a `and i.midiprogram is not None` ).
The error is essentially the same as in the last comment. I've searched every single issue here but never found the right one, so I am posting. 

[...]

Issue #150, opened 2016-03-20
PitchedInstrumentsPresentFeature 


supersational:
Hi, fantastic library! 

Currently trying to extract all features from a midi file:
```python
for f in file_list:
	print f
	s = converter.parse(f.strip())
	for f in features.jSymbolic.featureExtractors:
		print f.__name__
		fe = f(s)
		print getattr(fe, 'description', 'no description')
		print fe.extract().vector
```
currently fails at PitchedInstrumentsPresentFeature:
```
Traceback (most recent call last):
  File "C:\feature_extractors.py", line 21, in <module>
    print fe.extract().vector
  File "C:\Python27\lib\site-packages\music21\features\base.py", line 198, in extract
    self._process() # will set Feature object to _feature
  File "C:\Python27\lib\site-packages\music21\features\jSymbolic.py", line 2933, in _process
    self._feature.vector[i.midiProgram] = 1
TypeError: list indices must be integers, not NoneType
```

midi file here;
[009count.zip](https://github.com/cuthbertLab/music21/files/181381/009count.zip)

Is this expected behaviour? the midifile in question seems to have standard messages for setting instruments as far as I can tell (from mido analysis):
```
program_change channel=0 program=1 time=0
program_change channel=2 program=8 time=0
program_change channel=3 program=25 time=0
program_change channel=4 program=33 time=0
program_change channel=7 program=48 time=0
program_change channel=9 program=16 time=0
program_change channel=10 program=52 time=0
program_change channel=9 program=16 time=0.193661875
```

alexandermorgan:
This doesn't address your question but it is definitely less than ideal to assign f to the elements in both of your nested loops.

supersational:
Ah, agreed! Seems to be harmless in this instance at least.

mscuthbert:
Not expected at all -- can you trim down the MIDI file and the code until as small as possible while giving the same error, so that the programming team can debug faster.  The first thing I'd see is if either something in channel 9 (the only one that has a time message later) or channel 10 (which isn't a pitched instrument) is causing the problem.  I'm going to guess that channels 0-7 all behave the same.

supersational:
Hi, this should trigger it:
```python
import music21
f = "009count.mid"
s = music21.converter.parse(f)
e = music21.features.jSymbolic.PitchedInstrumentsPresentFeature
fe = e(s)
print fe.extract().vector
```

mscuthbert:
[...]

PR #980, opened 2021-04-22
Check if `Instrument.midiProgram` is None in feature extractors

jacobtylerwalls:
Generic `Instrument` objects (the fall back for several file parsers) have a `.midiProgram` of None, so we can't use that value as an index in the feature extractors.

```
>>> i = instrument.Instrument()
>>> s = stream.Stream()
>>> s.append(i)
>>> s.append(note.Note())
>>> fe2 = features.jSymbolic.PitchedInstrumentsPresentFeature(s)
>>> fe2.extract()
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/music21/features/base.py", line 249, in extract
    self.process()  # will set Feature object to _feature
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/music21/features/jSymbolic.py", line 3354, in process
    self.feature.vector[i.midiProgram] = 1
TypeError: list indices must be integers or slices, not NoneType
```

Since #571 we still want this to fail, just with a more informative error so that users can assess whether to fix up their data.

Happened to notice there was an old issue for this #150 (and mentioned also in #223).


mscuthbert:
Oh!  The #223 error was only recently fixed, with the template fixes.

mscuthbert (approved):
Looks great.  Thanks!

PR #922, opened 2021-03-18
Add classFilterList to makeTies and allow makeRests to put rests in Measures

jacobtylerwalls:
Noticed that we can solve 2 TODOs at once: 

- `makeRests()` makes rests inside Measures now
- in order to do that, we need to split rests at measure boundaries, which is what `makeTies()` does, but we don't want to touch the notes, therefore it was convenient to just handle the TODO declared there to create a class filter list.
- Update: now this also fixes some issues with MIDI notation making:
-- measures are now repositioned in the stream if they were enlarged due to gap-filling
-- ties are now made on notes and chords across barlines (vs. just rests, by virtue of this PR's new feature in makeRests())

PR #981, opened 2021-04-22
Prevent unnecessary deepcopies in `makeRests()`

jacobtylerwalls:
Missed in #922.

`returnObj` is already a deepcopy if need be, so just like the branch below for voices, we should just edit inPlace instead of doing more copies.


mscuthbert:
Oh, in fact it's more than just an extra deepcopy -- the result of makeTies is removed!

PR #982, opened 2021-04-23
getGraceDuration() now sets type on GraceDuration

### LINKED ISSUE #968: `getGraceDuration()` doesn't preserve unlinked type
**music21 version**

dev

**Problem summary**
Seems inappropriate to drop the unlinked type when creating a GraceDuration.

**Steps to reproduce**

```
>>> r = note.Rest(quarterLength=0.5)
>>> r.duration.type
'eighth'
>>> r2 = r.getGrace(inPlace=False)
>>> r2.duration.type
'zero'
>>> r2.duration
<music21.duration.GraceDuration unlinked type:zero quarterLength:0.0>
```

**More information**
Spotted when working on #941

jacobtylerwalls:
Fixes #968 

Reproduced the bug on 3.1.0, actually!

The issue was that you can't derive type from components once the duration is unlinked. And the constructor for a GraceDuration sets .linked = False shortly after calling the superclass constructor. So the GraceDuration constructor is your only chance to get a type set automatically. Figured this fix, deliberately setting type, was cleaner than some sort of (`if not self.components: self.linked = True`) logic in the GraceDuration constructor that might just cause bugs later.

### DIFF
--- a/music21/_version.py
+++ b/music21/_version.py
@@ -42,7 +42,7 @@ When changing, update the single test case in base.py.
 Changing this number invalidates old pickles -- do it if the old pickles create a problem.
 '''
 
-__version_info__ = (7, 0, 4)  # can be 4-tuple: (7, 0, 5, 'a2')
+__version_info__ = (7, 0, 5)  # can be 4-tuple: (7, 0, 5, 'a2')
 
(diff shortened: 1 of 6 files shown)


mscuthbert (approved):
Thanks!  Looks great!

PR #983, opened 2021-04-24
Provide default type when exporting grace durations to musicxml

jacobtylerwalls:
#982 was a little too strict when defining the new error condition for 'zero' duration.

If a user creates a vanilla GraceDuration() (i.e. without calling `getGrace()` to get a grace duration with a type), and it makes it to the MusicXML exporter, it seems harmless to give them a default of an eighth.


mscuthbert (approved):
Great! Thanks!

PR #984, opened 2021-04-24
Resolve Pylint consider using with and consider using max/min violations

jacobtylerwalls:
Suppressed the warnings for `consider-using-with` when there was a set of `open()`, `read()`, and ` write()` methods that would need a bigger refactor. Resolved the others.

multiprocessTest generated a huge diff, but it's just indentation and breaking lines on an f-string that would have gone over the line length limit.

Refactored `SubConverter.writeDataStream()` -- this is where I want eyes when you have time. (No rush!) There were too many branches before, IMO. Separated the decoding logic from the writing logic to reduce repetition.

### DIFF
--- a/music21/abcFormat/__init__.py
+++ b/music21/abcFormat/__init__.py
@@ -3235,7 +3235,7 @@ class ABCFile(prebase.ProtoM21Object):
         Open a file for reading
         '''
         # try:
-        self.file = io.open(filename, encoding='utf-8')
+        self.file = io.open(filename, encoding='utf-8')  # pylint: disable=consider-using-with
         # except
         # self.file = io.open(filename, encoding='latin-1')
         self.filename = filename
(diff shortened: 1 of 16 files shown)


jacobtylerwalls:
Decided to merge this because I have two PRs handling additional cases around recent dev, and I don't want them (or anyone else's PRs) to be slowed down on these new checks. If anyone has concerns, please open a new issue, and I'll take a look.

- `writeDataStream()` is overridden on the musicxml subconverter, so these changes affect writing MIDI, braille, etc. Only LilyPond and Braille have `codecWrite=True`. Tested `.write('braille')`. Looks good.
- Tested `python3 -m music21.test.multiprocessTest`. Works.
- The other changes are very low risk.

mscuthbert:
Sounds good -- yeah, usually better to have a second set of eyes, but when there are things waiting, go ahead and merge, since we can always unmerge if there are problems.

Happy to turn on `consider-using-with` as a pylint check.  And in general the max(x, 0) etc. is fine, though sometimes I like the explicit `if ql < 0:` check because it gives an opportunity later to turn that into a throw exception type of bug.

PR #985, opened 2021-04-26
Don't flatten Stream containers when passing to Stream()

jacobtylerwalls:
**Before**
`Score(Part(Measure(Note())))` gave Score > Measure > Note, because `common.isIterable(myPart)` evaluated True and we just took the `.elements` of `myPart`, effectively flattening the provided container away.

#946 didn't make this any worse, as far as I can tell, but by providing additional support for single elements, we would have nudged people toward this banana peel by encouraging them to send single elements to the constructor. So this is another case we should handle.

**Now**
We don't flatten the provided container when sending that single Stream container to `Stream()`, i.e. we wrap it in a list.

Spruced up some doctests that were previously demonstrating the antipattern.

There are some **very** deep problems with this previous antipattern in LilyPond and Variant that I'm not volunteering to resolve ATM, but I was able to get the tests to pass by deliberately asking for the result of the old antipattern--i.e. flattening the provided container--by sending `.elements` to `Stream()`.

### DIFF
--- a/music21/variant.py
+++ b/music21/variant.py
@@ -171,15 +171,15 @@ def mergeVariantScores(aScore, vScore, variantName='variant', *, inPlace=False):
 
     >>> aScore, vScore = stream.Score(), stream.Score()
 
-    >>> ap1 = stream.Part(converter.parse('tinynotation: 4/4   a4 b c d    e2 f2   g2 f4 g4 '
-    ...                                   ).makeMeasures())
-    >>> vp1 = stream.Part(converter.parse('tinynotation: 4/4   a4 b c e    e2 f2   g2 f4 a4 '
-    ...                                   ).makeMeasures())
+    >>> ap1 = converter.parse('tinynotation: 4/4   a4 b c d    e2 f2   g2 f4 g4 '
+    ...                                   ).makeMeasures()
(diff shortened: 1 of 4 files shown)


music21/variant.py
```diff
@@ -171,15 +171,15 @@ def mergeVariantScores(aScore, vScore, variantName='variant', *, inPlace=False):
...
     >>> aScore, vScore = stream.Score(), stream.Score()
 
-    >>> ap1 = stream.Part(converter.parse('tinynotation: 4/4   a4 b c d    e2 f2   g2 f4 g4 '
-    ...                                   ).makeMeasures())
-    >>> vp1 = stream.Part(converter.parse('tinynotation: 4/4   a4 b c e    e2 f2   g2 f4 a4 '
-    ...                                   ).makeMeasures())
+    >>> ap1 = converter.parse('tinynotation: 4/4   a4 b c d    e2 f2   g2 f4 g4 '
+    ...                                   ).makeMeasures()
+    >>> vp1 = converter.parse('tinynotation: 4/4   a4 b c e    e2 f2   g2 f4 a4 '
+    ...                                   ).makeMeasures()
```
mscuthbert, line 177: I think makeMeasures() can go -- it's done automatically since v5 or 6?

mscuthbert:
LGTM -- can we see if makeMeasures after tinyNotation is no longer needed?  THANKS!

PR #986, opened 2021-04-26
makeRests() gets bar duration context for voices from measures and recurses into Parts and checks for measure padding

jacobtylerwalls:
Handling some cases not handled by #922.

**Before**
This describes the first bullet point under "Now", below.
#922 was the first effort toward recursing in `makeRests()`. It wasn't quite enough -- final rests in Voices weren't created from the last element to the end of the bar, because the Voice objects don't have a `.barDuration`. Then, a loose rest was eventually created outside of the voices on the Measure object (see the regression test I included--the first test would have failed). Update: I'm actually only describing the MIDI parsing workflow. Reading voices as defined in a musicXML file is unaffected.

**Now**
- refactored `makeRests` to handle voices early enough that measure bar duration is accessible for voices
- recursed over parts, since that was the remaining nested stream container to handle
- defined and documented priority between keyword args `timeRangeFromBarDuration` and `refStreamOrTimeRange`
- made `timeRangeFromBarDuration` account for measure padding
- documented what was a missing feature on `makeVoices()` that now we handle creating trailing rests in voices when an incomplete measure is provided
- rests created by musicxml export AFTER makeNotation runs are now hidden

**Future**
<s>`timeRangeFromBarDuration` should probably default True. </s> Nope! Too presumptuous--don't want to "complete" every incomplete voice parsed from musicXML. <s>Some tests fail in interesting ways if I flip the default, suggesting possibly better results, but also that some work might be needed with `.paddingLeft`.</s> Handled the padding issues.

### DIFF
--- a/music21/stream/makeNotation.py
+++ b/music21/stream/makeNotation.py
@@ -705,7 +705,8 @@ def makeRests(
     time regions that have no active elements.
 
     If `timeRangeFromBarDuration` is True, and the calling Stream
-    is a Measure with a TimeSignature, the time range will be determined
+    is a Measure with a TimeSignature (or a Part containg them),
+    the time range will be determined
     by taking the :meth:`~music21.stream.Measure.barDuration` and subtracting
     :attr:`~music21.stream.Measure.paddingLeft` and
     :attr:`~music21.stream.Measure.paddingRight`.
--- a/music21/stream/makeNotation.py
+++ b/music21/stream/makeNotation.py
@@ -706,8 +706,11 @@ def makeRests(
 
     If `timeRangeFromBarDuration` is True, and the calling Stream
     is a Measure with a TimeSignature, the time range will be determined
-    based on the .barDuration property. This keyword takes priority
-    over `refStreamOrTimeRange`. If both are provided, `timeRangeFromBarDuration`
+    by taking the :meth:`~music21.stream.Measure.barDuration` and subtracting
+    :attr:`~music21.stream.Measure.paddingLeft` and
+    :attr:`~music21.stream.Measure.paddingRight`.
+    This keyword takes priority over `refStreamOrTimeRange`.
+    If both are provided, `timeRangeFromBarDuration`
     prevails, unless no TimeSignature can be found, in which case, the function
     falls back to `refStreamOrTimeRange`.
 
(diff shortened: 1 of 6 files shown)


jacobtylerwalls:
Ah, the .paddingLeft issues are a regression in #922 because we uncovered functionality and didn't handle this case. makeRests() *is* called in MusicXML export, just didn't do much if you already had measures (again, only streams made from MIDI were typically in that scenario before v7, but now even those have measures).

I'll push another commit to handle paddingLeft

music21/stream/makeNotation.py
```diff
@@ -792,7 +801,11 @@ def makeRests(
 
     Changed in v6 -- all but first attribute are keyword only
 
-    Changed in v7 -- `inPlace` defaults False.
+    Changed in v7:
+
+      - `inPlace` defaults False
+      - Recurses into parts, measures, voices
+      - Gave priority to `timeRangeFromBarDuration` over `refStreamOrTimeRange`
```
mscuthbert, line 808: btw -- I wonder if refStreamOrTimeRange can be removed -- is it ever used?

mscuthbert (approved):
I think it looks good -- I don't remember all the subtilties, but if you're confident, feel free to merge.  Otherwise I'll take a deep dive.

music21/stream/makeNotation.py
jacobtylerwalls: If we remove it, we'll be putting a constraint that you can only make rests in measures and voices, not flat streams. I'm not keen on taking that away.

After some thought I think it's time to default timeRangeFromBarDuration True. Especially since now we've given it priority.

jacobtylerwalls: Nope, on even more thought, it's too sensitive to turn this on, because now that we recurse, we will descend into every part and measure and voice and complete every voice, filling measures with all kinds of unwanted rests. Better to leave how currently configured -- in the musicxml parsers, measures should get timeRangeFromBarDuration=True in the musicxml parsers, voices should not, which is what we have without adding more changes here.

jacobtylerwalls:
> I think it looks good -- I don't remember all the subtilties, but if you're confident, feel free to merge. Otherwise I'll take a deep dive.

Thanks for having a look. I think it's time to merge this. Found another rest issue, but one that's piano-specific, can reproduce on v.3.1.0 and doesn't look any worse with this patch, so if you want to don your diving goggles we can do that then. In the meantime we have some output that's wrong we can fix by merging this and bumping the minor version number.

jacobtylerwalls:
This is worth a second (shallow) dive given that I added three more commits. Feel good about this now, but again, would appreciate a double-check.

jacobtylerwalls:
I'll be merging this shortly. Not trying to rush anything! Just have another PR waiting on this, and in hindsight, the last set of changes doesn't really seem to require another substantive review. Those last changes were:
- Avoiding redundant calls to makeRests given this PR's new recursion by part
- Hiding rests created on musicxml export, since we have decided spacer rests == hidden rests.

I'm squashing the trivial whitespace and typo fixes but leaving substantive commits for the independent pieces in the offchance we need to revert anything later. But this leaves the master branch in a better state to be merging this now, in order to resolve the issue with measure padding not being observed (causing extra rests on export).

PR #987, opened 2021-04-27
Set `Interval.noteStart` when `melodicIntervals()` encounters Chords

jacobtylerwalls:
I was using `melodicIntervals()` with the intent to find desired intervals and then extract the measure context by accessing .noteStart and then .derivation.origin, but I was encountering broken derivation chains because of the way `.noteStart` is set *after* the interval is created from `Pitch` objects:

```
>>> n = note.Note()
>>> n2 = note.Note()
>>> i = interval.notesToInterval(n.pitch, n2.pitch) # from pitches
>>> i.noteStart = n  # set the note afterward
>>> i.noteEnd = n2
>>> i.noteStart is n
False
>>> i.noteEnd is n2
True
```

The above example only happens in `melodicIntervals()` if chords are involved, so this PR just has melodicIntervals handle chords as it does notes and create intervals from Note objects, not Pitch objects. Took the opportunity to simplify a little bit, also.

### DIFF
--- a/music21/stream/base.py
+++ b/music21/stream/base.py
@@ -9798,7 +9798,6 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
         findConsecutiveNotes.
 
 
-
         >>> s1 = converter.parse("tinynotation: 3/4 c4 d' r b b'", makeNotation=False)
         >>> #_DOCS_SHOW s1.show()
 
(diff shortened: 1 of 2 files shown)


music21/stream/base.py
```diff
@@ -9834,33 +9833,26 @@ def melodicIntervals(self, *skipArgs, **skipKeywords):
...
-                                                   + firstNote.duration.quarterLength)
-                    returnInterval.duration = duration.Duration(opFrac(
-                        secondNote.offset - returnInterval.offset))
-                    returnStream.insert(returnInterval)
+            # returnList could contain None to represent a rest
+            if firstNote is None or secondNote is None:
+                continue
+            # Protect against empty chords
+            if not (firstNote.pitches and secondNote.pitches):
+                continue
```
mscuthbert, line 9841: NICE! Thank you!

mscuthbert (approved):
This is embarrassingly old code that predated having Notes in Chords.  This is definitely a step forward.

PR #989, opened 2021-04-28
Ignore grace notes when beaming

### LINKED ISSUE #345: beaming is broken in some cases where grace notes appear inside of tuplets
Beaming of a tuplet is broken whenever a grace note appears inside a tuplet.

Here is a possible fix, which skips grace notes when making beams.
See line `realNotes = [n for n in noteStream if n.duration.quarterLength > 0]`) and subsequent substitutions

```python

def makeBeams(s, *, inPlace=False):
    from music21 import stream
    returnObj = s if inPlace else copy.deepcopy(s)
    if 'Measure' in s.classes:
        mColl = [returnObj]  # store a list of measures for processing
    else:
        mColl = list(returnObj.iter.getElementsByClass('Measure'))  # a list of measures
        if not mColl:
            raise stream.StreamException('cannot process a stream that is neither a Measure nor has no Measures')

    lastTimeSignature = None
    for m in mColl:
        # this means that the first of a stream of time signatures will be used
        if m.timeSignature is not None:
            lastTimeSignature = m.timeSignature
        if lastTimeSignature is None:
            raise stream.StreamException('cannot process beams in a Measure without a time signature')
        noteGroups = []
        if m.hasVoices():
            for v in m.voices:
                noteGroups.append(v.notesAndRests.stream())
        else:
            noteGroups.append(m.notesAndRests.stream())

        for noteStream in noteGroups:
            if len(noteStream) <= 1:
                continue  # nothing to beam
            realNotes = [n for n in noteStream if n.duration.quarterLength > 0]
            durList = [n.duration for n in realNotes]
            # error check; call before sending to time signature, as, if this
            # fails, it represents a problem that happens before time signature
            # processing
            summed = sum([d.quarterLength for d in durList])
            durSum = opFrac(opFrac(summed))  # the double call corrects for tiny errors in adding
            # floats and Fractions in the sum() call -- the first opFrac makes it
            # impossible to have 4.00000000001, but returns Fraction(4, 1). The
            # second call converts Fraction(4, 1) to 4.0
            barQL = lastTimeSignature.barDuration.quarterLength
            if durSum > barQL:
                continue
            # getBeams
            offset = 0.0
            if m.paddingLeft != 0.0:
                offset = opFrac(m.paddingLeft)
            elif noteStream.highestTime < barQL:
                offset = barQL-noteStream.highestTime

            beamsList = lastTimeSignature.getBeams(realNotes, measureStartOffset=offset)
            for n, beams in zip(realNotes, beamsList):
                n.beams = beams if beams is not None else m21.beam.Beams()

    del mColl  # remove Stream no longer needed

    returnObj.streamStatus.beams = True
    return returnObj


```

jacobtylerwalls:
Fixes #345 

Applied equally with or without tuplets -- just an issue with makeBeams() in general.

**Before**
Screen Shot 2021-04-28 at 2 16 41 PM

**Patch**
Screen Shot 2021-04-28 at 2 16 45 PM

The variables `noteStream` and `noteGroups` are temp streams, so it's safe to edit.

### DIFF
--- a/music21/stream/makeNotation.py
+++ b/music21/stream/makeNotation.py
@@ -157,6 +157,7 @@ def makeBeams(
             for n in noteStream:
                 if n.duration.isGrace:
                     noteStream.remove(n)
+                    continue
                 durList.append(n.duration)
             # environLocal.printDebug([
             #    'beaming with ts', lastTimeSignature, 'measure', m, durList,
--- a/music21/stream/makeNotation.py
+++ b/music21/stream/makeNotation.py
@@ -90,6 +90,22 @@ def makeBeams(
     4 <music21.beam.Beams <music21.beam.Beam 1/continue>/<music21.beam.Beam 2/continue>>
     5 <music21.beam.Beams <music21.beam.Beam 1/stop>/<music21.beam.Beam 2/stop>>
 
+    Grace notes no longer interfere with beaming:
+
+    >>> m = stream.Measure()
+    >>> m.timeSignature = meter.TimeSignature('3/4')
+    >>> m.repeatAppend(note.Note(quarterLength=0.25), 4)
+    >>> m.repeatAppend(note.Rest(), 2)
+    >>> gn = note.Note(duration=duration.GraceDuration())
+    >>> m.insert(0.25, gn)
+    >>> m.makeBeams(inPlace=True)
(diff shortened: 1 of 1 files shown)


mscuthbert:
Approved!  Just added a little future-proofing `continue`.

Issue #988, opened 2021-04-28
Create more specific instruments from MusicXML import


retorquere:
[op-voor-op-achter.txt](https://github.com/cuthbertLab/music21/files/6390745/op-voor-op-achter.txt)

When I open this file directly with musicxml, this shows and plays as a drum score. When I open it using this:

```
from music21 import *
s = converter.parse('op-voor-op-achter.musicxml')
s.show()
```

it opens as a piano score, and the notes that used to be hihats (top of the bar) are now below the bar

retorquere:
Also this:

```
import music21 as m

s = m.converter.parse('op-voor-op-achter.musicxml')
for note in s.recurse().notes:
  print(note.name)
```

shows all notes to be `C` in the attached file? That isn't right, there's 3 different notes in the score.

jacobtylerwalls:
Hello, and welcome. In the future please use the bug template. For instance, what version of music21 are you using? On the latest version (6.7.1), MuseScore and Finale both show percussion clef and drumset:

![Screen Shot 2021-04-28 at 10 08 54 AM](https://user-images.githubusercontent.com/38668450/116418026-c51b3a00-a809-11eb-8557-b56aa8397edf.png)

Regarding the unpitched noteheads appearing below the staff, this is a known missing feature on the roadmap for v7 (#235).

No guarantees what features will make it in to v7, but stay tuned for announcements about the release of v7 on the mailing list if you like.  Cheers

retorquere:
Apologies - I didn't see a template, I just got a blank issue form.

retorquere:
I'm on music21 6.7.1 and musescore 3.6.2.538020600

jacobtylerwalls:
Thanks for the additional info. Can you clarify what you mean by piano score?

retorquere:
Sorry - I meant that it sounded like a piano when I played it.

jacobtylerwalls:
Thanks. If you install the dev version you might find that piece is already resolved -- see #961 .

retorquere:
Superb -- how do I install the dev version?

jacobtylerwalls:
```
pip3 uninstall music21
pip3 install git+https://github.com/cuthbertLab/music21.git
```

retorquere:
I don't know how I got to the issue submission screen previously but I got here through https://github.com/cuthbertLab/music21/issues/new/ rather than https://github.com/cuthbertLab/music21/issues/choose/. I believe that can be disabled somewhere.

[...]

PR #990, opened 2021-04-28
Catch MIDIPercussionException and just emit a warning

jacobtylerwalls:
Refs #988 (but doesn't close it, since we will track the MusicXML issues there)

cc @retorquere, thanks again for finding it!


mscuthbert (approved):
Thanks for quick diagnosis and fix!

PR #993, opened 2021-04-30
More miscellany

jacobtylerwalls:
Various cleanups and tiny fixes that didn't seem worth opening separate PRs for.

- Use pathlib in `Environment.__setitem__()`
- Fix module names in file headers
- Fix compileall command -- GitHub Actions log shows this wasn't compiling
- Typo in ch. 53
- Update test coverage
- Fix bug with intermittent metadata writing
- Enable providing path to saveFile arg in `runTranscribe()` (discovered when replying to a mailing list post)
- Better failure in audioSearch module if no pyaudio
- Remove unused EnumerationException
- Prevent crash in exporting MIDI from having MetronomeMark without a number
- Reduce indentation in `midiTrackToStream()` and remove redundant branch for file consisting of single note
- Don't export invalid StringIndications
- Remove check for hasattr 'software'
- Remove moot TODO in musicxml parser
- Add VS Code config files to .gitignore
- Avoid some calls to hyphenToCamelCase()
- Restore `- minPitch` for future safety (thought better of this after final changes in #965)


jacobtylerwalls:
Perhaps we shouldn't squash the commits on merge, since they're independent.

music21/environment.py
```diff
@@ -1268,8 +1268,8 @@ def __setitem__(self, key, value):
         if key in self._environment.getKeysToPaths():
             # try to expand user if found; otherwise return unaltered
             if value is not None and not str(value).startswith('/skip'):
-                value = common.cleanpath(value, returnPathlib=False)
-                if not os.path.exists(value):
+                value = common.cleanpath(value, returnPathlib=True)
+                if not value.exists():
```
mscuthbert, line 1272: Does `self._envrionment.__setitem__` work with a Pathlib object as a key?  Not sure.

music21/midi/translate.py
```diff
@@ -1860,91 +1868,73 @@ def midiTrackToStream(
...
-
-    elif len(notes) == 1:  # rare case of just one note
-        n = note.Note()
-        midiEventsToNote(notes[0], ticksPerQuarter, n)
-        # the time is the first value in the first pair
-        # need to round, as floating point error is likely
-        o = notes[0][0][0] / ticksPerQuarter
-        n.midiTickStart = notes[i][0][0]
-        s.coreInsert(o, n)
+    for i, notePair in enumerate(notes):
```
mscuthbert, line 1871: I'm not sure I understand why the special case of just one note can be removed?  I think we wouldn't have put it here if we didn't find it "in the wild"

music21/musicxml/m21ToXml.py
```diff
@@ -861,7 +861,7 @@ def setColor(self, mxObject, m21Object):
         '''
         Sets mxObject['color'] to a normalized version of m21Object.style.color
         '''
-        self.setStyleAttributes(mxObject, m21Object, 'color')
+        self.setStyleAttributes(mxObject, m21Object, 'color', 'color')
```
mscuthbert, line 864: Not sure why this change?  If the last attribute is omitted it should be the same as the former, no?

music21/musicxml/m21ToXml.py
```diff
@@ -5484,7 +5483,7 @@ def beamToXml(self, beamObject):
 
         # not to be done: repeater (deprecated)
         self.setColor(mxBeam, beamObject)
-        self.setStyleAttributes(mxBeam, beamObject, 'fan')
+        self.setStyleAttributes(mxBeam, beamObject, 'fan', 'fan')
```
mscuthbert, line 5486: as above.

mscuthbert (commented):
Looks really good to me.  Some questions on a few points.  Only one really relevant.

music21/musicxml/m21ToXml.py
jacobtylerwalls: No, setStyleAttributes runs hyphenToCamelCase on the musicxmlNames if m21Names is None, so we run it 10,000+ times, i.e. once per note. It's not an awful cost, since we stopped compiling a regex every time in an earlier PR, but just seemed like an easy win for heavily used parsers.

music21/midi/translate.py
jacobtylerwalls: No reason to dwell on this, so I'll just restore it. But yes, it does the same thing as far as I can tell. My guess it was just caution about not choking on IndexError in `for j in range(i+1...`, maybe that was different in Python2?

music21/environment.py
jacobtylerwalls: Yes, but it programmatically converts to string and then when retrieving programmatically converts back to pathlib. So there's no point to this change, I'll revert it. Thanks.

music21/musicxml/m21ToXml.py
mscuthbert: AH! Makes perfect sense.

mscuthbert: Maybe add in a notice why you're doing that or I'm libel to forget and remove it later!

PR #995, opened 2021-05-02
Generalize exceptions in duration module and remove try block in `xmlToDuration()`

### LINKED ISSUE #967: Investigate and remove `try` clause in `xmlToDuration()`
I touched this try/except in `xmlToDuration()` merely to unindent it, but I think it can be removed. Test suite doesn't exercise it. And I can't see how there could be a failure to update components on `d` when they have just been created on `durRaw`, since at the present, the setter for `components` clears the duration tuples first. All I can come up with is that maybe a bug is lurking here if tuplets on `d` aren't cleared by the components setter and then they interact with the new components to turn out non-power-of-2 uncomputable things, but that's a bug with regard to not clearing tuplets, not something we should try/except. Worth another PR and some **test cases** at some point. Probably a vestige of an earlier time with some infelicities with the components setter?

_Originally posted by @jacobtylerwalls in https://github.com/cuthbertLab/music21/pull/941#r616228817_

jacobtylerwalls:
- Fixes #967: Remove unneeded `try` block in `xmlToDuration()`
- Investigating this revealed there was a bug with tuplets not being cleared in that branch. Fixed and added a regression test. Luckily, since #941, we only execute that branch for grace notes, so the likelihood of a grace-note tuplet hitting this was vanishingly low. But adding some future-proofing.
- While investigating the exceptions, noticed that we should raise `ValueError` or `IndexError` instead of `DurationException` in cases of user error to distinguish from the inability to calculate durations. We would expect users to have wide catches for `DurationException`, but it would be shame for user errors to be caught. Added test coverage for these exceptions.

### DIFF
--- a/music21/duration.py
+++ b/music21/duration.py
@@ -1178,6 +1178,10 @@ class Tuplet(prebase.ProtoM21Object):
         Fraction(1, 3)
 
         Raises ValueError if `amountToScale` is negative.
+
+        >>> a.augmentOrDiminish(-1)
+        Traceback (most recent call last):
+        ValueError: amountToScale must be greater than zero
         '''
         if not amountToScale > 0:
             raise ValueError('amountToScale must be greater than zero')
@@ -1914,6 +1918,12 @@ class Duration(prebase.ProtoM21Object, SlottedObjectMixin):
         >>> gRetain.components
         (DurationTuple(type='eighth', dots=0, quarterLength=0.5),
          DurationTuple(type='eighth', dots=0, quarterLength=0.5))
+
+        Negative values raise ValueError:
+
+        >>> fDur.augmentOrDiminish(-1)
+        Traceback (most recent call last):
+        ValueError: amountToScale must be greater than zero
         '''
         if not amountToScale > 0:
             raise ValueError('amountToScale must be greater than zero')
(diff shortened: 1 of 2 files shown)


mscuthbert:
Looks really good.  But I worry about backwards compatibility for people who just try to set an exception and look for DurationException.  I found 7 places in code I use where this happened.  I wonder if we should do a migration path where we have 

```
class DurationException(Exception): pass
class DurationValueError(DurationException, ValueError): pass
class DurationIndexError(DurationException, IndexError): pass
class DurationTypeError(DurationException, TypeError): pass
```

and then at least there's a period where catching DurationException still works to catch all these, but people can also catch ValueError, etc.

jacobtylerwalls:
I guess I'm wondering how much value that adds. If you don't read the release note then we're relying on you discovering this while refactoring sometime between v7 and v8 and seeing unfamiliar things appear in a shell.

Or we could emit a UserWarning. I could be convinced of a UserWarning for one release cycle for some of the dicier ones where code might have untrusted user input or lacks a plausibility check, like sending -1 to augmentOrDiminish, but some of these exceptions definitely don't need one, like this one:

https://github.com/cuthbertLab/music21/blob/a6e379cb5bf264e2dced4dfa73ef9e54586224ee/music21/duration.py#L3053-L3055

Seems a little silly to say UserWarning: you passed 'None', valid values are True, False, None

I regret to be breaking your code! Also this is far from a priority, so no rush to respond.

jacobtylerwalls:
Forgive me if obvious -- I guess the motivation for this was from fixing #967 -- I suspect the issue is moot after a recent refactor, but I need to be sure, so I go read about all the possible failure points for rasising DurationException to find out if I no longer need to be catching it. Among actual failures to compute things I also ended up reading about user error that most projects shouldn't catch at the source (perhaps catching only by a top-level unhandled error log).

jacobtylerwalls:
I think to move this forward I should test whether any use of `Duration()` or setting `.components` will raise only `DurationException` rather than anything else. I think there should be a clean separation of scenario-unsupported-by-library versus bonehead-user-error.

As in, users should be able to catch library calculation failures without catching this user error:

```python
>>> d = duration.Duration()
>>> d.linked = None
  File "/Users/jwalls/music21/music21/duration.py", line 1773, in _setLinked
    raise DurationException(f'Linked can only be True or False, not {value}')
music21.duration.DurationException: Linked can only be True or False, not None
```

And not all library calculation failures are not currently caught and re-raised, as they should be. This one should raise `DurationException`:
```
>>> d = duration.Duration(float('nan'))
  File "/Users/jwalls/music21/music21/duration.py", line 1612, in __init__
    self.quarterLength = keywords['quarterLength']
  File "/Users/jwalls/music21/music21/duration.py", line 2798, in _setQuarterLength
    value = opFrac(value)
  File "/Users/jwalls/music21/music21/common/numberTools.py", line 279, in opFrac
    ir = num.as_integer_ratio()
ValueError: cannot convert NaN to integer ratio
```

mscuthbert:
v8 is a good time to get these wonderful changes in.

PR #996, opened 2021-05-02
Add `makeNotation=False` option to musicxml export and emit warnings on malformed streams

### LINKED ISSUE #870: Warnings and documentation about expected stream nesting
Starting a thread to stash some ideas for documenting better the expected (tested) nested stream hierachy of (Opus) - Score - Part - Measure - (Voice)

Here's a warning we could emit on show or write:

```diff --git a/music21/base.py b/music21/base.py
index c43979d30..12c9855d1 100644
--- a/music21/base.py
+++ b/music21/base.py
@@ -2575,6 +2575,9 @@ class Music21Object(prebase.ProtoM21Object):
 
         scClass = common.findSubConverterForFormat(regularizedConverterFormat)
         formatWriter = scClass()
+        if self.isStream and not self.isWellFormedNotation():
+            environLocal.warn('''Your operations on this stream have left it malformed. 
+Try running makeNotation(), or see docs for stream.isWellFormedNotation()''')
         return formatWriter.write(self,
                                   regularizedConverterFormat,
                                   fp=fp,
@@ -2642,6 +2645,9 @@ class Music21Object(prebase.ProtoM21Object):
 
         scClass = common.findSubConverterForFormat(regularizedConverterFormat)
         formatWriter = scClass()
+        if self.isStream and not self.isWellFormedNotation():
+            environLocal.warn('''Your operations on this stream have left it malformed. 
+Try running makeNotation(), or see docs for stream.isWellFormedNotation()''')
         return formatWriter.show(self,
                                  regularizedConverterFormat,
                                  app=app,
```

Somewhere in the user's guide wouldn't be a bad idea, either. Maybe a couple more examples in the doctests for `stream.isWellFormedNotation`.

This comes up constantly on the list, as you know.

jacobtylerwalls:
**Previously**
Showing or writing a stream always packed it in a score (if needed) and then made a deep copy and ran `makeNotation()` and `makeRests()` on that copy. (This was partially avoidable only by calling `ScoreExporter` directly, such as the testing tool `musicxml.m21ToXml.Test.getET()` does.)

**Now**

- `show()` and `write()` get a `makeNotation` argument that when set False, avoids making a deepcopy and making notation and making rests. Some deepcopies are still made during the packing of non-Scores into Scores, but makeNotation and makeRests calls (or their substitutes `fixupNotationMeasured`, etc.) are still avoided on the top level Score once packing is done.
- Fixes #870 : Emit a warning when streams fail `isWellFormedNotation()` on show or write

Refs #763

**Future**

We've discussed moving the static methods packing m21 objects into a Score to a more general location, such as`base.Music21Object.pack()` with overridden methods on Stream, GeneralNote, etc. I've started a bit of that, but I'm inclined to make it a separate PR.


gregchapman-dev (approved):

gregchapman-dev:
Can we get PR 996 merged into master soon, please?  I’ve been using it for over a month with great success.

I’d like to move to latest, to pick up other fixes (and be a good tester for other changes).

jacobtylerwalls:
Thanks to Michael for so many reviews last night. Thanks, Greg, for testing this one out, too.

Just to follow up on a conversation you and I had offline, Greg, since that time I've discovered a handy way to install from a local (i.e. git-tracked) directory into your pip-managed environment with `pip3 install -e` so that you can also be working from another project directory:

So let's say you needed to rebase this most excellent and salutary PR on top of latest like this:
```
cd music21
git checkout master
git pull upstream master
gh pr checkout 996
git rebase -i master
pip3 install -e music21
cd ../whatever-other-package-you-need-to-work-from
```

And it shows that pip will use my local install:
```
$ pip3 list
music21                            7.0.6a1             /Users/jwalls/music21
```

I just make sure I never EVER merge anything into my own local master branch so that I can always chuck everything local and pull cleanly from upstream.

music21/musicxml/m21ToXml.py
```diff
@@ -6437,7 +6476,10 @@ def testFromScoreNoParts(self):
         s.append(note.Note())
         s.append(note.Note())
         gex = GeneralObjectExporter(s)
-        tree = ET.fromstring(gex.parse().decode('utf-8'))
+
+        warn_msg = f'{s} is not well-formed; see isWellFormedNotation()'
+        with self.assertWarns(MusicXMLWarning, msg=warn_msg):
+            tree = ET.fromstring(gex.parse().decode('utf-8'))
```
jacobtylerwalls, line 6482: Looks like I made the same mistake here I cleaned up in ca6f533, so I'll push another commit.

music21/musicxml/m21ToXml.py
```diff
@@ -1579,14 +1600,15 @@ def parsePartlikeScore(self):
...
         '''
-        # self.parts is a stream of parts
-        # hide any rests created at this late stage, because we are
-        # merely trying to fill up MusicXML display, not impose things on users
-        self.parts.makeRests(refStreamOrTimeRange=self.refStreamOrTimeRange,
-                             inPlace=True,
-                             hideRests=True,
-                             timeRangeFromBarDuration=True,
-                             )
+        if makeNotation:
```
mscuthbert, line 1603: if self.makeNotation

music21/musicxml/m21ToXml.py
```diff
@@ -1602,7 +1624,7 @@ def parsePartlikeScore(self):
             pp.parse()
             self.partExporterList.append(pp)
 
-    def parseFlatScore(self):
+    def parseFlatScore(self, makeNotation: bool = True):
```
mscuthbert, line 1627: remove

music21/musicxml/m21ToXml.py
```diff
@@ -1625,7 +1647,7 @@ def parseFlatScore(self):
         '''
         s = self.stream
         pp = PartExporter(s, parent=self)
-        pp.parse()
+        pp.parse(makeNotation=makeNotation)
```
mscuthbert, line 1650: remove

music21/musicxml/m21ToXml.py
```diff
@@ -2398,14 +2420,28 @@ def __init__(self, partObj: Union[stream.Part, stream.Score, None] = None, paren
 
         self.spannerBundle = partObj.spannerBundle
 
-    def parse(self):
+    def parse(self, makeNotation: bool = True):
```
mscuthbert, line 2423: remove -- move to object

music21/converter/subConverters.py
```diff
@@ -1019,8 +1031,13 @@ def write(self, obj, fmt, fp=None, subformats=None,
             defaults.title = ''
             defaults.author = ''
 
+        dataBytes: bytes = b''
         generalExporter = m21ToXml.GeneralObjectExporter(obj)
-        dataBytes: bytes = generalExporter.parse()
+        if makeNotation is False and 'Score' in obj.classes:
```
mscuthbert, line 1036: Unclear to me why we should be searching for Score in classes -- if someone is using `makeNotation=False` they should know what they're doing and give a well-formed object.  This allows reverting the .parse(makeNotation=False) notes

music21/musicxml/m21ToXml.py
```diff
@@ -385,16 +392,28 @@ def parse(self, obj=None):
...
+
+        `makeNotation=True` (default) runs :meth:`~music21.stream.makeNotation`
+        on each of the parts, this time without making a copy, possibly leaving
+        side effects on `sc`.
+
+        Set `makeNotation=False` to avoid making notation.
+
+        Returns bytes.
         '''
+        if not sc.isWellFormedNotation():
```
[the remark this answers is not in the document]
mscuthbert: but I think it would still be better in `fromScore()` -- I don't like to make advanced-feature usage too hand-holding, otherwise people will use it when they shouldn't and then ask more Qs when they should be mastering the basic features first.

music21/musicxml/m21ToXml.py
```diff
@@ -385,16 +392,28 @@ def parse(self, obj=None):
...
+            warnings.warn(f'{sc} is not well-formed; see isWellFormedNotation()',
+                category=MusicXMLWarning)
+
         scoreExporter = ScoreExporter(sc)
```
mscuthbert, line 415: move makeNotation to constructor of ScoreExporter and set it here also.

mscuthbert (changes requested):
Looks good -- I like the concept.  However, I don't like adding a boolean to lots of signatures (maybe more later as we find them) for an advanced feature.  All of the Exporters are classes, so it would be better to set an attribute in the class after it's been set once.  (for sub exporters, they can even look to .parent.makeNotation instead).

Thanks!

music21/musicxml/m21ToXml.py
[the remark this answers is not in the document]
jacobtylerwalls: Ah, I had beginner users/list posters in mind here. Should have been a separate PR, maybe.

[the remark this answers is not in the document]
jacobtylerwalls: As in, my hand is dangling over the side of the boat reaching for the hands of `makeNotation=True` users here.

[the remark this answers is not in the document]
jacobtylerwalls: Or, I suppose, people who don't start with the User's Guide, i.e. github users in all likelihood, and who jump into the module documentation. ScoreExporter. Great! Let's export a score! Will jump them over all of the packing steps. Given that, this is probably better in ScoreExporter.

music21/converter/subConverters.py
jacobtylerwalls: So, in that case, if someone does myPart.show(makeNotation=False), should we just raise MusicXMLExportException?

mscuthbert: I'm trying to understand what the motivation for this attribute is except "I'm a super power user working with 7,000 scores and I want to speed something up"  in which case I hope they'll realize that the user's guide is there to help.

I was going to ask in fact that the documentation for "makeNotation=False" says that it is a very advanced feature that assumes you know the ins and outs of music21 and musicxml fully before running it.  Just trying to save support time in the future.  (It's why I immediately close issues for "I wanted to use _coreInsert() because it should be faster, but I couldn't get it to work, can you help?")

jacobtylerwalls: Ya, happy to add a caveat. I think the motivation was that makeNotation fiddles with beams, Tuplet brackets, accidentals etc., and so you can't necessarily depend on changes you made to the Python streams making it out the door. At least that's how I ended up here with tuplets long ago.

jacobtylerwalls: That said, those hiccups may have been minimized after #1005 which included some of what was in an earlier version of this PR -- reading from actual streamStatus instead of streamStatus of a dummy stream never exposed to the user. But, we've also talked about adding things to makeNotation/export so still good to have, I think.

jacobtylerwalls:
Alright, I have to look at this exception pretzel later. I can give up and just use a more general exception like music21Exception in the assert, or I can silence the pylint errors, but something fishy is going on -- I get different results on multiprocessTest vs the single core test runner, and my workarounds are a bit smelly. Clearly the import order is sensitive. But CPython could presumably be doing a better job of letting me know that a name in my namespace will be ignored by `assertRaises`.

music21/base.py
```diff
@@ -2558,6 +2558,10 @@ def write(self, fmt=None, fp=None, **keywords):  # pragma: no cover
         be used.  For most people that is musicxml.
 
         Returns the full path to the file.
+
+        Some formats, including .musicxml, create a copy of the stream, pack it into a well-formed
+        score if necessary, and run :meth:`~music21.stream.Score.makeNotation`. To
+        avoid this when writing .musicxml, use `makeNotation=False`.
```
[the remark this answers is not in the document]
jacobtylerwalls: Yep, we'd need a matrix for show and write on one axis and formats on the other.

music21/musicxml/m21ToXml.py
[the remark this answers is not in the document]
jacobtylerwalls: > but I think it would still be better in fromScore()

Good idea -- so we run it ONLY for makeNotation=True, given that makeNotation=False users probably don't want their output cluttered with warnings if they're doing idiosyncratic things.

Got it, I was just not following this for a minute because of the first comment about speed of the test.

jacobtylerwalls:
Great, I agree that an instance attribute is better. Diff is smaller now.

mscuthbert (approved):
LGTM!

For really uncommon attributes I usually don't even put a setter in `__init__` -- just let people set the attribute on the object itself (`ex = ScoreExporter(sc);  ex.makeNotation = False`), again just to not draw too much attention to it.  Python, not being Java, does not need getters and setters for everything.

jacobtylerwalls:
> For really uncommon attributes I usually don't even put a setter in __init__ -- just let people set the attribute on the object itself (ex = ScoreExporter(sc); ex.makeNotation = False), again just to not draw too much attention to it. Python, not being Java, does not need getters and setters for everything.

total sense = made. just doing as advised though :-) https://github.com/cuthbertLab/music21/pull/996#discussion_r659059487

mscuthbert:
> total sense = made. just doing as advised though :-) 

Ah, I meant setting `self.makeNotation = True` in the constructor `__init__` function body, not putting it in the constructor arguments.  How confusing that Python and other languages uses the same term to mean two different things.

jacobtylerwalls:
Ah, it was me who made the leap to the _signature_, then, got it.

Issue #1001, opened 2021-05-05
Convert hex to name when writing LilyPond


crsegerie:
**music21 version**
5.5.0

**Problem summary**
No color appearing when using the checkConsecutivePossibilities method in colab

**Steps to reproduce**
See here in this [minimum reproducible example](https://colab.research.google.com/drive/1pojsveGXTt6Xa8oH7hA7UVg7PQkByp46?usp=sharing) : 

**Expected vs. actual behavior**
No coloring of the parallel octave 

**More information**
I work on Colab. I'm aware only Mac and Windows are directly supported by music21 staff, but I'm willing to contribute if you suggest some directions where to hunt for the bug.

jacobtylerwalls:
Fortunately Colab is not the issue; the issue is using lily.png as the output format. Lilypond doesn't support hex codes. Your contrasting example of the red note appearing correctly results from using the string "red". Try setting "#FF0000" and observe it appear black. See Lilypond docs: 
https://lilypond.org/doc/v2.20/Documentation/notation/list-of-colors

show('musicxml.png') works fine, but requires MuseScore:

![tmpf40qill7-1](https://user-images.githubusercontent.com/38668450/117220209-c4b80b80-add4-11eb-9c11-b3a7d46bdad2.png)


***

In fact, on 7.0.6, I can't even run your example and write to `lily.png`:

```
>>> music21Stream.write('lily.png')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/jwalls/music21/music21/stream/base.py", line 317, in write
    return super().write(fmt=fmt, fp=fp, **keywords)
  File "/Users/jwalls/music21/music21/base.py", line 2580, in write
    return formatWriter.write(self,
  File "/Users/jwalls/music21/music21/converter/subConverters.py", line 452, in write
    conv.loadFromMusic21Object(obj)
  File "/Users/jwalls/music21/music21/lily/translate.py", line 304, in loadFromMusic21Object
    self.loadObjectFromScore(m21ObjectIn, makeNotation=False)
  File "/Users/jwalls/music21/music21/lily/translate.py", line 377, in loadObjectFromScore
    lpScoreBlock = self.lyScoreBlockFromScore(scoreIn)
  File "/Users/jwalls/music21/music21/lily/translate.py", line 406, in lyScoreBlockFromScore
    lpGroupedMusicList = self.lyGroupedMusicListFromScoreWithParts(scoreIn)
[...]

PR #1002, opened 2021-05-05
Fix LilyPond crash where Lyric.text is None

jacobtylerwalls:
Spotted when testing #1001 but doesn't address the reporter's issue.


jacobtylerwalls:
Crash first reproducible at daf4d42c522adedcc826214660ac90df8fe1ccb3.

mscuthbert (approved):
Great catch!

PR #1003, opened 2021-05-06
Stop storing id(x) as .id

mscuthbert:
The `.id` of the element allows for having the memory location only as a crutch so that it's not required.  It should not be stored but generated dynamically.

This was a mistake in architecture created before music21 v1 release and it's finally time to fix.

TODO: fix in beam and tie also.

### DIFF
--- a/music21/base.py
+++ b/music21/base.py
@@ -28,7 +28,7 @@ available after importing `music21`.
 <class 'music21.base.Music21Object'>
 
 >>> music21.VERSION_STR
-'7.0.6'
+'7.0.7'
 
 Alternatively, after doing a complete import, these classes are available
 under the module "base":
@@ -313,9 +313,6 @@ class Music21Object(prebase.ProtoM21Object):
 
     # documentation for all attributes (not properties or methods)
     _DOC_ATTR = {
-        'id': '''A unique identification string; not to be confused with the
-            default `.id()` method. However, if not set, will return
-            the `id()` number''',
         'groups': '''An instance of a :class:`~music21.base.Group`
             object which describes
             arbitrary `Groups` that this object belongs to.''',
@@ -359,6 +356,7 @@ class Music21Object(prebase.ProtoM21Object):
         # do not call super().__init__() since it just wastes time
         # None is stored as the internal location of an obj w/o any sites
         self._activeSite = None  # type: Optional['music21.stream.Stream']
+        self._id = None
         # offset when no activeSite is available
         self._naiveOffset = 0.0  # type: float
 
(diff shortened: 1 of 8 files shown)


music21/freezeThaw.py
```diff
@@ -758,12 +758,6 @@ def teardownSerializationScaffold(self, streamObj=None):
         >>> st = freezeThaw.StreamThawer()
         >>> st.teardownSerializationScaffold(a)
         '''
-        def _fixId(innerEl):
-            if (innerEl.id is not None
-                    and common.isNum(innerEl.id)
-                    and innerEl.id > defaults.minIdNumberToConsiderMemoryLocation):
-                innerEl.id = id(innerEl)
-
```
jacobtylerwalls, line 766: I think this will be a change in behavior if someone does this:
```
>>> from music21 import stream
>>> v = stream.Voice()
>>> v2 = stream.Voice()
>>> v2.id = v.id
>>> v.id
4569179616
>>> v2.id
4569179616
>>> freeze/thaw
>>> thawedVoice.id
4569179616
```
Is that tolerable? Or do we want to emit a warning when setting id when it's above the defaults.minimumToConsider...

jacobtylerwalls (approved):
Makes sense to me. Just wondering whether we need to help users not shoot themselves in the foot if they have code setting elem.id = other.id without happening to set .id first.

music21/base.py
```diff
@@ -408,6 +404,21 @@ def __init__(self, *arguments, **keywords):
         if 'editorial' in keywords:
             self.editorial = keywords['editorial']
 
+    @property
+    def id(self):
+        '''
+        A unique identification string; not to be confused with the
+        default `.id()` method. However, if not set, will return
```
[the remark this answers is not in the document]
mscuthbert: yes, indeed.  basically id should be as unique as people need in order for `getElementById()` to be acceptable if it returns the wrong id.  (in that way it's similar to DOM which was what it was intended to be).

music21/beam.py
```diff
@@ -158,6 +158,8 @@ def __init__(self, type=None, direction=None, number=None):  # type is okay @Res
         # represents which beam line referred to
         # 8th, 16th, etc represented as 1, 2, ...
         self.number = number
+
+        # this should be called something else.
```
mscuthbert, line 162: I think that this id and the beam id below can be removed.  I've never used them.

music21/tie.py
```diff
@@ -109,6 +109,7 @@ def __init__(self, type='start'):  # @ReservedAssignment
                 f'Type must be one of {self.VALID_TIE_TYPES}, not {type}')
         # naming this 'type' was a mistake, because cannot create a property of this name.
 
+        # this is not the correct way we want to do this, I don't think...
         self.id = id(self)
```
mscuthbert, line 113: no, I think not.  Can probably go.

PR #1012, opened 2021-05-11
Refactor catching of OSErrors when finding default temp directories, and improve the naming of fallback temp directories for Linux and Mac platforms

tanchihpin0517:
Hi,
I add two checks for read & write permission of the root temp directory.
I also modify the naming rule of alternative temp directory for ~Linux~ Linux and Mac platform with "music21-$UID" (which is "music21-$(8-digit hashed code)" originally). It can avoid music21 creating too many directories while using the alternative temp directory.

### DIFF
--- a/music21/environment.py
+++ b/music21/environment.py
@@ -432,6 +432,18 @@ class _EnvironmentCore:
             ]:
                 self.__setitem__(name, value)  # use for key checking
 
+    def _checkAccessibility(self, path: Union[str, pathlib.Path]) -> bool:
+        '''
+        Return True if the path exists, is readable and writable.
+        '''
+        if isinstance(path, (pathlib.Path, str)):
+            exists = os.path.exists(path)
+            readable = os.access(path, os.R_OK)
+            writable = os.access(path, os.W_OK)
+            return (exists and readable and writable)
+        else:
+            return False
+
     def toSettingsXML(self, ref=None):
         '''
         Convert a ref dictionary to an xml.etree.ElementTree.Element object
@@ -506,26 +518,46 @@ class _EnvironmentCore:
         True
 
         If failed to create the subdirectory (OSError is raised), this function
-        will return a temporary directory which is created by
+        will return (1), for Linux and Mac platforms, a subdirectory under the
+        system temp directory which is named with 'music21-userid-$UID' or (2),
+        for other platforms, a temporary directory which is created by
         tempfile.mkdtemp(prefix="music21-"), which is named with 'music21-'
-        plus 8 hashed codes. The location of this directory depends on OS.
+        plus 8 hashed codes, and (3) if failing for both above, it will return
+        the directory from tempfile.gettempdir(). The location of this directory
+        depends on OS.
 
-        Note the temporary dirctory will be different in each python session.
+        Note the temporary directory may be different in each python session.
         '''
-        if self.defaultRootTempDir is not None and self.defaultRootTempDir.exists():
+        if self._checkAccessibility(self.defaultRootTempDir):
             return self.defaultRootTempDir
 
         # this returns the root temp dir; this does not create a new dir
         self.defaultRootTempDir = pathlib.Path(tempfile.gettempdir()) / 'music21'
-        # if this path already exists, we have nothing more to do
-        if self.defaultRootTempDir.exists():
+
+        # if this path already exists, readable and writable, we have nothing more to do
+        if self._checkAccessibility(self.defaultRootTempDir):
             return self.defaultRootTempDir
         else:
             # make this directory as a temp directory
             try:
                 self.defaultRootTempDir.mkdir()
-            except OSError:
-                self.defaultRootTempDir = pathlib.Path(tempfile.mkdtemp(prefix="music21-"))
+            except OSError:  # directory already exists or permission denied
+                if common.getPlatform() in ['nix', 'darwin']:
+                    uid = os.getuid()
+                    dir_path = pathlib.Path(tempfile.gettempdir()) / f'music21-userid-{uid}'
+                else:
+                    dir_path = pathlib.Path(tempfile.mkdtemp(prefix="music21-"))
+
+                if self._checkAccessibility(dir_path):
+                    self.defaultRootTempDir = dir_path
+                else:
+                    try:
+                        dir_path.mkdir()
+                        self.defaultRootTempDir = dir_path
+                    except OSError:  # pragma: no cover
+                        # Give up and use /tmp
+                        self.defaultRootTempDir = pathlib.Path(tempfile.gettempdir())
+
             return self.defaultRootTempDir
 
     def getKeysToPaths(self):
@@ -638,23 +670,8 @@ class _EnvironmentCore:
         if suffix and not suffix.startswith('.'):
             suffix = '.' + suffix
 
-        try:
-            with tempfile.NamedTemporaryFile(dir=rootDir, suffix=suffix, delete=False) as ntf:
-                ntf_name = ntf.name
-        except PermissionError:
-            # On Linux, only the user who created /tmp/music21 has write access by default
-            # So create a new user-specific directory
-            import getpass
-            newDir = rootDir.parent / f'music21-{getpass.getuser()}'
-            if not newDir.exists():
-                try:
-                    newDir.mkdir()
-                except OSError:  # pragma: no cover
-                    # Give up and use /tmp
-                    newDir = rootDir.parent
-
-            with tempfile.NamedTemporaryFile(dir=newDir, suffix=suffix, delete=False) as ntf:
-                ntf_name = ntf.name
+        with tempfile.NamedTemporaryFile(dir=rootDir, suffix=suffix, delete=False) as ntf:
+            ntf_name = ntf.name
 
         if returnPathlib:
             return pathlib.Path(ntf_name)
@@ -1566,31 +1583,34 @@ class Test(unittest.TestCase):
         self.assertEqual(list(env['localCorpusSettings']), ['/a', '/b'])
 
     @unittest.skipUnless(
-        os.access(Environment().getDefaultRootTempDir(), stat.S_IRWXU),
-        'test will programmatically set read/write/exec permissions on this dir'
+        common.getPlatform() in ['nix', 'darwin'],
+        'os.getuid can be called only on Unix platforms'
     )
-    @unittest.skipIf(
-        common.getPlatform() == 'win',
-        'os.chmod does not have the intended effect on Windows'
-    )
-    def testGetTempFile(self):
-        import getpass
+    def testGetDefaultRootTempDir(self):
         import stat
+        import shutil
 
         e = Environment()
         oldScratchDir = e['directoryScratch']
         try:
             e['directoryScratch'] = None
+            oldTempDir = e.getDefaultRootTempDir()
+            oldPermission = oldTempDir.stat()[stat.ST_MODE]
             # Wipe out write, exec permissions on the default root dir
-            os.chmod(e.getDefaultRootTempDir(), stat.S_IREAD)
-            # Was the PermissionError caught and a new "music21-{user}" dir created?
-            tmp = e.getTempFile(returnPathlib=False)
-            self.assertIn('music21-' + getpass.getuser(), tmp)
+            os.chmod(oldTempDir, stat.S_IREAD)
+            newTempDir = e.getDefaultRootTempDir()
+            self.assertIn(f'music21-userid-{os.getuid()}', str(newTempDir))
         finally:
-            # Restore owner read/write/exec permissions and original path
-            os.chmod(e.getDefaultRootTempDir(), stat.S_IRWXU)
+            # Restore original permissions and original path
+            os.chmod(oldTempDir, oldPermission)
             e['directoryScratch'] = oldScratchDir
-        os.remove(tmp)
+
+            # If getting OSError while trying to create the directory on the first fallback,
+            # the default temp directory from tempfile.gettempdir() will be return on the second
+            # fallback. We don't want to delete the default temp directory. Therefore we check
+            # it before deleting.
+            if not newTempDir.samefile(tempfile.gettempdir()):
+                shutil.rmtree(newTempDir)
 
 
 # -----------------------------------------------------------------------------
--- a/music21/environment.py
+++ b/music21/environment.py
@@ -30,6 +30,7 @@ import tempfile
 import unittest
 
 from typing import Union
+from typing import Optional
 
 import xml.etree.ElementTree as ET
 from xml.sax import saxutils
@@ -432,7 +433,7 @@ class _EnvironmentCore:
             ]:
                 self.__setitem__(name, value)  # use for key checking
 
-    def _checkAccessibility(self, path: Union[str, pathlib.Path]) -> bool:
+    def _checkAccessibility(self, path: Optional[Union[str, pathlib.Path]]) -> bool:
         '''
         Return True if the path exists, is readable and writable.
         '''
--- a/music21/environment.py
+++ b/music21/environment.py
@@ -1589,10 +1589,12 @@ class Test(unittest.TestCase):
     )
     def testGetDefaultRootTempDir(self):
         import stat
-        import shutil
 
         e = Environment()
         oldScratchDir = e['directoryScratch']
+        oldTempDir = None
+        oldPermission = None
+        newTempDir = None
         try:
             e['directoryScratch'] = None
             oldTempDir = e.getDefaultRootTempDir()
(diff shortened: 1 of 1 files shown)


music21/environment.py
```diff
@@ -501,12 +501,20 @@ def getDefaultRootTempDir(self):
...
 
         >>> import os
+        >>> import sys
         >>> e = environment.Environment()
-        >>> e.getDefaultRootTempDir() == pathlib.Path(t) / 'music21'
+        >>> src_dir = e.getDefaultRootTempDir()
+        >>> tgt_dir= pathlib.Path(t) / 'music21'
+        >>> accessible = os.access(tgt_dir, os.R_OK) and os.access(tgt_dir, os.W_OK)
+        >>> if sys.platform.startswith('linux') and not accessible:\
+                tgt_dir = pathlib.Path(t) / f'music21-{os.getuid()}';
```
jacobtylerwalls, line 510: This is too complex for a doctest, which is supposed to show users how to use the API. But I understand you wanted to doctor it so that it would pass on CI running Linux. Unfortunately we need to fake the doctest a little bit with #_DOCS_HIDE or #_DOCS_SHOW to hide and show lines to run or not run (see examples elsewhere in this file), and then move this test to a unit test. You can add a decorator for skipUnless so that your test only runs on Linux platforms.

music21/environment.py
```diff
@@ -517,15 +525,25 @@ def getDefaultRootTempDir(self):
 
         # this returns the root temp dir; this does not create a new dir
         self.defaultRootTempDir = pathlib.Path(tempfile.gettempdir()) / 'music21'
-        # if this path already exists, we have nothing more to do
-        if self.defaultRootTempDir.exists():
+
+        # if this path already exists, readable and writable, we have nothing more to do
+        readable = os.access(self.defaultRootTempDir, os.R_OK)
+        writable = os.access(self.defaultRootTempDir, os.W_OK)
+        if self.defaultRootTempDir.exists() and readable and writable:
```
jacobtylerwalls, line 532: I like this, thanks. 👍🏻

music21/environment.py
```diff
@@ -517,15 +525,25 @@ def getDefaultRootTempDir(self):
...
             return self.defaultRootTempDir
         else:
             # make this directory as a temp directory
             try:
                 self.defaultRootTempDir.mkdir()
             except OSError:
-                self.defaultRootTempDir = pathlib.Path(tempfile.mkdtemp(prefix="music21-"))
+                if sys.platform.startswith('linux'):
+                    uid = os.getuid()
+                    dir_path = pathlib.Path(tempfile.gettempdir()) / f'music21-{uid}'
```
jacobtylerwalls, line 541: Ah, is this where we were creating too many directories? Each time User X lacks write access to the root temp dir, we catch `OSError` and then attempt to create more directories with new hashes each time? Makes sense to change this. But is there a reason you didn't want to use `prefix=` like below? I just feel like we should have both branches behave the same way, either construct the tempdir inline, or construct a path and then mkdir() after.

jacobtylerwalls (changes requested):
Thanks, this looks like a nice improvement. Left some comments to move this forward. Let me know if you have any questions (or if I've misunderstood something).

music21/environment.py
```diff
@@ -517,15 +525,25 @@ def getDefaultRootTempDir(self):
...
+        writable = os.access(self.defaultRootTempDir, os.W_OK)
+        if self.defaultRootTempDir.exists() and readable and writable:
             return self.defaultRootTempDir
         else:
             # make this directory as a temp directory
             try:
                 self.defaultRootTempDir.mkdir()
             except OSError:
-                self.defaultRootTempDir = pathlib.Path(tempfile.mkdtemp(prefix="music21-"))
+                if sys.platform.startswith('linux'):
```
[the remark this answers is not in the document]
tanchihpin0517: Yes. I didn't notice there are the common tools in this project. I'll fix it.

music21/environment.py
tanchihpin0517: Got it. But I'm not familiar with doc tests and unit tests. It may take some time :P

music21/environment.py
tanchihpin0517: > Ah, is this where we were creating too many directories? Each time User X lacks write access to the root temp dir, we catch OSError and then attempt to create more directories with new hashes each time?

Yes. It creates a new directory for each different python session.

> But is there a reason you didn't want to use prefix= like below? I just feel like we should have both branches behave the same way, either construct the tempdir inline, or construct a path and then mkdir() after.

I have tried to make code here more elegant. But tempfile doesn't provide suitable API to make both branches behave in the same way. Maybe I can change it to
```
...
dir_path = pathlib.Path(tempfile.gettempdir()) / f'music21-{uid}'
dir_path.mkdir()
...
```
and remove the if block
```
...
if not dir_path.exists():
    dir_path.mkdir()
...
```

music21/environment.py
jacobtylerwalls: No worries! The good news is that the doctest passes with your feature without having to alter it, so I think we should just leave this doctest as is. I verified [here--not-real PR for testing only](https://github.com/jacobtylerwalls/music21/pull/29).

Then finding a way to test this works in a unit test -- probably better to mock the results of `os.access()` if possible than change read/write permissions of `tmp/`. Maybe you can look into [unittest mocking](https://docs.python.org/3/library/unittest.mock.html)?

music21/environment.py
jacobtylerwalls: Sure. I was just wondering why we weren't using `tempfile.mkdtemp()`. I know `NamedTemporaryDirectory` is preferred over mkdtemp, but I wasn't sure why `gettempdir()` and then `mkdir()` is superior to just mkdtemp. Honestly just wondering if you had a reason or if I'm missing something.

music21/environment.py
jacobtylerwalls: I took the other approach in 10f097274e of trying to fiddle with permissions in a unittest when you and I were working in this area last, but because I was trying to be cautious about fiddling with people's systems, the test doesn't run on CI, and so we didn't notice that it needs updating with your changes:

https://github.com/cuthbertLab/music21/blob/10f097274e82199132f50073d6cd89ba2ca85411/music21/environment.py#L1647

So I guess I'm open to both possibilities, you can try to fiddle with permissions, but honestly it's probably better to just mock. But we should update the test I just pointed to.

jacobtylerwalls: As in, the test passes on my Mac because it's not Linux, but if this test ever runs on a Linux system where there is read/write access to tmp/ (apparently not on GitHub Actions) it will fail because `getpass.getuser()` != `os.getuid()`.

tanchihpin0517: I check [10f0972](https://github.com/cuthbertLab/music21/commit/10f097274e82199132f50073d6cd89ba2ca85411). I agree that it needs to be update. But I found these two part is used to deal with the same problem. It looks a little bit redundant. Should we integrate them into a single block? I mean remove the exception handling in `getTempFile()`, and only catch it in `getDefaultRootTempDir()`. Cause `getTempFile()` use `getDefaultRootTempDir()` to get the temp directory, we just need to make sure that `getDefaultRootTempDir()` provides a accessible and permission-allowed directory. Then we only need to maintain a single code if a new problem occurs in the future,

music21/environment.py
tanchihpin0517: The reason is `tempfile.mkdtemp()` create a directory with random name (even called with `prefix`). It can't be chased in different python session. If program can't get the last-created temp directory, then it have to create a new one. There are two solution as I know. The first one is storing the name of temp directory somewhere in the disk, and the second one is using a fixed name. I prefer the second one so I use `gettempdir()+suffix_name` and then `mkdir()`.

jacobtylerwalls: Thanks, that helps me understand! All good.

mscuthbert:
Our temp dir isn't supposed to be extremely temporary, which is what TempDir tries to do.  We'd like it to persist until the system decides it needs to reclaim the space.  

Agree w/ Jacob that these OS things shouldn't be in a doctest, better as a unittest that only runs on `if common.getPlatform() == 'nix'` -- and if possible, refactor out anything specific to unix distribution to a separate function/method.  As much as possible try to be explicit so people know what we are doing, so for instance instead of `'music21-{uid}'` use `music21-userid-{uid}`

music21/environment.py
jacobtylerwalls: Yes, that sounds great, let's simplify here and just have the permission checking in getDefaultRootTempDir().

tanchihpin0517:
Hi, I have finished the modification.
I follow your recommendations, using common tools to check platform and simplifying doc test by moving it to unit test. I also integrate the error checking in `getTempFile` into `getDefaultRootTempDir` and remove the unneeded unit test for `getTempFile`.
I push a new temporary commit to keep the edit history. I will integrate(with git rebase) all commits into a single one before merging.

jacobtylerwalls:
> I push a new temporary commit to keep the edit history. I will integrate(with git rebase) all commits into a single one before merging. But I'm not sure if it's correct cause I see the "Changes reuested" still marked. Should I integrate the new commit into the old one directly?

It's all good. We will squash or rebase when merging. And "requested changes" should go away when I can provide a second review in the next few days. Thanks for the updates!

music21/environment.py
```diff
@@ -1566,31 +1584,28 @@ def testEnvironmentA(self):
...
-            self.assertIn('music21-' + getpass.getuser(), tmp)
+            os.chmod(oldTempDir, stat.S_IREAD)
+            newTempDir = e.getDefaultRootTempDir()
+            self.assertIn(f'music21-userid-{os.getuid()}', str(newTempDir))
         finally:
-            # Restore owner read/write/exec permissions and original path
-            os.chmod(e.getDefaultRootTempDir(), stat.S_IRWXU)
             e['directoryScratch'] = oldScratchDir
-        os.remove(tmp)
+            # Restore owner read/write/exec permissions and original path
```
jacobtylerwalls, line 1606:
```suggestion
            # Restore original permissions and original path
```

music21/environment.py
```diff
@@ -1566,31 +1584,28 @@ def testEnvironmentA(self):
         self.assertEqual(list(env['localCorpusSettings']), ['/a', '/b'])
 
     @unittest.skipUnless(
-        os.access(Environment().getDefaultRootTempDir(), stat.S_IRWXU),
-        'test will programmatically set read/write/exec permissions on this dir'
-    )
-    @unittest.skipIf(
-        common.getPlatform() == 'win',
-        'os.chmod does not have the intended effect on Windows'
```
jacobtylerwalls, line 1574: Love that this test runs on GitHub actions now. But now it won't run on Macs. Can we add a different assertion for Macs inside this test so that we can still only skip this test if windows? Hoping it's just a matter of checking platform and then making a different assertion, not causing you to rewrite the permissions fiddling.

music21/environment.py
```diff
@@ -432,6 +432,19 @@ def _loadDefaults(self, forcePlatform=None):
             ]:
                 self.__setitem__(name, value)  # use for key checking
 
+    def _checkAccessibility(self, path):
+        '''
+        return True if the path exists, is readable and writable.
+        'path' must be a string or a subclass of pathlib.Path
+        '''
```
jacobtylerwalls, line 439: This is private, so not a big deal, but it's usually easier to annotate the type than to spell it out in English.
```suggestion
    def _checkAccessibility(self, path: Union[str, pathlib.Path]) -> bool:
        '''
        Return True if the path exists, and is readable and writable.
        '''
```

jacobtylerwalls (approved):
This is great to have both the behavior change as well as a better separation of concerns between `getTempFile()` and `getDefaultRootTempDir()`. Just some minor changes at this point. 

Windows test suite looks good: https://github.com/jacobtylerwalls/music21/pull/31 (existing failures known: musedata converter failing on Windows)

Michael also might have some comments. Thanks again.

music21/environment.py
tanchihpin0517: This test fail on Mac because the permissions fiddling behaves differently between Unix and other platforms (Mac and Windows). Should we include Mac into the routine of permissions fiddling same as Unix? Or just keep the current one?

jacobtylerwalls: I get this result if I run the test as written:
```
AssertionError: 'music21-userid-501' not found in '/var/folders/sd/zw9wtxl552j6wn4yvt1nhg5r0000gn/T/music21-2_cvthvx'
```
Can't we just assert something else if system.platform() == 'darwin'? Or is your failure different?

jacobtylerwalls: Also, `os.getuid()` is available on Mac, because when Python docs say UNIX they include Mac.

```
>>> import os
>>> os.getuid()
501
```

tanchihpin0517: >　Can't we just assert something else if system.platform() == 'darwin'?

Yeah. That's what I mean for "tinny modification". OK. I will do it.

tanchihpin0517:
Done. Ready for review.

music21/environment.py
jacobtylerwalls: I'm still confused -- I thought you were going to leave the behavior alone and just change the test to assert the getUid() version for Linux and the 8-hashed characters for Mac. But you left the assertions alone, widened the test to run on Macs also, and then changed the behavior to use getUid() for Linux and Mac. I think that's fine, it's just not what I expected you to do based on what you said. If we keep it this way -- again, fine with me -- then we should update the docs above, where you say only Linux platforms have the getUid() version and other platforms have 8 hashed characters. Thanks.

jacobtylerwalls: > and if possible, refactor out anything specific to unix distribution to a separate function/method

I agree with Michael's impulse to factor any Linux workarounds out so that they are easier to maintain (or ignore). But @tanchihpin0517 identified some two methods that weren't cleanly separated, so upon refactoring there really isn't much that's Linux specific anymore. What's different is that Windows doesn't have `getuid()`. Mac users are less likely than Linux users to hit permission errors, but we should still try to catch the permission errors and put user-friendly names on the fallback dirs we create. Just thinking out loud, not recommending any changes.

music21/environment.py
```diff
@@ -506,26 +518,46 @@ def getDefaultRootTempDir(self):
         True
 
         If failed to create the subdirectory (OSError is raised), this function
-        will return a temporary directory which is created by
+        will return (1), for Linux platforms, a subdirectory under the system
+        temp directory which is named with 'music21-userid-$UID' or (2), for
```
jacobtylerwalls, line 522: Linux and Mac

music21/environment.py
```diff
@@ -506,26 +518,46 @@ def getDefaultRootTempDir(self):
...
+        other platforms, a temporary directory which is created by
         tempfile.mkdtemp(prefix="music21-"), which is named with 'music21-'
-        plus 8 hashed codes. The location of this directory depends on OS.
+        plus 8 hashed codes, and (3) if failing for both above, it will return
+        the directory from tempfile.gettempdir(). The location of this directory
+        depends on OS.
 
-        Note the temporary dirctory will be different in each python session.
+        Note the temporary dirctory may be different in each python session.
```
jacobtylerwalls, line 529:
```suggestion
        Note the temporary directory may be different in each python session.
```

music21/environment.py
```diff
@@ -1566,31 +1583,30 @@ def testEnvironmentA(self):
...
         finally:
-            # Restore owner read/write/exec permissions and original path
-            os.chmod(e.getDefaultRootTempDir(), stat.S_IRWXU)
+            # Restore original permissions and original path
+            os.chmod(oldTempDir, oldPermission)
             e['directoryScratch'] = oldScratchDir
-        os.remove(tmp)
+
+            if not newTempDir.samefile(tempfile.gettempdir()):
+                shutil.rmtree(newTempDir)
```
jacobtylerwalls, line 1609: Great catch -- because the second fallback if we still get OSError trying to create the fallback might just be /tmp, and we don't want to try to delete that. Maybe add a comment to that effect? We want to make this easy for someone five years from now to see what's going on.

music21/environment.py
tanchihpin0517: > I'm still confused -- I thought you were going to leave the behavior alone and just change the test to assert the getUid() version for Linux and the 8-hashed characters for Mac.

It's my fault. I accepted your recommendation, doing permission fiddling on Unix (including Mac) rather than only on Linux, but didn't change the doc. I think that's why you feel confused.

tanchihpin0517:
Update here:

After discussing, we use fix-named alternative temp directory on Unix-like platforms (Linux and Mac), and use the one named with 8-digit random hashed codes on other platforms (Windows). Thanks for reminding.

tanchihpin0517:
Hi. This PR is suspended here for some time. I wrap up commits into a single one. This PR is ready for merging now.
Please remind me if there are something I miss. Thanks.

jacobtylerwalls:
Hi and thanks for the note. Nothing for you to do at the moment, we just need two reviewers to look at it. For FOSS projects with a significant volume of PRs patience is appreciated. Two weeks is actually a pretty quick turnaround for a FOSS project.

mscuthbert:
Hi @tanchihpin0517  -- just a note that force-pushing makes it so that reviewers have to start over from the beginning in reviewing the PR, so my "changes since last review" is absent.  This is part of what has slowed the review.  Looking now.

music21/environment.py
```diff
@@ -432,6 +432,18 @@ def _loadDefaults(self, forcePlatform=None):
             ]:
                 self.__setitem__(name, value)  # use for key checking
 
+    def _checkAccessibility(self, path: Union[str, pathlib.Path]) -> bool:
```
mscuthbert, line 435: path typing should include 'None' as a possibility, since rootTempDir can be None, and otherwise the isinstance check does not make sense below.

music21/environment.py
```diff
@@ -1566,31 +1583,34 @@ def testEnvironmentA(self):
...
+            # Restore original permissions and original path
+            os.chmod(oldTempDir, oldPermission)
             e['directoryScratch'] = oldScratchDir
-        os.remove(tmp)
+
+            # If getting OSError while trying to create the directory on the first fallback,
+            # the default temp directory from tempfile.gettempdir() will be return on the second
+            # fallback. We don't want to delete the default temp directory. Therefore we check
+            # it before deleting.
+            if not newTempDir.samefile(tempfile.gettempdir()):
```
mscuthbert, line 1612: because this is in `finally` code, cannot assume that `newTempDir` is defined at this point.  Define it as None outside the try and then test to see if it is not None first before running samefile.  Should also check that it is not the same file as 'directoryScratch' or oldScratchDir before running.

mscuthbert (changes requested):
The code for the examples is fine, but the test code is too dangerous to approve.  See comments below.

music21/environment.py
jacobtylerwalls: @tanchihpin0517 this was my fault, sorry, you just committed my suggestion.

```suggestion
    def _checkAccessibility(self, path: Optional[Union[str, pathlib.Path]]) -> bool:
```

mscuthbert:
Sorry -- I;m not seeing my main review comment appearing here -- it's about the shutil `rmtree` -- I think it needs to be removed and only the directories created during the creation deleted, otherwise the testcode could potentially wipe out a whole filesystem if not careful.

mscuthbert:
I can't find where I wrote an extensive comment, but basically I'm concerned that test code is using rmtree -- it is safe as it currently is, but the point of test code is that someone should be able to tinker with the underlying code in the future and run tests to see if it works.  What if someone changes the code or a later OS update changes returning getTempDir to return '/' for privacy?  Then the test code could potentially destroy an entire hard drive.  The test code in this case should individually remove whatever files it created one-by-one.

mscuthbert:
Converting to draft for safety to make sure it's not merged until that's fixed.

tanchihpin0517:
Sorry. I didn't notice recent updates here.
I make two changes:
1. make sure codes have been run in the `try` block before do something in the `final` block.
2. delete `newTempDir` only if it's empty. It can guarantee that the action of deleting is safe.

mscuthbert:
Looks GREAT! to me.  Thank you.  Merging.

PR #1018, opened 2021-05-13
Use `requests_html` to implement `ComposerPopularityFeature`

### LINKED ISSUE #907: Composer popularity feature always returns 0
**music21 version**

`7.0.1`

**Problem summary**
Composer popularity feature always returns 0.

**Steps to reproduce**
Can also reproduce using the doctest example with Mozart.
```
>>> x = converter.parse('tinynotation: d1')
>>> y = stream.Score([metadata.Metadata(), x])
>>> y.metadata.composer = 'Stevie Wonder'
>>> fe = features.native.ComposerPopularity(y)
>>> fe.extract().vector
[0]
```

Should probably fail more loudly if it's not working.

jacobtylerwalls:
Fixes #907 
This had stopped working because Google's response to the static UserAgent we were giving lacked result counts.

I was looking into this package for another reason today and realized it would solve this problem. It rotates user agents. It's owned by PSF. If you don't want to add it to requirements.txt then we can fiddle with the CI script to still have the test run on CI, or we can just deprecate the feature.

### DIFF
--- a/music21/features/native.py
+++ b/music21/features/native.py
@@ -16,7 +16,6 @@ import re
 import math
 from typing import Optional
 
-from urllib.request import Request, urlopen
 from urllib.parse import urlencode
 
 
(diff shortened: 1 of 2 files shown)


jacobtylerwalls:
#846 imports `requests`, so we should decide in tandem whether to use both packages. I tried this patch with `requests` instead of `requests_html`: no dice. Get back the same response from Google as status quo. Maybe a security through obscurity thing where Google is more inclined to filter and doctor requests coming from `requests` than `requests_html`.

mscuthbert:
<s>I'm not so thrilled to put a project with only 9 GitHub stars into the requirements.txt for music21.  It means that a project with thousands of users is dependent on one with only a few dozen to implement proper security features on something extremely...</s>

Never mind -- a google search put me on a fork of the project; the main branch has 11k stars.  But its sub requirements are still huge: 

```
# What packages are required for this module to be executed?
REQUIRED = [
    'requests', 'pyquery', 'fake-useragent', 'parse', 'beautifulsoup4', 'w3lib', 'pyppeteer>=0.0.14'
]
```

BS4 is a big file in itself.  I think better just to remove the Feature.

PR #1020, opened 2021-05-13
Remove ComposerPopularity

### LINKED ISSUE #907: Composer popularity feature always returns 0

jacobtylerwalls:
Fixes #907 

Open for suggestions on how best to do this. We shouldn't reuse the id, it seems to me. But we don't want it returned in `allFeaturesAsList()`.

### DIFF
--- a/music21/features/base.py
+++ b/music21/features/base.py
@@ -1160,7 +1160,7 @@ def extractorsById(idOrList, library=('jSymbolic', 'native')):
 
     >>> y = [x.id for x in features.extractorsById('all')]
     >>> y[0:3], y[-3:-1]
-    (['M1', 'M2', 'M3'], ['MD1', 'MC1'])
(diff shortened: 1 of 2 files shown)


mscuthbert (approved):
Thanks!  This was fun while it lasted, but there are better ways of doing this.  It's good that we left it in, because there may in the future be other ways of doing this (wikidata queries, etc.) so happy to take future PRs from anyone to restore it.

PR #1021, opened 2021-05-14
New colors for documentation background and code highlight

jacobtylerwalls:
Proposing a new color for the documentation background and code highlights. I love green, but this particular green has been there for a while.

Of course, I'm open to suggestions.

Screen Shot 2021-05-14 at 9 38 13 AM
Screen Shot 2021-05-14 at 9 38 35 AM


mscuthbert:
yeah these colors are better.  the whole docs system looks old, but never risen to that level of priority.

Issue #47, opened 2014-03-11
xmlHandler.py unit test fail : testOpen 


metalmike:
## ERROR: testOpen (music21.musicxml.xmlHandler.TestExternal)

Traceback (most recent call last):
  File "C:\Users\Michael\git\music21\music21\musicxml\xmlHandler.py", line 1124, in testOpen
    c.repr()
  File "C:\Users\Michael\git\music21\music21\musicxml\xmlHandler.py", line 1082, in repr
    self.reprTest()
  File "C:\Users\Michael\git\music21\music21\musicxml\xmlHandler.py", line 1086, in reprTest
    print('+'*20 + ' ' + self.getBestTitle())
TypeError: cannot concatenate 'str' and 'NoneType' objects

Issue #43, opened 2014-03-10
converter.py unit test fail : testParseURL


metalmike:
```
def testParseURL(self):
    urlA = 'http://kern.ccarh.org/cgi-bin/ksdata?l=users/craig/classical/schubert/piano/d0576&file=d0576-06.krn&f=xml'
    urlB = 'http://kern.ccarh.org/cgi-bin/ksdata?l=users/craig/classical/schubert/piano/d0576&file=d0576-06.krn&f=kern'
    urlC = 'http://kern.ccarh.org/cgi-bin/ksdata?l=users/craig/classical/bach/cello&file=bwv1007-01.krn&f=xml'
    for url in [urlA, urlB, urlC]:
        unused_post = parseURL(url)
```

shows the following error:
## ERROR: testParseURL (music21.converter.TestExternal)

Traceback (most recent call last):
  File "C:\Users\Michael\git\music21\music21\converter.py", line 1241, in testParseURL
    unused_post = parseURL(url)
  File "C:\Users\Michael\git\music21\music21\converter.py", line 1019, in parseURL
    v.parseURL(url, format=format)
  File "C:\Users\Michael\git\music21\music21\converter.py", line 937, in parseURL
    self._converter.parseFile(fp, number=number)
  File "C:\Users\Michael\git\music21\music21\converter.py", line 499, in parseFile
    raise ConverterException('score from file path (%s) no parts defined' % fp)
ConverterException: score from file path (c:\users\michael\appdata\local\temp\music21\m21-79fede18c19af95328cb65b5421c20d3.xml) no parts defined

This was the problematic link (urlB and urlC work well):
 urlA = 'http://kern.ccarh.org/cgi-bin/ksdata?l=users/craig/classical/schubert/piano/d0576&file=d0576-06.krn&f=xml'

PR #1023, opened 2021-05-18
Run most TestExternal classes on CI

### LINKED ISSUE #938: Explore running TestExternal classes on CI
The Coveralls stats are kind of meaningless because we're insisting that the `Test` classes not open browser windows, only `TestExternal`, which we don't run on CI--so everything except the helper methods is a missed line. The code will only be exercised by running `python3 -m music21.osmd` to run TestExternal.

Honestly I think ciMain() should also run TestExternal, if it doesn't double the time of the suite, in some way that wouldn't affect multiprocessTest and the other tools developers use where we don't to window-bomb them. But can be discussed separately.

_Originally posted by @jacobtylerwalls in https://github.com/cuthbertLab/music21/issues/326#issuecomment-809488356_

jacobtylerwalls:
**Motivation**

Running these on CI should find more bugs going forward. Examples (#953 , #443 - #446, #43 - #47)

**Proposal**

Use a class attribute on the TestExternal classes for whether or not to show files (png, pdf, etc), and use a keyword arg in the test runner so that `ciMain()` can set False without impacting manual tests. Then we don't have to install MuseScore or PNG readers on CI. All of the plot tests involve showing, so just skip them on CI by renaming to `TestExternalManual`.

**Fixes and cleanups found**

- a doctest wasn't setting the musicxml path back (test isolation issue)
- a failure we haven't addressed (will create issue, meanwhile added an @expectedFailure decorator)
- `Graph.__del__()` was hogging memory if doneAction is None
- some slow tests moved to TestSlow or otherwise made faster
- Prevent unraisable `ImportError` when shutting down Python after running a graph
- Improve TimeSignature construction performance

I know it was mentioned to move these over one at a time, but I wanted to demo this approach--rather than have to revisit this one at a time, which sounds possibly too tedious to prioritize, we can get most of it wired up this way. It adds about 40s to the suite (and adds 0.5% coverage).  Let me know what you think.

Fixes #938

### DIFF
--- a/music21/environment.py
+++ b/music21/environment.py
@@ -1111,6 +1111,7 @@ class Environment:
         Returns an xmlReaderType depending on the 'musicxmlPath'
 
         >>> a = environment.Environment()
+        >>> original = a['musicxmlPath']  #_DOCS_HIDE
         >>> a['musicxmlPath'] = '/Applications/Musescore.app'
         >>> a.xmlReaderType()
         'Musescore'
@@ -1130,6 +1131,8 @@ class Environment:
         >>> a['musicxmlPath'] = None
         >>> a.xmlReaderType() is None
         True
+
+        >>> a['musicxmlPath'] = original  #_DOCS_HIDE
         '''
         xp = self['musicxmlPath']
         if common.runningUnderIPython():
--- a/music21/graph/primitives.py
+++ b/music21/graph/primitives.py
@@ -162,6 +162,17 @@ class Graph(prebase.ProtoM21Object):
         if hasattr(self, 'figure') and self.figure is not None:
             self.plt.close(self.figure)
 
+    def __getstate__(self):
+        '''
+        The wrapper to matplotlib.pyplot stored as self.plt cannot be pickled/deepcopied.
+        '''
+        state = self.__dict__.copy()
+        del state['plt']
+        return state
+
+    def __setstate__(self, state):
+        self.__dict__.update(state)
+
     @property
     def doneAction(self):
         '''
--- a/music21/graph/primitives.py
+++ b/music21/graph/primitives.py
@@ -1590,7 +1590,7 @@ class Test(unittest.TestCase):
 
 
 # ------------------------------------------------------------------------------
-class TestExternal(unittest.TestCase):  # pragma: no cover
+class TestExternal(unittest.TestCase):
     show = True
 
     def testBasic(self):
--- a/music21/graph/primitives.py
+++ b/music21/graph/primitives.py
@@ -157,7 +157,7 @@ class Graph(prebase.ProtoM21Object):
         '''
         Matplotlib Figure objects need to be explicitly closed when no longer used...
         '''
-        if hasattr(self, 'figure') and self.figure is not None and self.doneAction is None:
+        if hasattr(self, 'figure') and self.figure is not None:
             etm = getExtendedModules()
             etm.plt.close(self.figure)
 
@@ -1599,14 +1599,10 @@ class TestExternal(unittest.TestCase):  # pragma: no cover
         a.data = data
         a.process()
 
-        del a
-
         a = GraphHistogram(doneAction=None, title='50 x with random(30) y counts')
         data = [(x, random.choice(range(30))) for x in range(50)]
         a.data = data
         a.process()
-
-        del a
 
         a = Graph3DBars(doneAction=None,
                                title='50 x with random values increase by 10 per x',
@@ -1643,7 +1639,7 @@ class TestExternal(unittest.TestCase):  # pragma: no cover
             z = random.choice(range(1, 20))
             data.append([x, y, z])
 
-        a = GraphScatterWeighted()
+        a = GraphScatterWeighted(doneAction=None)
         a.data = data
         a.process()
 
--- a/music21/graph/primitives.py
+++ b/music21/graph/primitives.py
@@ -104,7 +104,9 @@ class Graph(prebase.ProtoM21Object):
     )
 
     def __init__(self, *args, **keywords):
-        getExtendedModules()
+        extm = getExtendedModules()
+        self.plt = extm.plt  # wrapper to matplotlib.pyplot
+
         self.data = None
         self.figure = None  # a matplotlib.Figure object
         self.subplot = None  # an Axes, AxesSubplot or potentially list of these object
@@ -155,11 +157,10 @@ class Graph(prebase.ProtoM21Object):
 
     def __del__(self):
         '''
-        Matplotlib Figure objects need to be explicitly closed when no longer used...
+        Matplotlib Figure objects need to be explicitly closed when no longer used.
         '''
         if hasattr(self, 'figure') and self.figure is not None:
-            etm = getExtendedModules()
-            etm.plt.close(self.figure)
+            self.plt.close(self.figure)
 
     @property
     def doneAction(self):
--- a/music21/graph/primitives.py
+++ b/music21/graph/primitives.py
@@ -1639,7 +1639,11 @@ class TestExternal(unittest.TestCase):
             z = random.choice(range(1, 20))
             data.append([x, y, z])
 
-        a = GraphScatterWeighted(doneAction=None)
+        if self.show:
+            doneAction = 'write'
+        else:
+            doneAction = None
+        a = GraphScatterWeighted(doneAction=doneAction)
         a.data = data
         a.process()
 
--- a/music21/graph/primitives.py
+++ b/music21/graph/primitives.py
@@ -172,6 +172,8 @@ class Graph(prebase.ProtoM21Object):
 
     def __setstate__(self, state):
         self.__dict__.update(state)
+        extm = getExtendedModules()
+        self.plt = extm.plt
 
     @property
     def doneAction(self):
--- a/music21/graph/primitives.py
+++ b/music21/graph/primitives.py
@@ -1591,6 +1591,7 @@ class Test(unittest.TestCase):
 
 # ------------------------------------------------------------------------------
 class TestExternal(unittest.TestCase):  # pragma: no cover
+    show = True
 
     def testBasic(self):
         a = GraphScatter(doneAction=None, title='x to x*x', alpha=1)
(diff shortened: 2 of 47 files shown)


jacobtylerwalls:
Actions is down, so I'll rerun jobs tonight.

music21/meter/tests.py
```diff
@@ -47,7 +47,7 @@ def testCompound(self):
         meterStrDenominator = [1, 2, 4, 8, 16, 32]
         meterStrNumerator = [2, 3, 4, 5, 6, 7, 9, 11, 12, 13]
 
-        for i in range(30):
+        for i in range(8):
             msg = []
             for j in range(1, random.choice([2, 4])):
                 msg.append('%s/%s' % (random.choice(meterStrNumerator),
```
jacobtylerwalls, line 53: Up to 20 of the 40 seconds are taken by this test, even after I reduced the # of iterations. (sometimes takes as little as 3 seconds). TimeSignature creation is the chokepoint. Takes 4.5 seconds to create `meter.TimeSignature('9/4+6/8+13/1')`.

Could be moved to TestSlow.

jacobtylerwalls: This diff gets the creation time down to 3s from 4.2ish, by saving 261,120 opFrac calls, but not sure if you want to change the signature of `offsetToDepth()` over it.

len(meterSequence.flat) = 512
512 calls to offsetToDepth and offsetToIndex
for each call in offsetToIndex, 1 - 512 loop iterations, all but one with 2 calls to opFrac

261120 calls to opFrac saved

```diff
diff --git a/music21/meter/core.py b/music21/meter/core.py
index 6aee8858d..7382c5ed9 100644
--- a/music21/meter/core.py
+++ b/music21/meter/core.py
@@ -15,6 +15,7 @@ This module defines two component objects for defining nested metrical structure
 :class:`~music21.meter.core.MeterTerminal` and :class:`~music21.meter.core.MeterSequence`.
 '''
 import copy
+from typing import Optional
 
 from music21 import prebase
 from music21.common.numberTools import opFrac
@@ -1765,7 +1766,7 @@ class MeterSequence(MeterTerminal):
         iMatch = self.offsetToIndex(qLenPos)
         return opFrac(self[iMatch].weight)
 
-    def offsetToDepth(self, qLenPos, align='quantize'):
+    def offsetToDepth(self, qLenPos, align='quantize', index: Optional[int] = None):
         '''
         Given a qLenPos, return the maximum available depth at this position
 
@@ -1787,6 +1788,8 @@ class MeterSequence(MeterTerminal):
         >>> b.offsetToDepth(-1)
         Traceback (most recent call last):
         music21.exceptions21.MeterException: cannot access from qLenPos -1.0
+
+        `index` can be provided, if known, for some long MeterSequences to improve performance.
         '''
         qLenPos = opFrac(qLenPos)
         if qLenPos >= self.duration.quarterLength or qLenPos < 0:
@@ -1797,7 +1800,9 @@ class MeterSequence(MeterTerminal):
         # need to quantize by lowest level
         mapMin = self.getLevelSpan(self.depth - 1)
         msMin = self.getLevel(self.depth - 1)
-        qStart, unused_qEnd = mapMin[msMin.offsetToIndex(qLenPos)]
+        if index is None:
+            index = msMin.offsetToIndex(qLenPos)
+        qStart, unused_qEnd = mapMin[index]
         if align == 'quantize':
             posMatch = opFrac(qStart)
```

mscuthbert: That is fine -- this is a deep dive into the internals, and works fine if index isn't used, so that's great!

music21/environment.py
```diff
@@ -1111,6 +1111,7 @@ def xmlReaderType(self):
         Returns an xmlReaderType depending on the 'musicxmlPath'
 
         >>> a = environment.Environment()
+        >>> original = a['musicxmlPath']  #_DOCS_HIDE
```
mscuthbert, line 1114: this should be okay as is, since environment.Environment() does not write directly to disk unless `.save()` is called -- environment.UserSettings() does that.  (Why these two different versions is a long and boring old story.  Maybe they should be merged at some point.  But the notion was that people might want to temporarily change environment for different runs)

music21/graph/primitives.py
```diff
@@ -1642,7 +1639,7 @@ def testScatterWeighted(self):
             z = random.choice(range(1, 20))
             data.append([x, y, z])
 
-        a = GraphScatterWeighted()
+        a = GraphScatterWeighted(doneAction=None)
```
mscuthbert, line 1642: can this be dependent on self.show?

music21/romanText/translate.py
```diff
@@ -1133,22 +1133,18 @@ def romanTextToStreamOpus(rtHandler, inputM21=None):
...
             rth = rtf.readstr(tf)  # return handler, processes tokens
-            s = romanTextToStreamScore(rth)
-            s.show()
-
-
-class TestSlow(unittest.TestCase):  # pragma: no cover
-    '''
-    These tests are currently too slow to run every time.
-    '''
+            unused_s = romanTextToStreamScore(rth)
```
mscuthbert, line 1147: can this still be shown?  It's the only way the test can be determined to work is visual examination.

music21/test/testSingleCoreAll.py
```diff
@@ -122,7 +129,7 @@ def main(testGroup=('test',), restoreEnvironmentDefaults=False, limit=None, verb
 def ciMain():
     # the main call for ci tests.
     # exits with the returnCode
-    returnCode = main(verbosity=1)
+    returnCode = main(testGroup=('test', 'external'), verbosity=1, show=False)
```
mscuthbert, line 132: can we run this as two lines, with returnCode aggregated, so that all the `Test` are run first and then `TestExternal` after that.  In part so we can see the timing differences between the two.  I've been trying to get Test to run faster, and want to make sure we optimize the same with TestExternal

mscuthbert (commented):
This is a very good start to these tests -- I'm a little concerned though that by not showing the tests we're getting a false sense of security about them -- does our output crash lilypond is one of the important things to that TestExternal is checking.  This is good, but I think we eventually want to have TestExternal run things through lilypond or Musescore and make sure that they actually output a png or something.

Approved conditional on how much slower the tests are -- but w/ small changes requested if you agree.

music21/romanText/translate.py
jacobtylerwalls: Sure, given that I've moved these to TestSlow, but they are *very* slow!

music21/test/testSingleCoreAll.py
jacobtylerwalls: The test controller would end up loading the doctests twice. (Or, do you only want doctests run with `Test` vs `TestExternal`? If so, that's not too difficult, but just checking if you want to go that effort.)

jacobtylerwalls:
> This is a very good start to these tests -- I'm a little concerned though that by not showing the tests we're getting a false sense of security about them -- does our output crash lilypond is one of the important things to that TestExternal is checking. This is good, but I think we eventually want to have TestExternal run things through lilypond or Musescore and make sure that they actually output a png or something.

I experimented with doing this on CI, but it clogged the log output with verbose LilyPond messages I didn't know how to suppress. Same with MuseScore (well, there, the bigger issue was the 40s it takes to download and install it).

I still think it's worth taking a first step -- now we can have coverage diffs reported and also find bugs that don't have to do with crashing the viewing programs.

The sheer fact that they're called out separately and not just fully folded in to `Test` should indicate there's a reason they still need to loaded externally to be checked.

jacobtylerwalls:
Btw, if speed is a concern on GitHub Actions, we should move the Coveralls job to follow Python 3.8, which runs our suite much faster than 3.7 and slightly faster than 3.9.

music21/environment.py
jacobtylerwalls: That makes sense, but something strange is going on, then, because I only did this to get the tests to pass on Ubuntu during the period I was actually trying to run everything through MuseScore.

I just tried it again, and lo & behold -- failures if I revert this change (and try to open musescore):

compare [a run without this change](https://github.com/jacobtylerwalls/music21/pull/34/checks?check_run_id=2865340468)

to [restoring the change](https://github.com/jacobtylerwalls/music21/pull/34/checks?check_run_id=2865371614)

Except for the unrelated failures at the top of the log owing to my lack of effort to install a PNG reader, notice the failures all happen (reverse) alphabetically after the environment module, and we have code opening musescore well before that.

Do you want to debug what's going on or just merge it?

jacobtylerwalls:
> Approved conditional on how much slower the tests are

Before the change to `offsetToDepth()` this PR added about 30-40s to the Python 3.8/3.9 runs, I think. We have some nondeterministic tests in there.

After the change to `offsetToDepth()` this PR reduces (!) the time of the suite, even after adding all these tests.

[recent build on master](https://github.com/cuthbertLab/music21/runs/2592246422?check_suite_focus=true) (PR base): Ran 4673 tests in 318.421s

last run on this branch: 
Ran 4736 tests in 262.344s

I know we have nondeterministic tests in there, but this seemed to hold over several runs.

jacobtylerwalls:
Added another cleanup I found while in here; considered opening another PR for it, but there would have been merge conflicts and this branch already has unrelated cleanups on it.

Previously:
```
>>> bach= corpus.parse('bwv66.6')
>>> scatter2 = graph.plot.ScatterPitchClassOffset(bach, doneAction=None)
>>> scatter2.run()
>>> quit()
Exception ignored in: <function Graph.__del__ at 0x10c69bc10>
  File "/Users/jwalls/music21/music21/graph/primitives.py", line 161, in __del__
  File "/Users/jwalls/music21/music21/graph/utilities.py", line 46, in getExtendedModules
ImportError: sys.meta_path is None, Python is likely shutting down
```

music21/test/testSingleCoreAll.py
jacobtylerwalls: Done!

```
Ran 4684 tests in 261.671s

Ran 58 tests in 30.455s
```

But this will be somewhat variable based on the nondeterministic tests.

music21/environment.py
jacobtylerwalls: > Do you want to debug what's going on or just merge it?

Sorry I didn't see this immediately, I see now we're talking about two different things; you've correctly pointed out no change is necessary to avoid disturbing a developer's local system, but you can also see how without this change we spoil test isolation and any later unit tests run in the same session as this doc test will choke on nonexistent filepaths. So I think we need to keep this change.

PR #1027, opened 2021-05-23
Permit providing Chords to Stream constructor

### LINKED ISSUE #976: Opus objects should be sequences of appended scores.
**music21 version**

dev   

**Problem summary**

Opus objects should be documented to have Score objects follow sequentially.  Stream constructor should make the `not e.isMeasure` to also include `and not 'Score' in e.classSet`

**Steps to reproduce**

```
sc1 = corpus.parse('bwv66.6')
sc2 = stream.Score()
op = stream.Opus([sc1, sc2])
sc2.getOffsetBySite(op)
```

sc2's offset is 0.0.

sc2's offset should be the same as sc1's highestTime.

### LINKED ISSUE #1026: Providing a Chord to Stream constructor inserts each note separately
**music21 version**

`6.7.1` and master, i.e. we didn't make this worse, but we have perhaps left the impression that this is safe

https://github.com/cuthbertLab/music21/blob/e030bee3c401c45ab616949ad5c464f2d8eb8a9f/music21/stream/base.py#L163

**Problem summary**
Providing a chord to the Stream constructor without wrapping in an iterable will not insert the chord but rather insert each note.  See remarks here on "other Stream". Chords are not streams:

https://github.com/cuthbertLab/music21/blob/e030bee3c401c45ab616949ad5c464f2d8eb8a9f/music21/stream/base.py#L118-L125

**Steps to reproduce**

```
c = chord.Chord('C2 A2')
s = stream.Stream(c)
s.show('text')
```

Expected: 
{0.0} <music21.chord.Chord C2 A2>

Actual:
{0.0} <music21.note.Note C>
{0.0} <music21.note.Note A>

**Suggested fix**

We can either focus on the Stream constructor and not iterate over `givenElements` if isinstance base.Music21Object, OR we can decide if `isIterable` should return True for music21Objects:

```
>>> c = chord.Chord('A2 C2')
>>> common.isIterable(c)
True
```

jacobtylerwalls:
Fixes #1026: permit providing chords to Stream constructor

Fixes #976: append rather than insert when providing a list of Scores to stream constructor (i.e. `Opus()`)

### DIFF
--- a/music21/stream/base.py
+++ b/music21/stream/base.py
@@ -167,8 +167,8 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
     >>> s.first()
     <music21.meter.TimeSignature 4/4>
 
-    New in v7 -- providing a list of objects or Measures (but not other Stream
-    subclasses such as Parts) now positions sequentially, i.e. appends:
+    New in v7 -- providing a list of objects or Measures or Scores (but not other Stream
+    subclasses such as Parts or Voices) now positions sequentially, i.e. appends:
 
     >>> s2 = stream.Measure([note.Note(), note.Note(), bar.Barline()])
     >>> s2.show('text')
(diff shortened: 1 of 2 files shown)


mscuthbert:
Looks good to me!  Probably should also make Chord return False in common.isIterable.

jacobtylerwalls:
Maybe we should leave that alone--looks like the notes setter needs it.

https://github.com/cuthbertLab/music21/blob/49cc89bb7483ed04cff0bf4959ad3a5a594a046c/music21/chord/__init__.py#L4573-L4579

mscuthbert:
Ah, you're probably right.  we'll be adding special cases somewhere -- either in the Chord notes setter or in the Stream constructor, so probably better to do it where we've already done it.

PR #1029, opened 2021-05-23
Let `asKey()` solve for mode from tonic and sharps

jacobtylerwalls:
I built this for someone, and I'm offering it back upstream. It's what I had suggested in reviewing #781.

### DIFF
--- a/music21/stream/tests.py
+++ b/music21/stream/tests.py
@@ -2220,8 +2220,8 @@ class Test(unittest.TestCase):
         s.replace(n4, n1)
         self.assertEqual([s[0], s[1]], [n3, n1])
 
-        error_msg = f'{n3} already in {s}'
-        with self.assertRaises(StreamException, msg=error_msg):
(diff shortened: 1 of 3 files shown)


music21/stream/tests.py
```diff
@@ -2220,8 +2220,8 @@ def testReplace(self):
         s.replace(n4, n1)
         self.assertEqual([s[0], s[1]], [n3, n1])
 
-        error_msg = f'{n3} already in {s}'
-        with self.assertRaises(StreamException, msg=error_msg):
+        expected = f'{n3} already in {s}'
+        with self.assertRaisesRegex(StreamException, expected):
```
mscuthbert, line 2224: Thank you for these assert changes -- I had no idea that they weren't working!

mscuthbert:
Looks great to me!

music21/stream/tests.py
jacobtylerwalls: yeah, it was my bad...

PR #1030, opened 2021-05-24
Handle `<forward>` elements without `<voice>`

jacobtylerwalls:
After #1008 I've got MuseScore files emitting dozens of warnings:
```
WARNING: Cannot put in an element with a missing voice tag when no previous voice tag was given.  Assuming voice 1... 
```

Turns out they don't write `<voice>` tags on `<forward>` elements and just expect readers to infer the voice context. So extended a workaround already present for an issue with Sibelius.

The spec says `<voice>` tags are optional on forward tags but also that "The forward element is generally used within voices and staves."


mscuthbert (approved):
Wonderful.  Thanks for this dirty work!

PR #1031, opened 2021-05-28
New `Stream.getContextByClass()` in order to search itself at 0.0 before searching context

### LINKED ISSUE #1028: `getContextByClass()` doesn't find elements positioned in self at 0.0 by default
**Motivation**
I started the rest of the v7 deprecations. Figured my other contributions were distracting us from what we've already promised for v7.

While doing them, I noticed that if we are going to deprecate `getKeySignatures()` and `getClefs()` there will be a problem -- recursing is not really what they did; they searched context. These methods searched themselves first with gEBC(), and then, finding nothing, searched the context with gCBC().

This made replacing the calls in the test suite difficult: I had to let them fail, and then see whether the correct result could be had by calling either gEBC or gCBC. This is a sign that we are taking away some functionality.

**Feature summary**
`includeSelf=False` parameter to getContextByClass. Default False for BC. But in the deprecation notice tell users to replace `getClefs()` with `getContextByClass(clef.Clef, includeSelf=True)`.

Notice:

```
>>> s = stream.Stream()
>>> s.insert(0, clef.BassClef())
>>> s.getContextByClass(clef.Clef) is None
True
```

**Intent**

jacobtylerwalls:
Fixes #1028

Debatable whether this was supposed to be supported already, but I think this will line up with user expectations.

Needed for deprecations of `getClefs()` etc

### DIFF
--- a/music21/stream/base.py
+++ b/music21/stream/base.py
@@ -3737,7 +3737,7 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
         >>> p = converter.parse('tinyNotation: c1 d1')
         >>> m = p.measure(2)
         >>> m.insert(0, clef.BassClef())
-        >>> m.getContextByClass(clef.Clef, getElementMethod='getElementAtOrBefore')
+        >>> m.getContextByClass(clef.Clef)
         <music21.clef.BassClef>
 
         Searching before yields the clef in m. 1:
--- a/music21/stream/base.py
+++ b/music21/stream/base.py
@@ -3709,7 +3709,71 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
             sIterator = sIterator.getElementsByClass(classList)
         return sIterator
 
-    def getElementAtOrBefore(self, offset, classList=None):
+    def getContextByClass(
+        self,
+        className,
+        *,
+        getElementMethod='getElementAtOrBefore',
+        sortByCreationTime=False,
+        followDerivation=True,
+        priorityTargetOnly=False,
+    ) -> Optional[base.Music21Object]:
+        '''
+        Overrides :meth:`~music21.base.Music21Object.getContextByClass` in order
+        to look first in this stream's elements at offset 0.0 before executing the
+        rest of the context search, so long as `getElementMethod` is the default
+        `getElementAtOrBefore`.
+
+        These `getElementMethod` values are documented on the superclass
+        :meth:`~music21.base.Music21Object.getContextByClass`:
+
+        *    `'getElementBefore'`
+        *    `'getElementAfter'`
+        *    `'getElementAtOrBefore'` (Default)
+        *    `'getElementAtOrAfter'`
+
+        Ask for measure 2's clef context and get the clef at 0.0 in measure 2:
+
+        >>> p = converter.parse('tinyNotation: c1 d1')
+        >>> m = p.measure(2)
+        >>> m.insert(0, clef.BassClef())
+        >>> m.getContextByClass(clef.Clef, getElementMethod='getElementAtOrBefore')
+        <music21.clef.BassClef>
+
+        Searching before yields the clef in m. 1:
+
+        >>> m.getContextByClass(clef.Clef, getElementMethod='getElementBefore')
+        <music21.clef.TrebleClef>
+
+        Searching at-or-after, or even just plain after, since the next element
+        of this stream is the clef in question, yields the clef in m.2:
+
+        >>> m.getContextByClass(clef.Clef, getElementMethod='getElementAtOrAfter')
+        <music21.clef.BassClef>
+        >>> m.getContextByClass(clef.Clef, getElementMethod='getElementAfter')
+        <music21.clef.BassClef>
+
+        Failed searches return None:
+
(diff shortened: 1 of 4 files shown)


music21/stream/base.py
```diff
@@ -3709,7 +3709,71 @@ def getElementsByOffset(
...
 
-    def getElementAtOrBefore(self, offset, classList=None):
+    def getContextByClass(
+        self,
+        className,
+        *,
+        getElementMethod='getElementAtOrBefore',
+        sortByCreationTime=False,
+        followDerivation=True,
+        priorityTargetOnly=False,
```
mscuthbert, line 3719: can we make the new behavior optional with a keyword like `includeInternalContexts=SearchInternalContexts.RECURSE_ZERO` where SearchInternalContexts is an enum of `NEVER, AT_ZERO, and ALL, RECURSE_ZERO, and RECURSE_ALL`

I think that it would be strange that a Measure object getContextByClass('TimeSignature') will find it, but that Part object would not.  So RECURSE or not would determine what iterator to call, and ZERO vs ALL would put in the limitation of getElementsByOffsetInHierarchy at zero or not.

music21/stream/base.py
```diff
@@ -3709,7 +3709,71 @@ def getElementsByOffset(
...
+        *    `'getElementAfter'`
+        *    `'getElementAtOrBefore'` (Default)
+        *    `'getElementAtOrAfter'`
+
+        Ask for measure 2's clef context and get the clef at 0.0 in measure 2:
+
+        >>> p = converter.parse('tinyNotation: c1 d1')
+        >>> m = p.measure(2)
+        >>> m.insert(0, clef.BassClef())
+        >>> m.getContextByClass(clef.Clef, getElementMethod='getElementAtOrBefore')
```
mscuthbert, line 3740: remove `getElementMethod='getElementAtOrBefore'` to show more clearly that this is the default behavior.

mscuthbert (commented):
Looks great!  I just want to go a little bit further with the options for recursion and considering something in the stream but not at zero to also be a possibility.

THANKS!  once this is in, I think this is even more important to add to m21j to smooth over some early design problems there.

music21/stream/base.py
jacobtylerwalls: Wow, yes, I can see the utility of this. That will make this probably the most powerful method in m21, then--to be able to in one method call rip up, down, left, or right through the containers. I think this is manageable, will try to get to it in the next couple weeks. Once this is in I already have a followup patch prepared for the deprecations that depend on this change.

jacobtylerwalls: Despite video games not being a meaningful part of my life at any point, I've noticed we'll be creating music21's joystick controller.

mscuthbert: ha-ha-ha-ha!!!!

I think we should probably use one of the fancy new StringEnums to replace 'getElementAtOrBefore' etc. -- having routines that search for 'At' in the string and do different things based on it probably doesn't make sense...

mscuthbert: Shout out to Chris Ariza who initially implemented this when I decided I wanted it, and somehow figured it out how to get it done.  And Josiah Oberholtzer who figured out how to speed up the code immensely for large scores.

jacobtylerwalls: After thinking about this for a bit I'm having concerns about API surface area.

I think "ALL" is what would have solved the original problem with finding the clef at 0 (or midmeasure, at 2, etc.), and that can be had with either "ALL" or "gEBC" or whatnot:

**on main branch, demoing that we just need to document how to not limit by offsets:**
```python
>>> s = stream.Stream()
>>> s.insert(clef.BassClef())
>>> s.getContextByClass(clef.Clef) is None
True
>>> s.getContextByClass(clef.Clef, getElementMethod='dont_mind_me_just_a_synonym_for_all_or_gEBC')
<music21.clef.BassClef>
>>> s.getContextByClass(clef.Clef, getElementMethod='getElementsAtOrAfter')
<music21.clef.BassClef>
```

Then the other piece of this was whether to provide syntactic sugar to let people run two searches -- first, internal substreams, then context streams -- but then wouldn't it be possible a user would want different at/after/before for the two searches? Any clef in measure 20 including midmeasure clefs, so atOrAfter 0.0, but finding none, want any clef before 80.0, where that measure happens to sit in the context stream, which would be AtOrBefore. And so on--leading to me believe I shouldn't be adding this after all.

Really I think the original bug was just stumbling over the fact that in a tree search, I can't search other nodes at the same offset. If I happen to be at 0.0 but before the other two elements at 0.0, and I'm searching at-or-after: out of luck.

jacobtylerwalls:
Will open a follow up PR with just StringEnum cleanup and documenting a new option `getAllElements` that adds no offset filters.

music21/stream/base.py
jacobtylerwalls: Works equally well, and is used in the package, but just wasn't documented:

```python
>>> s.getContextByClass(clef.Clef, getElementMethod='getElementAtOrBeforeOffset')
<music21.clef.BassClef>
```

PR #1033, opened 2021-06-07
Factor `ChordBase` out of `Chord` to support `PercussionChord` objects and support MusicXML unpitched roundtrip

jacobtylerwalls:
I think this is a reviewable chunk. Input & output of `<unpitched>` MusicXML tags is working. Supersedes #756 

MIDI import/export would come next, probably in tandem with a `makeUnpitched()` function to take notes with unpitched instruments and replace the Notes with Unpitched (or Chord with PercussionChord, etc). Could flesh out the `getInstrument()` work at that time also (or back it out of this PR, your call).


mscuthbert:
Sorry I've been slow at reviewing this.  Should have time this weekend.

music21/instrument.py
```diff
@@ -2167,24 +2168,22 @@ def partitionByInstrument(streamObj):
     TODO: use proper recursion to make a copy of the stream.
     TODO: final barlines should be aligned.
     '''
-    from music21 import stream
```
jacobtylerwalls, line 2170: moved this to the top of the module.

music21/note.py
```diff
@@ -1190,6 +1192,22 @@ def _setVolume(self, value, setClient=True):
         [120, 80]
         ''')
 
+    def _getStoredInstrument(self):
+        return self._storedInstrument
+
+    def _setStoredInstrument(self, newValue):
+        self._storedInstrument = newValue
+
+    storedInstrument = property(_getStoredInstrument, _setStoredInstrument)
```
jacobtylerwalls, line 1201: I must have done this a while back when trying to emulate other m21 properties. This should probably just be an attribute.

jacobtylerwalls:
TODO: once decision is reached on whether _storedInstrument goes to NotRest or stays on Unpitched, factor that into musicxml export (#1051) so that context search is only a fallback.

music21/note.py
jacobtylerwalls: Oh I see, I merely moved this from Unpitched without changing it.

music21/chord/__init__.py
```diff
@@ -47,9 +47,339 @@ class ChordException(exceptions21.Music21Exception):
...
+        if 'beams' in keywords:
+            self.beams = keywords['beams']
+        else:
+            self.beams = beam.Beams()
+
+    def __eq__(self, other):
+        '''
+        True if the Chord passes all `super()`
+        equality tests and the pitches are the same
+        (possibly in a different order)
```
mscuthbert, line 135: This probably needs to be changed for BaseChord.  These docs look more like they belong in Chord.__eq__

music21/chord/__init__.py
```diff
@@ -47,9 +47,339 @@ class ChordException(exceptions21.Music21Exception):
...
+        for n in notes:
+            if isinstance(n, pitch.Pitch):
+                # assign pitch to a new Note
+                if useDuration:  # not False or None
+                    newNote = note.Note(n, duration=useDuration)
+                else:
+                    newNote = note.Note(n)
+                self._notes.append(newNote)
+                # self._notes.append({'pitch':n})
+            elif isinstance(n, Chord):
```
mscuthbert, line 222: probably BaseChord?  So that one Unpitched can merge another?

music21/chord/__init__.py
```diff
@@ -47,9 +47,339 @@ class ChordException(exceptions21.Music21Exception):
...
+        Traceback (most recent call last):
+        ValueError: Cannot remove ['C4', 'E4'] from a chord; try a Pitch or Note object
+        '''
+        if isinstance(removeItem, str):
+            for n in self._notes:
+                if n.pitch.nameWithOctave == removeItem:
+                    self._notes.remove(n)
+                    self.clearCache()
+                    return
+            raise ValueError('Chord.remove(x), x not in chord')
```
mscuthbert, line 358: This should be in Chord.remove() -- doesn't apply to Unpitched.

music21/chord/__init__.py
```diff
@@ -47,9 +47,339 @@ class ChordException(exceptions21.Music21Exception):
...
+        if not hasattr(removeItem, 'classes'):
+            raise ValueError('Cannot remove {} from a chord; try a Pitch or Note object'.format(
+                removeItem))
+        if 'Pitch' in removeItem.classes:
+            for n in self._notes:
+                if n.pitch == removeItem:
+                    self._notes.remove(n)
+                    self.clearCache()
+                    return
+            raise ValueError('Chord.remove(x), x not in chord')
```
mscuthbert, line 369: As above.

music21/instrument.py
```diff
@@ -63,10 +63,11 @@ def unbundleInstruments(streamIn, *, inPlace=False):
     if inPlace is True:
         s = streamIn
     else:
-        s = copy.deepcopy(streamIn)
+        s = streamIn.coreCopyAsDerivation('unbundleInstruments')
```
mscuthbert, line 66: :-)

music21/instrument.py
mscuthbert: ah!  This used to have recursion problems.  Probably fixed with the base/core separation!

music21/musicxml/m21ToXml.py
```diff
@@ -3863,6 +3887,49 @@ def pitchToXml(self, p: pitch.Pitch):
...
 
+    def unpitchedToXml(self,
+                       up: note.Unpitched,
+                       noteIndexInChord: int = 0,
+                       chordParent: chord.ChordBase = None) -> Element:
+        '''
+        Convert a :class:`~music21.note.Unpitched` to a <note>
+        with an <unpitched> subelement.
+
+        >>> up = note.Unpitched('D5')
```
mscuthbert, line 3898: for constructing Unpitched objects, can we specify `displayName='D5'`  -- I think that otherwise it might be an issue later where people are surprised that displayName='C4' appears one line below the staff no matter what the clef is.

music21/musicxml/xmlToM21.py
```diff
@@ -2646,11 +2647,22 @@ def xmlToChord(self, mxNoteList):
         >>> c = MP.xmlToChord([a, b])
         >>> c.getNotehead(c.pitches[0])
         'diamond'
+
+        >>> a = EL('<note><unpitched><display-step>A</display-step>'
+        ...        + '<display-octave>3</display-octave></unpitched>'
+        ...        + qnDuration
+        ...        + '<notehead>diamond</notehead></note>')
+        >>> MP.xmlToChord([a, b])
+        <music21.percussion.PercussionChord unpitched[A3] B3>
```
mscuthbert, line 2656: wow!  Thanks!

music21/note.py
mscuthbert: yeah -- I didn't know back then how easy it was to change an attribute to a property later.

music21/note.py
```diff
@@ -1190,6 +1192,22 @@ def _setVolume(self, value, setClient=True):
...
+
+    def getInstrument(self):
+        '''
+        TOOD: doc and test
+        '''
+        if self.storedInstrument is not None:
+            return self.storedInstrument
+        return self.getContextByClass('Instrument', followDerivation=False)
```
mscuthbert, line 1209: not sure whether followDerivation should be True or False here.

music21/note.py
```diff
@@ -1445,7 +1462,7 @@ def _setOctave(self, value: int):
     @property
     def pitches(self) -> Tuple[pitch.Pitch]:
         '''
-        Return the :class:`~music21.pitch.Pitch` object in a tuple.
+        Return the :class:`~music21.pitch.Pitch` objects in a tuple.
```
mscuthbert, line 1465: maybe clearer "Return the single Pitch object within a tuple" -- there can't be more than one pitch object here.

music21/note.py
```diff
@@ -1611,9 +1627,6 @@ class Unpitched(NotRest):
     A General class of unpitched objects which appear at different places
     on the staff.  Examples: percussion notation.
 
-    The `Unpitched` object does not currently do anything and should
-    not be used.
-
```
mscuthbert, line 1616: can we change docs below to "which shows where they should be displayed as if the staff were a 5-line staff in treble clef"?

music21/musicxml/xmlToM21.py
```diff
@@ -2995,6 +3014,42 @@ def xmlToPitch(self, mxNote, inputM21=None):
...
+            unp = note.Unpitched()
+        else:
+            unp = inputM21
+
+        mxDisplayStep = mxUnpitched.find('display-step')
+        mxDisplayOctave = mxUnpitched.find('display-octave')
+        if textStripValid(mxDisplayStep):
+            unp.displayStep = mxDisplayStep.text.strip()
+        if textStripValid(mxDisplayOctave):
+            unp.displayOctave = int(mxDisplayOctave.text.strip())
```
mscuthbert, line 3049: I'd like to make a TODO to compensate for MusicXML number of staff-lines affects the display-step behavior, which I don't want to adopt in music21.  In a one line staff, B4 should equal center line in music21 (as it does in most other notation settings).  In MusicXML which line an Unpitched is displayed on depends on the number of stafflines currently active.

See https://github.com/w3c/musicxml/issues/351

music21/note.py
```diff
@@ -878,8 +878,9 @@ class NotRest(GeneralNote):
     '''
     Parent class for Note-like objects that are not rests; that is to say
     they have a stem, can be tied, and volume is important.
-    Basically, that's a `Note` or `Chord` or
-    `Unpitched` object for now.
+    Basically, that's a :class:`Note` or :class:`~music21.chord.Chord`
+    (or their subclasses such as :class:`~music21.harmony.ChordSymbol`), or
```
mscuthbert, line 882: :-)

music21/note.py
mscuthbert: but let's leave it.  We might want to change the displayStep etc. based on the instrument.

music21/note.py
```diff
@@ -1630,10 +1643,16 @@ class Unpitched(NotRest):
...
 
-    def __init__(self):
-        super().__init__()
-        self.displayStep = 'C'
-        self.displayOctave = 4
+    def __init__(self, displayName=None, **keywords):
+        super().__init__(**keywords)
+
+        self.displayStep: str = 'C'
+        self.displayOctave: int = 4
```
mscuthbert, line 1650: If possible, let's change the default to 'B', 4.  Any maybe put in some sort of "displayStepIsInferred" if it's not set, so that it can be affected by Instrument.

music21/note.py
```diff
@@ -1671,6 +1690,32 @@ def displayPitch(self) -> pitch.Pitch:
...
+        Returns the MIDI pitch corresponding to the stored display pitch.
+
+        >>> unp = note.Unpitched('D4')
+        >>> unp.midi
+        62
+        >>> unp2 = note.Unpitched()
+        >>> unp2.midi
+        60
+        '''
+        return self.displayPitch().midi
```
mscuthbert, line 1705: Not sure I want this in yet -- is there a reason it needs to be here?   I think people would expect the MIDI not to be dependent on the displayPitch but on the storedInstrument's percussion MIDI mapping.

music21/percussion.py
```diff
@@ -0,0 +1,95 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+# Name:         percussion.py
+# Purpose:      music21 classes for representing unpitched events
+#
+# Authors:      Michael Scott Cuthbert
+#               Christopher Ariza
+#
```
mscuthbert, line 8: Definitely add your name here.  :-)

music21/percussion.py
```diff
@@ -0,0 +1,95 @@
...
+    object.
+
+    >>> pChord = percussion.PercussionChord([note.Unpitched('D4'), note.Note('E5')])
+    >>> pChord.isChord
+    False
+
+    Has notes, just like any ChordBase:
+
+    >>> pChord.notes
+    (<music21.note.Unpitched object at 0x...>, <music21.note.Note E>)
```
mscuthbert, line 33: next TODO -- improve repr for Unpitched (probably with Instrument name rather than displayStep)

music21/test/testRunner.py
```diff
@@ -49,13 +49,10 @@ def addDocAttrTestsToSuite(suite,
...
     >>> s1TestsAfter = len(s1._tests)
     >>> s1TestsAfter - s1TestsBefore
-    1
+    2
     >>> t = s1._tests[-1]
     >>> t
     isRest ()
-
-    >>> 'hi'
-    'hi'
```
mscuthbert, line 58: this used to be a test for unicode.  :-)

mscuthbert (approved):
Looks great to me!  Thanks -- totally the right direction

I think that some of the things that should move from Chord to ChordBase are `tie`, `volume` and anything else that finds the appropriate `_notes` object but doesn't deal with Pitch.

This is probably a good enough work through musicxml.  Where people really want this is in MIDI, so we can look at the midi.percussion next -- PercussionMapper is the skeleton to work with there.

jacobtylerwalls:
Any opinion on the wisdom of moving _storedInstrument to NotRest from Unpitched? Would turn out to be useful for the multiple instrument musicxml export patch.

music21/note.py
jacobtylerwalls: I need a diatonic note number to sort the notes from musicxml documents here:

https://github.com/cuthbertLab/music21/blob/c69968d44213a1915d1e578f7509f3a7642c0e3b/music21/musicxml/xmlToM21.py#L2680-L2682

but I should probably just implement `diatonicNoteNum` instead.

jacobtylerwalls: Nope! I see what you mean, I was accessing the displayPitch() here directly anyway, so I don't need this implemented here.

music21/note.py
jacobtylerwalls: display-step and display-octave are required, so I don't want to build extra layers on top we might not need just yet.

jacobtylerwalls: Oh I see you mean starting from the m21 side.

jacobtylerwalls: Still inclined to defer until we have a way of changing display pitch based on instrument

music21/musicxml/m21ToXml.py
jacobtylerwalls: I didn't quite follow this.

music21/chord/__init__.py
jacobtylerwalls: Not sure I follow -- I thought the point of `PercussionChord` was so it could mix anything, unpitched or pitched, so we can cleanly import multi-sticked percussion chords, or chordified orchestral scores, or whatever. So we still want remove(), but this is great you've pointed this out, because there's a bug:

```
>>> pChord = percussion.PercussionChord([note.Unpitched('D4'), note.Note('E5')])
>>> pChord.remove(pitch.Pitch('E5'))
  File "/Users/jwalls/music21/music21/chord/__init__.py", line 365, in remove
    if n.pitch == removeItem:
AttributeError: 'Unpitched' object has no attribute 'pitch'
```

music21/note.py
jacobtylerwalls: what are the most likely derivations? stripTies? partitionByInstrument? I think those keep instruments around, so I think we're alright. We just added instruments to voicesToParts.

music21/chord/__init__.py
jacobtylerwalls: I rewrote it slightly.

music21/chord/__init__.py
jacobtylerwalls: Great catch!

music21/note.py
jacobtylerwalls: I'm thinking it should be False. I just went on a goose chase in #1051 dealing with instruments being found that weren't in the stream being exported and getting mismatches with the `<part-list>`.

jacobtylerwalls: In other words, it's not really "stored" if the instrument is not in the stream. So, maybe: If you want derivations, then opt-in with `gCBC(follow...`)

music21/note.py
jacobtylerwalls: (But I did change C4 to B4.)

music21/chord/__init__.py
jacobtylerwalls: The bug was fixed, so I think we're good now.

music21/instrument.py
```diff
@@ -63,10 +63,11 @@ def unbundleInstruments(streamIn, *, inPlace=False):
...
 
     for thisObj in s:
-        if 'Unpitched' in thisObj.classes:
+        if 'NotRest' in thisObj.classes:
+            # TODO(QUESTION) -- also unbundle each note of chord?
```
mscuthbert: yes, but that will be hard because we will need Voice objects to move each instrument.

music21/musicxml/m21ToXml.py
mscuthbert: Watch the next diff to see what I meant.

mscuthbert: Just let's never create:

```
note.Unpitched('D5')
```

always instead

```
note.Unpitched(displayName='D5')
```

because even if the current clef is Bass clef, displayName refers to the location as if the staff were in treble clef no matter what.

music21/musicxml/xmlToM21.py
```diff
@@ -2686,11 +2687,22 @@ def xmlToChord(self, mxNoteList):
...
+        >>> MP.xmlToChord([a, b])
+        <music21.percussion.PercussionChord unpitched[A3] B3>
         '''
         notes = []
         for mxNote in mxNoteList:
             notes.append(self.xmlToSimpleNote(mxNote, freeSpanners=False))
-        c = chord.Chord(notes)
+
+        if any(mxNote.find('unpitched') for mxNote in mxNoteList):
+            c = percussion.PercussionChord(notes)
```
mscuthbert, line 2703: Super!

music21/musicxml/xmlToM21.py
```diff
@@ -2779,10 +2791,17 @@ def xmlToSimpleNote(self, mxNote, freeSpanners=True):
...
         d = self.xmlToDuration(mxNote)
-        n = note.Note(duration=d)
 
-        # send whole note since accidental display not in <pitch>
-        self.xmlToPitch(mxNote, n.pitch)
+        n = None
+
+        mxUnpitched = mxNote.find('unpitched')
+        if mxUnpitched is None:
+            # send whole note since accidental display not in <pitch>
```
mscuthbert, line 2799: should change that to "send entire note" since I didn't understand what wholenotes had to do with it.

music21/note.py
```diff
@@ -1675,6 +1693,18 @@ def displayPitch(self) -> pitch.Pitch:
...
+    def displayName(self) -> str:
+        '''
+        Returns the `nameWithOctave` of the :meth:`displayPitch`.
+
+        >>> unp = note.Unpitched('B2')
+        >>> unp.displayName
+        'B2'
+        '''
+        display_pitch = self.displayPitch()
+        return display_pitch.nameWithOctave
```
mscuthbert, line 1706: why is this not just `self.displayStep + str(self.displayOctave)`?

mscuthbert (commented):
Looks really good to me -- feel free to merge after reading, fixing any broken tests, and responding if needed.

Issue #881, opened 2021-02-27
music21's import conventions prevent using PyInstaller


keeganNull:
I have a program that uses the music21. When I execute the program of .py , it's OK. But when I use the pyinstaller to pack the program and execute it. It shows that 
![image](https://user-images.githubusercontent.com/69446006/109382107-5741b780-7919-11eb-848f-cdff0038085c.png)
and the code:
![image](https://user-images.githubusercontent.com/69446006/109382123-6d4f7800-7919-11eb-97a8-28669fe37d80.png)

keeganNull:
And I just try a program like this.
![image](https://user-images.githubusercontent.com/69446006/109382446-74778580-791b-11eb-9aa2-1c13e491c1f7.png)
still like this:
![image](https://user-images.githubusercontent.com/69446006/109382455-7fcab100-791b-11eb-9b6a-5c514b7e12f5.png)

jacobtylerwalls:
Hi -- in the future, please use the provided issue template. For instance, what version of music21 are you using? There was a fix for `importlib` issues in 6.7.0.

keeganNull:
> Hi -- in the future, please use the provided issue template. For instance, what version of music21 are you using? There was a fix for `importlib` issues in 6.7.0.
Sorry for I didn't use the template.
My music21 version is 6.7.1. python version in 3.7.4. 
As for I'm not very good at python,  I don't understand what are you talking about. Should I update the 'importlib' or change the way I import 'music21'?

jacobtylerwalls:
Try this:

'pip install numpy==1.19.3'

Source:


https://stackoverflow.com/questions/64772965/pyinstaller-executable-file-throws-an-error-when-opening-on-another-pc

keeganNull:
> Try this:
> 
> 'pip install numpy==1.19.3'
> 
> Source:
> 
> https://stackoverflow.com/questions/64772965/pyinstaller-executable-file-throws-an-error-when-opening-on-another-pc

Sorry. I tried it. But it didn't work.  According to your source, it's the problem when the program executes on different PCs.But my error may is caused by the 'music21'.
And the error : module 'music21' has no attribute 'alpha'.

mscuthbert:
I don't know what pyinstaller is -- we recommend using `pip` which is the default Python installer.

keeganNull:
> I don't know what pyinstaller is -- we recommend using `pip` which is the default Python installer.

[...]

PR #1034, opened 2021-06-09
June miscellany

jacobtylerwalls:
Highlights are removing the "mx" file format and adding some test coverage for the `splitAtDurations()` edits in #1005.


documentation/source/usersGuide/usersGuide_12_music21object.ipynb
```diff
@@ -1454,24 +1454,18 @@
  ],
  "metadata": {
   "anaconda-cloud": {},
+  "interpreter": {
+   "hash": "aee8b7b246df8f9039afb4144a1f6fd8d2ca17a180786b69acc140d282b71a49"
+  },
   "kernelspec": {
-   "display_name": "Python 3",
-   "language": "python",
+   "display_name": "Python 3.10.0b1 64-bit",
```
mscuthbert, line 1461: oh -- good to know that music21 is working on 3.10-beta.  I hadn't tested it yet.

music21/corpus/__init__.py
```diff
@@ -53,13 +53,8 @@
 
 ]
 
-import re
-import os
-import unittest
```
mscuthbert, line 58: really shocked that this didn't cause a pylint error...

music21/stream/base.py
```diff
@@ -2908,6 +2908,14 @@ def splitAtDurations(self, *, recurse=False) -> base._SplitTuple:
         >>> result = m.splitAtDurations(recurse=True)
         >>> list(result[0][note.Rest])
         [<music21.note.Rest whole>, <music21.note.Rest quarter>]
+
+        >>> r2 = note.Rest(quarterLength=5.0)
```
mscuthbert, line 2912: Can we add a note explaining that 'fullMeasure rests are not split at durations, since they will be expressed as a single whole note in most scores anyhow'

music21/stream/base.py
```diff
@@ -2908,6 +2908,14 @@ def splitAtDurations(self, *, recurse=False) -> base._SplitTuple:
...
+        >>> r2.fullMeasure = True
+        >>> m2 = stream.Measure(r2)
+        >>> m2.insert(0, meter.TimeSignature('6/4'))
```
mscuthbert, line 2915: I think that `5/4` would make this clearer, no?

mscuthbert (commented):
Looks great!  One little change requested -- then feel free to merge.

documentation/source/usersGuide/usersGuide_12_music21object.ipynb
jacobtylerwalls: yeah, the only pain point is some enum repr()s have to be changed in doctests. more later...

music21/stream/base.py
jacobtylerwalls: Needs a comment to explain, clearly. This is showing that when the rest doesn't fill the measure but we set `.fullMeasure` = True anyhow, we still treat as full measure and don't split it.

music21/corpus/__init__.py
jacobtylerwalls: We have it turned off because we have wildcard imports. And speaking of which, that must have been how identified these, I was looking at pylint reports when poking at #881 (which BTW I think I have a solution for)

jacobtylerwalls:
We haven't really discussed our policy on squashing vs. rebasing/merge commits. I think this one's too eclectic to squash. It also has the benefit of letting you see the last three commits I pushed.

PR #1038, opened 2021-06-15
Prevent writing `<part-group>` tags on multi-staff parts

### LINKED ISSUE #1037: Superfluous `<part-group>` tags on joinable groups are written
**music21 version**

dev

**Problem summary**
[Spec says](https://w3c.github.io/musicxml/musicxml-reference/elements/part-group/) not to use `<part-group>` tags to represent multi-staff parts. This was reworked as part of the part staff exporter feature in v6.5, but apparently I should have jettisoned it completely:

> A `<part-group>` element is not needed for a single multi-staff part. By default, multi-staff parts include a brace symbol and (if appropriate given the `<bar-style>`) common barlines. The symbol formatting for a multi-staff part can be more fully specified using the `<part-symbol>` element.

**Steps to reproduce**

- load score containing piano into music21
- observe one `layout.StaffGroup` created to join the piano staves in music21
- export musicxml
- observe `<part-group>` tag written: usually harmless but noncompliant
- do stuff with another package that might be altering the staff-group tags
- load that exported musicxml file back into music21
- observe there are two `StaffGroup` objects now
- export THAT

Screen Shot 2021-06-15 at 3 48 33 PM

**More info**
Haven't debugged what the "other package" is doing with the staff-group tags, because I can (theoretically) moot that by not writing out the superfluous tags.

jacobtylerwalls:
Fixes #1037

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -1968,18 +1968,20 @@ class ScoreExporter(XMLExporterBase, PartStaffExporterMixin):
           <part-group number="1" type="stop" />
         </part-list>
 
-        And with a piano staff, to observe that only one <part> tag has both "start" and "stop":
+        Multi-staff parts (such as piano staves), should NOT receive `<part-group>` tags,
+        since they are joined by `<staff>` tags:
 
         >>> cpe = corpus.parse('cpebach')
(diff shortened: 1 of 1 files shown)


mscuthbert (approved):
lgtm

mscuthbert:
We should probably make a note that `<part-symbol>` objects are not currently generated.  All good!

PR #1040, opened 2021-06-16
Fixes to manipulation and export of scores in concert pitch

### LINKED ISSUE #1039: Transposing instrument part (atSoundingPitch==True) not shown in written pitch
**music21 version**

`7.0.6a1+PR996`   

**Problem summary**
score.show('musicxml.pdf') of a score containing transposing instruments (with part.atSoundingPitch set to True) exports with sounding pitch being written instead of written pitch.

See code example below...

```

import music21

inst = music21.instrument.Instrument('madeUpInstrument')

# instrument transposes by 1 "letter" (generic 2nd), 2 semitones (2 chromatic steps)
inst.transposition = music21.interval.intervalFromGenericAndChromatic(2, 2)
part = music21.stream.Part()
part.atSoundingPitch = True
part.insert(0, music21.clef.TrebleClef())
part.insert(0, music21.key.KeySignature(sharps=-1)) # sounding key = 1 flat
part.insert(0, music21.note.Note('C4')) # sounding note = middle C
part.insert(0, inst)

score = music21.stream.Score()
score.insert(0, part)

# the following should display a B-flat (key = 3 flats).
# Instead it displays a C (key = 1 flat).
score.show('musicxml.pdf')

part.toWrittenPitch(inPlace=True)

# now it's right
score.show('musicxml.pdf')
```

See code snippet.

**More information**

macOS Big Sur, using MuseScore for showing scores.

jacobtylerwalls:
Fixed #1039

<s>FYI -- I like the pattern I started using in #845 to rely on `obj.recurse(streamsOnly=True, includeSelf=True)` to handle recursion instead of trying to recursively call the function in specific nesting cases (part-like, etc).</s>

<s>There could be more edge cases to test (what if a measure has soundingPitch=False but the part has soundingPitch=True), but this is the chunk I'm committing to ATM.</s> got it!

MusicXML import doesn't really present this question because all scores were written until 4.0. So it makes sense Greg found this with Humdrum where all scores are sounding.

@gregchapman-dev, want to have a look?


jacobtylerwalls:
Looked into adding coverage for finding the property via sites: but this became difficult because of the way that `extendDurationAndGetBoundaries()` swallows exceptions, and it's marked for deprecation anyway, which we may turn to here also before end of the summer.

jacobtylerwalls:
Still not quite right -- need to restore the functionality where a score's parts have mixed statuses and calling toWrittenPitch on the score still works out. And add tests for it, which we don't have. Think I can do this without reverting to the part-like streams pattern, but just not going to get to it today.

music21/stream/base.py
```diff
@@ -4932,6 +4932,20 @@ def _transposeByInstrument(self,
             # print(k, i.transposition)
         return returnObj
 
+    def _shouldBeAtSoundingPitch(self, contextStream):
```
mscuthbert, line 4935: I get this and why it needs to be here, but maybe a little more docs about why internal Streams might have atSoundingPitch = 'unknown' (Measures, etc.) while the Score has a transposition set

music21/stream/base.py
```diff
@@ -4966,51 +4980,29 @@ def toSoundingPitch(self, *, inPlace=False):
...
-        Traceback (most recent call last):
-        music21.exceptions21.StreamException: atSoundingPitch is unknown: cannot transpose
-        >>> s.atSoundingPitch = False
-        >>> sp = p.toSoundingPitch()
-        >>> sp
-        <music21.stream.Part partEmpty>
-        >>> sp.derivation.origin is p
-        True
+        Changed in v.7 -- no longer raises `StreamException` if `atSoundingPitch`
+        is `'unknown'`.
```
mscuthbert, line 4984: can you specify what happens in that case?

music21/stream/base.py
```diff
@@ -5043,40 +5035,30 @@ def toWrittenPitch(self, *, inPlace=False):
         False
         >>> scWritten.recurse().notes[0].nameWithOctave
         'A4'
+
+        Changed in v.7 -- no longer raises `StreamException` if `atSoundingPitch`
+        is `'unknown'`.
```
mscuthbert, line 5040: As above -- needs details about what is returned instead.

music21/stream/tests.py
```diff
@@ -7034,19 +7035,31 @@ def testTransposeByPitchA(self):
...
+                post = m._transposeByInstrument(inPlace=False, reverse=reverse)
+                for n in post.notes:
+                    demoPitches.extend(n.pitches)
+            return demoPitches
 
-        test = p1._transposeByInstrument(inPlace=False, reverse=True)
-        self.assertEqual([str(p) for p in test.pitches], ['G4', 'G4', 'D4', 'D4'])
+        self.assertEqual([str(p) for p in p1.pitches], ['C4', 'C4', 'C4', 'C4'])
+        self.assertEqual([str(p) for p in simulateRecursing(p1)], ['F3', 'F3', 'B-3', 'B-3'])
+        self.assertEqual([str(p) for p in simulateRecursing(p1, reverse=True)],
```
mscuthbert, line 7061: might it be better to make transposeByInstrument recurse?  I thought it was supposed to, or is it only the private method that doesn't?  Calling private methods on anything but `self` is slightly frowned upon, since it doesn't give flexibility to change internal behavior in the futre.

music21/stream/tests.py
```diff
@@ -7106,13 +7131,29 @@ def testTransposeByPitchB(self):
...
+        p = converter.parse('tinyNotation: c1 d1')
+        p.insert(0, instrument.Horn())
+        s = Score(p)
+        s.atSoundingPitch = True
+        self.assertEqual(p.atSoundingPitch, 'unknown')
+
+        # Can still perform toWrittenPitch by getting the value of atSoundingPitch
+        # from the score
+        p.toWrittenPitch(inPlace=True)
+        self.assertEqual(p.flat.notes[0].nameWithOctave, 'G4')
```
mscuthbert, line 7153: SUPER!

mscuthbert (commented):
Looking great!  Some Qs about calling private methods on non-self objects

music21/stream/tests.py
jacobtylerwalls: `_transposeByInstrument` essentially gets instruments and then transposes. It doesn't recurse. Tranposing recurses, sure, but the instrument search does not recurse. The instrument search is done with `extendDurationsAndGetBoundaries('Instrument')`, which is marked for deprecation, so maybe it's time to rewrite the instrument search and then drop the recursing from toSoundingPitch/toWrittenPitch, which will simplify them even further.

Another benefit of doing that is that it will prevent edge cases from there being superfluous instruments at multiple levels. Bb Clarinet in part and Bb Clarinet in measure == Clarinet in Ab?

Another another benefit is that the helper method to treatAtSoundingPitch won't have to be called on non-self objects and so it can stay private.

jacobtylerwalls: This probably means I have to restore the special-case for part-like streams. Calling a score should not search downward to find all part instruments and transpose the entire score-slice by that value. Calling a part should search downward to find all measure and voice instruments and tranpsose the entire part-slice by that value.

jacobtylerwalls: Ended up restoring the special-case for has part-like streams.

mscuthbert (approved):
Looks great, thanks!

mscuthbert:
Needed to fix a conflict.  Feel free to merge if all tests pass.

PR #1041, opened 2021-06-16
Support and recommend `.iter()` instead of `.iter`

### LINKED ISSUE #1025: `.iter` -> `.iter()` transition
Toyed with implementing this today. Some questions:

- If we continue supporting an `.iter` property for one version, unfortunately this will interfere with the intended new usage of `.iter()`, which will call the property rather than any new `iter()` method and then fail with `TypeError: 'StreamIterator' object is not callable`.
- Workaround would be to implement `StreamIterator.__call__()` for one version to just return self, which solves the TypeError but not the deprecation warnings. Assuming we want to emit deprecation warnings once/session with @common.deprecated, every call to `.iter()` is going to hit `.iter` and emit a warning, potentially confusing users.
- Workaround for that is for all internal code to call `iter(self)`--or equivalently, `self.__iter__()`--instead of `self.iter()`, but that causes me to question whether the new intended use should just be `iter(self)` instead of `self.iter()`, which would be slightly more Pythonic anyway.

jacobtylerwalls:
Fixes #1025 

I had this lying around from when I wrote up the issue. Just tweaked the doc string on the `__call__()` method a bit and it was ready to go.

On second glance, some of these existing calls to `.iter` look unnecessary, <s>but I didn't stop to audit them all.</s> so I spruced them up to get Pylint to pass (two extra characters knocked a couple of these over the line limit!)

### DIFF
--- a/music21/stream/iterator.py
+++ b/music21/stream/iterator.py
@@ -317,7 +317,7 @@ class StreamIterator(prebase.ProtoM21Object):
         >>> s = stream.Stream()
         >>> s.insert(0, note.Note('F#'))
         >>> s.repeatAppend(note.Note('C'), 2)
-        >>> sI = s.iter
+        >>> sI = s.iter()
         >>> sI
         <music21.stream.iterator.StreamIterator for Stream:0x104743be0 @:0>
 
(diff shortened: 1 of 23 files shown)


music21/stream/iterator.py
```diff
@@ -164,6 +164,21 @@ def _reprInternal(self):
...
+        (recommended) will retrieve the property `.iter` and necessitate
+        this workaround.
+
+        Returns `self` without any changes.
+
+        TODO: manage and emit DeprecationWarnings in v.8
+        TODO: remove in v.9
+        '''
+        return self
+
```
mscuthbert, line 181: PERFECT!!!

mscuthbert (approved):
Looks GREAT to me!

PR #1044, opened 2021-06-25
Implemented jazz minor scale and relative key functionality on harmonic, melodic, and jazz minors

### LINKED ISSUE #955: Melodic minor asc/desc
**music21 version**

5.7.2

**Problem summary**
noob request
I have helped to write a fretboard exercise switches scale periodically.  MelodicMinorScale respect the tradition of alterin 2 notes depending whether ascending or descending, but jazz musicians generally regard melodic minor as just the "ascending" mode (ie major 6th major 7th) regardless of direction. If there is a scale within Music21 that gives me this result, I cannot find it. Would I need to define my own scale?

**Steps to reproduce**

```

myStream1 = converter.parse('tinynotation: 4/4 c4 e4 g2')
myStream2 = stream.Stream()
myStream2.insert(note.Rest(0.5))
```

descending scales give undesirable notes

**More information**

Phantasmogenesis:
Closes #955 and also allows getRelativeMinor/getRelativeMajor functions to be used on minor scales other than natural minor, as per the [documentation](http://web.mit.edu/music21/doc/moduleReference/moduleScale.html).

### DIFF
--- a/music21/key.py
+++ b/music21/key.py
@@ -1078,7 +1078,7 @@ class Key(KeySignature, scale.DiatonicScale):
         To use the harmonic form, change `.abstract` on the key to
         another abstract scale:
 
-        >>> minorKey.abstract = scale.AbstractHarmonicMinorScale()
+        >>> minorKey.abstract = key.Key(mode='harmonic minor').abstract
         >>> minorKey.deriveByDegree(7, 'E')
         <music21.key.Key of f minor>
         >>> minorKey.deriveByDegree(6, 'G')
--- a/music21/scale/__init__.py
+++ b/music21/scale/__init__.py
@@ -50,14 +50,14 @@ __all__ = [
     'TERMINUS_LOW', 'TERMINUS_HIGH',
     'ScaleException', 'Scale',
     'AbstractScale', 'AbstractDiatonicScale', 'AbstractOctatonicScale',
-    'AbstractHarmonicMinorScale', 'AbstractMelodicMinorScale',
+    'AbstractHarmonicMinorScale', 'AbstractJazzMinorScale', 'AbstractMelodicMinorScale',
     'AbstractCyclicalScale', 'AbstractOctaveRepeatingScale',
     'AbstractRagAsawari', 'AbstractRagMarwa', 'AbstractWeightedHexatonicBlues',
     'ConcreteScale', 'DiatonicScale', 'MajorScale',
     'MinorScale', 'DorianScale', 'PhrygianScale', 'LydianScale', 'MixolydianScale',
     'HypodorianScale', 'HypophrygianScale', 'HypolydianScale', 'HypomixolydianScale',
     'LocrianScale', 'HypolocrianScale', 'HypoaeolianScale',
-    'HarmonicMinorScale', 'MelodicMinorScale',
+    'HarmonicMinorScale', 'JazzMinorScale', 'MelodicMinorScale',
     'OctatonicScale', 'OctaveRepeatingScale', 'CyclicalScale', 'ChromaticScale',
     'WholeToneScale', 'SieveScale', 'ScalaScale', 'RagAsawari',
     'RagMarwa', 'WeightedHexatonicBlues',
@@ -847,13 +847,16 @@ class AbstractHarmonicMinorScale(AbstractScale):
     A true bi-directional scale that with the augmented
     second to a leading tone.
 
-    This is the only scale to use the "_alteredDegrees" property.
+    The harmonic and jazz minor scales are the only to
+    use the "_alteredDegrees" property.
     '''
 
     def __init__(self, mode=None):
         super().__init__()
         self.type = 'Abstract Harmonic Minor'
         self.octaveDuplicating = True
+        self.relativeMinorDegree = 1
+        self.relativeMajorDegree = 3
         self.dominantDegree: int = -1
         self.buildNetwork()
 
@@ -872,6 +875,57 @@ class AbstractHarmonicMinorScale(AbstractScale):
             'interval': interval.Interval('a1')
         }
 
+    def getRelativeMajor(self):
+        return MajorScale(self.tonic)
+
+    def getRelativeMinor(self):
+        return MinorScale(self.tonic)
+
+
+class AbstractJazzMinorScale(AbstractScale):
+    '''
+    A true bi-directional scale containing the raised
+    6th and 7th scale degrees.
+
+    The harmonic and jazz minor scales are the only to
+    use the "_alteredDegrees" property.
+    '''
+
+    def __init__(self, mode=None):
+        super().__init__()
+        self.type = 'Abstract Jazz Minor'
+        self.relativeMinorDegree = 1
+        self.relativeMajorDegree = 3
+        self.octaveDuplicating = True
+        self.dominantDegree: int = -1
+        self.buildNetwork()
+
+    def buildNetwork(self):
+        intervalList = ['M2', 'm2', 'M2', 'M2', 'm2', 'M2', 'M2']  # a to A
+        self.tonicDegree = 1
+        self.dominantDegree = 5
+        self._net = intervalNetwork.IntervalNetwork(intervalList,
+                                                    octaveDuplicating=self.octaveDuplicating,
+                                                    pitchSimplification=None)
+
+        # raise the sixth and seventh in all directions
+        # 6 and 7 here are both scale step/degree, not node id
+        self._alteredDegrees[6] = {
+            'direction': intervalNetwork.DIRECTION_BI,
+            'interval': interval.Interval('a1')
+        }
+
+        self._alteredDegrees[7] = {
+            'direction': intervalNetwork.DIRECTION_BI,
+            'interval': interval.Interval('a1')
+        }
+
+    def getRelativeMajor(self):
+        return MajorScale(self.tonic)
+
+    def getRelativeMinor(self):
+        return MinorScale(self.tonic)
+
 
 class AbstractMelodicMinorScale(AbstractScale):
     '''
@@ -882,6 +936,8 @@ class AbstractMelodicMinorScale(AbstractScale):
         super().__init__()
         self.type = 'Abstract Melodic Minor'
         self.octaveDuplicating = True
+        self.relativeMinorDegree = 1
+        self.relativeMajorDegree = 3
         self.dominantDegree: int = -1
         self.buildNetwork()
 
@@ -893,6 +949,11 @@ class AbstractMelodicMinorScale(AbstractScale):
             pitchSimplification=None)
         self._net.fillMelodicMinor()
 
+    def getRelativeMajor(self):
+        return MajorScale(self.tonic)
+
+    def getRelativeMinor(self):
+        return MinorScale(self.tonic)
 
 class AbstractCyclicalScale(AbstractScale):
     '''
@@ -2830,9 +2891,9 @@ class HarmonicMinorScale(DiatonicScale):
     '''
     The harmonic minor collection, realized as a scale.
 
-    (The usage of this collection as a scale, is quite ahistorical for
+    (The usage of this collection as a scale is quite ahistorical for
     Western European classical music, but it is common in other parts of the
-    world, but where the term "HarmonicMinor" would not be appropriate).
+    world, where the term "Harmonic Minor" would not be appropriate).
 
     >>> sc = scale.HarmonicMinorScale('e4')
     >>> [str(p) for p in sc.pitches]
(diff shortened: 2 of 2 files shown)


music21/scale/__init__.py
```diff
@@ -2864,6 +2925,38 @@ def __init__(self, tonic=None):
         # self._abstract.buildNetwork()
 
 
+class JazzMinorScale(DiatonicScale):
+    '''
+    The jazz minor scale.
```
jacobtylerwalls, line 2930: Thanks! I would think a short description in words would be nice, something like the ascending form of melodic minor but in both directions.

music21/scale/__init__.py
```diff
@@ -2864,6 +2925,38 @@ def __init__(self, tonic=None):
...
+    def __init__(self, tonic=None):
+        super().__init__(tonic=tonic)
+        self.type = 'jazz minor'
+
+        # note: this changes the previously assigned AbstractDiatonicScale
+        # from the DiatonicScale base class
+
+        self._abstract = AbstractJazzMinorScale()
+        # network building happens on object creation
+        # self._abstract.buildNetwork()
```
jacobtylerwalls, line 2957: IMO we can remove these boilerplate comments copied from other scales

music21/scale/__init__.py
```diff
@@ -872,6 +875,57 @@ def buildNetwork(self):
             'interval': interval.Interval('a1')
         }
 
+    def getRelativeMajor(self):
+        return MajorScale(self.tonic)
+
+    def getRelativeMinor(self):
+        return MinorScale(self.tonic)
```
jacobtylerwalls, line 882: Can you say a bit more about the problem you are trying to solve here?

Works for me without any changes:

```
>>> from music21 import scale
>>> gm = scale.MinorScale('g')
>>> gm.getRelativeMajor()
<music21.scale.MajorScale B- major>
```

Also, at a glance the implementation looks wrong (it's getting parallel scales).

music21/scale/__init__.py
```diff
@@ -847,13 +847,16 @@ class AbstractHarmonicMinorScale(AbstractScale):
...
+    The harmonic and jazz minor scales are the only to
+    use the "_alteredDegrees" property.
     '''
 
     def __init__(self, mode=None):
         super().__init__()
         self.type = 'Abstract Harmonic Minor'
         self.octaveDuplicating = True
+        self.relativeMinorDegree = 1
+        self.relativeMajorDegree = 3
```
jacobtylerwalls, line 859: Likewise, can you explain why this is needed?

jacobtylerwalls (commented):
Thanks, Jazz minor will be a nice addition as I see it. Some questions about the rest. Cheers, Jacob

music21/scale/__init__.py
Phantasmogenesis: Indeed, the natural minor scale was already able to use the relative minor/major functionality, but the harmonic/melodic/jazz minor scales could not. With my changes, your test could work on these other scales as well:

`>>> from music21 import scale`
`>>> gm = scale.HarmonicMinorScale('g')`
`>>> gm.getRelativeMajor()`
`<music21.scale.MajorScale B- major>`

Although it does look strange how the functions use self.tonic rather than self.relativeMinorDegree/self.relativeMajorDegree, that is what I saw other similar scales using to implement the relative key functionality. Nevertheless, the output to the user seems to correctly display relative major and minor scales unless I am missing something.

music21/scale/__init__.py
Phantasmogenesis: Without this code, an error message like this is given when using the getRelativeMinor/getRelativeMajor functions on the harmonic/melodic/jazz minor scales:

`AttributeError: 'AbstractHarmonicMinorScale' object has no attribute 'relativeMinorDegree'`

jacobtylerwalls: Ah, got it, any of the non-natural minors, thanks. Yes, this should work because it's a method on `DiatonicScale`.

Can we move these attribute assignments to `buildNetwork()`? That's where they seem to be in all other cases.

music21/scale/__init__.py
jacobtylerwalls: Okay, got it now. Still, I don't think we need to define the functions here on the subclasses, they're already defined on `DiatonicScale`. Actually, now that I look, these are methods on ConcreteScales, so we definitely shouldn't define them on AbstractScales.

Speaking of which, it looks like the (untested) implementation doesn't quite work:

PR:
```
>>> from music21 import *
>>> ms = scale.MelodicMinorScale()
>>> ms.getRelativeMajor()
  File "/Users/jwalls/music21/music21/scale/__init__.py", line 2660, in getRelativeMajor
    return MajorScale(self.pitchFromDegree(self.abstract.relativeMajorDegree))
  File "/Users/jwalls/music21/music21/scale/__init__.py", line 1742, in pitchFromDegree
    post = self._abstract.getPitchFromNodeDegree(
  File "/Users/jwalls/music21/music21/scale/__init__.py", line 487, in getPitchFromNodeDegree
    post = self._net.getPitchFromNodeDegree(
  File "/Users/jwalls/music21/music21/scale/intervalNetwork.py", line 2619, in getPitchFromNodeDegree
    realizedPitch, realizedNode = self.realize(
  File "/Users/jwalls/music21/music21/scale/intervalNetwork.py", line 1714, in realize
    raise IntervalNetworkException('pitchReference cannot be None')
music21.scale.intervalNetwork.IntervalNetworkException: pitchReference cannot be None
```

music21/scale/__init__.py
```diff
@@ -2878,6 +2940,36 @@ def __init__(self, tonic=None):
...
         self._abstract = AbstractMelodicMinorScale()
+        
+        
+class JazzMinorScale(DiatonicScale):
+    '''
+    The jazz minor scale.
+    
+    A synthetic scale identical to the ascending melodic minor. Useful in
+    jazz, especially over harmonies that employ minor-major seventh chords.
+    >>> sc = scale.JazzMinorScale('e4')
```
jacobtylerwalls, line 2951: Great explanation, thanks. But a potential gotcha is that you need a blank line between the paragraph and the doctests (beginning with the Python prompt) for Sphinx to render this correctly.

music21/scale/__init__.py
jacobtylerwalls: In fact, that's probably why the new lines are unexercised on the Coveralls report.

mscuthbert: right should not be on abstract scales and also returns the parallel, not relative.

music21/scale/__init__.py
```diff
@@ -893,7 +904,58 @@ def buildNetwork(self):
             pitchSimplification=None)
         self._net.fillMelodicMinor()
 
+    def getRelativeMajor(self):
+        return MajorScale(self.tonic)
 
+    def getRelativeMinor(self):
+        return MinorScale(self.tonic)
```
mscuthbert, line 911: this also gets the parallels.  But minor scales shouldn't getRelativeMinor anyhow -- it doesn't make sense for a minor to have a parallel minor or a relative minor.  They already are a form of minor.

music21/scale/__init__.py
```diff
@@ -893,7 +904,58 @@ def buildNetwork(self):
...
+        self._alteredDegrees[7] = {
+            'direction': intervalNetwork.DIRECTION_BI,
+            'interval': interval.Interval('a1')
+        }
+
+    def getRelativeMajor(self):
+        return MajorScale(self.tonic)
+
+    def getRelativeMinor(self):
+        return MinorScale(self.tonic)
```
mscuthbert, line 956: as above

music21/scale/__init__.py
```diff
@@ -2830,9 +2892,9 @@ class HarmonicMinorScale(DiatonicScale):
     '''
     The harmonic minor collection, realized as a scale.
 
-    (The usage of this collection as a scale, is quite ahistorical for
+    (The usage of this collection as a scale is quite ahistorical for
     Western European classical music, but it is common in other parts of the
-    world, but where the term "HarmonicMinor" would not be appropriate).
```
mscuthbert, line 2835: thanks for the space, but the "but" should remain

music21/scale/__init__.py
mscuthbert, line 2951: space between comment and example.

music21/scale/__init__.py
```diff
@@ -2878,6 +2940,36 @@ def __init__(self, tonic=None):
...
+    >>> sc = scale.JazzMinorScale('e4')
+    >>> [str(p) for p in sc.pitches]
+    ['E4', 'F#4', 'G4', 'A4', 'B4', 'C#5', 'D#5', 'E5']
+    >>> sc.getTonic()
+    <music21.pitch.Pitch E4>
+    >>> sc.getDominant()
+    <music21.pitch.Pitch B4>
+    >>> sc.pitchFromDegree(1)  # scale degree 1 is treated as lowest
+    <music21.pitch.Pitch E4>
+    >>> sc = scale.JazzMinorScale()
```
mscuthbert, line 2960: space above and explain what's happening here.

music21/scale/__init__.py
```diff
@@ -2878,6 +2940,36 @@ def __init__(self, tonic=None):
...
+    A synthetic scale identical to the ascending melodic minor. Useful in
+    jazz, especially over harmonies that employ minor-major seventh chords.
+    >>> sc = scale.JazzMinorScale('e4')
+    >>> [str(p) for p in sc.pitches]
+    ['E4', 'F#4', 'G4', 'A4', 'B4', 'C#5', 'D#5', 'E5']
+    >>> sc.getTonic()
+    <music21.pitch.Pitch E4>
+    >>> sc.getDominant()
+    <music21.pitch.Pitch B4>
+    >>> sc.pitchFromDegree(1)  # scale degree 1 is treated as lowest
```
mscuthbert, line 2958: usually no need for comments in doc tests, write it as documentation just above.

mscuthbert (changes requested):
Making great progress.  Happy to add this to music21!

music21/scale/__init__.py
Phantasmogenesis: > But minor scales shouldn't getRelativeMinor anyhow -- it doesn't make sense for a minor to have a parallel minor or a relative minor. They already are a form of minor.

I respectfully disagree on this point, as I think there are edge cases in which it would be helpful for users to be able to getRelativeMinor/getParallelMinor when they need to determine a key signature or simply not receive an error message when using these functions. 
Also, the basic MajorScale class, for example, allows getRelativeMajor much in the same way as the basic MinorScale class allows getRelativeMinor. For the sake of consistency, I think it is important to have these functions work for the non-natural minor scales as well. Or, perhaps, would you say that these other scales should not have such functionality either?

Nevertheless, I greatly appreciate the feedback and plan to continue working on making the improvements that you and jacobtylerwalls have mentioned sometime in the near future.

jacobtylerwalls: > there are edge cases in which it would be helpful for users to be able to getRelativeMinor/getParallelMinor when they need to determine a key signature or simply not receive an error message when using these functions.

I agree, we shouldn't bother trying to remove any of this functionality, it should just work as currently documented, and you have demonstrated it doesn't work for melodic/harmonic/jazz minor. I think this is neither here nor there, though; we should revert the new function definitions. Your patch works with just setting the attributes for relativeMajorDegree and relativeMinorDegree, nothing else needed.

music21/scale/__init__.py
Phantasmogenesis: `<music21.pitch.Pitch B4>`
`>>> sc.pitchFromDegree(1)  # scale degree 1 is treated as lowest`
`<music21.pitch.Pitch E4>`
`>>> sc = scale.JazzMinorScale()`

> explain what's happening here.
usually no need for comments in doc tests, write it as documentation just above.

To clarify, these comments are largely borrowed from other diatonic scales (mainly the harmonic minor) that I had used as a template. Should I rework the comments on these other scales as well?

music21/scale/__init__.py
Phantasmogenesis: In my latest commit, while addressing this feedback I removed the individual abstract minor scale classes so that all minor scales would instead be implemented directly within the AbstractDiatonicScale class. In making this change, I hope to increase consistency and user experience slightly, but recognize that there could be other concerns I am not aware of; as such, if y'all think I should leave the abstract harmonic/melodic/jazz minor classes, I can easily do so in a future commit.

music21/scale/__init__.py
```diff
@@ -842,58 +857,6 @@ def buildNetwork(self, mode=None):
         # might also set weights for tonic and dominant here
 
 
-class AbstractHarmonicMinorScale(AbstractScale):
-    '''
-    A true bi-directional scale that with the augmented
```
mscuthbert, line 847: We're losing two documented classes without a deprecation process.  Why are these being removed?

Phantasmogenesis: My thought was that it would be better to be able to create non-natural minor keys in one line, especially considering the fact that in the existing documentation, abstract scales are described as "rarely created or manipulated directly by most users." For example, with the changes in this most recent commit, in order to create a harmonic minor key, users would type:
`minorKey = key.Key(mode='harmonic minor')`
vs. what was previously required and seems unnecessary:
`minorKey = key.Key(mode='minor')`
`minorKey.abstract = AbstractHarmonicMinor()`
Additionally, as far as I can tell, no functionality that a user would manipulate is lost by moving the abstract harmonic/melodic/jazz minor scale classes entirely into the AbstractDiatonicScale class. However, this is my first PR _ever_; would it be better to leave those abstract classes?

music21/key.py
```diff
@@ -1078,7 +1078,7 @@ def deriveByDegree(self, degree, pitchRef):
         To use the harmonic form, change `.abstract` on the key to
         another abstract scale:
 
-        >>> minorKey.abstract = scale.AbstractHarmonicMinorScale()
+        >>> minorKey.abstract = key.Key(mode='harmonic minor').abstract
```
jacobtylerwalls, line 1081: This strikes me as a strange way of deriving a scale, i.e., creating a Key object, throwing it away, and then getting the .abstract. Either way we're exposing the user to the concept of abstract scales, so I'm not sure this is an improvement.

music21/scale/__init__.py
jacobtylerwalls: Smaller PRs are usually better for this very reason. This started as a bugfix and then added a proposal for a refactor. But that's two different motivations, and the bugfix probably could have been merged already were it isolated.

We would ordinarily wait a whole version to deprecate documented classes, with notice on the list and issuing warnings in the code for people still using the old pattern. For such an unclear benefit I'm -1. You can certainly wait to hear from Michael before changing the PR, though.

Thanks for the PR and welcome aboard 👍🏻

music21/scale/__init__.py
jacobtylerwalls: Right, because "where" is not delimiting certain parts of the world where the term is/is not meaningful.

music21/scale/__init__.py
jacobtylerwalls: hiya @Phantasmogenesis, absolutely no rush here or anything, and I hope I haven't sounded short, I am genuinely pleased to see more folks lend a hand to m21. It's what I was hoping for when I marked this as a good first issue!

So please do say if you're waiting on anyone to offer guidance or direct any part of the implementation. All the best, and thanks again, it will be great to get JazzMinor and the bugfix you identified into the system. 🥇

mscuthbert:

mscuthbert:
Just as an FYI -- I've been back for two years but I'm not sure where we were on this.  Thanks!

PR #1045, opened 2021-06-25
Fix repr() of scales without tonics

jacobtylerwalls:
Fixes this issue (and there are tonic-less uses like this in the test suite, for instance, calling `getPitches()` will assume a tonic of C4):

```
>>> sc = scale.MinorScale()
>>> sc
  File "/Users/jwalls/music21/music21/prebase.py", line 236, in __repr__
    strRepr = self._reprInternal()
  File "/Users/jwalls/music21/music21/scale/__init__.py", line 1381, in _reprInternal
    return f'{self.tonic.name} {self.type}'
AttributeError: 'NoneType' object has no attribute 'name'
```

The repr() was simplified in f58395e45b to just be the same as the .name when a tonic exists, so let's just DRY it out and grab the .name.

### DIFF
--- a/music21/scale/__init__.py
+++ b/music21/scale/__init__.py
@@ -1378,7 +1378,7 @@ class ConcreteScale(Scale):
             return ' '.join([self.tonic.name, self.type])
 
     def _reprInternal(self):
-        return f'{self.tonic.name} {self.type}'
+        return self.name
 
     # --------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


mscuthbert:
Does it still print the tonic on say a C major scale?  if so go ahead and merge.  Thanks!

jacobtylerwalls:
Covered 🥇 

https://github.com/cuthbertLab/music21/blob/a4594ae9e4c48d9c3f0a8e5edd6d1d47203d302d/music21/scale/__init__.py#L1444-L1447

PR #1046, opened 2021-06-30
Pin pylint==2.8.3

jacobtylerwalls:
see most recent PR run and https://github.com/PyCQA/astroid/issues/1080

edit: it's probably much simpler to simply disable pylint checks on the one method having trouble in text.py, come to think of it.

edit edit: with limited time before v7, how about we just also pin pylint until after v7

### DIFF
--- a/.github/workflows/pythonpylint.yml
+++ b/.github/workflows/pythonpylint.yml
@@ -23,8 +23,7 @@ jobs:
                   python -m pip install --upgrade pip
                   pip install -r requirements.txt
                   pip install IPython
-                  pip install pylint
-                  pip install astroid==2.5.7
+                  pip install pylint==2.8.3
             - name: Lint with pylint
               run: |
                   pylint -j0 music21
--- a/.github/workflows/pythonpylint.yml
+++ b/.github/workflows/pythonpylint.yml
@@ -24,6 +24,7 @@ jobs:
                   pip install -r requirements.txt
                   pip install IPython
                   pip install pylint
+                  pip install astroid==2.5.7
             - name: Lint with pylint
               run: |
                   pylint -j0 music21


jacobtylerwalls:
Amazing to see the crash patched in astroid so quickly. Now that we can get a full run on pylint 2.9.1, we still have about 85 messages to clean up or suppress. Still think pinning makes sense; no reason to rush these decisions before v7.

consider-using-dict-items
unnecessary-dict-index-lookup
use-maxsplit-arg
unused-private-member
arguments-renamed
deprecated-module
consider-using-from-import

mscuthbert:
> arguments-renamed
I'm fine with renaming arguments.  I need an argument against.  :-)

> consider-using-dict-items
I wonder if this will end up being like "consider-using-enumerate" which I believe is purposefully disabled, because I find the first example below cleaner than the second:

```
a = [...list_of_things...]
b = [...list_of_things_same_length_as_a...]

# 1
for i in range(len(a)):
     m = a[i]
     n = b[i]

# 2
for i, m in enumerate(a):
    n = b[i]
```
the latter doesn't show the parallel operations being done on `a` and `b` nearly so well.

jacobtylerwalls:
> the latter doesn't show the parallel operations being done on a and b nearly so well.

I get that, but if the index doesn't matter so much and is just being used to iterate through two iterables, there I would say the usage should be zip, or in 3.10, zip(strict=True).

Looking at the consider-using-dict-items violations, I don't think it's worth touching old code, there isn't much of a readability benefit unless the variables are renamed at the same time. If we just rename a lot of dict[k] to "value" we haven't accomplished much.

I'd probably want new contributions to follow it, though, as part of choosing good names.

jacobtylerwalls:
E.g. this would have been a _great_ opportunity for items() so that we could know what in the world packetStorage[trackId] consists of (presumably something like "trackPackets", but that ship has sailed at this point ⛵ 

https://github.com/cuthbertLab/music21/blob/cb7df6362f73d92c0fa803b9b5ee6ae8f295eb24/music21/midi/translate.py#L2448-L2450

PR #1049, opened 2021-06-30
Support and recommend `.sorted()` instead of `.sorted`

jacobtylerwalls:
As mentioned in v6.7 release notes.


mscuthbert (approved):
LGTM -- deemphasizing unsorted streams is pretty important.  This can have a small deprecation time.  The property version could be gone in v.8

Btw -- with the easy way of recurse[GetElementsByClass], I'm not as 100% sure that we need to go through the painful transition of `.flat` to `.flatten()` but still thinking about it.

PR #1051, opened 2021-07-01
Support mid-part instrument changes in musicxml export

jacobtylerwalls:
Needed this for a project and was faster to implement in music21 (hooray!)

[Spec](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/instrument/):

> The `<instrument>` element distinguishes between <score-instrument> elements in a `<score-part>`. If multiple `<score-instrument>` elements are specified in a `<score-part>`, there should be an `<instrument>` element for each note in the `<part>`. Notes that are shared between multiple `<score-instrument>`s can have more than one <instrument> element.

Playback respected in MuseScore, but perhaps not in Finale.

<s>This creates a TODO for #1033 -- when that implements `getInstrument()`, use that before falling back to a context search.</s> Done ✔️ 

I know ideally we would tackle import at the same time, but this is all I can commit to presently.


music21/musicxml/m21ToXml.py
```diff
@@ -2717,7 +2766,7 @@ def instrumentToXmlScoreInstrument(self, i):
         </score-instrument>
         '''
         mxScoreInstrument = Element('score-instrument')
-        mxScoreInstrument.set('id', str(i.instrumentId))
```
mscuthbert, line 2720: I don't understand this removal.  It just makes something break that has been working in the past in case someone sets the id to a number or None.

music21/musicxml/m21ToXml.py
```diff
@@ -2746,7 +2795,7 @@ def instrumentToXmlMidiInstrument(self, i):
         </midi-instrument>
         '''
         mxMidiInstrument = Element('midi-instrument')
-        mxMidiInstrument.set('id', str(i.instrumentId))
```
mscuthbert, line 2749: as above.

music21/musicxml/m21ToXml.py
```diff
@@ -3472,7 +3521,23 @@ def noteToXml(self, n: note.GeneralNote, noteIndexInChord=0, chordParent=None):
...
+                assert closest_inst in self.parent.instrumentStream
+            instance_to_use = self.parent.instrumentStream[type(closest_inst)].first()
+            if instance_to_use is None:
+                # exempt coverage, because this is only for safety/unreachable
+                raise MusicXMLExportException(f'Could not find {closest_inst} for note {n}'
+                    + 'in instrumentStream')  # pragma: no cover
+            mxInstrument = SubElement(mxNote, 'instrument')
+            assert instance_to_use.instrumentId, id(instance_to_use)
+            mxInstrument.set('id', instance_to_use.instrumentId)
+
```
mscuthbert, line 3540: Can this be pulled out into a separate routine and documented a bit more.

also I don't like bare-asserts in code because if it fails then someone just gets AssertionError and a random id with nothing they can do to fix it.

music21/musicxml/m21ToXml.py
```diff
@@ -6552,6 +6617,29 @@ def testFromSoundingPitch(self):
         self.assertEqual(len(root.findall('.//transpose')), 1)
         self.assertEqual(root.find('.//step').text, 'D')
 
+    def testMultipleInstruments(self):
```
mscuthbert, line 6620: what are the "Multiple Instruments" being tested?  I'm not really clear on this.

music21/musicxml/m21ToXml.py
```diff
@@ -6552,6 +6617,29 @@ def testFromSoundingPitch(self):
...
+        ])
+        p2 = stream.Part([
+            stream.Measure([instrument.Oboe(), note.Note(type='whole')]),
+            stream.Measure([instrument.Flute(), note.Note(type='whole')]),
+            stream.Measure([instrument.Oboe(), note.Note(type='whole')]),
+        ])
+        s = stream.Score([p1, p2])
+        scEx = ScoreExporter(s)
+        tree = scEx.parse()
+        self.assertEqual(len(tree.findall('.//score-instrument')), 4)
```
mscuthbert, line 6636: should we be testing which instruments were the 4 that are retained?

mscuthbert: This code is complex enough and I don't fully understand it, that I'd rather have the instrumentIds be A, B, C, D etc. rather than randomized (unless necessary) in order to be able to visualize what is happening here by looking at the whole Tree (or at least relevant parts of it).

music21/musicxml/partStaffExporter.py
```diff
@@ -243,6 +248,13 @@ def joinableGroups(self) -> List[StaffGroup]:
                 deduplicatedGroups.append(jg)
             permutations.add(containedParts)
 
+        # But forbid overlapping, spaghetti StaffGroups
+        joinable_components_list = flattenList(deduplicatedGroups)
+        if len(set(joinable_components_list)) != len(joinable_components_list):
+            warnings.warn(
+                MusicXMLWarning('Got overlapping StaffGroups; will not merge ANY groups.'))
+            return []
```
mscuthbert, line 256: this is good.

music21/musicxml/m21ToXml.py
```diff
@@ -1422,6 +1426,10 @@ def parse(self):
         self.scorePreliminaries()
 
         if s.hasPartLikeStreams():
+            # Prepopulate partExporterList so that joinable groups can be identified
+            # before attempting to identify and count instruments
+            self._populatePartExporterList()
+            self.joinedGroups = self.joinableGroups()
```
mscuthbert, line 1432: should the name `joinedGroups` be changed since they're not actually joined yet.  In case they're queried.

music21/musicxml/m21ToXml.py
```diff
@@ -2529,7 +2541,10 @@ def instrumentSetup(self):
         {4.0} <music21.instrument.BassClarinet 'Bass clarinet'>
         '''
         # get a default instrument if not assigned
-        self.instrumentStream = self.stream.getInstruments(returnDefault=True, recurse=True)
+        if self.parent is not None and id(self.stream) in self.parent.instrumentsByStream:
+            self.instrumentStream = self.parent.instrumentsByStream[id(self.stream)]
```
mscuthbert, line 2545: when would this take place?

mscuthbert: ah -- I get it from below.  Maybe a Note.  "Will be True if this is the second or subsequent PartStaff in a Group"  -- help people who are reading the code for a general case read it.

music21/musicxml/m21ToXml.py
```diff
@@ -2541,7 +2556,34 @@ def instrumentSetup(self):
...
+                    # where subsequent staffs entering this method will find and use it
+                    if self.parent:
+                        self.parent.instrumentsByStream[id(subseq_staff)] = self.instrumentStream
+            elif self.stream in joined_group:
+                # This stream was already (or will be) processed
+                # UNLESS there is a spaghetti case where
+                # this stream is second in groupB, but first in groupA,
+                # but PartStaffExporterMixin guarantees that won't happen
+                return
+
```
mscuthbert, line 2580: Call out to a separate routine to explain a little more what's happening and keep the general case simpler.

music21/musicxml/m21ToXml.py
```diff
@@ -2541,7 +2556,34 @@ def instrumentSetup(self):
...
+        seenInstrumentClasses = set()
         for thisInstrument in self.instrumentStream:
+            if type(thisInstrument) in seenInstrumentClasses:
+                continue
+            else:
+                seenInstrumentClasses.add(type(thisInstrument))
```
mscuthbert, line 2586: what if two instruments of the same class have different MIDI numbers or channels?

music21/musicxml/m21ToXml.py
```diff
@@ -2554,16 +2596,18 @@ def instrumentSetup(self):
...
-                self.parent.instrumentList.append(thisInstrument)
-            # force this instrument into this part
-            # meterStream is only used here if there are no measures
-            # defined in this part
-
-            if thisInstrument.instrumentId is None:
+            # Enforce uniqueness
+            if (thisInstrument.instrumentId is None
+                or (self.parent
+                    and thisInstrument.instrumentId in self.parent.instrumentIdList)):
```
mscuthbert, line 2602: maybe here's the place to check for numerical ids.

music21/musicxml/m21ToXml.py
```diff
@@ -2676,17 +2720,22 @@ def getXmlScorePart(self):
...
+        for inst in self.instrumentStream:
+            # only use the first instance of this class
+            if inst is not self.instrumentStream[type(inst)].first():
+                continue
+            if (inst.instrumentName is not None
+                    or inst.instrumentAbbreviation is not None
+                    or inst.midiProgram is not None):
+                mxScorePart.append(self.instrumentToXmlScoreInstrument(inst))
+
+        for inst in self.instrumentStream:
```
mscuthbert, line 2732: this looks like it's going through the stream twice.  Is there another way of making it clear that we're doing different code for the first instrument of each type in the stream, by iterating once over self.instrumentStream and getting the first of each one?

mscuthbert: Ah, I see there's ScoreInstrument and MidiInstrument -- perhaps a comment could draw attention to that.  Such as:

"The first instrument of each Class appears as a Score instrument."  And later "Subsequent instruments appear as MIDI instruments"

music21/musicxml/m21ToXml.py
```diff
@@ -3472,7 +3521,23 @@ def noteToXml(self, n: note.GeneralNote, noteIndexInChord=0, chordParent=None):
             for t in mxTieList:
                 mxNote.append(t)
 
-        # TODO: instrument
+        # instrument tags are necessary when there is more than one
```
mscuthbert, line 3524: when there is more than one -- on the same Note?  (That's a musicxml 4.0 addition) or more than one on the same musicxml Part (or music21 part?)

music21/musicxml/m21ToXml.py
```diff
@@ -3472,7 +3521,23 @@ def noteToXml(self, n: note.GeneralNote, noteIndexInChord=0, chordParent=None):
...
+        if self.parent is not None and len(self.parent.instrumentStream) > 1 and not n.isRest:
```
mscuthbert, line 3525: Does this mean that we'll be doing a getContextByClass for each note or chord being exported if there is more than one instrument in a score?  This might be much slower.

music21/musicxml/m21ToXml.py
```diff
@@ -3472,7 +3521,23 @@ def noteToXml(self, n: note.GeneralNote, noteIndexInChord=0, chordParent=None):
...
+            if chordParent:
```
mscuthbert, line 3526: Maybe:  

```
searchingObject = chordParent if chordParent else n
closest_inst = searchingObject.getContextByClass(Instrument, followDerivation=False)
```

to remove the redundancy.

music21/musicxml/m21ToXml.py
```diff
@@ -3472,7 +3521,23 @@ def noteToXml(self, n: note.GeneralNote, noteIndexInChord=0, chordParent=None):
...
+                closest_inst = chordParent.getContextByClass(Instrument, followDerivation=False)
+                assert closest_inst in self.parent.instrumentStream
+            else:
+                closest_inst = n.getContextByClass(Instrument, followDerivation=False)
+                assert closest_inst in self.parent.instrumentStream
+            instance_to_use = self.parent.instrumentStream[type(closest_inst)].first()
```
mscuthbert, line 3532: As before -- if the first instance of each Instrument were cached, this would be both faster and cleaner:

```
instance_id_to_use = self.getScoreInstrumentId(closest_inst)
```

or something like that.

mscuthbert (changes requested):
I think I need a little more context on in what way this will be supporting "multiple" instruments in musicxml export -- can't we already do that?  Or is there a problem with multiple instruments within a PartStaff-group or the same instrument appearing in different parts?  

There are some naming things that I'm unclear on.  <s>I wouldn't expect the main parsing of a score to take place in a private function `_populatePartExporterList` rather than `parse()`, but now it does!  I would think that populatePartExporterList would just put the PartExporters into the self.partExporterList, not actually run parsing.</s>  Never mind!  I read the `pp.parse()` line as added rather than removed.  My bad.

There are places where you can get a bit of a performance boost but also a lot more clarity for someone who needs to edit this later (usually me) if more helper routines and subroutines are used.

music21/musicxml/m21ToXml.py
jacobtylerwalls: More than one anywhere on the musicxml `<part>`

music21/musicxml/m21ToXml.py
jacobtylerwalls: good idea!

music21/musicxml/m21ToXml.py
jacobtylerwalls: Currently music21 only exports the first instrument in a stream.Part to musicXML. There are a handful of code comments referring to this, but nowhere in docs, I don't think.  I'll point out the existing implementation with a couple comments in this diff.

jacobtylerwalls: i.e., you cannot export a mid-part Instrument change ("to Oboe" etc.)

music21/musicxml/m21ToXml.py
jacobtylerwalls: Sure, I'll give this another look and make it as easy as possible to understand. This is a score for two woodwind players both doubling on flute and oboe. They both switch to flute and then back to oboe. There are six m21 instruments to represent this, but the `<score-instrument>` tags need just four, since no musicXML `<part>` needs two oboes in it, etc., unless there is a patch change/MIDI instrument change. Which kind of gets to your other question about whether we should deduplicate instruments that differ in MIDI program but resolve to same m21 instrument. I'll have another look.

music21/musicxml/m21ToXml.py
```diff
@@ -2676,17 +2720,22 @@ def getXmlScorePart(self):
 
         # TODO: part-abbreviation-display
         # TODO: group
-
-        # TODO: unbounded...
-        i = self.firstInstrumentObject
```
jacobtylerwalls, line 2681: This is the only instrument object that was used in the existing implementation

music21/musicxml/m21ToXml.py
```diff
@@ -2676,17 +2720,22 @@ def getXmlScorePart(self):
...
-
-        if (i.instrumentName is not None or i.instrumentAbbreviation is not None
-                or i.midiProgram is not None):
-            mxScorePart.append(self.instrumentToXmlScoreInstrument(i))
-
-        # TODO: midi-device
-        if i.midiProgram is not None:
-            mxScorePart.append(self.instrumentToXmlMidiInstrument(i))
```
jacobtylerwalls, line 2689: `i` was the first instrument only

music21/musicxml/m21ToXml.py
```diff
@@ -2504,12 +2519,9 @@ def parse(self):
 
     def instrumentSetup(self):
         '''
-        sets self.instrumentStream and self.firstInstrumentObject for the stream,
+        Sets self.instrumentStream and self.firstInstrumentObject for the stream,
         checks for a unique midiChannel and then blocks it off from future use.
 
-        Note that there's a deficiency currently that only the first instrument is fully
-        converted.
```
jacobtylerwalls, line 2511: "Only the first instrument..."

jacobtylerwalls (commented):
Pardon if this was not your question, but I took you to be asking whether m21 supported exporting multiple instruments out of a generic `stream.Part`, so here are the parts of the existing implementation.

music21/musicxml/m21ToXml.py
jacobtylerwalls: Happy to remove. I think it was because `.set()` handles that--in fact, better in case of None, see:

```
>>> mxArt = EL('<spiccato placement="above"/>')
>>> mxArt.set('lucky_number', 3.14159)
>>> mxArt.get('lucky_number')
3.14159
>>> mxArt.set('lucky_number', None)
>>> mxArt.get('lucky_number') is None
True
>>> mxArt.set('lucky_number', str(None))
>>> mxArt.get('lucky_number')
'None'
```

jacobtylerwalls: Actually, the None case may have been why I did this edit, I can't remember at the moment.

music21/musicxml/m21ToXml.py
jacobtylerwalls: We should probably hold this until we handle note storage on `NotRest` in #1033, then I can replace this block with the `getStoredInstrument` lookup which will be cached.

jacobtylerwalls:
OH I see my poor phrasing now -- my mind is too stuck in the m21 ontology to see the failure to say instrument "CHANGE". In m21 instrument changes are represented not with inserting "changes" per se, as in MuseScore/Finale ("To Oboe") etc., you just insert a new oboe at some offset. So in m21 "multiple instruments in a stream.Part" is how you represent _instrument changes_.

music21/musicxml/m21ToXml.py
mscuthbert: Supe.  I did not know that `set` took an any.

mscuthbert (commented):
Okay -- I've reviewed everything here and #1033 is in.  Now we just need to make it so that the instrument search isn't extremely slow.

I actually think that given the Tree-based context search, it won't be that slow BUT note.getInstrument() and storedInstrument is still not tested or doc'd so those need to go in in order to be sure.  But I'll bet that the timing addition of the context search is much less than you might think.

mscuthbert (dismissed):

music21/musicxml/m21ToXml.py
jacobtylerwalls: Same implementation as getInstrument(), but with the ability to store a different instrument, so I was planning on using that now that #1033 is in:

https://github.com/cuthbertLab/music21/blob/fe2586ad956fa68e048b979bba2d15eb6369bf9c/music21/note.py#L1205-L1211

music21/musicxml/m21ToXml.py
jacobtylerwalls: think this is better now -- two loops purely for the sake of writing the tags in the right order. these are tiny streams of just a few elements. (unless you have instrument changes every single note, which we aren't optimizing for)

music21/musicxml/m21ToXml.py
jacobtylerwalls: I think we should iterate on that in future versions. Not least because I would want to make sure I was getting the musicxml document prepped correctly for that.

music21/musicxml/m21ToXml.py
jacobtylerwalls: I see the point that it needs to be tested first.

music21/musicxml/m21ToXml.py
jacobtylerwalls: thanks for tackling the refactor!

music21/musicxml/m21ToXml.py
jacobtylerwalls: Yeah, I think the risk of "None" being written out is worth the change.

jacobtylerwalls: Nononono. I'll revert this. Better to write out "None" than to crash:

```
        >>> i.instrumentId = None
        >>> PEX = musicxml.m21ToXml.PartExporter()
        >>> mxMidiInstrument = PEX.instrumentToXmlMidiInstrument(i)
        >>> PEX.dump(mxMidiInstrument)
```
```
File "/Users/jwalls/music21/music21/musicxml/m21ToXml.py", line 2847, in __main__.PartExporter.instrumentToXmlMidiInstrument
Failed example:
    PEX.dump(mxMidiInstrument)
Exception raised:
      File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/xml/etree/ElementTree.py", line 1052, in _escape_attrib
        if "&" in text:
    TypeError: argument of type 'NoneType' is not iterable

    During handling of the above exception, another exception occurred:

      File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/doctest.py", line 1336, in __run
        exec(compile(example.source, filename, "single",
      File "<doctest __main__.PartExporter.instrumentToXmlMidiInstrument[9]>", line 1, in <module>
        PEX.dump(mxMidiInstrument)
      File "/Users/jwalls/music21/music21/musicxml/m21ToXml.py", line 791, in dump
        return helpers.dump(obj)
      File "/Users/jwalls/music21/music21/musicxml/helpers.py", line 69, in dump
        print(dumpString(obj))
      File "/Users/jwalls/music21/music21/musicxml/helpers.py", line 46, in dumpString
        xStr = ET.tostring(xmlEl, encoding='unicode')
      File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/xml/etree/ElementTree.py", line 1109, in tostring
        ElementTree(element).write(stream, encoding,
      File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/xml/etree/ElementTree.py", line 748, in write
        serialize(write, self._root, qnames, namespaces,
      File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/xml/etree/ElementTree.py", line 906, in _serialize_xml
        v = _escape_attrib(v)
      File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/xml/etree/ElementTree.py", line 1075, in _escape_attrib
        _raise_serialization_error(text)
      File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/xml/etree/ElementTree.py", line 1029, in _raise_serialization_error
        raise TypeError(
    TypeError: cannot serialize None (type NoneType)
```

jacobtylerwalls: I think the changes in this PR already randomize when we encounter None, but adding the safety back is good.

music21/musicxml/m21ToXml.py
jacobtylerwalls: My bad, but I didn't follow this comment.

music21/musicxml/m21ToXml.py
jacobtylerwalls: AH, you already pushed that change. Got it. Just needs a doctest.

music21/musicxml/m21ToXml.py
```diff
@@ -1404,7 +1404,7 @@ def __init__(self, score: Optional[stream.Score] = None, makeNotation: bool = Tr
 
         self.partExporterList: List['PartExporter'] = []
 
-        self.joinedGroups: List['StaffGroup'] = []
+        self.groupsToJoin: List['StaffGroup'] = []
```
jacobtylerwalls, line 1407: This attribute was introduced during v7 dev (cc09afcde9283538d74aef6bb5adf2fc7130c432) so that's why we are taking the opportunity to rename before final release to avoid implying the joining was already done at any given point.

music21/musicxml/m21ToXml.py
```diff
@@ -3691,8 +3691,9 @@ def setNoteInstrument(self,
             return
 
         searchingObject = chordParent if chordParent else n
-        closest_inst = searchingObject.getInstrument()
-        if closest_inst is None:
+        closest_inst = searchingObject.getInstrument(returnDefault=True)
+        if closest_inst is None:  # pragma: no cover
+            # unreachable; purely for extra safety
```
jacobtylerwalls: Should we add a `returnDefault=` argument to match `Stream.getInstrument()`. That's how I ended up misinterpreting this.

mscuthbert (approved):
Thanks for being so patient.  Merging.  This is huge and difficult code but no longer difficult to understand.

jacobtylerwalls:
Thank you for helping refactor it!

Issue #1036, opened 2021-06-14
Empty `<text/>` on lyric tag imports as .text = None and causes TypeError on access


jacobtylerwalls:
**music21 version**

dev

**Problem summary**
When trying to horizontally align rows of lyrics/annotations, it's helpful to have an empty value for a dummy lyric. I've resorted to using `' '`, but I keep accidentally using `''` and causing TypeErrors. `None` is no good because `.addLyric()` turns `None` -> `"None"`.

**Steps to reproduce**

```
>>> n = note.Note()
>>> n.addLyric('')  # want to insert a dummy lyric somehow
>>> n.lyrics  # so far so good
[<music21.note.Lyric number=1 syllabic=single text=''>]
>>> out = n.write()  # writes empty <text/> tag
>>> inn = converter.parse(out)  # imports as .text = None
>>> inn.recurse().notes[0].lyrics
[<music21.note.Lyric number=1 identifier='1' syllabic=single>]
>>> inn.recurse().notes[0].lyric  # can't handle .text = None
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/music21/note.py", line 587, in _getLyric
    return '\n'.join(allText)
TypeError: sequence item 0: expected str instance, NoneType found
```

**Suggested fix**
We should decide if `None` or `''` represents the empty value and then fix the importer or possibly `_getLyric()`.

**More information**
It might be possible I'm supposed to be aligning lyrics with the number attribute instead of trying to insert dummy empty lyrics. Not certain.

jacobtylerwalls:
Crash fixed in #1053

jacobtylerwalls:
Agree with Michael that we can live with two values for empty lyrics for now, as there is just as much ambiguity in musicxml space as music21 space (and insisting on one or the other might catch some folks by surprise): https://github.com/cuthbertLab/music21/pull/1053#discussion_r663293971

PR #1053, opened 2021-07-02
Fix common Lyric/Text problems

mscuthbert:


music21/musicxml/xmlToM21.py
```diff
@@ -4363,9 +4363,11 @@ def findM21VoiceFromXmlVoice(self, mxVoice=None):
...
             useVoice = self.lastVoice
             if useVoice is None:
-                environLocal.warn('Cannot put in an element with a missing voice tag when '
-                                  + 'no previous voice tag was given.  Assuming voice 1... '
-                                  )
+                # # Commenting this out -- this is very common with SmartScore OMR software.
+                environLocal.printDebug(
+                    'Cannot put in an element with a missing voice tag when '
+                    + 'no previous voice tag was given.  Assuming voice 1... '
+                )
```
jacobtylerwalls, line 4370: Is this still a problem after #1030 ? If there's another failure case, we could look into why.

mscuthbert: Ah!  I thought that this might be obsolete -- will take it out.

This was a left-over uncommitted set of changes from a long time ago that I wanted to get in so I had a clean working directory for #1054

music21/note.py
```diff
@@ -584,7 +584,7 @@ def _getLyric(self) -> Optional[str]:
             return None
 
         allText = [ly.text for ly in self.lyrics]
-        return '\n'.join(allText)
+        return '\n'.join([t for t in allText if t is not None])
```
jacobtylerwalls, line 587: This solves the crashy part of #1036 but would still be worth a small think over how empty values should roundtrip, i.e. as empty string or None, because right now = inconsistent. Just mentioning in case you're feeling opinionated.

jacobtylerwalls: (in a future PR)

mscuthbert: in this case, it's a short-hand so it's not a big deal if it doesn't roundtrip perfectly, since it also doesn't preserve which lyric each syllable (or composite) comes from.

PR #1056, opened 2021-07-04
Reclassify first instrument in MusicXML import

### LINKED ISSUE #988: Create more specific instruments from MusicXML import

jacobtylerwalls:
Reclassify first instrument of part in MusicXML import

**Previously**
Previously, every first instrument was imported as an `instrument.Instrument`. The test updated here to check for oboe, clarinet, and horn instruments fails on master.

**Now**
Rely on the MIDI program rather than, say, a fromString() operation on the name, although this could be added later (for instance to change Piano to Soprano when a file used piano sounds rather than Choir Aahs).

Fixes #988

**Future**
import all instruments from part


music21/_version.py
```diff
@@ -42,7 +42,7 @@
 Changing this number invalidates old pickles -- do it if the old pickles create a problem.
 '''
 
-__version_info__ = (7, 0, 7)  # can be 4-tuple: (7, 0, 5, 'a2')
+__version_info__ = (7, 0, 7, 'b1')  # can be 4-tuple: (7, 0, 5, 'a2')
```
mscuthbert, line 45: will need to be changed to merge w/ other recent merges.

mscuthbert: During Alpha/Beta phases, it's okay not to update this immediately.  Only when released.  Or you know it'll break everything for everyone.

music21/converter/subConverters.py
```diff
@@ -551,7 +551,7 @@ class ConverterTextLine(SubConverter):
 
     >>> s = corpus.parse('bwv66.6')
     >>> s.measures(1, 4).show('textline')
-    "{0.0} <music21.stream.Part Soprano> / {0.0} <music21.instrument.Instrument '... 1'>..."
+    "{0.0} <music21.stream.Part Soprano> / {0.0} <music21.instrument.Piano '... 1'>..."
```
mscuthbert, line 554: I'll get to this below...

music21/musicxml/xmlToM21.py
```diff
@@ -1586,8 +1587,30 @@ def _adjustMidiData(mc):
...
+            mxMidiProgram = mxMIDIInstrument.find('midi-program')
+            if textStripValid(mxMidiProgram):
+                try:
+                    i = instrument.instrumentFromMidiProgram(_adjustMidiData(mxMidiProgram.text))
+                except instrument.InstrumentException as ie:
+                    warnings.warn(MusicXMLWarning(ie))
+                    i = instrument.Instrument()
+                seta(i, mxMIDIInstrument, 'midi-channel', transform=_adjustMidiData)
+        if i is None:
+            i = instrument.Instrument()
```
mscuthbert, line 1612: Seeing all the Sopranos/Altos/etc. with `music21.instrument.Instrument` tags makes me think that something's missing here.  Some possibilities:

We should be reclassifying based on the Instrument name in preference to the midi-program, if possible.  That would solve the Sopranos, etc. (we might need to create some new Instrument objects for Soprano, Alto, Tenor, Baritone, Bass though).  Or we can special-case midi-program 0, unless piano is in the text OR max-staves >= 2.   MIDI 0 is often used just to mean I didn't bother to specify the instrument.

For me, seeing a mislabeled "Piano" on things that are clearly not pianos is worse than a generic Instrument. tag.

Does MIDI-Program ever get set in case no instrument can be found?  I think there are some midi-programs that don't map to one of our instruments [on purpose] such as FX-Goblin or TremoloStrings, and we should preserve the number, even if we can only get i = instrument.Instrument()

music21/musicxml/xmlToM21.py
```diff
@@ -1874,8 +1881,9 @@ def updateTransposition(self, newTransposition):
                 # no need for a change of instrument
                 pass
                 # environLocal.warn('Put trans on active instrument')
-            else:
-                # We have an activeInstrument, so this change of transposition
+            elif self.activeInstrument.transposition != newTransposition:
```
mscuthbert, line 1884: Thanks!

music21/stream/base.py
```diff
@@ -4423,7 +4423,7 @@ def template(self,
         >>> sopr = b.parts[0]
         >>> soprEmpty = sopr.template()
         >>> soprEmpty.show('text')
-        {0.0} <music21.instrument.Instrument 'P1: Soprano: Instrument 1'>
+        {0.0} <music21.instrument.Piano 'P1: Soprano: Instrument 1'>
```
mscuthbert, line 4426: This is what I was referring to as the problem.

mscuthbert (changes requested):
One-staff pianos are unlikely to be correct -- need to look at instrument name a bit more to get more plausible instruments.

music21/_version.py
jacobtylerwalls: Yes and this change would actually take the version number backward, so needs to be backed out.

music21/musicxml/xmlToM21.py
jacobtylerwalls: > For me, seeing a mislabeled "Piano" on things that are clearly not pianos is worse than a generic Instrument. tag.

I was thinking we could iterate on this. Just wanted to get a patch going for discussion to get the implementation agreed on. Yes, we could special case MIDI number 0 (1 in mxml).

Our fromString methods handle all the easy cases, but it seems more discrete and reliable to privilege the MIDI data if we can, but yes, absolutely special case 0 to go to the instrument name instead.

In terms of being worse than a generic instrument, it _would_ get worse once we do more `if isinstance(inst, instrument.UnpitchedPercussion)`, etc., which I think we already do (or should be doing) for MIDI export.

jacobtylerwalls: > Does MIDI-Program ever get set in case no instrument can be found? I think there are some midi-programs that don't map to one of our instruments [on purpose] such as FX-Goblin or TremoloStrings, and we should preserve the number, even if we can only get i = instrument.Instrument()

They all have fallbacks like `Sampler` #757 and even the generic fallback would make `i` not None; `if i is None` is just paranoia to catch if `mxMidiInstrument` is None (encoding error?) or if `textStripValid` rejected some garbage -- clearly a comment is in order!

jacobtylerwalls: > a fromString() operation on the name, although this could be added later (for instance to change Piano to Soprano when a file used piano sounds rather than Choir Aahs)

music21/midi/translate.py
```diff
@@ -3711,7 +3711,6 @@ def procCompare(mf_inner, match_inner):
                  (0, 'SET_TEMPO', None),
                  (1024, 'END_OF_TRACK', None),
                  (0, 'SEQUENCE_TRACK_NAME', None),  # Music track
-                 (0, 'PROGRAM_CHANGE', None),
```
jacobtylerwalls, line 3714: This looked like an improvement (same message duplicated after the pitch bend message). Must be doing a better job of deduplicating generic instruments?

Issue #377, opened 2019-03-16
rename TextExpressions.positionPlacement to .placement


mscuthbert:
Per JoeW on 15-Mar-2019 on music21list.

mscuthbert:
should probably be in Style?

PR #1057, opened 2021-07-07
add a placement attribute for MetronomeMark 

gesellkammer:
This PR adds a placement' attribute for MetronomeMark. Without that some musicxml applications (like musescore) misinterpret this and place metronome marks below the staff

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -5331,7 +5331,7 @@ class MeasureExporter(XMLExporterBase):
                 te.offset = ti.offset
                 unused_mxDirectionText = self.textExpressionToXml(te)
 
-        if ti.placement is not None:
+        if hasattr(ti, 'placement') and ti.placement is not None:
             assert ti.placement in {'above', 'below'}
             mxDirection.set('placement', ti.placement)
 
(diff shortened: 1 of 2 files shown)


jacobtylerwalls:
Thanks @gesellkammer for identifying this missing feature and providing a patch.

I had a quick look, and I think the root cause is that `MetronomeMark` lacks `.positionPlacement` because it doesn't inherit from `TextExpression`. In #377 we want to rename that to `.position` anyway. I mention this merely to point out I think you can just simply create the attribute and no changes to the exporter will be necessary, since the exporter already looks for `.positionPlacement` (see `placeInDirection()`). I think there _will_ be a change needed for the importer, though, something like:

```diff
diff --git a/music21/musicxml/xmlToM21.py b/music21/musicxml/xmlToM21.py
index 65a7d3725..3e443cbb9 100644
--- a/music21/musicxml/xmlToM21.py
+++ b/music21/musicxml/xmlToM21.py
@@ -4809,6 +4809,8 @@ class MeasureParser(XMLParserBase):
         elif tag == 'metronome':
             mm = self.xmlToTempoIndication(mxDir)
             # SAX was offsetMeasureNote; bug? should be totalOffset???
+            _setAttributeFromAttribute(mm, mxDirection,
+                            'placement', 'positionPlacement')
             self.insertCoreAndRef(totalOffset, staffKey, mm)
             self.setEditorial(mxDirection, mm)
```

If you're up for it, it would be great to do #377 at the same time.

gesellkammer:
Thanks for the tip. I will close this and send a new PR

gregchapman-dev:
Thanks for finding this!  I was tearing my hair out trying to figure out why my humdrum importer was making badly placed MetronomeMarks.

PR #1058, opened 2021-07-07
add positionPlacement to MetronomeMark

gesellkammer:
As suggested, this PR adds a positionPlacement to MetronomeMark. It also enables the corresponding import from xml, converts xml placement to positionPlacement.

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -6603,11 +6603,16 @@ class Test(unittest.TestCase):
         c.useSymbol = False
         f = repeat.Fine()
         mm = tempo.MetronomeMark(text='Langsam')
+        mm.positionPlacement = 'above'
         s.measure(1).storeAtEnd([c, f, mm])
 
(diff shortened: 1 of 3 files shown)


jacobtylerwalls:
Thanks, this is getting closer. I added a test, and I believe it shows the new implementation is still not quite there (isn't position an attribute on `<direction>` rather than `<metronome>`? I think the original suggestion does work, though.

gesellkammer:
I don't really know how to solve the tranlation of placement inside direction to a positionPlacement in MetronomeMark, since the implementation in music21 (the class structure) is the actual problem.

jacobtylerwalls:
I'm inclined to merge, since it's a trivial change with both import and export tested. If there's a request for more documentation we can handle it in #377 --> another reason to get this in so that any fix for #377 can be based on top of this change.

Appreciate the PR, @gesellkammer !

mscuthbert:
this is good -- but at some point positionPlacement should be moved over to Style for all Expressions, etc.

PR #1059, opened 2021-07-09
Rename `positionPlacement` to `placement` on Dynamic, TextExpression, and MetronomeMark

### LINKED ISSUE #377: rename TextExpressions.positionPlacement to .placement

jacobtylerwalls:
Fixes #377 

Also, document the attribute for the first time.

cc @gesellkammer
cc @gregchapman-dev

### DIFF
--- a/music21/dynamics.py
+++ b/music21/dynamics.py
@@ -212,6 +212,7 @@ class Dynamic(base.Music21Object):
             >>> d.englishName
             'very soft'
             ''',
+        'placement': "Staff placement: 'above', 'below', or None.",
     }
 
     def __init__(self, value=None):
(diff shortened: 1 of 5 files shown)


mscuthbert:
This is good.  Thanks!

mscuthbert:
Maybe move to Style at some point, since I don't think that where it's placed is important for anything except visual output.   But maybe wait until we can do Literal typing or a StrEnumOrNone for 'above', 'below', or None  (None is better as something like 'UNSPECIFIED')

jacobtylerwalls:
What about setting the default to "above" for now? For MIDI files, etc., where we insert defaults. Looks a little silly below the staff. Or would you prefer the default not be set on the object but merely defaulted during musicxml export?

PR #1060, opened 2021-07-10
`ElementSearch` enum for `getContextByClass` and related deprecations

### LINKED ISSUE #1028: `getContextByClass()` doesn't find elements positioned in self at 0.0 by default

jacobtylerwalls:
Fixes #1028 with documentation changes.

Adds `ElementSearch` enum and cleanup for `getContextByClass()`.

Various v. 7 deprecations per release notes in v.6.7.0:
 - `getClefs()`
 - `getKeySignatures()`
 - `makeChords()`
 - `extendDurationsAndGetBoundaries()`
 - `tree.toStream.chordified()`
 - `tree.toStream.partwise()`
 - `VoiceLeading.color`
 - `isClassOrSubclass()`

### DIFF
--- a/music21/base.py
+++ b/music21/base.py
@@ -60,7 +60,7 @@ from typing import (
 
 from music21.sites import SitesException
 from music21.sorting import SortTuple, ZeroSortTupleLow, ZeroSortTupleHigh
-from music21.common.enums import OffsetSpecial
+from music21.common.enums import ElementSearch, OffsetSpecial
 from music21.common.numberTools import opFrac
 from music21.common.types import OffsetQL, OffsetQLIn
 from music21 import style  # pylint: disable=unused-import
@@ -1158,7 +1158,7 @@ class Music21Object(prebase.ProtoM21Object):
         self,
         className,
         *,
-        getElementMethod='getElementAtOrBefore',
+        getElementMethod=ElementSearch.AT_OR_BEFORE,
         sortByCreationTime=False,
         followDerivation=True,
         priorityTargetOnly=False,
@@ -1237,14 +1237,16 @@ class Music21Object(prebase.ProtoM21Object):
         need a flat representation, the caller needs to be the source Stream,
         not its Sites reference.
 
-        The `getElementMethod` is a string that selects which Stream method is
-        used to get elements for searching. These strings are accepted:
+        The `getElementMethod` is an enum value (new in v7) from
+        :class:`~music21.common.enums.ElementSearch` that selects which
+        Stream method is used to get elements for searching. (The historical form
+        of specifying a string is also accepted):
 
         *    'getElementBefore'
         *    'getElementAfter'
         *    'getElementAtOrBefore' (Default)
         *    'getElementAtOrAfter'
-        *    'all'
+        *    'all' (new in v. 7)
 
         The "after" do forward contexts -- looking ahead.
 
@@ -1261,9 +1263,12 @@ class Music21Object(prebase.ProtoM21Object):
         >>> s.getContextByClass(clef.Clef, getElementMethod='getElementAtOrAfter')
         <music21.clef.BassClef>
 
-        This can be remedied by removing the offset constraint, that is, searching
-        all elements contained at some given level of the hierarchy, no matter
-        their temporal position.
+        This can be remedied by explicitly searching by offsets:
+
+        >>> s.getContextByClass(clef.Clef, getElementMethod='getElementAtOrBeforeOffset')
+        <music21.clef.BassClef>
+
+        Or by not limiting the search by temporal position at all:
 
         >>> s.getContextByClass(clef.Clef, getElementMethod='all')
         <music21.clef.BassClef>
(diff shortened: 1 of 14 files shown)


music21/base.py
```diff
@@ -1278,6 +1289,29 @@ def getContextByClass(
...
+        <music21.clef.BassClef>
+
+        This can be remedied by explicitly searching by offsets:
+
+        >>> s.getContextByClass(clef.Clef, getElementMethod='getElementAtOrBeforeOffset')
+        <music21.clef.BassClef>
+
+        Or by not limiting the search by temporal position at all:
+
+        >>> s.getContextByClass(clef.Clef, getElementMethod='all')
```
mscuthbert, line 1312: These should be updated to the new Enum, along with the doc on line 1293.

mscuthbert: (we can test the string version in a unittest, but better to get it out of the doc.

music21/stream/tests.py
```diff
@@ -1846,11 +1836,9 @@ def testContextNestedC(self):
         self.assertEqual(s2.elementOffset(s1), 0.0)
         self.assertRaises(sites.SitesException, s2.elementOffset, s1.flat)
 
-        # getContextByClass will not work; the clef is in s2; its not in a context of s2
-        post = s2.getContextByClass(clef.Clef)
-        self.assertIsNone(post)
+        post = s2.getContextByClass(clef.Clef, getElementMethod='all')
```
mscuthbert, line 1839: use Enum here

mscuthbert (commented):
Really like this.  Thanks for a TON of work.

Now though I'm getting second thoughts about deprecating `isClassOrSubclass()`.  I was thinking only of the `if clef.Clef in n.classSet` type which seemed to be just as good or better than 'if n.isClassOrSubclass(clef.Clef)`,  but the `if not n.classSet.isdisjoint([clef.Clef, note.Note])` form seems much harder for novices to grasp than `n.isClassOrSubclass([clef.Clef, note.Note])` -- and also requires doing something different than `if not n.classSet.isdisjoint(clef.Clef)` which will raise a not-an-iterable problem.  (Which is the same problem that comes from doing `if n.classSet.intersection([clef.Clef, note.Note])` which is probably better since its return value can be ignored).  

So I'm afraid right now I think that isClassOrSubclass() might get a reprieve

jacobtylerwalls:
> Really like this. Thanks for a TON of work.
Now though I'm getting second thoughts about deprecating isClassOrSubclass().

Sure, take some time. I didn't do this all this summer, I had the easy cleanup bits done a long time ago and it rustled up the interesting questions (that then made work on #1040 a lot easier because I knew we planned to get rid of `extendDurationsAndGetBoundaries()`...)

jacobtylerwalls:
> The multiple class invocation suddenly seems very complex when it appears inside code.

I had a thought; see how it strikes you.

On Python 3.10b4:
```
>>> timeit("'NotRest' in r.classes", setup='from music21 import note; r = note.Rest()')
0.29005374199914513
>>> timeit("'NotRest' in r.classSet", setup='from music21 import note; r = note.Rest()')
0.21796941299544415
>>> timeit("note.NotRest in r.classSet", setup='from music21 import note; r = note.Rest()')
0.24427016799745616
>>> timeit("r.isClassOrSubclass((note.NotRest,))", setup='from music21 import note; r = note.Rest()')
0.4091015360027086
>>> timeit("r.isClassOrSubclass(('NotRest',))", setup='from music21 import note; r = note.Rest()')
0.35828373900585575
>>> timeit("isinstance(r, note.NotRest)", setup='from music21 import note; r = note.Rest()')
0.12698115999955917
```

`isinstance()` is the clear winner: low maintenance, Pythonic, fast, we don't have to maintain anything--and it handles multiple classes (`[note.Note, note.Chord]`) -- but the value of having the m21 properties like `.classes` and `.classSet`, is, I take it, mainly for library developers (meaning, you) who would like to design permissive APIs that don't crash when given strings instead of class names to be friendly to beginners. Plus I think you said at one point it was faster than `isinstance` back in the day.

So, I'm thinking:

Myke's students -- could be taught to prefer `isinstance(className)` but can use `my_string in el.classes` to save keystrokes

most folks -- should just use `isinstance(classList)`

Myke's work on m21 -- readers of the code could tolerate looking at `not el.classSet.isdisjoint(classList)` with the understanding that it's only here for backwards compatibility and to be as permissive as possible. In fact, making it look cumbersome and ugly might remind folks "oh right, I should just tighten up my code to use `isinstance(classList)`." Maybe the docs for classSet could say something to that effect, that you probably only need it if you're at an interactive prompt and the saved keystrokes matter.

jacobtylerwalls:
Not too make too much out of nothing, but just because the thought just sprang up:  if we're talking about a user reading library code to figure out how to use the library, `not el.classSet.isdisjoint(...` might look verbose, but there is also a verbosity to having five or six ways to get class membership. There's a kind of API-surface area problem where a user thinks, "I have to go read each set of docs to make sure I'm using the right one." I'm sure I would probably fail a closed-book pop quiz over what all these methods accept. Not a big deal either way, but just mentioning if trying to forecast user understanding barriers.

mscuthbert:
it's amazing how much faster `isinstance` has become.  In Python 2 it was between 1 and 2 orders of magnitude slower than before -- it's the main reason why the whole 'classes' structure exists.

Now, I won't be removing it completely because it's absolutely necessary in music21j (because of minification and the problem of having more than one copy of the library loaded because of node_modules), so it should be there as a parallel interface, but I'm happy to think more about it.  You're right that outside of the library it is easy to use `isinstance()` instead and that what I've encouraged.

We can probably slate `.isRest`, `.isStream` and `.isNote` etc. for removal in v8 as well, since they were just there to get around the formerly huge isinstance bottlenecks from Python 2.

Now I'm curious when they sped this up.

mscuthbert:
I wonder also if inside stream.py,  `from music21 import Stream as S2;  s = S2(); isinstance(s, Stream)` still reports false because one was defined in the module and the other was imported from disk?  This used to be another huge Python headache.

mscuthbert:
Okay, I'm merging -- we'll leave any more class discussion for Version 8 -- it's August 7, and I need a v7 release with whatever we've gotten in very soon.  Other features will need to wait to v8 if they're backwards incompatible or v7.2 if they just add new features.

PR #1061, opened 2021-07-12
Handle more expressionIsInferred toggling on Durations

jacobtylerwalls:
#908 didn't handle toggling `.expressionIsInferred` back to False when setting type/dots after construction.

No change when setting tuplets, since part of the point is that for numeric durations such as 4/3, they still bear .expressionIsInferred = True even when expressed with tuplets.

While here, I noticed that the attribute should likely be set True when setting numeric qL on unlinked durations, not just on linked durations.

### DIFF
--- a/music21/duration.py
+++ b/music21/duration.py
@@ -2593,6 +2593,7 @@ class Duration(prebase.ProtoM21Object, SlottedObjectMixin):
         for i, dt in enumerate(self._components):
             self._components[i] = durationTupleFromTypeDots(dt.type, value)
         self._quarterLengthNeedsUpdating = True
+        self.expressionIsInferred = False
         self.informClient()
 
     @property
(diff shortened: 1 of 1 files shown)


music21/duration.py
```diff
@@ -2786,6 +2787,7 @@ def _getQuarterLength(self) -> OffsetQL:
         return self._qtrLength
 
     def _setQuarterLength(self, value: OffsetQLIn):
+        self.expressionIsInferred = True
```
mscuthbert, line 2790: if a duration is unlinked that's almost certainly because it has an dedicated expression.  So if I do this:

```
d = duration.Duration()
d.linked = False
d.quarterLength = 1.95
d.type = 'half'
```

I expect that to be the same as:

```
d = duration.Duration()
d.linked = False
d.type = 'half'
d.quarterLength = 1.95
```

Also I don't think that we should change expressionIsInferred if the quarterLength does not change:

```
d = duration.Duration(type='quarter')
d.quarterLength = 1.0
```

shouldn't set expressionIsInferred.

mscuthbert: Requesting this to be changed back, but THANKS for other parts!

jacobtylerwalls: good point!

jacobtylerwalls: then that probably needs to be handled in the cases I added elsewhere, in the qL setter

mscuthbert:
This is great for me!

PR #1062, opened 2021-07-14
Avoid gaps that don't fit quantization scheme

### LINKED ISSUE #811: Prevent so many gaps from quantizing durations
**Feature request**
A MIDI note with a duration of, say, 0.385 QL will, with the default quantization units of 0.25 and 0.33QL, have its duration shortened to 0.33QL. Later, a 0.16QL rest will be needed to fill the gap from 0.33 to 0.5. This is likely not what users want.

**Steps to reproduce**

```
s = stream.Stream()
n1 = note.Note(quarterLength=0.385)
n2 = note.Note(quarterLength=0.385)
s.insert(0, n1)
s.insert(0.5, n2)
s.quantize(inPlace=True)
s.makeRests(fillGaps=True, inPlace=True)
s.show('t')
```

**Result**
{0.0} <music21.note.Note C>
{0.3333} <music21.note.Rest 23/100ql>
{0.5} <music21.note.Note C>

**What should be done?**

- Unlike offsets, should durations only round up rather than up or down? Or only when there is a gap to be filled?

jacobtylerwalls:
Fixes #811

Avoid gaps that don't fit quantization scheme (1/6 QL rests, etc.) by looking ahead to see the divisor that will be used to quantize the subsequent offset.

***

Example file: https://github.com/cuthbertLab/music21/blob/master/music21/midi/testPrimitive/test01.mid

**Finale (reference)**
Screen Shot 2021-07-14 at 12 49 42 PM

**m21 roundtrip Before**
Screen Shot 2021-07-14 at 12 48 01 PM

**m21 roundtrip After**

Screen Shot 2021-07-14 at 12 47 52 PM

### DIFF
--- a/music21/stream/base.py
+++ b/music21/stream/base.py
@@ -29,6 +29,7 @@ import pathlib
 import unittest
 import sys
 
+from collections import namedtuple
 from fractions import Fraction
 from math import isclose
 from typing import Union, List, Optional, Set, Tuple, Sequence
@@ -70,6 +71,7 @@ environLocal = environment.Environment('stream')
 StreamException = exceptions21.StreamException
 ImmutableStreamException = exceptions21.ImmutableStreamException
 
+BestQuantizationMatch = namedtuple('BestQuantizationMatch', ['error', 'match', 'signedError', 'divisor'])
 
 class StreamDeprecationWarning(UserWarning):
     # Do not subclass Deprecation warning, because these
@@ -9013,8 +9015,8 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
             found = []
             for div in divisors:
                 match, error, signedErrorInner = common.nearestMultiple(target, (1 / div))
-                # reverse match, error for sorting and add divisor expressed as qL
-                found.append((error, match, signedErrorInner, 1 / div))
+                # reverse match, error for sorting
+                found.append(BestQuantizationMatch(error, match, signedErrorInner, div))
             # get first, and leave out the error
             bestMatchTuple = sorted(found)[0]
             return bestMatchTuple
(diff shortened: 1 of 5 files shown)


music21/stream/base.py
```diff
@@ -8988,46 +9013,58 @@ def bestMatch(target, divisors):
...
         useStreams = [returnStream]
         if recurse is True:
-            for obj in returnStream.recurse():
-                if 'Stream' in obj.classes:
-                    useStreams.append(obj)
+            useStreams = returnStream.recurse(streamsOnly=True, includeSelf=True)
 
         for useStream in useStreams:
-            for e in useStream._elements:
+            for i, e in enumerate(useStream._elements):
```
mscuthbert, line 9035: in general, we shouldn't be looking into another object's private attributes, would .elements work here?

mscuthbert: oh -- that was already an existing bug.  Nevermind!

music21/stream/base.py
```diff
@@ -8988,46 +9013,58 @@ def bestMatch(target, divisors):
...
+                    # Check that any gaps from this quantized duration to the next onset
+                    # are at least as large as the smallest quantization unit (largest divisor)
+                    # If not, then re-quantize this duration with the divisor
+                    # that will be used to quantize the next element's offset
+                    if processOffsets and i + 1 < len(useStream._elements):
+                        next_element = useStream._elements[i + 1]
+                        next_offset = useStream.elementOffset(next_element)
+                        look_ahead_result = bestMatch(float(next_offset), quarterLengthDivisors)
+                        next_offset = look_ahead_result[1]
+                        next_divisor = 1 / look_ahead_result[3]
```
mscuthbert, line 9062: Because of this[1] , [3], I'd prefer that bestMatch returned a NamedTuple, so that (as above), we could avoid all the "unused" and see where `inverseDivisor` is actually being used.  It does seem though that the only time it is used, it is immediately flipped back, so should it just be stored as divisor rather than inverse?

music21/stream/base.py
```diff
@@ -8988,46 +9013,58 @@ def bestMatch(target, divisors):
...
+                        if 0 < next_offset - (e.offset + qlNew) < 1 / max(quarterLengthDivisors):
+                            unused_error, qlNew, signedError, unused_inverse_divisor = bestMatch(
+                                float(ql), (next_divisor,))
                     # Enforce nonzero duration for non-grace notes
+                    # TODO: what about rests? remove rests with ql=0?
```
mscuthbert, line 9067: PLEASE!

mscuthbert: (I was going to say, remove all zero-length notes, but I suppose all zero-length notes should become part of a Chord with the next note, etc.  but that's too much to ask here).

mscuthbert (commented):
Really great.  Just some comments about existing bugs and namedtuple for style.

music21/stream/base.py
jacobtylerwalls: in the next block, all `GeneralNotes` are given a minimum nonzero duration (usually 0.25)-- believe that was #641. 

But my idea here was that's too aggressive when it comes to rests: we should let 0.0 rests that come out of quantization just get deleted instead of beefing them up. Since this is mostly used with MIDI and MIDI doesn't have a notated absence that should be preserved, it just has ... absences. Shouldn't be an effortful addition.

jacobtylerwalls: Ah, but this would probably only happen when either quantizing a stream with notated rests already in it, since in MIDI parsing quantizing comes before rest creation.

music21/stream/base.py
jacobtylerwalls: I needed it to sort with the larger divisor/smaller resulting QL _first_, so a little lazy/hacky, but could do with a code comment.

jacobtylerwalls: nope... wrong... we already have `match` as 1/divisor and that's early enough in the sort, so I can ditch the fourth component and just take 1/match when I need divisor

mscuthbert:
Thanks!  Looks great!

PR #1063, opened 2021-07-15
`fromGeneralNote()` stream packer creates multiple measures

### LINKED ISSUE #1022: Inexpressible durations crash export to musicxml
**music21 version**

v7 dev

**Problem summary**
Inexpressible durations are not divided into tied versions of themselves when converting to MusicXML, but instead crash.

**Steps to reproduce**
note.Note(quarterLength=80).write('musicxml')

Expected:  Note to be divided at every breve length or so into tied notes.

V6 behavior:  Incorrect musicxml is generated (with type "inexpressible") but remainder of score is outputted.

Actual behavior: Crash on musicxml export

jacobtylerwalls:
Fixes #1022 (unless there is another use case you'd like to present)

(force-pushed to correct commit msg)


mscuthbert (approved):
Looks great!  Thanks!

PR #1068, opened 2021-07-23
Set paddingRight on MusicXML import and fix beaming of quarter + eighth in final incomplete 6/8 measure

jacobtylerwalls:
- Set paddingRight on musicXML import
- `makeBeams()` now observes a difference between paddingLeft and paddingRight instead of treating all incomplete measures as having paddingLeft
- `getBeams()` for 6/8 fixed to handle quarter + eighth + eighth:
  - should be None, None, None
  - was None, 'start', 'stop' if incompleteness programmatically interpreted as paddingLeft (see above)
  - was None, None, 'stop' once above issue was fixed

### DIFF
--- a/music21/_version.py
+++ b/music21/_version.py
@@ -42,7 +42,7 @@ When changing, update the single test case in base.py.
 Changing this number invalidates old pickles -- do it if the old pickles create a problem.
 '''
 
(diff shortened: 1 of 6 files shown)


mscuthbert (approved):
Looks great! Thanks!

PR #1072, opened 2021-08-04
Do not completely skip Nested Tuplets

mscuthbert:
Also add some typing, and remove obsolete code from when DurationTuples could have their own tuplets.


jacobtylerwalls (commented):
Changes look good and will make debugging easier in future, thanks!

PR #1075, opened 2021-08-04
Do not import pygame/pyaudio until needed

mscuthbert:
A recent change to run TestExternal had the drawback that it imported pygame and pyaudio every time music21 was started (if it was installed), adding half a second to music21 start. 

All parts of a Class definition that are not in a `def` are run on music21 startup.  Therefore setup commands, etc. for test modules / rarely run modules need to be fast, or music21 startup will slow to a crawl.  (hence lazy importing of matplotlib, etc.)

### DIFF
--- a/music21/audioSearch/recording.py
+++ b/music21/audioSearch/recording.py
@@ -20,6 +20,7 @@ users of 64-bit windows but 32-bit python should download the win32 port
 
 users of 64-bit windows and 64-bit python should download the amd64 port
 '''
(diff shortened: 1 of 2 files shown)


jacobtylerwalls (approved):
Thanks for catching my oversight here (and for the `pygame` -> `pyaudio` change).

jacobtylerwalls:
The extra pragmas are harmless, even desirable -- since someday we might mass remove the ones on the class level.

PR #1077, opened 2021-08-04
Avoid so much context searching in makeRests()

jacobtylerwalls:
Some of the changes to makeRests in v7 introduced expensive context searching by calling `barDuration` on Measure objects, which have to look left to get a time signature, which can be a long distance.

UPDATE: fixed test failure.

### DIFF
--- a/music21/stream/makeNotation.py
+++ b/music21/stream/makeNotation.py
@@ -14,7 +14,8 @@
 
 import copy
 import unittest
-from typing import List, Generator, Optional
+from typing import List, Generator, Optional, Union
+from fractions import Fraction  # typing only
 
 from music21 import beam
 from music21 import clef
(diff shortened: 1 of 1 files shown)


music21/stream/makeNotation.py
```diff
@@ -840,34 +840,46 @@ def makeRests(
...
+        Needed for timeRangeFromBarDuration.
+        Returns 0.0 if no meter can be found.
+        """
+        bar_duration: float = 0.0
+        if ts is not None:
+            bar_duration = ts.barDuration
+        elif m is not None:
+            # More expensive context search
+            bar_duration = m.barDuration
+        post = bar_duration.quarterLength
```
mscuthbert, line 857: barDuration is typed as a float but floats don't have quarterLength.  Better as:

```
post: float = 0.0
if ts is not None:
     post = ts.barDuration.quarterLength
elif m is not None:
    # more expensive context search
    post = m.barDuration.quarterLength
```

note that the quarterLength can be a fraction (as can paddingLeft/paddingRight) so either type the func and post as Union[float, Fraction]) (best) or return float(post).

jacobtylerwalls: Thanks, I missed this because I haven't tested the (unreachable, I think) error condition when both args are None.

PR #1078, opened 2021-08-04
Return UnpitchedPercussion instance when MIDI instrument lookup fails on channel 10

jacobtylerwalls:
The basic idea is that we should create a better fallback instrument when we have bad MIDI data on channel 10, because this information is starting to be used to create and manage `note.Unpitched` instances.

And a side benefit: reduce indentation


mscuthbert:
We should rewrite PercussionMapper to use the same indexing scheme as the rest of the system.

PR #1080, opened 2021-08-05
Enable Sphinx link-checking and use HTTPS in hyperlinks

### LINKED ISSUE #1079: Link in description: change http -> https
I just noticed that the link in the description takes to [http://web.mit.edu/music21/](http://web.mit.edu/music21/).

It seems that [https://web.mit.edu/music21/](https://web.mit.edu/music21/) works well and avoids the usual "unsafe site" browser warnings.

Maybe consider replacing. I have not checked whether this happens in other places as well other than the description.

jacobtylerwalls:
Enable Sphinx's `linkcheck` mode and fix (most of) the redirected links. Some broken links remain.

Thanks @napulen for the push; I had this lying around from a few weeks ago and hadn't opened a PR yet. And you pointed out some that Sphinx didn't find.  (EDIT: duh, because the music21 site doesn't redirect from HTTP to HTTPS.) Does my third commit get the ones you were thinking of? Fixes #1079.


napulen:
Hey @jacobtylerwalls, thanks for the quick reply!

That commit gets the `CONTRIBUTING.md` markdown and two sources.

I think the `README` ones are still missing.

Here is how the proposed patch for those looks like:

image

napulen:
The last one is in the description of the repo. I don't think that has a source code file anywhere reachable. Possibly it can only be changed if an admin goes and edits the description in the settings of the repo.

jacobtylerwalls:
Thanks, I wasn't sure I got 'em all, so that's why I asked! And now I see what you're referring to about description. Sounds like a Michael task. Nice eagle eye!

napulen:
A maybe even better solution would be to make `web.mit.edu` return a redirect to https if people go with http, but I think that's above all of our paygrades :).

mscuthbert:
MIT made two choices years ago that were decent bets at the time, but that have come back to bite them.  One was that web.mit.edu and not www.mit.edu would be their main address and that www.mit.edu would be student run only.  The second was that https://web.mit.edu would only be used for sites that required authorization to access, and http://www.mit.edu would be for general use.  Labs would need to assure that they didn't use any authorization resources in order to get https: access to their public sites.  I never bothered to go through the cumbersome process of ensuring that I didn't use any authorization tokens at cuthbertLab or music21 so last time I checked before now, https://web.mit.edu/music21 didn't work, but just checked now and it seems to work.  So merging.  I'm less bothered by bare http links since they do still tend to get upgraded without a problem, so I'll get to the second half of this PR when I have time.  MERGING.

jacobtylerwalls:
Sounds good. The reason for fixing the redirects was just to get less noisy output of `linkcheck`, that is, so that it's easier to work on the broken links later.

PR #1086, opened 2021-08-10
Set correct MIDI program when constructing instrument from name given in MIDI file

### LINKED ISSUE #1085: Separate handling of instrument name and patch change MIDI messages to avoid false program values
> The weirdest is that some of the tracks are assigned "Choir Ooos" which wasn't even in the original file. Must be some sort of logic currently in place based on track name?

This should probably be tracked separately (program, rather than channel). What's happening is that the same routine is used to process both `PROGRAM_CHANGE` messages and `INSTRUMENT_NAME` or `SEQUENCE_TRACK_NAME` messages, and that routine attempts to set the midiProgram if the message has that data, which a ` SEQUENCE_TRACK_NAME` message will not. From the `SEQUENCE_TRACK_NAME` messages in your file, we normalized each of "Soprano", "Alto", etc. to an `instrument.Soprano`, etc., instance with default midiProgram = 53 (0-indexed, so = 54: Choir Oohs). We should be smarter about that.

_Originally posted by @jacobtylerwalls in https://github.com/cuthbertLab/music21/issues/465#issuecomment-895267105_

jacobtylerwalls:
Fixes #1085 

Thanks to @Meekohi for the report. Oversight in #589.

Bumped the minor version so that parsed files won't accidentally be loaded when testing the fix.

Regression test fails on current version.

### DIFF
--- a/music21/midi/translate.py
+++ b/music21/midi/translate.py
@@ -1728,7 +1728,8 @@ def getMetaEvents(events):
     from music21.midi import MetaEvents, ChannelVoiceMessages
 
     metaEvents = []  # store pairs of abs time, m21 object
-    for i, eventTuple in enumerate(events):
+    last_program: int = 0
+    for eventTuple in events:
         t, e = eventTuple
         metaObj = None
         if e.type == MetaEvents.TIME_SIGNATURE:
@@ -1739,9 +1740,14 @@ def getMetaEvents(events):
         elif e.type == MetaEvents.SET_TEMPO:
             metaObj = midiEventsToTempo(e)
         elif e.type in (MetaEvents.INSTRUMENT_NAME, MetaEvents.SEQUENCE_TRACK_NAME):
+            # midiEventsToInstrument() WILL NOT have knowledge of the current
+            # program, so set it here
             metaObj = midiEventsToInstrument(e)
+            metaObj.midiProgram = last_program
         elif e.type == ChannelVoiceMessages.PROGRAM_CHANGE:
+            # midiEventsToInstrument() WILL set the program on the instance
             metaObj = midiEventsToInstrument(e)
+            last_program = e.data
         elif e.type == MetaEvents.MIDI_PORT:
             pass
         else:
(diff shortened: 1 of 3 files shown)


music21/midi/translate.py
```diff
@@ -1728,7 +1728,8 @@ def getMetaEvents(events):
     from music21.midi import MetaEvents, ChannelVoiceMessages
 
     metaEvents = []  # store pairs of abs time, m21 object
-    for i, eventTuple in enumerate(events):
+    last_program: int = 0
```
mscuthbert, line 1731: set to -1 to start... see below.

music21/midi/translate.py
```diff
@@ -1739,9 +1740,14 @@ def getMetaEvents(events):
         elif e.type == MetaEvents.SET_TEMPO:
             metaObj = midiEventsToTempo(e)
         elif e.type in (MetaEvents.INSTRUMENT_NAME, MetaEvents.SEQUENCE_TRACK_NAME):
+            # midiEventsToInstrument() WILL NOT have knowledge of the current
+            # program, so set it here
             metaObj = midiEventsToInstrument(e)
+            metaObj.midiProgram = last_program
```
mscuthbert, line 1746: `if last_program != -1`  -- so we're trusting midiEventsToInstrument if no program was set.

music21/midi/translate.py
```diff
@@ -3885,6 +3891,34 @@ def testEmptyExport(self):
...
+        # the default midiProgram, we want 0.
+        event2.data = b'Soprano'
+        event2.channel = 2
+        event2.type = midiModule.MetaEvents.SEQUENCE_TRACK_NAME
+
+        DUMMY_DELTA_TIME = None
+        meta_event_pairs = getMetaEvents([(DUMMY_DELTA_TIME, event1), (DUMMY_DELTA_TIME, event2)])
+        # Second element of the tuple is the instrument instance
+        self.assertEqual(meta_event_pairs[0][1].midiProgram, 0)
+        self.assertEqual(meta_event_pairs[1][1].midiProgram, 0)
```
mscuthbert, line 3920: this is good.  but without a PROGRAM_CHANGE before it, should give the default midiProgram for Soprano.

mscuthbert (changes requested):
Looks good -- but I think that in the rare cases (but existing in the wild, I believe) where a program change is not specified but a track name is, we should trust the default instrument.

mscuthbert:
(Feel free to merge after putting the change in)

mscuthbert:
GitHub's problems seem to be preventing Actions from running, but it all looks good to me.  Merging

PR #1088, opened 2021-08-12
Enable braille keywords in show() and write()

jacobtylerwalls:
Regardless of whether we make all the keywords explicit in #872, there's no reason not to permit them to work when calling `show()` or `write()`!

Makes debugging easier.


mscuthbert:
Thanks!

PR #1090, opened 2021-08-13
Silence MuseScore command-line output in case of success

### LINKED ISSUE #1089: Suppress messages from MuseScore CLI
**music21 version**

`music21==6.7.1`   

**Problem summary**
Tiny notation + show() produces an error, which nonetheless does not affect the output, but is very annoying in Notebooks.

(The actual problem is that the tinynotation parser produces something that the XML parser does not like, and this is not a notebook specific problem.)

**Steps to reproduce**

Using the exact example from the beginning of the documentation:

```
music21.converter.parse("tinynotation: 3/4 c4 d8 f g16 a g f#").show()
```

Expected: see the music example in cell output

Actual: see the music example, preceded by an error message:

```
convert </var/folders/n2/3fzlvz9j4xl3mthwygc5n67m0000gn/T/music21/tmp14amnsf_.xml>...
Error at line 24 col 47: no instrument found for part 'Pbb8d9a620e009425ce86d3f720af6859'
	to </var/folders/n2/3fzlvz9j4xl3mthwygc5n67m0000gn/T/music21/tmp14amnsf_.png>
... success!
```

Screen Shot 2021-08-12 at 4 01 54 PM

**More information**

Fixing the XML created by tinynotation.  
or.  
Fixing the XML parser so that it doesn't expect something that it apparently doesn't need.  
are both good options, and one of them is probably the right one, but...

Suppressing a non-blocking error from output when calling `.show()` would also be equally useful (at least for me), and might be easier.

jacobtylerwalls:
Fixes #1089 -- although silencing everything might seem heavy-handed, printing 7 messages for every success seems much.

### DIFF
--- a/music21/converter/subConverters.py
+++ b/music21/converter/subConverters.py
@@ -23,7 +23,6 @@ import io
 import os
 import pathlib
 import subprocess
-import sys
 import unittest
 
 from typing import Union
--- a/music21/converter/subConverters.py
+++ b/music21/converter/subConverters.py
@@ -944,7 +944,13 @@ class ConverterMusicXML(SubConverter):
         if completed_process.returncode != 0:
             # Raise same exception class as findNumberedPNGPath()
             # for backward compatibility
-            raise SubConverterFileIOException(completed_process.stderr)
+            stderr_bytes = completed_process.stderr
+            try:
+                import locale
+                stderr_str = stderr_bytes.decode(locale.getpreferredencoding(do_setlocale=False))
+            except UnicodeDecodeError:
+                stderr_str = stderr_bytes  # not really a str, but best we can do.
+            raise SubConverterFileIOException(stderr_str)
 
         if common.runningUnderIPython() and common.getPlatform() == 'nix':
             # Leave environment in original state
--- a/music21/converter/subConverters.py
+++ b/music21/converter/subConverters.py
@@ -941,12 +941,7 @@ class ConverterMusicXML(SubConverter):
 
             musescoreRun.extend(['-r', str(defaults.ipythonImageDpi)])
 
-        storedStrErr = sys.stderr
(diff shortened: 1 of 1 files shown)


jacobtylerwalls:
I haven't tested to be sure, but looking at the inoperative lines I was deleting that attempt to capture stderr, there were probably rendered inoperative by 141dfecb1149405c612795aa4fb6522bfbb74d19. So I think silencing is right-track; probably a regression to have these messages appearing again.

mscuthbert:
What happens if there's a bug/error?  Is there an opportunity to review the code if there is something wrong?

I'm okay with silencing on success, but on error, people will want verbose messages

jacobtylerwalls:
Yeah, why not raise the stderr content if we can. Too bad MuseScore writes success messages to stderr.

Here is the output in case of failure with this PR (failure case described in https://github.com/cuthbertLab/music21/issues/890#issuecomment-899083110):

music21.converter.subConverters.SubConverterFileIOException: b'dlopen error : dlopen(libjack.0.dylib, 1): image not found \ndlopen error : dlopen(/usr/local/lib/libjack.0.dylib, 1): image not found \nCreating main window\xe2\x80\xa6\nZoomBox::setLogicalZoom(): Formatting logical zoom level as 100% (rounded from 1.000000)\nReading translations\xe2\x80\xa6\nconvert </var/folders/_c/qsz9054d71v9z4wt9p38wc9c0000gn/T/music21/tmpxmsarix2.musicxml>...\nError at line 1359 col 14: rounding up measure duration 13/12 to 70/64\ncannot add rest at tick 4/4 (1920) track 8: element already present\ncannot add rest at tick 4/4 (1920) track 24: element already present\n'

jacobtylerwalls:
Latest changes to improve error look good!

PR #1093, opened 2021-08-20
Reiterate accidentals each measure in Stream.makeAccidentals()

jacobtylerwalls:
Fixes issue demonstrated in regression test. Need to reiterate accidentals in subsequent measures when making accidentals on a score.

Surely escaped notice in the past by musicxml export pipeline running makeNotation() and ultimately `Part.makeAccidentals()` rather than `Stream.makeAccidentals()`.


mscuthbert (approved):
Not entirely sure how this error crept in -- surely would have been noticed before?

PR #1095, opened 2021-08-21
NotImplemented fixes for subclass `__eq__()` methods

jacobtylerwalls:
Fixes Python DeprecationWarnings:

`DeprecationWarning: NotImplemented should not be used in a boolean context`.

This is because `NotImplemented` is truthy. Haven't checked to see if this has been causing logic issues.


mscuthbert (approved):
thanks!

PR #1098, opened 2021-08-22
V7 pre release

mscuthbert:
Errors etc. fixed in making sure that the corpus and docs got updated.


jacobtylerwalls:
When it comes to release notes, just wanted to make sure on the radar the two biggest breaking changes:

```python
for note of midi_song.parts[0]:
    # notes are in measures now, so this won't execute

for note of stream.notes:
    if not isinstance(note, chord.Chord):
        # cannot assume these are Notes anymore, might be Unpitched
        # don't access n.pitch
```

PR #1100, opened 2021-08-22
Prevent AttributeError when making accidentals

jacobtylerwalls:
```
  File "/Users/jwalls/music21/music21/stream/makeNotation.py", line 1689, in makeAccidentalsInMeasureStream
    tiePitchSet = getTiePitchSet(previousNoteOrChord)
  File "/Users/jwalls/music21/music21/stream/makeNotation.py", line 1625, in getTiePitchSet
    tiePitchSet.add(n.pitch.nameWithOctave)
AttributeError: 'Unpitched' object has no attribute 'pitch'
```

***
The other way to do it is to check `hasattr(n, 'pitch')`, which we do elsewhere, but this is slightly more explicit to say why the check is needed.

### DIFF
--- a/music21/expressions.py
+++ b/music21/expressions.py
@@ -421,8 +421,11 @@ class Ornament(Expression):
         transposeInterval
     ):
         '''
-        Used by trills and mordants to fill out their realization
+        Used by trills and mordents to fill out their realization.
         '''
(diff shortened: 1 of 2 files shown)


mscuthbert:
Thanks!

PR #1101, opened 2021-08-23
Restore original instrument lengths after _transposeByInstrument

mscuthbert:
By not restoring original instrument lengths we were creating instruments with quarter lengths like 40.0, which is a complex duration that was being split by  `splitAtDurations`.  Then we'd end up with two Instrument objects in measure 1, one of which was FAR FAR later, so even though it was being removed in output, it caused a massive `<forward>` tag in the musicxml output which caused spacing to be terrible.

Also saves making a second deepcopy of a score for 98% of show() cases which don't have complex durations.

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -2492,14 +2492,8 @@ class PartExporter(XMLExporterBase):
 
         self.xmlRoot.set('id', str(self.firstInstrumentObject.partId))
 
-        # Split complex durations in place if needed
-        has_complex_durations = bool(
-            self.stream.recurse().getElementsNotOfClass(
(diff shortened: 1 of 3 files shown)


mscuthbert:
@jacobtylerwalls  -- you might like this...

jacobtylerwalls:
Nice fix. I think I missed it because I was testing score roundtrips, not part roundtrips.

```
>>> b = corpus.parse('beach', forceSource=True)
>>> b.parts[0].atSoundingPitch
True
>>> b.atSoundingPitch
'unknown'
```

Flipping the property to `unknown` on the part doesn't reproduce the issue!

jacobtylerwalls:
Ah! I knew why this seemed familiar. I had a different fix for this on #1051 , see 0a42b4d

PR #1102, opened 2021-08-23
Clear tuplets correctly in xmlToDuration()

jacobtylerwalls:
Cherry-picked from #885 so we can separate the bug fix from the design question about exceptions.

Tuplets might not be cleared correctly if a value is provided for the `inputM21` argument.

m21's own use of `xmlToDuration()` renders it unlikely to encounter this bug, since it only provides `inputM21` for grace note corrections, but fixing since the function is public.

Coveralls shows that the `except` clause doesn't run. I don't think in the modern duration module there's a way to get durations to raise when setting `.components`: `quarterConversion()` catches them all and provides fallbacks. (If that's wrong, and we see this in the wild, I'd like to get a bug report so we can fix or document with a test.)

Can wait for next version; no rush.

### DIFF
--- a/music21/musicxml/xmlToM21.py
+++ b/music21/musicxml/xmlToM21.py
@@ -12,7 +12,6 @@
 import copy
 import fractions
 import io
-import math
 # import pprint
 import re
 # import sys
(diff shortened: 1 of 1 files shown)


mscuthbert (approved):
THANKS!

PR #1104, opened 2021-08-25
Make files compress automatically with fp='x.mxl'

mscuthbert:


music21/converter/subConverters.py
```diff
@@ -848,8 +848,8 @@ def findNumberedPNGPath(inputFp: Union[str, pathlib.Path]) -> str:
         path_without_extension = inputFp[:-1 * len('.png')]
 
         for search_extension in ('1', '01', '001', '0001', '00001'):
-            search_path = path_without_extension + '-' + search_extension + '.png'
-            if os.path.exists(search_path):
+            search_path = pathlib.Path(path_without_extension + '-' + search_extension + '.png')
+            if search_path.exists():
                 return search_path
```
jacobtylerwalls, line 853: Thanks 👍🏻

jacobtylerwalls (approved):

jacobtylerwalls:
Coveralls stats appear wrong -- notice these "uncovered" lines are uncovered on latest build on main branch, see:
https://coveralls.io/builds/42488358/source?filename=music21%2Fmusicxml%2Fm21ToXml.py#L535
and
https://coveralls.io/builds/42488358/source?filename=music21%2Fmusicxml%2FpartStaffExporter.py#L228

jacobtylerwalls:
We already had auto compression for `fmt='mxl'` from `Music21Object.write()`, just not from `ConverterMusicXML.write()`, which is good to have, also--but renaming just to not get confusion on this point, since the former is the usual use case.

https://github.com/cuthbertLab/music21/blob/f97789a54afcbff60036a54a2a1b72074e1f7490/music21/base.py#L2652-L2653

PR #1109, opened 2021-08-28
Allow making beams in individual measures without time signatures

jacobtylerwalls:
Allow making beams on individual measures that may lack a `TimeSignature` by getting the time signature from the context.

Useful for operating on, say, just the final measure of a part.

### DIFF
--- a/music21/stream/base.py
+++ b/music21/stream/base.py
@@ -6522,7 +6522,7 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
             classFilterList=classFilterList,
         )
 
-    def makeBeams(self, *, inPlace=False, setStemDirections=True):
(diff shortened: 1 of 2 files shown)


jacobtylerwalls:
Marking as work in progress because increasing the severity of the message from debug to warning made warnings emanate from garden-variety makeNotation calls like this:

https://github.com/cuthbertLab/music21/blob/172a81c4daa99822e007f77dbac53a0a07fc4049/music21/converter/subConverters.py#L1516-L1529

That doesn't seem good. We shouldn't be failing to make beams because a time signature is interpolated later. Need to test to verify that's what's happening.

mscuthbert (approved):
Wonderful!  Thank you!

PR #1120, opened 2021-09-08
Mxl direct writing cleanups

mscuthbert:
Jacob and I were working on the same thing at the same time.  His branch got almost everything, but there were some small cleanups that we also would like to get in.


mscuthbert:
Doesn't matter.

Issue #992, opened 2021-04-30
Converter midi to musicxml with one sixteenth note  and dotted eighth note and dotted half note(tie pre eightth note) grouping one 4/4 bar


Hwanggang:
**music21 version**

`6.7.1`   

**Problem summary**

Converter midi to musicxml with one 4/4 bar ,which grouped by one sixteenth note  and dotted eighth note and dotted 
![test](https://user-images.githubusercontent.com/24386723/116670698-9f8d5c80-a9d2-11eb-8d6f-a8d01503970c.png)
half note(tie pre eightth note) . 

**Steps to reproduce**

**Expected vs. actual behavior**

`Expected in Measure:
<note  dynamics="111.11">
    <pitch>
     <step>D</step>
     <octave>4</octave>
    </pitch>
    <duration>2520</duration>
    <type>16th</type>
    <beam number="1">begin</beam>
</note>
<note  dynamics="111.11">
    <pitch>
     <step>A</step>
     <octave>4</octave>
    </pitch>
    <duration>7560</duration>
    <tie type="start"/>
    <type>eighth</type>
    <dot/>
    <beam number="1">end</beam>
     <tied type="start"/>
</note>
<note  dynamics="111.11">
    <pitch>
     <step>A</step>
     <octave>4</octave>
    </pitch>
    <duration>30240</duration>
    <tie type="stop"/>
    <type>half</type>
    <dot/>
     <tied type="stop"/>
</note>

Actual in Measure :
<note dynamics="111.11">
        <pitch>
          <step>D</step>
          <octave>4</octave>
        </pitch>
        <duration>2520</duration>
        <type>16th</type>
</note>
<note dynamics="111.11">
        <pitch>
          <step>A</step>
          <octave>4</octave>
        </pitch>
        <duration>37800</duration>
        <type>half</type>
        <dot />
        <dot />
        <dot />
 </note>`

**More information**

jacobtylerwalls:
Greetings, and thanks for taking the time to write. This is a known missing feature. Can be reproduced without external files like this:

```
>>> from music21 import *
>>> n = note.Note(quarterLength = 0.25)
>>> n2 = note.Note(quarterLength = 3.75)
>>> m = stream.Measure([n, n2])
>>> m.show()
```

![Screen Shot 2021-04-30 at 8 28 01 AM](https://user-images.githubusercontent.com/38668450/116695022-ffaddf80-a98d-11eb-9147-5c27c6502d49.png)


[...]

PR #1124, opened 2021-09-11
Add preserveExisting=True option to `partitionByCount()`

### LINKED ISSUE #1115: Subdivide MeterSequences without destroying original beat partitions
I worked through the issues you presented, and I think they all have to do with the destructive-ness of `partitionByCount()`, which is also called by `subdivideNestedHierarchy()`. (Both methods mention destructiveness in the docs.) I think there should be a non-destructive way to subdivide, i.e., to retain the original partitions.

_Originally posted by @jacobtylerwalls in https://github.com/cuthbertLab/music21/issues/992#issuecomment-913689245_

jacobtylerwalls:
Fixes #1115 

(Although I would like to also be able to use this someday in `subdivideNestedHierarchy()`, this is enough to get it working in `partitionByCount()`.)

@MarkGotham could you give this a spin and report back as to whether it helps with the issues you were working on in #992? Thanks!

### DIFF
--- a/music21/meter/core.py
+++ b/music21/meter/core.py
@@ -666,7 +666,7 @@ class MeterSequence(MeterTerminal):
         music21.exceptions21.MeterException: Cannot set partition by 11 (5/8)
 
         This method will destroy any established structure in the stored partition
-        unless `preserveExisting` is True (new in v7.3):
+        unless `preserveExisting` is True (new in v.7.3):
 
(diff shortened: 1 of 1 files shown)


jacobtylerwalls:
Closing per https://github.com/cuthbertLab/music21/issues/992#issuecomment-917468003.

PR #1125, opened 2021-09-12
Fix improper call to locale library in multiprocess test text output

jacobtylerwalls:
- `getdefaultencoding()` is on `sys`, not `locale`. whoops! mistake in #1117.
- Just remove needless complexity and write in 'utf-8', since this is just a library developer tool.


jacobtylerwalls:
We should enable the `no-member` pylint check, which could have caught this. I have an old branch where I started that a few months ago and added pragmas where there were false positives (not that onerous, really). I'll open a PR this fall.

PR #1126, opened 2021-09-13
Fix `stripTies()` crash in certain scenario with matchByPitch=False

jacobtylerwalls:
`stripTies()` crashed in a scenario where:
  - `matchByPitch=False` (only currently used in MIDI writing; default is otherwise True)
  - a Chord with "stop" or "continue" ties follows a Rest
  (that would seem malformed, but for the fact that stripTies() flattens the part, so you can have a perfectly valid measure that when flattened gives a stop tie following a rest in virtue of there being voices)

Regression in b7bcad9914 (v7.1)

Luckily, `matchByPitch` defaults True in v7, so this only affects MIDI writing by default.

Reported in StackOverflow [post](https://stackoverflow.com/questions/69140830/music21-write-function-giving-attributeerror-rest-object-has-no-attribute).

***

I'm not certain why `matchByPitch` is still False in the MIDI writer. I suppose I just was unwilling to change something I didn't fully understand. EDIT: opened #1127 to change this.

I expect `mypy` would have caught this. (The variable name `nLast` coaxed out the coding error.) I'll be developing with mypy more.

### DIFF
--- a/music21/stream/base.py
+++ b/music21/stream/base.py
@@ -7136,7 +7136,7 @@ class Stream(core.StreamCoreMixin, base.Music21Object):
         # need to just get .notesAndRests, as there may be other objects in the Measure
         # that come before the first Note, such as a SystemLayout object
         f = returnObj.flatten()
-        notes = f.notesAndRests.stream()
+        notes_and_rests = f.notesAndRests.stream()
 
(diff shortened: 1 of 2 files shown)


mscuthbert:
Looks great!  I don't understand everything but I understand the scenario and why I didn't check for it but also why it's important.

PR #1127, opened 2021-09-13
Use matchByPitch=True in MIDI export

jacobtylerwalls:
Another way of preventing the issue in #1126 is to just use the safer method of stripping ties during MIDI export.

matchByPitch was False since the inception of the MIDI exporter, and since then the speed vs. accuracy tradeoff has swung way, way in favor of accuracy. Essentially a follow-up to #818.


mscuthbert:
Thanks!

PR #1130, opened 2021-09-16
harmonicFunction. maps between RN figures and function labels

### LINKED ISSUE #1103: Harmony: map between function and figure
**Motivation**
music21 supports Roman numeral analysis in many ways, but currently has nothing on function (T, S, D ... ).

**Feature summary**
Provide a simple mapping between figure and function.

**Proposed implementation**
Store pairs of corresponding figure and function labels on the `roman.py` module:
```
functionFigureTuples = (
    # Scale degrees (major)
    # 1:
    ('T', 'I'), 
    ('t', 'i'),
    # b2:
    ('sL', 'bII'),
    # 2:
    ('Sp', 'ii'),
    # b3:
    ('dL', 'III'),
    ('tP', 'III'),  # Note last: tP generally preferred over dL
    # 3:
    ('Tl', 'iii'),
    ('Dp', 'iii'),  # Note last: Dp generally preferred over tL
    # 4:
    ('S', 'IV'),
    ('s', 'iv'),
    # 5:
    ('D', 'V'),
    ('D7', 'V7'),
    ('d', 'v'),
    # b6:
    ('tL', 'VI'),
    ('sP', 'VI'),  # Note last: sP generally preferred over tL
    # 6:
    ('Sl', 'vi'),
    ('Tp', 'vi'),  # Note last: Tp generally preferred over Sl
    # b7:
    ('dP', 'VII'),
    # 7:
    ('Đ7', 'viio'),
    ('Dl', '#VII'),
    )
```
Either read directly from that, or store this in two dicts `functiontoFigureDict` and `figureToFunctionDict`.
In any case, offer two functions for mapping `functiontoFigure()` and `figureToFunction()`.
Probably static functions rather than methods on the `roman.RomanNumeral` class, because of the asymmetry: both accept `str`; the `figureToFunction()` probably also accepts a full `roman.RomanNumeral` object (retrieving `.figureAlone` for the equivalent input).

**Intent**
I'm happy to finish up an implement of this, integrating any comments on the above, and/or any preference for where in the `roman.py` module it should go.

MarkGotham:
Resolves #1103


MarkGotham:
#1103

mscuthbert:
Thanks for the contribution Mark -- I think I really do want them to be Enums -- we're moving that way throughout the system.  They can have a `str()` form that turns them into the labels here, but I'm trying to create more ways that people can see what the possible choices are and what they mean.  Sort of like we did with the various #6, #7 Enums in RomanText.

music21/analysis/harmonicFunction.py
```diff
@@ -0,0 +1,290 @@
...
+from music21 import key
+from music21 import roman
+from music21 import scale
+
+from music21 import environment
+_MOD = 'analysis.harmonicFunction'
+environLocal = environment.Environment(_MOD)
+
+
+class FUNKTION(common.enums.StrEnum):
```
mscuthbert, line 25: Should be `Funktion` Enum classes are in TitleCase.  Members are in ALL_CAPS

music21/analysis/harmonicFunction.py
```diff
@@ -0,0 +1,290 @@
...
+
+    TONIKA_DUR = 'T'
+    TONIKA_DUR_PARALLELKLANG_MOLL = 'Tp'
+    TONIKA_DUR_GEGENKLANG_MOLL = 'Tg'
```
mscuthbert, line 29: Where there is a common English equivalent (TONIC_MAJOR)  I much prefer English.

music21/analysis/harmonicFunction.py
mscuthbert: And HarmonicFunction not German (as below)

music21/analysis/harmonicFunction.py
```diff
@@ -0,0 +1,290 @@
...
+
+    DOMINANT_MOLL = 'd'
+    DOMINANT_MOLL_PARALLELKLANG_DUR = 'dP'
+    DOMINANT_MOLL_GEGENKLANG_DUR = 'dG'
+
+
+_functionFigureTuplesKeyNeutral = (
+
+    ('T', 'I'),
+    ('t', 'i'),
```
mscuthbert, line 55: The value of the key should be a `HarmonicFunction.TONIC_MINOR` type entry.

mscuthbert: Why is this a tuple of tuples and not a Dict?

mscuthbert (commented):
A few comments and requested changes before looking more closely at the code.

The code should also use English where possible.  Only in a few cases where there is no commonly used English equivalent ("Hauptstimme", "Augenmusik") should German or French ("jetté") be used.  Italian is sort of grandfathered in...

music21/analysis/harmonicFunction.py
MarkGotham: Seemed a better representation as they map in both directions equally (both entries are used as both key and value). 

Anyway, no problem – happy to recast as a dict. Done.

music21/analysis/harmonicFunction.py
```diff
@@ -0,0 +1,305 @@
...
+
+    referenceTuples = functionFigureTuplesMajor
+    if keyOrScale.mode == 'minor':
+        referenceTuples = functionFigureTuplesMinor
+
+    for thisKey, thisValue in referenceTuples.items():
+        if thisHarmonicFunction == thisKey:
+            return roman.RomanNumeral(thisValue, keyOrScale)
+
+    return False
```
jacobtylerwalls, line 205: A small performance improvement to avoid walking every item of `referenceTuples`. You could do something similar below. And return `None` instead of mixing booleans with romans.
```suggestion
try:
    figure = referenceTuples[thisHarmonicFunction]
except KeyError:
    return None

return roman.RomanNumeral(figure, keyOrScale)
```

music21/analysis/harmonicFunction.py
```diff
@@ -0,0 +1,305 @@
...
+}
+
+functionFigureTuplesMinor = {
+    **functionFigureTuplesMinor,
+    **_functionFigureTuplesKeyNeutral,
+}
+
+
+def functionToRoman(thisHarmonicFunction: HarmonicFunction,
+                    keyOrScale: Union[key.Key, scale.Scale, str] = 'C'):
```
jacobtylerwalls, line 116:
```suggestion
def functionToRoman(thisHarmonicFunction: HarmonicFunction,
                    keyOrScale: Union[key.Key, scale.Scale, str] = 'C') -> Optional[roman.RomanNumeral]:
```

music21/analysis/harmonicFunction.py
```diff
@@ -0,0 +1,305 @@
...
+    for thisKey, thisValue in referenceTuples.items():
+        if thisHarmonicFunction == thisKey:
+            return roman.RomanNumeral(thisValue, keyOrScale)
+
+    return False
+
+
+def romanToFunction(rn: roman.RomanNumeral,
+                    onlyHauptHarmonicFunction: bool = False
+                    ):
```
jacobtylerwalls, line 210: Sim.

music21/analysis/harmonicFunction.py
```diff
@@ -0,0 +1,305 @@
...
+            referenceTuples = functionFigureTuplesMinor
+
+    for thisKey, thisValue in referenceTuples.items():
+        if rn.romanNumeral == thisValue:
+            if onlyHauptHarmonicFunction:
+                return HarmonicFunction(str(thisKey)[0])
+            else:
+                return thisKey
+
+    return False
```
jacobtylerwalls, line 264: sim.

music21/analysis/harmonicFunction.py
```diff
@@ -0,0 +1,305 @@
...
+        fakeExample = 'TPG'
+        self.assertRaises(ValueError, HarmonicFunction, fakeExample)
+
+    def testFunctionToRoman(self):
+        self.assertEqual(functionToRoman('T').figure, 'I')
+
+    def testSimplified(self):
+        rn = roman.RomanNumeral('III', 'f')
+        self.assertEqual(str(romanToFunction(rn)), 'tP')
+        self.assertEqual(str(romanToFunction(rn, onlyHauptHarmonicFunction=True)), 't')
```
jacobtylerwalls, line 295: Casting to `str` should be unnecessary, since a `StrEnum` already compares equal with a string. Maybe one of the tests could compare with the enum value e.g. `assertIs(..., HarmonicFunction.GEGENKLANG)`.

music21/analysis/harmonicFunction.py
```diff
@@ -0,0 +1,305 @@
...
+import unittest
+
+from music21 import common
+from music21 import key
+from music21 import roman
+from music21 import scale
+
+from music21 import environment
+_MOD = 'analysis.harmonicFunction'
+environLocal = environment.Environment(_MOD)
```
jacobtylerwalls, line 22: @mscuthbert might be able to speak to this, but I think we can chop these MOD lines going forward.

jacobtylerwalls (commented):
Thanks Mark, I left a few comments. Myke may be hoping to merge #1298 before this, so there might be a mypy tweak to do when that's in.

music21/analysis/harmonicFunction.py
MarkGotham: Good shout re including a comparison with the enum value. Thanks!

music21/analysis/harmonicFunction.py
MarkGotham: `None` = now done, but I don't immediately see how to do the / a more efficient version this way round.

music21/analysis/harmonicFunction.py
MarkGotham: Righto. Let me know.

music21/analysis/harmonicFunction.py
jacobtylerwalls: Oops, you're right, I didn't look closely. Sorry!

jacobtylerwalls (approved):

music21/analysis/harmonicFunction.py
mscuthbert: Yeah, I'm just putting it in the environment.Environment('analysis.harmonicFunction') initializer going forward.

Also to note: I'm starting to import typing as t and using t.Optional, etc.  Remembering what has already been imported from typing and what hasn't was beginning to be too much of a mental stress.

mscuthbert:
Thanks.  Great work.

music21/analysis/harmonicFunction.py
MarkGotham: Thanks @mscuthbert and duly noted x2

PR #1131, opened 2021-09-16
Stop exporting superfluous meters and keys in multi-staff parts

### LINKED ISSUE #1129: MusicXML importer: PartStaffs end up with too many (2) TimeSignatures in a non-first Measure
**music21 version**

`7.1`   

**Problem summary**

Attached is a MusicXML file that music21 wrote from a stream containing two PartStaffs (and a StaffGroup relating them).  It's an Organ score, with 5 measures, each of which has a new time signature: 7/8, 8/8, 7/8, 8/8, and 11/16, respectively.  I believe the MusicXML is correct, but I might be wrong.  It displays correctly in any case.  When I import this MusicXML file back into music21, however, while the first measure's time signature looks right, the second measure's time signature (and I'm guessing the others as well) shows up in the measure twice (in each PartStaff)

If I export a different MusicXML file from an otherwise identical Score containing Parts instead of PartStaffs, and then import _that_, all is well (I get one time signature at the beginning of each Measure in each Part).

I would note that neither MusicXML file appears to have the extra time signatures in it, so I'm guessing this is a bug in MusicXML import, that only happens when a part has more than one staff.

I have attached both MusicXML files for your enjoyment (zipped).

[LesCorps_Parts.musicxml.zip](https://github.com/cuthbertLab/music21/files/7173870/LesCorps_Parts.musicxml.zip)
[LesCorps_PartStaffs.musicxml.zip](https://github.com/cuthbertLab/music21/files/7173871/LesCorps_PartStaffs.musicxml.zip)

jacobtylerwalls:
Fixes #1129

@gregchapman-dev would you like to test?

### DIFF
--- a/music21/mei/test_base.py
+++ b/music21/mei/test_base.py
@@ -3897,7 +3897,7 @@ class Test(unittest.TestCase):
         # ensure the right number and @n of parts
         self.assertEqual(4, len(actual.keys()))
         for eachN in expectedNs:
-            self.assertTrue(eachN in actual.keys())
+            self.assertTrue(eachN in actual)
         # ensure the measure number is set properly,
         #        there is one voice with one note with its octave set equal to the staff's @n,
         #        the right barline was set properly
@@ -4020,8 +4020,8 @@ class Test(unittest.TestCase):
         # ensure the right number and @n of parts (we expect one additional key, for the "rptboth")
         self.assertEqual(5, len(actual.keys()))
         for eachN in expectedNs:
-            self.assertTrue(eachN in actual.keys())
-        self.assertTrue('next @left' in actual.keys())
+            self.assertTrue(eachN in actual)
+        self.assertTrue('next @left' in actual)
         # ensure the measure number is set properly,
         #        there is one voice with one note with its octave set equal to the staff's @n,
         #        the right barline was set properly
--- a/music21/musicxml/helpers.py
+++ b/music21/musicxml/helpers.py
@@ -6,11 +6,14 @@
 # Authors:      Michael Scott Cuthbert
 #               Jacob Tyler Walls
 #
-# Copyright:    Copyright © 2013-2020 Michael Scott Cuthbert and the music21 Project
+# Copyright:    Copyright © 2013-2021 Michael Scott Cuthbert and the music21 Project
 # License:      BSD, see license.txt
 # ------------------------------------------------------------------------------
 import copy
+from typing import List
+
 from xml.etree.ElementTree import tostring as et_tostring
+from xml.etree.ElementTree import Element  # for typing
 
 def dumpString(obj, *, noCopy=False) -> str:
     r'''
@@ -178,6 +181,42 @@ def measureNumberComesBefore(mNum1: str, mNum2: str) -> bool:
     else:
         sortedSuffixes = sorted([m1Suffix, m2Suffix])
         return m1Suffix is sortedSuffixes[0]
+
+def childrenEqual(e1: Element, e2: Element, children: List[str]) -> bool:
+    '''
+    Tests whether the given child tags exist under each given element
+    and the texts of those child tags compare equal.
+
+    >>> from xml.etree.ElementTree import fromstring as El
+    >>> from xml.etree.ElementTree import SubElement
+    >>> a = El('<clef />')
+    >>> b = El('<clef />')
+    >>> a_sign = SubElement(a, 'sign')
+    >>> a_sign.text = 'G'
+    >>> b_sign = SubElement(b, 'sign')
+    >>> b_sign.text = 'F'
+
+    >>> from music21.musicxml.helpers import childrenEqual
+    >>> childrenEqual(a, b, children=['sign'])
+    False
+
+    >>> b_sign.text = 'G'
+    >>> childrenEqual(a, b, children=['sign'])
+    True
+
+    >>> childrenEqual(a, b, children=['sign', 'line'])
+    False
+    '''
+    for child in children:
+        e1_child = e1.find(child)
+        if e1_child is None:
+            return False
+        e2_child = e2.find(child)
+        if e2_child is None:
+            return False
+        if e1_child.text != e2_child.text:
+            return False
+    return True
 
 
 if __name__ == '__main__':
--- a/music21/musicxml/partStaffExporter.py
+++ b/music21/musicxml/partStaffExporter.py
@@ -629,6 +629,7 @@ class PartStaffExporterMixin:
         bump voice numbers if they conflict;
         account for <backup> and <forward> tags;
         skip <print> tags;
+        skip superfluous <time> and <key> tags;
         set "number" on midmeasure clef changes;
         replace existing <barline> tags.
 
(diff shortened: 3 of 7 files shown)


gregchapman-dev:
I'll try this out later today, thanks!

gregchapman-dev:
Your fix fixes my issue with doubled-up time signatures.  And gets me past my failed assertion, and now I can see more doubled stuff: a doubled MetronomeMark, and every Slur is doubled as well (in both staves).  It's possible this is just more fixes needed, or perhaps its a regression due to your fix (less likely).  Attached is the musicxml file produced with this fix in place.
[LesCorps_PartStaffs_2.musicxml.zip](https://github.com/cuthbertLab/music21/files/7180542/LesCorps_PartStaffs_2.musicxml.zip)

jacobtylerwalls:
Rebased to make cosmetic updates to comments and squash certain commits but not others, including the ones on the pylint-upgrades branch. Hoping that doing so was groovy since there was not a pending set of review comments.

Thanks for testing, Greg. I'll extract your comment about metronome marks into a new issue. The slurs, I can't reproduce. I parse the file you posted in your last comment and I get five slurs, which looks correct. Am I missing something?

gregchapman-dev:
I'll test again and watch where the extra slurs come from. Perhaps that's an unrelated issue in my code...

gregchapman-dev:
I will test again today.  Can we bump the version for this change, please (7.2.1?).  My humdrum importer will need to know whether the fix is there.

gregchapman-dev:
Tested latest multimeter-export branch (as of a few minutes ago).  Looks good; no extra timesig or metronome mark.

jacobtylerwalls:
Thanks for the retest.

I find this bit a bit surprising, though:

> no extra timesig or metronome mark.

I would have thought the extra metronome mark depended on #1136, which I just posted on another branch.

gregchapman-dev:
Sorry, my mistake.  Doubled metronome mark is still there.  (sigh)

music21/mei/test_base.py
```diff
@@ -3897,7 +3897,7 @@ def testMeasureIntegration1(self):
         # ensure the right number and @n of parts
         self.assertEqual(4, len(actual.keys()))
         for eachN in expectedNs:
-            self.assertTrue(eachN in actual.keys())
+            self.assertTrue(eachN in actual)
```
mscuthbert, line 3900: Please, let's get out of the habit of putting fixes unrelated to the issue at hand in the same PR.  My amount of review time is severely constrained in the first month or so of term and I want to make sure I can get through as many in half a day on Saturdays as I can.

music21/musicxml/helpers.py
```diff
@@ -179,6 +182,42 @@ def splitSuffix(measureNumber):
         sortedSuffixes = sorted([m1Suffix, m2Suffix])
         return m1Suffix is sortedSuffixes[0]
 
+def childrenEqual(e1: Element, e2: Element, children: List[str]) -> bool:
```
mscuthbert, line 185: this is great -- but I feel more that we should be removing in music21 before generating musicxml -- when musicxml is manipulated after generation it becomes much harder to debug -- "There were two <key> tags...oh, now there's one"

music21/musicxml/partStaffExporter.py
```diff
@@ -705,11 +706,48 @@ def moveMeasureContents(measure: Element, otherMeasure: Element, staffNumber: in
...
                 for midMeasureClef in elem.findall('clef'):
                     midMeasureClef.set('number', str(staffNumber))
+
+                # Determine how much of the <attributes> tag to move
+                target_measure_meters_or_none = otherMeasure.findall('attributes/time')
+                target_measure_keys_or_none = otherMeasure.findall('attributes/key')
+                source_measure_meter_or_none = elem.find('time')
+                source_measure_key_or_none = elem.find('key')
+                skippable_meter = False
+                skippable_key = False
```
mscuthbert, line 718: this works, but I don't like how much everything is moving into musicxml/Element manipulation.  I had qualms about it before when the partStaffExporter started doing it, but I want to stop before it becomes a norm in the musicxml manipulation.  We should identify skippable and identical music21 elements and remove them from the stream before outputting, not remove things from the XML DOM.

mscuthbert (changes requested):
I'd much prefer that these changes happen in the music21 stream before going out to XML Elements, not after it's been generated.  THANKS!

mscuthbert:
Let me clarify a bit more why I like the deepcopied stream to be manipulated before export rather than the XML output.  Suppose we find that MEI or if-it-ever-comes-out MNX needs something similar; then we have the same code -- "removeDuplicatePartElements()" that we can call to do it for them too.   Same with for Lilypond.  But this way, each exporter's duplicate finder has to be written separately and it becomes more code maintenance.

music21/mei/test_base.py
jacobtylerwalls: Wholeheartedly agree; I try to stay hygienic about such things. This happened because this was the first branch to fail CI on a new pylint version, and I think this part of the diff would have disappeared if I had, after those changes were merged, pulled again from master? Sorry! I'll keep the branches up to date!

jacobtylerwalls:
Makes total sense. I'll come up with a proposal for which layer to have this logic in. First guess: promote `joinableGroups()` out of the partStaffExporter (leaving a shim) into the stream module, and then have a makeNotation routine for removing duplicate elements, and then various exporters can use that routine at the beginning of their export.

mscuthbert:
> Makes total sense. I'll come up with a proposal for which layer to have this logic in. First guess: promote `joinableGroups()` out of the partStaffExporter (leaving a shim) into the stream module

It doesn't have to come out of musicxml at this point, because those other exporters are just hypothetical.  Unless it's really easy to do.  I just don't want more refactoring in the future than is necessary.

By `stream module` I hope to either makeNotation or a new module in the stream directory.  Stream is still huge!  My dream is to start reducing max file length by 1000 lines every few months until there's nothing over 3000 lines.

(for instance, what's Lyric doing in note.py?  It's just there because it was there in music21 v0.0.0.1 -- it'll need to be available as note.Lyric for a LONG time, but the actual code should be in text or its own lyric module etc.)

jacobtylerwalls:
Now I'm wondering how feasible it will be strip the meters and keys out.

The original problem is a key or meter change in the left hand in any measure after the first -- how to ensure we don't write it into the `<attributes>` tag.

The now current thought is to just remove the m21 object after its deepcopy is made during export (ignoring for the moment what to do about makeNotation=False) and before initial conversion to xml elements.

But I think that information will be essential to exporting the second staff correctly. All the time and padding and forward and backup -- will all be corrupted if there is a missing time signature object. The alternative would be, in every export routine that affects time and meter and key to check if the parent part participates in a joinable group and then run a context search to see if there is important information in the other partstaff? Just thinking aloud before I dig in (some point next month or so, likely). Not sure how we would do this.

jacobtylerwalls:
I guess following https://github.com/cuthbertLab/music21/pull/1136#issuecomment-927163129 we could have a "drop it on export" attribute, which we could set early in the export, leave the object in, and then check for it when the `<attributes>` tag is first written, or very late in the partStaffExporter in `moveMeasureContents()`. If we do it too early, I don't want to make it too easy to write out invalid musicmxl.

jacobtylerwalls:
Actually given the discussion on #1136 I'm thinking my first diagnosis is wrong and rather that the musicxml being emitted is now legitimate and this is, as Greg originally suggested, an issue in the importer using template() regardless of whether another mark exists in the other part.

jacobtylerwalls:
Closing per https://github.com/cuthbertLab/music21/issues/1129#issuecomment-927204102.

PR #1132, opened 2021-09-16
Postpone consider-using-f-string

jacobtylerwalls:
Postpone `consider-using-f-string` until existing codebase is updated (53 warnings). Leave `consider-iterating-dictionary` and `consider-using-in` enabled.

Cherry-picked from #1131 so that it can be merged faster and unblock other PRs.

### DIFF
--- a/music21/abcFormat/testFiles.py
+++ b/music21/abcFormat/testFiles.py
@@ -838,7 +838,7 @@ class Test(unittest.TestCase):
             s = translate.abcToStreamScore(ah)
             part = s.parts[0]
             self.assertFalse(part.getElementsByClass(chord.Chord),
-                             'Empty chord "%s" in Score' % abc_chord)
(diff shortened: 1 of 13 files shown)


music21/abcFormat/testFiles.py
```diff
@@ -838,7 +838,7 @@ def testAbc21Chords(self):
             s = translate.abcToStreamScore(ah)
             part = s.parts[0]
             self.assertFalse(part.getElementsByClass(chord.Chord),
-                             'Empty chord "%s" in Score' % abc_chord)
+                             f'Empty chord "{abc_chord}" in Score')
```
mscuthbert, line 841: great.  Most of these should be `{abc_chord!r}` (without the quotation marks) next time we refactor.

mscuthbert (commented):
Thanks for doing this.  All improvements.

PR #1136, opened 2021-09-18
MusicXML import: Stop duplicating MetronomeMarks in multi-staff parts

### LINKED ISSUE #1133: [Decision needed] Musicxml import produces MetronomeMark objects for each staff of multistaff part
**music21 version**
dev

**Problem summary**
Importing multistaff parts from musicxml (e.g. keyboard), produces MetronomeMark objects on each `PartStaff`. Great for querying. Exporting, though, will print the marks on both staves.

**Steps to reproduce**

Insert the tag @greg-chapman-dev mentioned in https://github.com/cuthbertLab/music21/pull/1131#issuecomment-921190744 into corpus/demos/two-parts.xml
```
   </direction>
      <direction placement="above">
        <direction-type>
          <metronome parentheses="no">
            <beat-unit>eighth</beat-unit>
            <per-minute>82</per-minute>
      </metronome>
    </direction-type>
    <staff>1</staff>
    <sound tempo="41" />
```
```
>>> s = corpus.parse('demos/two-parts', forceSource=True)
>>> len(s[tempo.MetronomeMark])
2
```

A design decision is needed here. Since inception music21 has been prioritizing extraction and querying of piano music over exporting it. I got export working in v.6.5 about a year ago, and now we're growing into the edge cases.

From a totally pedantic stream representation standpoint, there is no MetronomeMark in the second PartStaff, so maybe it shouldn't be there. I just tested if we can still query for it from the stream context. We can, using some of the new tools in v7, but it's involved:

I deleted the second metronome mark:
```python
>>> len(s.parts.last()["Measure"][0]["MetronomeMark"])  # I removed the superfluous tempo mark
0
>>> s.parts.last()["Measure"][0].getContextByClass('MetronomeMark') is None
True
>>> s.parts.last()["Measure"][0].getContextByClass('MetronomeMark', getElementMethod='all') 
<music21.tempo.MetronomeMark andantino Eighth=82.0>
```

So we need to decide on the stream representation.

jacobtylerwalls:
Fixes #1133

### DIFF
--- a/music21/corpus/demos/two-parts.xml
+++ b/music21/corpus/demos/two-parts.xml
@@ -146,6 +146,16 @@
     <voice>1</voice>
     <staff>1</staff>
    </direction>
+   <direction placement="above">
+    <direction-type>
+     <metronome parentheses="no">
+      <beat-unit>eighth</beat-unit>
+      <per-minute>82</per-minute>
+     </metronome>
+    </direction-type>
+    <staff>1</staff>
+    <sound tempo="41" />
+   </direction>
    <note default-x="88">
     <rest />
     <duration>128</duration>
--- a/music21/musicxml/xmlToM21.py
+++ b/music21/musicxml/xmlToM21.py
@@ -1710,6 +1710,7 @@ class PartParser(XMLParserBase):
             'GeneralNote',
             'KeySignature',
             'StaffLayout',
+            'TempoIndication',
             'TimeSignature',
         ]
 
@@ -6172,6 +6173,13 @@ class Test(unittest.TestCase):
         # has only sound tempo=x tag
         # s = converter.parse(testPrimitive.articulations01)
         # s.show()
+
+    def testImportMetronomeMarksC(self):
+        '''Import tempo into only the first PartStaff'''
+        from music21 import corpus
+        s = corpus.parse('demos/two-parts')
+        self.assertEqual(len(s.parts.first()[tempo.MetronomeMark]), 1)
+        self.assertEqual(len(s.parts.last()[tempo.MetronomeMark]), 0)
 
     def testImportGraceNotesA(self):
         # test importing from musicxml


jacobtylerwalls:
@greg-chapman-dev I know you are probably more in favor of removing the mark altogether, but there may be other projects depending on it being there, so I was thinking at least preventing display is a good compromise. (And consistent with MIDI import.) You could write to code to filter or delete marks with this attribute. Any thoughts?

gregchapman-dev:
Yeah, you’re right about my preference… 

In my Humdrum exporter I normally output hidden items (marked as hidden in the resulting Humdrum file, of course), since editors of library versions of historical (often handwritten) scores often put hidden things in the scores for librarian-ish reasons, and they really want them there.  They represent things that were implied by the score (perhaps because of some notational idiosyncrasies of a particular historical era), but not visible in the score.

So I’m not particularly happy to add humdrum exporter code that removes hidden MetronomeMarks that may have been put there on purpose by library folks.

I’m also not particularly happy to export hidden MetronomeMarks that were only put there in a MusicXML file by music21 for backward compatibility reasons.

I want to make sure I am understanding exactly what you are doing with these marks:

You are only adding hidden MetronomeMarks during export to MusicXML files from PartStaff-containing Scores, right?

So this isn’t really a music21 API/stream backward-compatibility issue (music21 doesn’t add these hidden MetronomeMarks during import, or during makeNotation(), or…).  It’s that you are trying to create MusicXML files that meet some peoples' possibly-existing expectations… But music21 is not the only creator of MusicXML files, so no matter what you do, those expectations are unreasonable, unless Finale, MuseScore, and others also have added these hidden MetronomeMarks for a while now.

OK, that’s my whole argument, best I can do. :-)

If you really can’t remove the mark altogether, I guess I could do something like “remove hidden MetronomeMarks from PartStaffs (not Parts) that are exactly the same as a visible MetronomeMark at the exact same time offset in another PartStaff (not Part)”.  Does that match what you’re currently doing?

Greg

jacobtylerwalls:
Ah, I made a faulty assumption. The "bug" of creating duplicate `MetronomeMark`s is not since inception, and not since the part staff exporter introduction in 6.5, it's only since the optimization 01597856f0ad7e4f113c05aab27706625aed82f8 (6.7). So only about nine months of code that might be depending on it.

The original proposal on this branch was to fix the bug but then add back some "implicit" `MetronomeMark` objects in the left hand to help folks query, but now I think there's no reason to have them. Folks can always use the snippet from #1133 if they need to query the tempo from the context.

I'll push a commit dropping the "implicit" marks I was inventing and then, I think we'll both be satisfied. Let's also see what Myke thinks. 👍🏻

jacobtylerwalls:
## Squash this PR ##
to avoid misleading commit messages

mscuthbert:
I think that the current (post 6.7) approach is better for analysis functions.   How long do measures 22-27 in the left hand of the piano last?  getContextByClass() does not know to look at the r.h. to get the current metronome mark.  I think that it needs to be in the music21 stream because of the way we use PartStaffs.  But I'm happy to have an attribute that says "drop it on output" or to have it dropped on output if it's identical to the same element in the previous PartStaff.

There's definitely no guarantee for any file format that loading into music21 and exporting will produce identical code.  Not producing two metronome marks for the piano would be very nice though.

jacobtylerwalls:
> getContextByClass() does not know to look at the r.h. to get the current metronome mark. 

Actually, maybe, I think it does:

```
>>> len(s.parts.last()["Measure"][0]["MetronomeMark"])  # I removed the superfluous tempo mark
0
>>> s.parts.last()["Measure"][0].getContextByClass('MetronomeMark') is None
True
>>> s.parts.last()["Measure"][0].getContextByClass('MetronomeMark', getElementMethod='all') 
<music21.tempo.MetronomeMark andantino Eighth=82.0>
```

And so my thought would be, maybe we make this easier to do with a property or something?

Otherwise I'm not sure I see a solution -- numberImplicit is least bad, though we're still adding things; and I think hideObjectOnPrint doesn't even work (we ran this by Michael Good IIRC).

mscuthbert:
Ah, yes, if hideObjectOnPrint doesn't work for musicxml output then it IS better to not duplicate.  So is the PR good to merge then?

jacobtylerwalls:
Yeah, it's ready for merge.

Issue #909, opened 2021-03-10
Articulations of notes in chords are uncorrectly written in MusicXML


louisbigo:
**music21 version**

`6.6.0`

**Problem summary**
It seems that string/fret articulations are uncorrectly written (or not written at all) in musicXML when applied to a Note which is in a Chord (although it works well for single notes). Maybe its because in case of chords, articulations are likely to concern the whole chord (but I'm not sure of that, in any case this obviously does not apply for string/fret informations which are distinct for each note of a chord).

I found that the problem can be solved by adding the following lines in the function `noteToNotations` in `m21ToXml.py`. This is really ad-hoc to string/fret articulations and I'm sure it can be corrected in a much cleaner way. I still put the code here in case it might be useful to someone, and in case it can help to improve the code properly.

```
if chordParent is not None:
    for a in n.articulations:
        if 'TechnicalIndication' in a.classes:
            if 'StringIndication' in a.classes or 'FretIndication' in a.classes:
                if mxTechnicalMark is None:
                    mxTechnicalMark = Element('technical')
                mxTechnicalMark.append(self.articulationToXmlTechnical(a))
```

jacobtylerwalls:
https://github.com/cuthbertLab/music21/blob/60d91ddaf4a7398337181a6168c9ec45abcb93b3/music21/articulations.py#L455-L460

Speaking only for myself, I agree it would be wise to loosen this at least as far as StringIndication and FretIndication. I would try something like this:

```diff
diff --git a/music21/musicxml/m21ToXml.py b/music21/musicxml/m21ToXml.py
index 7d06c39c1..295821392 100644
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -4006,11 +4006,12 @@ class MeasureExporter(XMLExporterBase):
                 else:
                     notations.append(mxExpression)
 
-        # apply all articulations apart from fingerings only to first note of chord
+        # allow all articulations on first note of chord
+        # else, only allow certain ones (Fingering, FretIndication, StringIndication)
         applicableArticulations = []
         fingeringNumber = 0
         for a in chordOrNote.articulations:
-            if 'Fingering' in a.classSet:
+            if not a.classSet.isdisjoint(('Fingering', 'FretIndication', 'StringIndication')):
                 if fingeringNumber == noteIndexInChord:
                     applicableArticulations.append(a)
                 fingeringNumber += 1
```

[...]

PR #1142, opened 2021-09-28
Hammer-on and Pull-off as Spanners

louisbigo:


music21/musicxml/m21ToXml.py
```diff
@@ -4823,7 +4893,7 @@ def articulationToXmlTechnical(self, articulationMark):
             musicXMLTechnicalName = 'other-technical'
 
         # TODO: support additional technical marks listed above
-        if musicXMLTechnicalName in ('hammer-on', 'pull-off', 'bend', 'hole', 'arrow'):
+        if musicXMLTechnicalName in ('bend', 'hole', 'arrow'):
```
jacobtylerwalls, line 4896: Since this is now unreachable ('bend' etc are not in the dict):

```suggestion
        if musicXMLTechnicalName in ('bend', 'hole', 'arrow'):  # pragma: no cover
```

This will exempt from test coverage. But leaving this in is good in case we add these classes to the dict but don't fully implement in future.

jacobtylerwalls (commented):
Thanks for this contribution.

About the failing test: I think we can leave the example for translating a hammer-on, but remove the remarks about "Same with technical marks not yet supported." which I added in #780. We don't need an example for unsupported marks since there shouldn't be a way to produce `<other-technical />` now since Arrow, Hole, and Bend are not even in `TECHNICAL_MARKS_REV`.

So in sum, I suggest keeping this much:

```python
        >>> h = articulations.HammerOn()
        >>> mxOther = MEX.articulationToXmlTechnical(h)
        >>> MEX.dump(mxOther)
```

because we don't want either `<fret>0</fret>` or `<other-technical>`. This might require returning early from `articulationToXmlTechnical()` -- which we reach if the articulation is in the note's array of articulations (which I think is reasonable to expect). I left some comments that should help guide us toward that.

Thanks again!

music21/musicxml/xmlToM21.py
```diff
@@ -3939,6 +3939,12 @@ def xmlNotationsToSpanners(self, mxNotations, n):
                                     )
             self.setColor(mxObj, slur)
 
+        for mxObj in mxNotations.findall('technical/hammer-on'):
+            self.xmlOneSpanner(mxObj, n, articulations.HammerOn)
+
+        for mxObj in mxNotations.findall('technical/pull-off'):
+            self.xmlOneSpanner(mxObj, n, articulations.PullOff)
+
```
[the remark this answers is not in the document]
louisbigo: Ok I understand we might want to add them in the articulations. But we can not use `xmlTechnicalToArticulation` as they are not in TECHNICAL_MARKS anymore. So I think we would like hammer-on to be in TECHNICAL_MARKS for reading (so it would automatically benefit from the standard way to be appended to the articulations list) but not for writing (as we don't want it to be written as standard technicals). Maybe the ultimate perspective would be to have 2 distinct lists TECHNICAL_MARKS, one for reading and one for writing. But I guess this would require some important code modifications which are unnecessary at this stage.

So for now when reading I'll explicitly  append them in the articulation list which seems to be the most reasonable choice at the moment.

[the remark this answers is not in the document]
jacobtylerwalls: Thanks, that seems like a wise course of action. When CI passes this will be mergeable, I think!

music21/musicxml/m21ToXml.py
```diff
@@ -3403,6 +3403,69 @@ def objectAttachedSpannersToNotations(self, obj, objectSpannerBundle=None):
...
+        technicals = []
+        if objectSpannerBundle is not None:
+            sb = objectSpannerBundle
+        else:
+            sb = self.objectSpannerBundle
+
+        if not sb:
+            return technicals
+
+        for su in sb.getByClass('articulations.HammerOn'):
```
jacobtylerwalls, line 3443:
```suggestion
        for su in sb.getByClass(articulations.HammerOn):
```

Ah, sorry I meant this, the evaluated class name, not a string. I suggested it because I supposed it would run every note in musicxml export and needed to be fast, but looking just above we have a short-circuit, so I'm not even concerned about it at this point. But I think this change is was caused more tests to fail.

music21/musicxml/m21ToXml.py
```diff
@@ -3403,6 +3403,69 @@ def objectAttachedSpannersToNotations(self, obj, objectSpannerBundle=None):
...
+                mxHammerOn.set('type', 'start')
+                mxHammerOn.text = 'H'
+            elif su.isLast(obj):
+                mxHammerOn.set('type', 'stop')
+            else:
+                continue
+            mxHammerOn.set('number', "1")
+            technicals.append(mxHammerOn)
+
+        for su in sb.getByClass('articulations.PullOff'):
```
jacobtylerwalls, line 3455:
```suggestion
        for su in sb.getByClass(articulations.PullOff):
```

music21/musicxml/m21ToXml.py
```diff
@@ -4823,7 +4889,7 @@ def articulationToXmlTechnical(self, articulationMark):
             musicXMLTechnicalName = 'other-technical'
```
jacobtylerwalls, line 4889: We will need to return early to avoid getting the `<other-technical>`.

music21/musicxml/m21ToXml.py
```diff
@@ -4800,13 +4870,9 @@ def articulationToXmlTechnical(self, articulationMark):
         >>> MEX.dump(mxOther)
         <other-technical>unda maris</other-technical>
 
-        Same with technical marks not yet supported.
-        TODO: support HammerOn, PullOff, Bend, Hole, Arrow.
-
         >>> h = articulations.HammerOn()
         >>> mxOther = MEX.articulationToXmlTechnical(h)
         >>> MEX.dump(mxOther)
-        <other-technical />
```
jacobtylerwalls, line 4809:
```suggestion
        >>> MEX.dump(mxOther) is None
        True
```
Cosmetic thing I forgot to mention: so that it doesn't look in the Sphinx-built version of docs that something is missing or someone forgot to write the result.

music21/musicxml/m21ToXml.py
jacobtylerwalls: Let's do it. Maybe the hammer-on doctest can also be moved to `noteToNotations` if not too much effort.

jacobtylerwalls: Ah. I see. The doctest is useless now because we can't export unless we have a Stream object with the spanner in its spanner bundle. Okay. This is getting too complex for a doctest; we're just trying to see that the `<other-technical>` went away. Maybe you can add a couple lines to `testArticulationSpecialCases()`: add the HammerOn to the note, and then assert
```python
        self.assertIsNone(tree.find('.//other-technical'))
```
Thanks for bearing with me; once we get this we'll have CI passing and should be good.

music21/musicxml/xmlToM21.py
```diff
@@ -3939,6 +3939,14 @@ def xmlNotationsToSpanners(self, mxNotations, n):
                                     )
             self.setColor(mxObj, slur)
 
+        for mxObj in mxNotations.findall('technical/hammer-on'):
+            self.xmlOneSpanner(mxObj, n, articulations.HammerOn)
+            n.articulations.append(articulations.HammerOn())
+
+        for mxObj in mxNotations.findall('technical/pull-off'):
+            self.xmlOneSpanner(mxObj, n, articulations.PullOff)
+            n.articulations.append(articulations.PullOff())
```
jacobtylerwalls, line 3948: Ah. We need to append the same articulation object returned by xmlOneSpanner, not a fresh object.

jacobtylerwalls:
Green CI!

I was expecting some weirdness with chords, perhaps, since we haven't done #909. But I think this is too basic and too easy to fix:

```python
>>> n = note.Note()
>>> n2 = note.Note()
>>> h = articulations.HammerOn([n, n2])
>>> n.articulations.append(h)
>>> n2.articulations.append(h2)
>>> c = chord.Chord([n])
>>> c2 = chord.Chord([n2])
>>> s = stream.Stream([h, c, c2])
>>> s.show()  # no hammer on
```

Haven't tested, but hoping something like...
```diff
diff --git a/music21/musicxml/m21ToXml.py b/music21/musicxml/m21ToXml.py
index 8848efc3b..b658a0f71 100644
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -4456,7 +4456,12 @@ class MeasureExporter(XMLExporterBase):
         # TODO: glissando
         # TODO: slide
 
-        spannerTechnicals = self.objectAttachedSpannersToTechnicals(n)
+        if isSingleNoteOrFirstInChord and chordParent is not None:
+            spannerTechnicals = self.objectAttachedSpannersToTechnicals(chordParent)
+        elif chordParent is not None:
+            spannerTechnicals = None
+        else:
+            spannerTechnicals = self.objectAttachedSpannersToTechnicals(n)
         if spannerTechnicals:
             if mxTechnicalMark is None:
                 mxTechnicalMark = Element('technical')
```

jacobtylerwalls:
> Could we consider linking hammer-on/pull-off to notes (possibly inside chords) exclusively, but not to chords ? I'm not sure if it would make sense to include a chord within a (spanner) HammerOn.

Yes. I don't think we need to support chord-level hammerOns, certainly not in this PR. Just being able to export them from a contained note.

I re-read #909 and saw that we are not currently supporting exporting of articulations from individual chord notes. But we are allowing slurs (spanners) from individual notes. So I'm hoping there's a not-invasive way to support it as a spanner, and not deal with the articulations issues. If not, feel free to say you think it's out of scope for now. Thanks again.

jacobtylerwalls:
> Should we investigate further in that direction ?

Thanks for looking. I now see that slurs between individual chord notes aren't exported (which makes sense, since it's nonstandard to have to slurs on separate notes of a chord), but with hammer-ons we have more of a valid use case to export spanners found on individual notes. (Is that correct?)

I think that should be a follow-up PR to deal with collecting the spanners from chord notes.

Thanks again for this great contribution! Will see if Myke has any comments.

music21/articulations.py
```diff
@@ -583,10 +584,16 @@ class FrettedPluck(FretIndication, Fingering):
     '''
     pass
 
-class HammerOn(FretIndication):
+class HammerOn(spanner.Spanner, TechnicalIndication):
```
mscuthbert, line 587: Should these also be FretIndications as well?

music21/musicxml/m21ToXml.py
```diff
@@ -3403,6 +3402,80 @@ def objectAttachedSpannersToNotations(self, obj, objectSpannerBundle=None):
...
+        '''
+        return a list of <technical> from spanners related to the object that should appear
+        in the technical tag of the notations tag (hammer-on, pull-off etc.)
+
+        >>> n0 = note.Note('C')
+        >>> n1 = note.Note('D')
+        >>> n2 = note.Note('F')
+        >>> n3 = note.Note('E')
+        >>> hammerOn01 = articulations.HammerOn([n0,n1])
+        >>> pullOff23 = articulations.PullOff([n2,n3])
```
mscuthbert, line 3415: spaces after commas

music21/musicxml/m21ToXml.py
```diff
@@ -4799,14 +4888,6 @@ def articulationToXmlTechnical(self, articulationMark):
...
         >>> MEX.dump(mxOther)
         <other-technical>unda maris</other-technical>
-
-        Same with technical marks not yet supported.
-        TODO: support HammerOn, PullOff, Bend, Hole, Arrow.
-
-        >>> h = articulations.HammerOn()
-        >>> mxOther = MEX.articulationToXmlTechnical(h)
-        >>> MEX.dump(mxOther)
-        <other-technical />
```
mscuthbert, line 4809: If Bend, Hole, Arrow are still not supported, then we should leave this but just change the example to one of those.

music21/musicxml/m21ToXml.py
```diff
@@ -4823,7 +4904,7 @@ def articulationToXmlTechnical(self, articulationMark):
             musicXMLTechnicalName = 'other-technical'
 
         # TODO: support additional technical marks listed above
-        if musicXMLTechnicalName in ('hammer-on', 'pull-off', 'bend', 'hole', 'arrow'):
+        if musicXMLTechnicalName in ('bend', 'hole', 'arrow'):  # pragma: no cover
```
mscuthbert, line 4907: remove pragma: no cover -- following comment above.

louisbigo:
Ok so that would mean removing (in xmlToM21.py, l.3942) the two commented lines below (or should we remove the whole block as you mentioned @jacobtylerwalls ? but in this case I don't see how hammerOn/PullOff could be included in the stream, even as spanners, correct me if I'm wrong)

```
        for mxObj in mxNotations.findall('technical/hammer-on'):
            ho = self.xmlOneSpanner(mxObj, n, articulations.HammerOn)
            # n.articulations.append(ho)

        for mxObj in mxNotations.findall('technical/pull-off'):
            po = self.xmlOneSpanner(mxObj, n, articulations.PullOff)
            # n.articulations.append(po)
```

jacobtylerwalls:
You're quite right--I misspoke!

jacobtylerwalls:
Hi @mscuthbert any concerns if I squash and merge sometime these next couple weeks?

We left the chord behavior undefined (status quo), and didn't import ho/po's into a note's articulations or expressions array, and have not moved these classes out of the articulations module. (I get that to do so is probably a tolerable break of Backwards Compatibility given that the classes did nothing, but I'm not sure I see the payoff of breaking BC to move it into spanners, either. Something to be said for the fact that users might look in the articulations module to find them!)

mscuthbert:
This discussion got lost in my Thanksgiving-New Year crush.  Looking at it again soon.

mscuthbert:
I've read this again (thanks for working on first just the single-note hammer on first; we'll get to chords later) and looked through the literature on hammer-on/pull-off in general.  I'm still not sure that putting spanners inside articulations-list on a Note is the right way to do this -- we don't do this with multiple-note Tremolos, for instance -- they're in Spanners stored in the measure or Part object.  

(keeping everything simple here and referring to Notes alone and not yet Chords, etc.)

Spanners contain (span) Notes.  Notes should not contain spanners.  Articulations in particular are objects attached to a single Note.

I think that there are multiple things happening in the playing and notation of a hammer on or pull off and, despite being more complex to create, should probably be represented by multiple objects:

Notes that are to be represented as slurred together, showing a hammer-on/pull-off chain should be in a Spanner (probably a HammerPullChain subclass of Slur [or HammerPullSlur]).  And that Spanner goes in the Measure/Part object (spannerBundles can figure out the right place).   Any Note can figure out if it's in a HammerPullChain by doing n.getSpannerSites('HammerPullChain').  Etc.   This should probably go in the spanner module.

The actual HammerOn, PullOff non-spanners can go in articulations.  Probably the musicxml way of specifying start and stop variants of the same articulation is the best way, but I'm not entirely sure.  Definitely this is where the configuration about whether to display H or P (or something else) should go.

---
For Chords where only some notes are HammerOn/Off, we could put in the HammerPullChain object a dictionary mapping Chords in the spanner to a list of notes (or note indexes) that the grouping applies to.

The HammerPullChain object could have a helper method that adds appropriate articulation objects to the Notes/Chords automatically.

Thanks for giving me the time to think through this more.  Spanners are already one of the most complex parts of music21, and a solution that makes all this possible without making Spanners even more complex is important to me.

mscuthbert (changes requested):
See previous comment about HammerPullChain etc. for how I think this should be implemented.

jacobtylerwalls:
> I'm still not sure that putting spanners inside articulations-list on a Note is the right way to do this -- we don't do this with multiple-note Tremolos, for instance -- they're in Spanners stored in the measure or Part object.

Thanks for the careful review -- just want to flag a potential point of crossed wires. The above (⬆️ ) is not what the PR currently does; I do spot one test that _suggests_ that it might, but it's to just "be permissive" during export on behalf of users who might have tried to execute that pattern themselves. A comment over the test is probably in order to clarify.

jacobtylerwalls:
> Notes that are to be represented as slurred together, showing a hammer-on/pull-off chain should be in a Spanner (probably a HammerPullChain subclass of Slur [or HammerPullSlur]). And that Spanner goes in the Measure/Part object (spannerBundles can figure out the right place). Any Note can figure out if it's in a HammerPullChain by doing n.getSpannerSites('HammerPullChain'). Etc. This should probably go in the spanner module.

The PR does roughly this now, just without renaming the classes "chain", without moving the classes to the spanner module, and without subclassing Slur. But those seem easily done.

> The actual HammerOn, PullOff non-spanners can go in articulations. Probably the musicxml way of specifying start and stop variants of the same articulation is the best way, but I'm not entirely sure. Definitely this is where the configuration about whether to display H or P (or something else) should go.

Do you have a sense of what the other objects would represent? I was thinking start and stop would just naturally be expressed by the start and stop notes of the spanner, and text on the spanner also, since different text on different start/stop objects would add complexity to have to diff them. Let me know if I'm missing something obvious!

mscuthbert:
> > I'm still not sure that putting spanners inside articulations-list on a Note is the right way to do this -- we don't do this with multiple-note Tremolos, for instance -- they're in Spanners stored in the measure or Part object.
> 
> Thanks for the careful review -- just want to flag a potential point of crossed wires. The above (⬆️ ) is not what the PR currently does;

Screen Shot 2022-01-11 at 15 38 20

These are spanners attached to notes, not notes attached to spanners.

jacobtylerwalls:
Right, that's the "permissive export" test case I was referring to: intended for the case that _users_ have stuck HammerOns on a note's articulation array--not that the documentation says to, or that the musicxml importer did so, but just accounting for the fact that they might have _tried_ to. In that case, this branch prevents stray `<other-technical>` tags from being written out. (This branch is "if not hammer on... keep on doing stuff"). Of course, we're not responsible for making the exporter safe against all kinds of things a user might have done against the documented patterns--they're on their own if they put a Stream in a `.articulations`--but this just seemed like an easy, non-far-fetched thing to do. (For instance, some discussion above about whether an m21 user would know where to find in the docs that slurs don't go in these arrays.)

But the value of the "do no harm" branch has already "done harm" by slowing us down here, so I would be glad to volunteer myself to rewrite the comment for clarity!

music21/musicxml/m21ToXml.py
```diff
@@ -6893,16 +6974,19 @@ def testFullMeasureRest(self):
         self.assertEqual(rest.get('measure'), 'yes')
 
     def testArticulationSpecialCases(self):
-        n = note.Note()
+        n1 = note.Note()
         a = articulations.StringIndication()
-        n.articulations.append(a)
+        n1.articulations.append(a)
+        h = articulations.HammerOn()
+        n1.articulations.append(h)
```
jacobtylerwalls, line 6981: This will need a blaring warning/docstring saying *don't do this*, but if you did...

music21/musicxml/m21ToXml.py
```diff
@@ -6893,16 +6974,19 @@ def testFullMeasureRest(self):
...
 
         # Legal values for StringIndication begin at 1
         self.assertEqual(a.number, 0)
         # Use GEX to go through wellformed object conversion
-        gex = GeneralObjectExporter(n)
+        gex = GeneralObjectExporter(n1)
         tree = et_fromstring(gex.parse().decode('utf-8'))
         self.assertIsNone(tree.find('.//string'))
+        self.assertIsNone(tree.find('.//other-technical'))
```
jacobtylerwalls, line 6989: ... then at least we didn't write out `<other-technical>`.

music21/musicxml/m21ToXml.py
```diff
@@ -4341,7 +4415,9 @@ def noteToNotations(self, n, noteIndexInChord=0, chordParent=None):
                     applicableArticulations.append(a)
                 fingeringNumber += 1
             elif isSingleNoteOrFirstInChord:
-                applicableArticulations.append(a)
+                # hammer-on/pull-off are written from their Spanner representation
+                if not isinstance(a, (articulations.HammerOn, articulations.PullOff)):
+                    applicableArticulations.append(a)
```
jacobtylerwalls, line 4420: Should add a blaring comment something like "safe to ignore any HammerOn or PullOff a user might have put in this array"

music21/spanner.py
```diff
@@ -2684,6 +2684,29 @@ def testGetSpannedElementIds(self):
...
+    def testHammerOnPullOff(self):
+        from music21 import converter
+        from music21.musicxml import testPrimitive
+
+        s = converter.parse(testPrimitive.notations32a)
+
+        num_hammer_on = len(s.flatten().getElementsByClass('HammerOn'))
+        num_pull_off = len(s.flatten().getElementsByClass('PullOff'))
+        self.assertEqual(num_hammer_on, 1)
+        self.assertEqual(num_pull_off, 1)
```
jacobtylerwalls, line 2696: This is how importing currently works.

music21/musicxml/xmlToM21.py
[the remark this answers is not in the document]
jacobtylerwalls: We reversed course in 72b79c4 and decided not to put these objects in the articulations array.

jacobtylerwalls:
@mscuthbert I just fixed merge conflicts and added some comments. I'm happy to "take over" and push any requested changes to get this in.

Big picture: I think we're on the same page these should be spanners (they are), and that they shouldn't be put on the `.articulations` array of a note (they aren't--the importer doesn't do it--but there _is_ some code to guard against the possibility of a user doing that _themselves_)

The new `HammerOn` and `PullOff` classes could be moved to the spanner module if wanted. Easy find/replace. They do inherit from `articulations.TechnicalIndication` though. Can't figure out the best place for them. I documented more clearly that they don't go on `.articulations`.

mscuthbert:
Thanks!  @jacobtylerwalls -- finally had a chance to review and looks great.  Merging.

mscuthbert (approved):

PR #1148, opened 2021-10-14
Accidentals never and if-absolutely-necessary

mscuthbert:
Add more logic for makeAccidentals for displayType "never"

Add "if-absolutely-necessary" displayType

Simplify some parts of the accidental making.

Do not merge until I have a chance to write tests.

(Ignore branch name -- did not get to it in this PR)


music21/pitch.py
```diff
@@ -4673,62 +4725,66 @@ def updateAccidentalDisplay(
...
+                    acc.displayStatus = True
                     return
                 else:  # names are the same, skip this line of questioning
                     break
+
         # nope, no previous pitches in this octave and register, now more complex things...
+        # but first handle 'if-absolutely-necessary' case.
+        if acc and acc.displayType == 'if-absolutely-necessary':
+            acc.displayStatus = False
+            return
```
jacobtylerwalls, line 4768: I think this is too early to short-circuit, if `if-absolutely-necessary` also needs to show a necessary natural contradicting the ks. If it's not what it's supposed to do, then maybe reword `only-for-cancelling-in-register`? (not that I love that phrasing). Just that cancelling a ks seems absolutely necessary.

pr:
```python
>>> m = stream.Measure([key.Key('C-'), note.Note('G#'), note.Note('Cn')])
>>> p = stream.Part(m)
>>> m.notes[-1].pitch.accidental.displayType = 'if-absolutely-necessary'
>>> p.show()
```

Screen Shot 2021-10-20 at 10 09 34 PM


***
Maybe this means organizing the cases differently so we can short-circuit somewhere.

music21/pitch.py
```diff
@@ -4650,20 +4699,23 @@ def updateAccidentalDisplay(
...
-            elif self.accidental.displayStatus is None:  # not set; need to set
+            elif acc.displayStatus is None:  # not set; need to set
                 # configure based on displayStatus alone, continue w/ normal
                 pass
-            elif self.accidental.displayStatus in (True, False):
+            elif acc.displayStatus in (True, False):
                 return  # exit: already set, do not override
 
+        if acc and acc.displayType == 'never':
+            acc.displayStatus = True
```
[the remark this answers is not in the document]
mscuthbert: Because I'm an idiot! :-D

music21/pitch.py
mscuthbert: THANKS!

PR #1151, opened 2021-10-16
Avoid calling globals in `instrument.fromString()`

jacobtylerwalls:
Just a small modernization task since this code predated `importlib` in Python 3.


mscuthbert:
thanks!

PR #1158, opened 2021-10-30
MIDI import: create Unpitched and PercussionChord instances

jacobtylerwalls:
Part of the percussion story.

- 3d9a8ed: `ChordBase.add()` took the argument `runSort`, but could lead to errors for PercussionChord (no attribute `sortAscending`). Removed `runSort` from `ChordBase.add()` signature and implemented a `Chord.add()` to get it back where desired -- with some docs duplication. If the docs duplication is not desired, we can remove the examples from ChordBase.add() and just have them on Chord.add().
- 86c7952: Raise TypeError if Unpitched provided to Chord
- 8279cd0: Create Unpitched and PercussionChord instances from MIDI import; move tie and volume and setVolume to ChordBase (needed for PercussionChord)


music21/midi/translate.py
```diff
@@ -267,7 +269,60 @@ def music21ObjectToMidiFile(
...
+        raise TypeError(f'Expected subclass of note.NotRest; got {preferredClass}')
+
+    if eOn.channel == 10:
+        if preferredClass in (chord.Chord, percussion.PercussionChord):
+            preferredClass = percussion.PercussionChord
+        else:
+            preferredClass = note.Unpitched
+
+    if inputM21 is not None:
+        if not isinstance(inputM21, preferredClass):
```
mscuthbert, line 299: if we have other PRs open to speeding up MIDI, is this isinstance check needed for the rare case that someone gives an incorrect class?  It's now a private routine.

music21/midi/translate.py
```diff
@@ -349,28 +411,23 @@ def midiEventsToNote(eventList, ticksPerQuarter=None, inputM21=None) -> note.Not
...
-    n.volume.velocityIsRelative = False  # not relative coming from MIDI
+    nr = _constructOrUpdateNotRest(
+        eOn, tOn, tOff, ticksPerQuarter, inputM21=inputM21, preferredClass=note.Note)
+
+    if isinstance(nr, note.Note):
+        nr.pitch.midi = eOn.pitch
+    elif isinstance(nr, note.Unpitched):
+        display_pitch = pitch.Pitch(eOn.pitch)
+        nr.displayStep = display_pitch.step
+        nr.displayOctave = display_pitch.octave
```
mscuthbert, line 422: this does not seem correct -- we would rarely want to display a percussion note at MIDI pitch.  I think that somewhere in music21 there's a mapper from instruments to positions on staves.

music21/midi/translate.py
```diff
@@ -522,25 +583,35 @@ def midiEventsToChord(eventList, ticksPerQuarter=None, inputM21=None):
         <music21.midi.MidiEvent NOTE_ON, track=1, channel=None, pitch=46, velocity=94>]
 
     Changed in v.7 -- Uses the last DeltaTime in the list to get the end time.
+    Changed in v.7.3 -- Returns a :class:`~music21.percussion.PercussionChord` if
+    any event is on channel 10. Returns None if `inputM21` provided.
+
+    >>> me2.channel = 10
+    >>> midi.translate.midiEventsToChord([dt1, me1, dt2, me2, dt3, me3, dt4, me4])
+    <music21.percussion.PercussionChord unpitched[A2] unpitched[B2]>
```
mscuthbert, line 591: just as a further refinement -- instead of showing display step on PercussionChord it should be a list of instruments:

<music21.percussion.PercussionChord [Bass Drum, Snare]>

or if mixed with notes:

<music21.percussion.PercussionChord [Vibraslap, D4, D8, Snare]>

music21/midi/translate.py
```diff
@@ -522,25 +583,35 @@ def midiEventsToChord(eventList, ticksPerQuarter=None, inputM21=None):
...
     '''
     tOn: int = 0  # ticks
     tOff: int = 0  # ticks
 
     if ticksPerQuarter is None:
         ticksPerQuarter = defaults.ticksPerQuarter
 
-    from music21 import pitch
```
mscuthbert, line 532: thank you!  There was a very old reason for this, now long gone.

music21/midi/translate.py
```diff
@@ -522,25 +583,35 @@ def midiEventsToChord(eventList, ticksPerQuarter=None, inputM21=None):
...
     # this is a format provided by the Stream conversion of
     # midi events; it pre groups events for a chord together in nested pairs
     # of abs start time and the event object
     if isinstance(eventList, list) and eventList and isinstance(eventList[0], tuple):
         # pairs of pairs
         for onPair, offPair in eventList:
             tOn, eOn = onPair
+            if firstOn is None:
+                firstOn = eOn
+            any_channel_10 |= eOn.channel == 10
```
mscuthbert, line 614: Please rewrite to not use bitwise in-place with a comparison:

```
if not any_channel_10 and eOn.channel == 10:
     any_channel_10 = True
```
or even just 
```
if eOn.channel == 10: 
    any_channel_10 = True
```

setting a variable multiple times is faster than trying to figure out what this code does.

mscuthbert (changes requested):
LGTM!  Feel free to merge after making the changes below.  (The repr for PercussionChord does not need to be in this PR; just wanted to comment now)

music21/midi/translate.py
jacobtylerwalls: Thanks, looks like it's the `.percMapPitch` of an `UnpitchedPercussion` instance. So there could be a cleanup routine that could live elsewhere that MIDI parsing calls at the end to set displayPitch from the percMapPitch of the closest instrument. I'll track on another issue.

jacobtylerwalls: Never mind, I think I can do it quickly on this branch.

jacobtylerwalls: Actually, no I can't find staff positions anywhere. `percMapPitch` _isn't_ it.

PR #1159, opened 2021-10-30
Export style attributes from chord members

jacobtylerwalls:
Previously, only color information was exported from individual chord members (Note instances), not other style attributes.

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -3508,8 +3508,8 @@ class MeasureExporter(XMLExporterBase):
             chordOrN = n
         else:
             chordOrN = chordParent
-            # Ensure color is read from `n`, since only `chordOrN` is handled below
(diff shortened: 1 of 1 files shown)


music21/musicxml/m21ToXml.py
```diff
@@ -3508,8 +3508,8 @@ def noteToXml(self, n: note.GeneralNote, noteIndexInChord=0, chordParent=None):
             chordOrN = n
         else:
             chordOrN = chordParent
-            # Ensure color is read from `n`, since only `chordOrN` is handled below
-            self.setColor(mxNote, n)
+            # Ensure style is read from `n` before reading from `chordOrN`
+            self.setPrintStyle(mxNote, n)  # sets color
```
jacobtylerwalls, line 3512: Thus, anything that should be overwritten by the chord's style, e.g. stem style, will be.

PR #1162, opened 2021-11-07
Clean up temporary attributes set by braille translator

jacobtylerwalls:
Fixes side-effects on streams left by braille translator, and removes related workarounds in test suite.

Simple fix for now without committing to any refactoring.


mscuthbert (approved):
Wonderful.  Thank you!

PR #1166, opened 2021-11-13
Fix export of chord symbols with realized durations

### LINKED ISSUE #963: harmony.realizeChordSymbolDurations() causes misalignment when XML exporting.
harmony realizeChordurations OFF-1
harmony RealizeChordDurations ON-1
[ChordSymbol Durations Script and MXML file.zip](https://github.com/cuthbertLab/music21/files/6327941/ChordSymbol.Durations.Script.and.MXML.file.zip)

:
Per thread from the mailing list:

I had a quick look -- this is reproducible without involving templates. If you set writeAsChord = False (which harmony.realizeChordSymbolDurations() does) then musicxml export gets gummed up if there are gaps between them and no rests to fill the gap. Please open a GitHub issue.

A workaround would be to fill with rests (use the argument provided when making the template, or just makeRests(fillGaps=True)), and then recurse over the stream and set .style.hideObjectOnPrint = True on each rest if you don't want to see them.

makeNotation() is run during musicxml export, and fixes lots of things that might gum up musicxml display, but making rests isn't one of them. Which is a neat feature for someone making a worksheet. Tangent: I've been toying with the idea of adding a makeNotation=False argument to show/write so that you can "freeze" what you have and not let musicxml export make more choices for you. But irrelevant here because makeNotation doesn't (currently) make rests.

Thanks for the report.
Jacob

jacobtylerwalls:
Fixes #963

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -3059,7 +3059,7 @@ class MeasureExporter(XMLExporterBase):
             root.append(sp)
 
         classes = obj.classes
-        if 'GeneralNote' in classes:
+        if 'GeneralNote' in classes and 'Harmony' not in classes:
             self.offsetInMeasure += obj.duration.quarterLength
 
         # turn inexpressible durations into complex durations (unless unlinked)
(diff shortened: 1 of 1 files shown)


mscuthbert:
This seems correct for "writeAsChord = False" but not for "writeAsChord = True"  I don't know that we have output tests for that in musicxml.  But skipping those Chords' durations seems incorrect

It also makes me realize that N.C. should allow for "writeAsChord = True" but it should make a rest instead of a Chord.

music21/musicxml/m21ToXml.py
```diff
@@ -3059,7 +3059,7 @@ def parseOneElement(self, obj):
             root.append(sp)
 
         classes = obj.classes
-        if 'GeneralNote' in classes:
+        if 'GeneralNote' in classes and 'Harmony' not in classes:
```
mscuthbert, line 3062: Likely to fail if writeAsChord = False.

mscuthbert (changes requested):
Check with writeAsChord = True or add tests for that.  Does this also impact RomanNumerals?

jacobtylerwalls:
> But skipping those Chords' durations seems incorrect

Thanks for the catch.

> Does this also impact RomanNumerals?

yes, if anyone has been setting duration on a RomanNumeral, then they should be getting better musicxml output, no stray offset and backup tags corrupting the output.

> It also makes me realize that N.C. should allow for "writeAsChord = True" but it should make a rest instead of a Chord.

Smart! See next PR.

PR #1167, opened 2021-11-14
Import/export unpitched percussion instruments <--> MusicXML

jacobtylerwalls:
Read and write `<midi-unpitched>` tag so as to import and export unpitched percussion instruments in MusicXML.

Also resolve confusion over whether m21's percussion mapper is 1-indexed (it's not). (Fixed up the confusion over this I introduced to docs and MIDI import.)

### DIFF
--- a/music21/midi/translate.py
+++ b/music21/midi/translate.py
@@ -29,7 +29,6 @@ from music21 import environment
 from music21 import stream
 
 from music21.instrument import Conductor, deduplicate
-from music21.midi import percussion
 
 _MOD = 'midi.translate'
 environLocal = environment.Environment(_MOD)
--- a/music21/midi/translate.py
+++ b/music21/midi/translate.py
@@ -774,8 +774,8 @@ def midiEventsToInstrument(eventList):
             i = instrument.fromString(decoded)
         elif event.channel == 10:
             pm = percussion.PercussionMapper()
-            # PercussionMapper.midiPitchToInstrument() is 1-indexed
-            i = pm.midiPitchToInstrument(event.data + 1)
+            # PercussionMapper.midiPitchToInstrument() is 0-indexed
+            i = pm.midiPitchToInstrument(event.data)
             i.midiProgram = event.data
         else:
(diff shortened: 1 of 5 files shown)


music21/midi/translate.py
```diff
@@ -750,7 +749,7 @@ def midiEventsToInstrument(eventList):
     >>> me.channel = 10
     >>> i = midi.translate.midiEventsToInstrument(me)
     >>> i
-    <music21.instrument.Tambourine 'Tambourine'>
+    <music21.instrument.UnpitchedPercussion 'Percussion'>
```
mscuthbert, line 752: Okay, I get the change

music21/musicxml/xmlToM21.py
```diff
@@ -1614,7 +1614,17 @@ def _adjustMidiData(mc):
...
+            mxMidiUnpitched = mxMIDIInstrument.find('midi-unpitched')
+            if textStripValid(mxMidiUnpitched):
+                pm = PercussionMapper()
+                try:
+                    i = pm.midiPitchToInstrument(_adjustMidiData(mxMidiUnpitched.text))
+                except MIDIPercussionException as mpe:
+                    # objects not yet existing in m21 such as Cabasa
+                    warnings.warn(MusicXMLWarning(mpe))
+                    i = instrument.UnpitchedPercussion()
+                    i.percMapPitch = int(_adjustMidiData(mxMidiUnpitched.text))
```
mscuthbert, line 1626: No need for `int(...)` here, `_adjustMidiData` always returns an int.

mscuthbert (approved):
Looks good -- one little redundancy that can be fixed later.

We really should add Cabasa instead of adding more tests that it fails.  Apparently GM Percussion has extended a bit in either direction: https://usermanuals.finalemusic.com/SongWriter2012Win/Content/PercussionMaps.htm  -- instrument 27 (Laser) is unlikely to be supported.  But those things should probably not raise an error, but instead have a MiscellaneousPercussion instrument object of some sort.

PR #1169, opened 2021-11-18
fixes #778

### LINKED ISSUE #778: String/Fret indications uncorrectly written in MusicXML files
**music21 version**

`6.5.0`   

**Problem summary**
Hi,

It seems that the musicXML writer can produce mistakes when the m21 stream corresponds to a tablature.

The parsing seems ok (string/fret informations seem correctly stored in the m21 stream in articulations.FretIndication / articulations.StringIndication). However, the written musicXML file is invalid (as indicated when opening with MuseScore for example). 

When looking into it, it seems that unexpected fret tags (more precisely `<fret>0</fret>`) have been added in place of technical indications (like hammer-on, pull-off) - see example below. I would not be suprised that multiple (and distinct) fret indications for one single note induces this invalidity.

I guess that guitar specific technics such as hammer-on/pull-off are not yet implemented in m21. I think that simply omitting them (instead of replacing them by non-sense `<fret>0</fret>`) would produce a valid musicXML and might be a good option until that time.

Thanks again for this fantastic tool !

**Steps to reproduce**

The musicXML page dedicated to tab (https://www.musicxml.com/tutorial/tablature/fret-and-string/) provides a simple test file (tab1.musicxml).

```
from music21 import *
tabmxml = converter.parse('./tab1.musicxml')
tabmxml.write('musicxml','./tab1-m21.musicxml')
```

In tab1.musicxml we have (l.321) :

```
<technical>
  <pull-off number="1" type="stop"/>
  <pull-off number="1" type="start">P</pull-off>
  <string>3</string>
  <fret>7</fret>
</technical>
```

and in the analogue place of the file written by m21 (tab1-m21.musicxml) we have (l.288) : 

```
<technical>
  <fret>0</fret>
  <fret>0</fret>
  <string>3</string>
  <fret>7</fret>
</technical>
```

louisbigo:
I followed your recommendations in #778 (thanks for this). I realized that I will have to use `StaffLayout` as 

```
In music21, the <staff-layout> and <staff-details> are
intertwined in a StaffLayout object.
```

so I think this will need to be handled in `xmlStaffLayoutFromStaffDetails`. But therefore I can not  drop the identified fretboard in the beginning of the part as you suggested, so I tried to simply add it as a new attribute in the StaffLayer. Hope it makes sense.

I still haven't added any test, I wanted to first make sure we agree on the direction.

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -6062,6 +6062,16 @@ class MeasureExporter(XMLExporterBase):
             mxStaffLines.text = str(staffLayout.staffLines)
 
         # TODO: staff-tuning
+        if hasattr(staffLayout, 'fretboard'):
+            tuning_pitches = staffLayout.fretboard.tuning
+            for i in range(len(tuning_pitches)):
+                mxStaffTuning = SubElement(mxStaffDetails, 'staff-tuning')
+                mxStaffTuning.set('line', str(i + 1))
+                mxTuningStep = SubElement(mxStaffTuning, 'tuning-step')
+                mxTuningStep.text = tuning_pitches[i].name
+                mxTuningOctave = SubElement(mxStaffTuning, 'tuning-octave')
+                mxTuningOctave.text = str(tuning_pitches[i].octave)
+
         # TODO: capo
         # TODO: staff-size
         return mxStaffDetails
--- a/music21/musicxml/xmlToM21.py
+++ b/music21/musicxml/xmlToM21.py
@@ -5691,7 +5691,20 @@ class MeasureParser(XMLParserBase):
                 stl.staffType = stream.enums.StaffType(xmlText)
             except ValueError:
                 environLocal.warn(f'Got an incorrect staff-type in details: {mxStaffType}')
+
         # TODO: staff-tuning*
+        mxStaffTuning = mxDetails.findall('staff-tuning')
+        if mxStaffTuning is not None:
+            tuning_pitches = []
+            for i in range(len(mxStaffTuning)):
+                staff_tuning = mxStaffTuning[i]
+                line = int(staff_tuning.get('line'))
+                tuning_step = staff_tuning.find('tuning-step').text
+                tuning_octave = int(staff_tuning.find('tuning-octave').text)
+                tuning_pitches.append(pitch.Pitch(tuning_step + str(tuning_octave)))
+            fretboard = tablature.FretBoard.getFretBoardFromTuning(tuning_pitches)
+            setattr(stl, 'fretboard', fretboard)
+
         # TODO: capo
         # TODO: show-frets
         # TODO: print-spacing
--- a/music21/tablature.py
+++ b/music21/tablature.py
@@ -244,6 +244,13 @@ class FretBoard(prebase.ProtoM21Object):
 
         return pitchList
 
+    @staticmethod
+    def getFretBoardFromTuning(pitches):
+        standard_fret_boards = [GuitarFretBoard, UkeleleFretBoard, BassGuitarFretBoard, MandolinFretBoard]
+        for standard_fret_board in standard_fret_boards:
+            if pitches == standard_fret_board().tuning:
+                return standard_fret_board()
+        return CustomFretBoard(pitches)
 
 class FirstFret:
     '''
@@ -344,6 +351,17 @@ class MandolinFretBoard(FretBoard):
         super().__init__(numStrings, fretNotes, displayFrets)
 
         self.tuning = [pitch.Pitch('G3'), pitch.Pitch('D4'), pitch.Pitch('A4'), pitch.Pitch('E5')]
+
+class CustomFretBoard(FretBoard):
+    '''
+    A custom fretboard tuned with any list of pitches
+    '''
+
+    def __init__(self, pitches, fretNotes=None, displayFrets=4):
+        numStrings = len(pitches)
+        super().__init__(numStrings, fretNotes, displayFrets)
+
+        self.tuning = pitches.copy()
 # ------------------------------------------------------------------------------
 
 


music21/musicxml/xmlToM21.py
```diff
@@ -5691,7 +5691,19 @@ def xmlStaffLayoutFromStaffDetails(
...
+        mxStaffTuning = mxDetails.findall('staff-tuning')
+        if mxStaffTuning is not None:
+            tuning_pitches = []
+            for i in range(len(mxStaffTuning)):
+                staff_tuning = mxStaffTuning[i]
+                tuning_step = staff_tuning.find('tuning-step').text
+                tuning_octave = int(staff_tuning.find('tuning-octave').text)
+                tuning_pitches.append(pitch.Pitch(tuning_step + str(tuning_octave)))
+            fretboard = tablature.FretBoard.getFretBoardFromTuning(tuning_pitches)
+            setattr(stl, 'fretboard', fretboard)
```
jacobtylerwalls, line 5705: > But therefore I can not drop the identified fretboard in the beginning of the part as you suggested, so I tried to simply add it as a new attribute in the StaffLayer.

Can we just insert into the current measure? `staffLayoutToXmlStaffDetails` is a method on the MeasureExporter, so we have access to `self.stream`.

music21/tablature.py
```diff
@@ -344,6 +352,17 @@ def __init__(self, fretNotes=None, displayFrets=4):
...
+class CustomFretBoard(FretBoard):
+    '''
+    A custom fretboard tuned with any list of pitches
+    '''
+
+    def __init__(self, pitches, fretNotes=None, displayFrets=4):
+        numStrings = len(pitches)
+        super().__init__(numStrings, fretNotes, displayFrets)
+
+        self.tuning = pitches.copy()
```
jacobtylerwalls, line 365: I'm not certain we need a new class. The only thing I see different compared to `FretBoard` is being able to pass pitches to the constructor instead of doing:

```python
fb = FretBoard(numStrings=len(pitches))
fb.pitches = pitches  # or pitches.copy()
```

And at this point I would be reluctant to create a new public class for that, when we just need to solve musicxml translation. But let me know if there are some details I've neglected here.

music21/tablature.py
```diff
@@ -244,6 +244,14 @@ def getPitches(self):
 
         return pitchList
 
+    @staticmethod
```
jacobtylerwalls, line 247: This feels like a module-level function to me.

music21/musicxml/xmlToM21.py
```diff
@@ -5691,7 +5691,19 @@ def xmlStaffLayoutFromStaffDetails(
                 stl.staffType = stream.enums.StaffType(xmlText)
             except ValueError:
                 environLocal.warn(f'Got an incorrect staff-type in details: {mxStaffType}')
+
         # TODO: staff-tuning*
+        mxStaffTuning = mxDetails.findall('staff-tuning')
+        if mxStaffTuning is not None:
+            tuning_pitches = []
+            for i in range(len(mxStaffTuning)):
```
jacobtylerwalls, line 5699: I find `enumerate(mxStaffTuning)` more readable.

music21/musicxml/xmlToM21.py
```diff
@@ -5691,7 +5691,19 @@ def xmlStaffLayoutFromStaffDetails(
...
+                staff_tuning = mxStaffTuning[i]
+                tuning_step = staff_tuning.find('tuning-step').text
```
jacobtylerwalls, line 5701: You might verify that using `textStripValid()` makes sense here. (See related uses). Helps guard against None/bad data, so that we don't try to create a pitch from "FNone" or something.

music21/musicxml/m21ToXml.py
```diff
@@ -6062,6 +6062,16 @@ def staffLayoutToXmlStaffDetails(self, staffLayout):
             mxStaffLines.text = str(staffLayout.staffLines)
 
         # TODO: staff-tuning
```
jacobtylerwalls, line 6064: are there further scenarios involving `staff-tuning` that won't be handled? let's update or remove the comment.

music21/musicxml/m21ToXml.py
```diff
@@ -6062,6 +6062,16 @@ def staffLayoutToXmlStaffDetails(self, staffLayout):
...
+        if hasattr(staffLayout, 'fretboard'):
+            tuning_pitches = staffLayout.fretboard.tuning
+            for i in range(len(tuning_pitches)):
```
jacobtylerwalls, line 6067: enumerate

music21/musicxml/m21ToXml.py
```diff
@@ -6062,6 +6062,16 @@ def staffLayoutToXmlStaffDetails(self, staffLayout):
...
+                mxStaffTuning = SubElement(mxStaffDetails, 'staff-tuning')
+                mxStaffTuning.set('line', str(i + 1))
+                mxTuningStep = SubElement(mxStaffTuning, 'tuning-step')
+                mxTuningStep.text = tuning_pitches[i].name
+                mxTuningOctave = SubElement(mxStaffTuning, 'tuning-octave')
+                mxTuningOctave.text = str(tuning_pitches[i].octave)
```
jacobtylerwalls, line 6073: need to guard against None.

jacobtylerwalls (commented):
Thanks for taking this on! I think we shouldn't couple the fretboards too tightly to m21 `StaffDetails` objects just because MusicXML couples them (what about other formats, etc.)

music21/musicxml/xmlToM21.py
louisbigo: `self.stream.append(fretboard)` does not work because `fretboard` is not a music21 object (the err message recommends to create a `music21.ElementWrapper`, should we ?)

music21/tablature.py
louisbigo: agreed.

music21/tablature.py
louisbigo: I agree, I moved it as a module-level function (I initially put it into a class because I didn't see any other module function in `tablature.py`.

music21/musicxml/xmlToM21.py
louisbigo: ok.

music21/musicxml/xmlToM21.py
jacobtylerwalls: Goodness, alright. My mistake. Looking now, what do you suppose about setting `.stringPitches` on `self.activeInstrument`, if it is a `StringInstrument`? We don't even need to make or check fretboards, I guess.

jacobtylerwalls: This is probably worth a ping for @mscuthbert before you end up wasting too much motion! Myke, any thoughts about these three options:

1a -- use FretBoards, and put them on StaffLayout objects
1b -- use FretBoards, but make them Music21Objects so they can go in streams
2 -- use `Instrument.stringPitches`

jacobtylerwalls: One benefit of the instrument approach is that if we have ambiguous data for an instrument but we have staff-tuning tags, we could possibly reclassify the object to be a `StringInstrument`, or possibly even more specific.

music21/musicxml/xmlToM21.py
louisbigo: Ok. I can do the following, but maybe we would prefer to give up the whole inclusion of tuning in the case one of the pitch is not valid (in the solution below a tuning of (n-1) pitches will be validated if one pitch is incorrect, which seems unlikely useful). What do you think ?

```
        mxStaffTuning = mxDetails.findall('staff-tuning')
        if mxStaffTuning is not None:
            tuning_pitches = []
            for staff_tuning in enumerate(mxStaffTuning):
                mxTuningStep = staff_tuning.find('tuning-step') 
                mxTuningOctave = staff_tuning.find('tuning-octave')
                if textStripValid(mxTuningStep) and textStripValid(mxTuningOctave):
                    tuning_step = staff_tuning.find('tuning-step').text
                    tuning_octave = int(staff_tuning.find('tuning-octave').text)
                    tuning_pitches.append(pitch.Pitch(tuning_step + str(tuning_octave)))
```

music21/musicxml/xmlToM21.py
louisbigo: Thanks for the suggestions. For my part, using `self.activeInstrument` seems the most relevant. Of course this would require to identify the instrument of the part as guitar (I'm not sure how it's done but in case it might help I'm dropping here the musicXML score-part block of an [example tab file provided by MusicXML](https://www.w3.org/2021/06/musicxml40/musicxml-reference/examples/tutorial-tablature/) ):

```
      <score-part id="P2">
         <part-name print-object="no">Guitar [TAB]</part-name>
         <part-abbreviation print-object="no">Gtr.</part-abbreviation>
      </score-part>
```

The reason why I'm putting this here is that I'm a little bit afraid that the absence of `score-instrument` or `midi-instrument` tag could limit to identify the instrument as Guitar. Unless we can identify it from `part-name` ?

I think the reason why instrument informations are missing in part P2 is because they are already included in part P1 - Part P2 here is only intended to provide additional informations (fret, strings, etc.) on part P1 as it is usually the case in tablature notation (2 stacked scores telling the same, the first with notes the second one with fret/strings). But in term of instrument and track, P1 and P2 are one single part. I don't know if there is any way to link the 2 parts such that P2 will refer to the instrument of P1. I guess this goes beyond the current issue.

mscuthbert: > I don't know if there is any way to link the 2 parts such that P2 will refer to the instrument of P1.

In musicxml both staves should be within the same part and the second staff should have a staff-details `<staff-type>alternate</staff-type>` tag.

This is not something that music21 supports right now, but it's something we want to do, so test files with an implementation would be helpful.

https://www.w3.org/2021/06/musicxml40/musicxml-reference/examples/staff-type-element/

jacobtylerwalls: If that's the ideal MusicXML representation, then I would think the tablature tutorial @louisbigo linked to should be updated to conform to it.

In https://github.com/w3c/musicxml/issues/205 they confirmed they wanted to move in this direction and update the documentation, so they would probably welcome a new issue for it.

That issue is also useful for its summary of the state of play.

jacobtylerwalls: > In musicxml both staves should be within the same part and the second staff should have a staff-details <staff-type>alternate</staff-type> tag.

In that case, then these MusicXML documents are clearly representing optional notation for a single part, and then I think we can adopt @louisbigo's initial proposal of putting FretBoards on the `StaffLayout` object, possibly mixing in `StyleMixin` to FretBoard so that one can set `hideObjectOnPrint` False to suppress the alternate notation.

My thought of updating the instrument's stringPitches seems doable, but perhaps should be done in a separate PR.

> we shouldn't couple the fretboards too tightly to m21 StaffDetails objects just because MusicXML couples them (what about other formats, etc.)

Just to reiterate, this discussion is helping me no longer be as worried about this.

music21/musicxml/xmlToM21.py
jacobtylerwalls: I checked and `tuning-step` and `tuning-octave` are both required on any `staff-tuning`, so if we lack one, I would either let the rest of the data through, or if I wanted to abort, I would raise `MusicXMLImportException`, since as a machine-coded input format we don't need to be that permissive about bad encodings. If we remove all the pitches and pass the file, then it will look like music21 didn't have a feature to translate it (and would be difficult for users to debug in Python).

jacobtylerwalls: You might also include the optional `<tuning-alter>` element.

music21/musicxml/xmlToM21.py
mscuthbert: I don't know that Fretboards belong on the StaffLayout object -- they're not about the layout of the staff, and if a fretboard gets moved to a different location it shouldn't change how the staff around it is laid out (unless it needs extra space for it; but the same could be said for notes on ledger lines, etc.)  -- they seem to me to be a kind of Harmony object subclass, like ChordSymbols.

jacobtylerwalls: > they seem to me to be a kind of Harmony object subclass, like ChordSymbols.

Well, for _that_ we already have `tablature.ChordWithFretBoard`, but I didn't take that to be relevant here because we don't have chord symbols at all.

> I don't know that Fretboards belong on the StaffLayout object -- they're not about the layout of the staff, and if a fretboard gets moved to a different location it shouldn't change how the staff around it is laid out

Now I fear I've confused things. I think the issue Louis set out to solve really _is_ about the layout of the staff, i.e. alternate notation on a tab staff, not fretboard symbols that can be "moved" (what would it mean for staff lines of a tablature staff to "move"). But please do correct me if I'm off base here.

mscuthbert: I think that I'm a bit confused also.  We can have ChordWithFretBoard, but we can also have the same thing without a Chord symbol.  That's what I'm talking about that I call a FretBoard.  I think I'll need to see an image of what everyone is talking about to catch up.  But it doesn't seem like any of this belongs on StaffLayout unless it affects the size or amount of space around a staff or the change in the visual appearance.  If it's about the tuning of different lines, that's a different object.  I don't want to put things into one object that do not belong with it.

louisbigo: The `FretBoard` definition in the music21 doc says : _A FretBoard represents a displayed fretboard (i.e. used in chord symbols)._. So I think that the `FretBoard` class has initially be thought to handle the displaying of string/frets positions accompanying chord symbols (see image below).


freatboard

So the question is whether the class `FretBoard` could also be used to describe the string instrument linked to a whole tablature *staff*. Looking at the code of `FretBoard`, it seems to me that it is too related to *positions* (see for examples methods `getFretNoteByString` which does not make sense in the context of a whole tablature staff). So I'm starting to think that we might not want to use `Fretboard` here (I'm also wondering if to avoid futur similar confusion, we could consider renaming the class `FretBoard` in something like `FretBoardPosition`, but it would require changing the name of many related classes / methods with similarly ambiguous names in `tablature.py` like `FretNote`, etc.).

On the other hand, I agree that the tuning of the string instrument is not supposed to affect the layout of the staff so using `StaffLayout` might not seem justified here. Unfortunately it does not seem that  `<staff-tuning>` , as a `<staff-details>` child, could be easily be processed outside `xmlStaffLayoutFromStaffDetails`. Any idea of where it could be processed ?

jacobtylerwalls: > So I'm starting to think that we might not want to use Fretboard here

I agree, we need some other data structure in music21 for tab staves. The image of what we're talking about can be found from the tutorial Louis linked to--we need to capture the pitches of the lines of a tab-staff, which may or may not (although it should) be in the same musicxml `<part>`:

image


I'm thinking the `StaffType` enum needs a new member TAB, and StaffTypes do currently exist on StaffLayout objects. Myke, what do you think of that?

Where to encode the actual pitches, I'm not sure. We could add attributes to `TabClef` or we could keep `StringInstrument.stringPitches` up to date with the pitches of the tab staff. 🤷🏻‍♂️ 

> Any idea of where it could be processed ?

`handleStaffDetails()` could call a new method to read the `<staff-tuning>` tag.

jacobtylerwalls: Maybe this means breaking out into two PartStaff objects when both are in one `<part>` just as we do for piano music.

louisbigo: > we need to capture the pitches of the lines of a tab-staff, which may or may not (although it should) be in the same musicxml `<part>`

I think the tuning pitches will allways appear on the tab-staff (unless the file is uncorrect) - however the instrument information (guitar in that case) that might only appear in an other musicxml `<part>`.

> Where to encode the actual pitches, I'm not sure. We could add attributes to `TabClef` or we could keep `StringInstrument.stringPitches` up to date with the pitches of the tab staff.

Both make sense to me. But I guess using `TabClef` is more safe as `StringInstrument.stringPitches` might not work correctly in case instrument / parts are not indicated properly (problem above). So I'll go for `TabClef` is everyone is ok.

> `handleStaffDetails()` could call a new method to read the `<staff-tuning>` tag.

Good for me !

mscuthbert:
how are we progressing on this?

mscuthbert:
Pinging @louisbigo -- wondering if you can address Jacob's comments.

mscuthbert:

PR #1170, opened 2021-11-20
Braille: export bowing symbols

jacobtylerwalls:
In February I was doing some work with MusicXML treatment of fingerings/TechnicalIndications, and I was learning how to debug the braille module to get CI jobs to pass. At the time I looked into the combination of the two, and then I rebased at some point in September but never opened a PR.

This PR:
 - adds bowing symbols to braille export
 - tightens up an edge case involving TechnicalIndications beyond Fingerings that should not be exported
 - adds some coverage
 - small docs fixes

### DIFF
--- a/music21/articulations.py
+++ b/music21/articulations.py
@@ -12,10 +12,10 @@
 
 '''
 Classes for representing and processing articulations.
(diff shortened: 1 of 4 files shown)


mscuthbert (commented):
LGTM Thanks!

PR #1175, opened 2021-11-27
MIDI import optimizations

### LINKED ISSUE #1174: MIDI: reading performance can be improved
**Motivation**
The code that reads in MIDIs uses a lot of inefficient code, causing in rather slow reads.

**Feature summary**

I tried some straightforward approaches to make it more efficient. See https://github.com/kwikwag/music21/tree/feature/midi-read-performance

**Proposed implementation**

* Replace `bytes` with `memoryview` for the reading process.
* Use a pre-calculated `frozenset()` for the `_ContainsEnum` base class

**Intent**

kwikwag:
This fixes issue #1174

### DIFF
--- a/music21/midi/__init__.py
+++ b/music21/midi/__init__.py
@@ -915,7 +915,7 @@ class MidiEvent(prebase.ProtoM21Object):
         elif SysExEvents.hasValue(byte0):
             self.type = SysExEvents(byte0)
             length, midiBytesAfterLength = getVariableLengthNumber(midiBytes)
-            self.data = midiBytesAfterLength[:length]
+            self.data = bytes(midiBytesAfterLength[:length])
             return midiBytesAfterLength[length:]
 
         # SEQUENCE_TRACK_NAME and other MetaEvents are here
@@ -926,7 +926,7 @@ class MidiEvent(prebase.ProtoM21Object):
                 raise MidiException(f'Unknown midi event type: FF {byte1:02X}')
             self.type = MetaEvents(byte1)
             length, midiBytesAfterLength = getVariableLengthNumber(midiBytes[1:])
-            self.data = midiBytesAfterLength[:length]
+            self.data = bytes(midiBytesAfterLength[:length])
             # return remainder
             return midiBytesAfterLength[length:]
         else:
@@ -1293,7 +1293,7 @@ class MidiTrack(prebase.ProtoM21Object):
 
         # all event data is in the track str
         trackData = midiBytes[:length]
-        self.data = trackData
+        self.data = bytes(trackData)
 
         remainder = midiBytes[length:]
         self.processDataToEvents(trackData)
--- a/music21/midi/__init__.py
+++ b/music21/midi/__init__.py
@@ -350,7 +350,12 @@ class _ContainsEnum(IntEnum):
 
     @classmethod
     def hasValue(cls, val):
-        return val in cls.__members__.values()
+        # using a stored set() here improves performance by ~5%
+        try:
+            return val in cls.__values__
+        except AttributeError:
+            cls.__values__ = frozenset(cls.__members__.values())
+            return val in cls.__values__
 
 
 class ChannelVoiceMessages(_ContainsEnum):
--- a/music21/midi/__init__.py
+++ b/music21/midi/__init__.py
@@ -720,7 +720,7 @@ class MidiEvent(prebase.ProtoM21Object):
         self.parameter1 = d2
         self.parameter2 = d1  # d1 is most significant byte here
 
-    def parseChannelVoiceMessage(self, byte0: int, midiBytes: bytes) -> bytes:
+    def parseChannelVoiceMessage(self, midiBytes: bytes, byte0: int = None) -> bytes:
         r'''
         Take a set of bytes that represent a ChannelVoiceMessage and set the
         appropriate enumeration value and data, returning the remaining bytes.
@@ -743,7 +743,13 @@ class MidiEvent(prebase.ProtoM21Object):
 
         Now show how the midiBytes changes the event:
 
-        >>> remainder = me1.parseChannelVoiceMessage(midBytes[0], midBytes[1:])
+        >>> remainder = me1.parseChannelVoiceMessage(midBytes)
+        >>> me1
+        <music21.midi.MidiEvent NOTE_ON, track=1, channel=1, pitch=60, velocity=120>
+
+        It works also if you specify the status byte separately.
+
+        >>> remainder = me1.parseChannelVoiceMessage(midBytes[1:], byte0=midBytes[0])
         >>> me1
         <music21.midi.MidiEvent NOTE_ON, track=1, channel=1, pitch=60, velocity=120>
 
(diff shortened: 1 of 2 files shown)


kwikwag:
@jacobtylerwalls I ran pylint locally and it seems to be fine after last commit. Workflows can be re-run 🤞

kwikwag:
To benchmark, I used the following code in IPython, making use of the MIDI files that are part of the music21 codebase:

```python
from pathlib import Path
import music21

paths = list(Path(music21.__file__).parent.parent.rglob('*.mid'))

def test_func():
    for path in paths:
        mf = music21.midi.MidiFile()
        mf.open(path)
        try:
            mf.read()
        finally:
            mf.close()
```

On `master` I get:
```pycon
In [2]: %timeit -r10 -n5 test_func()
351 ms ± 7.4 ms per loop (mean ± std. dev. of 10 runs, 5 loops each)
```

Only with the `_ContainsEnum` fix:

```pycon
In [2]: %timeit -r10 -n5 test_func()
320 ms ± 5.29 ms per loop (mean ± std. dev. of 10 runs, 5 loops each)
```

Only with the `memoryview` fix:

```pycon
In [2]: %timeit -r10 -n5 test_func()
247 ms ± 5.12 ms per loop (mean ± std. dev. of 10 runs, 5 loops each)
```

With both changes:
```pycon
In [2]: %timeit -r10 -n5 test_func()
225 ms ± 5.3 ms per loop (mean ± std. dev. of 10 runs, 5 loops each)
```

kwikwag:
I note here as well what I've mentioned in the issue - the `memoryview` change can break implementations which rely on directly accessing `MidiEvent`'s `data` member. For me this is the case - as I want to directly access track name information. There are several ways to go here: 

  1. document this and have callers explicitly use `bytes` when they need them
  2. implement a getter for `data` which returns `bytes` over the `memoryview`
  3. store `bytes` in `data` members in the first place

I suspect 3 might hurt performance on read time, but unsure.

mscuthbert:
The ContainsEnum speedup, is a definite win (except that I wouldn't name the special variable `__values__` since it's not a builtin Python value; call it `._cachedValues`).

I'm impressed with the speedup on the second, though I'd want to know how many real MIDI files we have to read before it makes a significant difference worth making the code more complex.  I'd want at the least to have typing on all objects and renamed variables (i.e., not calling something "midiBytes" if it's not a bytes object) -- it's something we're going for in the codebase anyhow, but a moment when something changes type (like from a bytes object to a pure int) is the time to do so.  Instead of memoryview should we be doing bytearray to keep the internal formats the same?

kwikwag:
@mscuthbert The whole problem with using `bytes` (or `bytearray` for that matter) is that slices are copies, and there is no need to do memory copies while reading the MIDI file. To keep the changes local, I avoided doing variable renames.

The `memoryview` change is actually not that complex. It involves two places that need to be changed. One is the `MidiFile.read()` method, simply wrapping `midiBytes` with `memoryview`. The second involves the methods `MidiEvent.read()` and `MidiEvent.parseChannelVoiceMessage()` - `MidiEvent.read()` contained some complex logic that added an artificial running-status byte (rsb) in front of the byte buffer; this would again result in a copy if it were to remain as-is. Instead, I opted in keeping the running status byte aside (as `byte0`, same name that was used before), and had to pass it as a separate argument to `MidiEvent.parseChannelVoiceMessage()`, which expected the first byte to be the RSB. 

Since the MIDI code is anyway not very clearly written and not supposed to be performant, I don't see this as a step towards cleaning up the code, but rather as a simple fix that doesn't leave the code more or less damaged, just more performant.

I need to be running code on a large dataset of MIDI files (~100K). It takes about 24 hours to go through the dataset with the fixes. You can image it would take several more without. I don't expect many passes on the dataset, and would love a pure Python implementation. I am considering moving towards a C or Rust based workflow (and perhaps this would be a good solution for music21 as well down the road), but in the meantime thought to make this contribution.

I'll make some name changes (e.g. `_cached_values`) and try to find an elegant solution for the typing of `MidiEvent.data`.

kwikwag:
@mscuthbert I changed the name in `_ContainsEnum` and fixed it so that `MidiEvent.data` is now `bytes`, as before. It results in a few extraneous copies but this doesn't hurt performance in any noticable way (get same results as before). I'll be glad if this pull request could be merged. If you would like, I can add more test cases, if you feel it still does not have enough coverage.

P.S. The underlying data in the various `midiBytes` variables is still byte data, so I don't think `midiBytes` is a bad name for it.

music21/midi/__init__.py
```diff
@@ -715,7 +720,7 @@ def setPitchBend(self, cents, bendRange=2):
         self.parameter1 = d2
         self.parameter2 = d1  # d1 is most significant byte here
 
-    def parseChannelVoiceMessage(self, midiBytes: bytes) -> bytes:
+    def parseChannelVoiceMessage(self, byte0: int, midiBytes: bytes) -> bytes:
         r'''
         Take a set of bytes that represent a ChannelVoiceMessage and set the
         appropriate enumeration value and data, returning the remaining bytes.
```
jacobtylerwalls, line 726:
```suggestion
        appropriate enumeration value and data, returning the remaining bytes.

        Changed in v7.3 -- a description of the changes (we don't have sphinx versionchanged blocks yet, just some text)
```

jacobtylerwalls (commented):
Just a small request for a line describing the changes in `parseChannelVoiceMessage`.

I'm glad the backwards compatibility concerns are smaller now that self.data is still bytes. Thanks.

Myke, do we need to go the list to see if folks are willing to accept a break in the signature of parseChannelVoiceMessages before v.8? We could also release v.7.3 upon merging a handful of open PRs and then move master over to v8 like last year.

music21/midi/__init__.py
kwikwag: I'll write here rather than committing. If it seems good, I'll add a commit.

```
Changed in v7.3 -- the status byte is now accepted as the parameter `byte0`. Previously the first byte of `midiBytes` was expected to carry the status. Previous version calls to `MidiEvent.parseChannelVoiceMessage(midiBytes)` should be replaced with `MidiEvent.parseChannelVoiceMessage(midiBytes[0], midiBytes[1:])`
```

kwikwag: Alternatively, I can add `byte0` as an optional last parameter, and have `parseChannelVoiceMessage()` read it from `midiBytes` if it isn't specified. That way it won't be a breaking change.

kwikwag: ...and that's what I did.

jacobtylerwalls: Fantastic, I was also going to suggest that as an option. I think a changed notice explaining when the new parameter was added would still be nice. We don't have different sets of docs by version, so someone could be reading this and still be on 5.7.2...

`Changed in v.7.3 -- Added optional `byte0` argument for better performance.`

music21/midi/__init__.py
```diff
@@ -742,6 +747,12 @@ def parseChannelVoiceMessage(self, midiBytes: bytes) -> bytes:
         >>> me1
         <music21.midi.MidiEvent NOTE_ON, track=1, channel=1, pitch=60, velocity=120>
 
+        It works also if you specify the status byte separately.
```
jacobtylerwalls, line 750:
```suggestion
        Added in v.7.3 -- The optional `byte0` argument to separately supply the
        status byte was added to improve performance.
```

mscuthbert:
I'm currently still -1 on everything but the ContainsEnum -- 17 years of maintaining music21 has led me to not accept changes to complex systems where I no longer have an intuitive sense of how the system works anymore.  I wouldn't know how to add more features to the MIDI system after this would be merged.   I no longer have a sense of what is an int and what is a byte-string or where our parse position is, or what I can optimize further.

jacobtylerwalls:
Closing in favor of #1207, and inviting comments/benchmarking there.

PR #1176, opened 2021-11-28
Explict imports in init.py; support 3.10

mscuthbert:
Adds all music21 modules by name to `__init__.py` imports, hopefully making life easier for certain IDEs.  For a long time, the music21 core modules were changing almost weekly, so `import *` within `__init__.py` was the best approach.  Now we add or rename a core module maybe once a year, so this should be a good approach.

Also add Python 3.10 to the github matrix.  All tests currently passing without warnings.

### DIFF
--- a/music21/__init__.py
+++ b/music21/__init__.py
@@ -29,10 +29,8 @@ Music and Theater Arts section and the School of Humanities, Arts,
 and Social Sciences.  Portions of music21 were originally part of
 the PMusic (Perl) library, developed by Cuthbert prior to arriving at MIT.
 
-music21 outputs a subset of XML data defined by the  MusicXML 3.1
-standard, Copyright © Recordare LLC;  License available at
-http://www.recordare.com/dtds/license.html, transferred to MakeMusic
-now transferred to W3C
+music21 outputs a subset of XML data defined by the MusicXML 4.0
+standard, Copyright © 2004-2021 the Contributors to the MusicXML Specification.
 
 The corpus files have copyrights retained by their
 owners who have allowed them to be included with music21.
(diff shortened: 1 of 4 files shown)


music21/__init__.py
```diff
@@ -191,7 +192,80 @@
...
+from music21 import braille  # noqa: E402
+from music21 import capella  # noqa: E402
+from music21 import chord  # noqa: E402
+from music21 import common  # noqa: E402
+from music21 import converter  # noqa: E402
+from music21 import corpus  # noqa: E402
+from music21 import features  # noqa: E402
+from music21 import figuredBass  # noqa: E402
+from music21 import graph  # noqa: E402
+from music21 import humdrum  # noqa: E402
```
jacobtylerwalls, line 209: This is an improvement over the wildcard, so great. I do think we can get rid of these entirely, though. I have a branch wiping these imports out in order to solve #881, but I haven't rebased it on v.7 yet. 

For instance, I don't see a reason we need to support `humdrum` as a name if a super-user is doing a super-user-python-friend-style `import music21`. If someone wants a clean namespace, they should import `music21.humdrum`, or `music21.humdrum as humdrum` or just the modules they need. I think these imports are the cause of #881, but they are a little tedious to fix because fixing them involves cleaning up other circular imports that are hidden by the deterministic import right here.

mscuthbert: I'm pretty enamored of the notion that when you import a package you get the package, not a random set of modules that happen to be imported because of dependencies.  I'm not sure that changing music21's behavior to fix a Pyinstaller bug is a tradeoff I'm willing to make.

jacobtylerwalls: Oh for sure, that would be a terrible outcome. I wasn't proposing that. It looks like I was confused -- I thought that `import music21` was adding `humdrum` to the global namespace. I guess I was wrong.

Your change here was enough to solve the pyinstaller doo-hickey after all! 🎉

PR #1188, opened 2021-12-19
Fix parsing of add, alter, omit, and subtract in chord symbol figures

jacobtylerwalls:
Fix the failing tests for add, alter, omit, and subtract in chord figures.


mscuthbert:
Thanks!

PR #1192, opened 2022-01-09
Fix dissonance score for compound intervals

### LINKED ISSUE #1191: Dissonance score treats compound intervals incorrectly
**music21 version**
dev

**Problem summary**
The dissonance score used by `pitch.simplifyMultipleEnharmonics()` et al. might give a "triad award" to a compound diminished seventh, or other compound intervals enharmonically equivalent to triadic intervals (thirds, sixths).

**Steps to reproduce**

```python
In [1]: from music21 import key, pitch

In [2]: pitch.simplifyMultipleEnharmonics([pitch.Pitch('B5')], keyContext=key.KeySignature(2))
Out[2]: [<music21.pitch.Pitch C-6>]

```

Should have returned B5 (original pitch).

**More information**
The dissonance score algorithm is doing a mod 8 operation that breaks for compound intervals. The default "D" in the two-sharps key signature is D4, so we have a compound diminished seventh (diminished fourteenth) from D4 -> Cb6 getting a "triad award" incorrectly.

Thanks to Brian Lee for reporting on Google Groups.

jacobtylerwalls:
Fixes #1191

Issue since introduction in #128 (v.5.0.3.a1). Thanks to Brian Lee for the report.

### DIFF
--- a/music21/pitch.py
+++ b/music21/pitch.py
@@ -454,19 +454,14 @@ def _dissonanceScore(pitches, smallPythagoreanRatio=True, accidentalPenalty=True
 
     if triadAward:
         # score_triad = number of thirds per pitch (avoid double-base-thirds)
-        triad_bases = []
         for p1, p2 in itertools.combinations(pitches, 2):
             this_interval = interval.Interval(noteStart=p1, noteEnd=p2)
-            generic_interval_value = abs(this_interval.generic.value) % 8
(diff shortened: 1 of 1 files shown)


mscuthbert (approved):
Looks great to me.  We should do a search for "% 8" in the code, because it's almost always wrong if it's talking about intervals.

PR #1193, opened 2022-01-10
Respect Accidental.displayType option 'never'

jacobtylerwalls:
Portion of #1148 to get this small piece in faster.


mscuthbert:
Thanks!

PR #1197, opened 2022-01-14
Fix accidental display after note intervenes between identical chromatic pitches

### LINKED ISSUE #1196: Superfluous accidentals in edge case with overrideStatus=True
**music21 version**
dev, but not caused recently

**Problem summary**
displayStatus miscalculated

**Steps to reproduce**

```python
from music21 import converter, key
p = converter.parse('tinyNotation: f#4 e f#')
p.measure(1).insert(0, key.Key('G'))
p.recurse().notes.last().pitch.accidental.displayStatus = False
p.makeAccidentals(inPlace=True, overrideStatus=True)
for el in p.recurse().notes:
    if el.pitch.accidental:
        print(el, el.pitch.accidental.displayStatus)
```

Expected False and False
```
<music21.note.Note F#> False
<music21.note.Note F#> True
```

jacobtylerwalls:
Fixes #1196

### DIFF
--- a/music21/pitch.py
+++ b/music21/pitch.py
@@ -4713,7 +4713,7 @@ class Pitch(prebase.ProtoM21Object):
                     return
                 else:  # names are the same, skip this line of questioning
                     break
-        # nope, no previous pitches in this octave and register, now more complex things...
+        # nope, no conflicting accidentals at this name and octave in past...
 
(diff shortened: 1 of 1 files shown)


mscuthbert (commented):
LGTM

PR #1199, opened 2022-01-16
Fix chord symbol duration realization edge case

jacobtylerwalls:
Found while testing #1166

### DIFF
--- a/music21/harmony.py
+++ b/music21/harmony.py
@@ -2549,7 +2549,7 @@ def realizeChordSymbolDurations(piece):
                 lastChord = cs
         return pf
     elif len(onlyChords) == 1:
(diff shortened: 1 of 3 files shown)


music21/harmony.py
```diff
@@ -2549,7 +2549,7 @@ def realizeChordSymbolDurations(piece):
                 lastChord = cs
         return pf
     elif len(onlyChords) == 1:
-        onlyChords[0].duration.quarterLength = pf.highestOffset - pf.elementOffset(onlyChords[0])
+        onlyChords[0].duration.quarterLength = pf.highestTime - pf.elementOffset(onlyChords[0])
```
jacobtylerwalls, line 2552: Same as branch above.

PR #1200, opened 2022-01-16
Realize NoChords as rests if writeAsChord = True

jacobtylerwalls:
See https://github.com/cuthbertLab/music21/pull/1166#issuecomment-1010504878.


mscuthbert (approved):
Thanks!

PR #1201, opened 2022-01-19
January typing

jacobtylerwalls:
All for now, just wanted to see what other "little" stuff there was that didn't involve the huge modules.


.github/workflows/pythonpylint.yml
```diff
@@ -63,5 +63,5 @@ jobs:
...
+                  mypy --follow-imports=silent music21/capella music21/common music21/corpus music21/features music21/figuredBass music21/humdrum music21/ipython21 music21/languageExcerpts music21/lily music21/mei music21/metadata music21/musedata music21/noteworthy music21/omr music21/romanText music21/test music21/vexflow
+                  mypy --follow-imports=silent music21/articulations.py music21/bar.py music21/beam.py music21/clef.py music21/configure.py music21/defaults.py music21/derivation.py music21/duration.py music21/dynamics.py music21/editorial.py music21/environment.py music21/exceptions21.py music21/expressions.py music21/freezeThaw.py music21/harmony.py music21/instrument.py music21/interval.py music21/layout.py music21/percussion.py music21/prebase.py music21/repeat.py music21/roman.py music21/serial.py music21/sieve.py music21/sites.py music21/sorting.py music21/spanner.py music21/style.py music21/tablature.py music21/tempo.py music21/text.py music21/tie.py music21/tinyNotation.py music21/variant.py music21/voiceLeading.py music21/volpiano.py music21/volume.py
```
jacobtylerwalls, line 67: Just need pitch, key, note, base to make this into `music21/*.py`

music21/base.py
```diff
@@ -608,6 +607,26 @@ def __setstate__(self, state: Dict[str, Any]):
         # defining self.__dict__ upon initialization currently breaks everything
         self.__dict__ = state  # pylint: disable=attribute-defined-outside-init
 
+    def _reprInternal(self) -> str:
```
jacobtylerwalls, line 610: Moved from ProtoM21Object because mypy really, really doesn't like `hasattr`!

music21/clef.py
```diff
@@ -148,26 +148,12 @@ def octaveChange(self) -> int:
...
-        24
         '''
         return self._octaveChange
 
     @octaveChange.setter
     def octaveChange(self, newValue: int):
-        oldOctaveChange = self._octaveChange
         self._octaveChange = newValue
-        if hasattr(self, 'lowestLine') and self.lowestLine is not None:
-            self.lowestLine += (newValue - oldOctaveChange) * 7
```
jacobtylerwalls, line 170: Moved to PitchClef.octaveChange

music21/expressions.py
```diff
@@ -870,6 +870,7 @@ def realize(
                                            inPlace=True)
 
             if self._setAccidentalFromKeySig:
+                assert currentKeySig is not None
```
jacobtylerwalls, line 873: Incredibly unlikely that `self._setAccidentalFromKeySig` was toggled from False to True as a side effect of the last few lines since it was just checked a minute ago to get a value . But mypy doesn't know that.

music21/interval.py
```diff
@@ -318,10 +318,7 @@ def _extractPitch(nOrP: Union['music21.note.Note', 'music21.pitch.Pitch']):
     True
 
     '''
-    if hasattr(nOrP, 'pitch'):
-        return nOrP.pitch
-    else:
-        return nOrP
+    return getattr(nOrP, 'pitch', nOrP)
```
jacobtylerwalls, line 321: Mypy really hates `hasattr`. But if this feels illegible, do say so.

music21/interval.py
```diff
@@ -721,6 +720,14 @@ def transposePitch(self, pitch1, *, inPlace=False):
         '''
         raise NotImplementedError
 
+    @abc.abstractmethod
+    def reverse(self):
+        '''
+        IntervalBase does not know how to do this, so it must be overridden in
+        derived classes.
```
jacobtylerwalls, line 727: Copied from above

music21/sorting.py
```diff
@@ -31,14 +31,14 @@
...
-
 
 class SortingException(exceptions21.Music21Exception):
     pass
 
 
-class SortTuple(namedtuple('SortTuple', _attrList)):
+class SortTuple(namedtuple('SortTuple', (
+    'atEnd', 'offset', 'priority', 'classSortOrder', 'isNotGrace', 'insertIndex'
+))):
```
jacobtylerwalls, line 41: Somewhere on an issue mypy/guido said they're never going to support dynamic namedtuple fields.

music21/spanner.py
```diff
@@ -429,14 +429,16 @@ def addSpannedElements(self,
         True
         '''
         # presently, this does not look for redundancies
-        if not common.isListLike(spannedElements):
-            spannedElements = [spannedElements]
+        if isinstance(spannedElements, base.Music21Object):
+            spannedEls = [spannedElements]
+        else:
+            spannedEls = spannedElements
```
jacobtylerwalls, line 435: If this feels too homonymous, I can just add a disable?

music21/duration.py
```diff
@@ -2929,6 +2938,7 @@ def type(self) -> str:
         0.25
         '''
         if self.linked is False:
+            assert self._unlinkedType is not None, 'linked setter failed to set _unlinkedType'
```
mscuthbert, line 2941: I'm not sure what this is for; I'm again not happy with these bare asserts -- it doesn't tell the user what they should do if they see this message.  I'd rather have a mypy ignore than to have extra asserts.

music21/expressions.py
mscuthbert: Better for this to assign a local variable `setAccidentalFromKeySig = self._setAccidentalFromKeySig` and then check against the local version; avoids another assert.

music21/instrument.py
```diff
@@ -79,6 +78,8 @@ def unbundleInstruments(streamIn: StreamType, *, inPlace=False) -> Optional[Stre
 
     if inPlace is False:
         return s
+    else:
+        return None
```
mscuthbert, line 82: can we disable this setting?  One of my least liked parts of TypeScript is having to return undefined if I return something else somewhere.  None is Python's default return value.  I'd rather not have a lot of "return None" at the end of files.

music21/interval.py
mscuthbert: yes, that feels illegible, sorry to say.  I'd rather disable mypy for the function than have something that isn't clear what it does.  I still use music21 code as a teaching tool, and I don't like to encourage my students to write something like this.

The approach in https://github.com/python/mypy/issues/1424#issuecomment-894179881  is acceptable if commented.  But I really don't understand why getattr is acceptable but hasattr is not.

music21/prebase.py
```diff
@@ -264,26 +263,8 @@ def _reprInternal(self) -> str:
         >>> p = prebase.ProtoM21Object()
         >>> p._reprInternal()
         'object at 0x112590380'
-
-        If an object has `.id` defined and `x.id` is not the same as `id(x)`
```
mscuthbert, line 268: Would prefer a note here saying that "a more complex `_reprInternal` that handles the case of objects with .id defined is found in Music21Object"  Thanks!

music21/sorting.py
mscuthbert: <s>Can _attrList be renamed '_sortAttributes' and be made a tuple and then consulted below? This is so performance critical that I don't want to have the extra method resolution step in the methods below</s> -- never mind: the methods below are rarely called, so it's okay.  Still I think it's somewhat clearer than self.fields.

music21/spanner.py
mscuthbert: I'd prefer the disable here -- there's no difference between spannedElements and spannedEls and it's just more confusing to me.  I don't consider rewriting an argument to be a problem.  And we'll have a lot more of this when we get to Stream etc.

I think I wrote "List" where Sequence is fine.

music21/style.py
```diff
@@ -498,7 +498,7 @@ class StyleMixin(common.SlottedObjectMixin):
     '''
     _styleClass = Style
 
-    __slots__ = ('_style', '_editorial')
+    __slots__: Tuple[str, ...] = ('_style', '_editorial')
```
mscuthbert, line 501: does this need to be here, since it derives from a typed class?

music21/voiceLeading.py
```diff
@@ -518,7 +518,9 @@ def parallelMotion(
         >>> vl.parallelMotion(gi, allowOctaveDisplacement=True)
         True
         '''
-
+        # .generic could only be None if intervals were constructed without diatonic components
+        assert self.vIntervals[0].generic is not None
+        assert self.vIntervals[1].generic is not None
```
mscuthbert, line 523: Let's find a way to work around this.  I **hate** bare asserts.

music21/voiceLeading.py
```diff
@@ -1111,6 +1113,8 @@ def isProperResolution(self) -> bool:
             n1degree = None
             n2degree = None
 
+        # .generic could only be None if intervals were constructed without diatonic components
+        assert self.vIntervals[1].generic is not None
```
mscuthbert, line 1117: Same here.  Work around or disable.

music21/voiceLeading.py
```diff
@@ -1176,6 +1180,9 @@ def leapNotSetWithStep(self) -> bool:
         if self.noMotion():
             return False
 
+        # .generic could only be None if intervals were constructed without diatonic components
+        assert self.hIntervals[0].generic is not None
+        assert self.hIntervals[1].generic is not None
```
mscuthbert, line 1185: Same here.  Work around or disable.  I'm okay assigning to a singleton GenericInterval in the class definition if need be.  

Users of an object should not be responsible for checking corner cases of the definition of a lower-level object.  And VoiceLeadingQuartet is a user of Interval.

music21/voiceLeading.py
```diff
@@ -1230,6 +1237,8 @@ def opensIncorrectly(self) -> bool:
         c2 = chord.Chord([self.vIntervals[1].noteStart, self.vIntervals[1].noteEnd])
         r1 = roman.identifyAsTonicOrDominant(c1, self.key)
         r2 = roman.identifyAsTonicOrDominant(c2, self.key)
+        # TODO in Py3.8+, remove when identifyAsTonicOrDominant returns Union[str | Literal[False]]
+        assert r1 is not True and r2 is not True
```
mscuthbert, line 1241: Remove, disable, or replace with if and raise normal exception

music21/voiceLeading.py
```diff
@@ -1275,6 +1284,9 @@ def closesIncorrectly(self) -> bool:
             raisedMinorCorrectly = True
         preClosings = (6, 3)
         closingPitches = [self.v1n2.pitch.name, self.v2n2.name]
+        # .generic could only be None if intervals were constructed without diatonic components
+        assert self.vIntervals[0].generic is not None
+        assert self.vIntervals[1].generic is not None
```
mscuthbert, line 1289: As above, remove, work around in interval.py or disable.

music21/volume.py
```diff
@@ -254,14 +259,15 @@ def getRealized(
...
             if useArticulations is not False:
-                if common.isIterable(useArticulations):
-                    am = useArticulations
-                elif (hasattr(useArticulations, 'classes')
-                       and 'Articulation' in useArticulations.classes):
+                am: Iterable[articulations.Articulation]
+                if isinstance(useArticulations, articulations.Articulation):
                     am = [useArticulations]  # place in a list
+                elif common.isIterable(useArticulations):
+                    assert useArticulations is not True
```
mscuthbert, line 269: remove or disable -- using classes are not responsible for corner cases on main classes.

mscuthbert (changes requested):
Thanks for the great work.  It'll need changes though to be accepted.

I can't overstress how much I dislike bare asserts and how I don't want them in music21.   (I hate them more than mypy hates hasattr).   In nearly every case where they appear, I would prefer to have type checking disabled on the parent function (return "Any") than have users need to deal with asserting that things not supported in this version of Python (like Literal[False]) don't apply.  

Maybe we can put this into a m21 v8 branch, which won't support Python 3.7, and that will have fewer cases of teaching the system about something explained in the docs, but here I count one possible bug that mypy found, compared to at least 9 workarounds for bugs/deficiencies in their own system.

Mypy conformance might have been something easy to get in if music21 were being started from scratch in 2022, but at this point there are so much code history that cannot cleanly be brought into line with what mypy wants that I'm not sure it's going to be possible to get everything into the typechecking system until they become more flexible on what they're able to read.

music21/voiceLeading.py
jacobtylerwalls: It's all good. Not a problem to get rid of em.

In this case, my adding an assert here was just lazy, because there _is_ a possible path for a user to get here. They could create voice leading quartets made up of totally empty intervals. So mypy found a corner case bug, and I was too lazy to add a proper exception. As opposed to the one below about `common.isIterable` where the so-called failure case is not reachable because of what we know about `isIterable()`, that a boolean should never pass that function (it's music21's very own type-narrowing function).

To think aloud just to check my own understanding. (Not to persuade, because let's definitely not add things that you find annoying!)

The `assert` doesn't bother me as an articulation of an assumption the author relies on down the line. ("Somewhere else in the module is responsible for making this not None. But that's forty lines away, and if it ever gets refactored, let's add safety here, because I'm about to start accessing attributes on it.") As opposed to a failure path that is part of normal execution that a user might actually need to know about, care about, and catch. And I think part and parcel with `no-implication-that-users-should-know-or-care` is why you can run python in `-O` to skip them. They're not necessarily even net negatives for user feedback, if the author is proven wrong and the "never supposed to happen" turns out to happen--as it's better to get `AssertionError: self.stream is None`, which articulates the root cause, instead of `AttributeError: self.stream has no attribute __iter__` 400 lines later, which obscures it. Yes, more messages for the user is always nice, but this is the tool for those gray areas "no, this can't fail, but it wouldn't be safe later if something else changes". If there is a path for a user to reach it, then assert is not the tool.

In a few weeks I can spruce these up to disable/raise instead.

I think the value of `mypy` is not just "did it find bugs in my functionality" but maybe more importantly for us "did it find bugs in my type annotations".  On a project with community contributions, it should save a future maintainer time to make sure folks aren't adding in new bad annotations.

jacobtylerwalls: None of this is new to you, certainly! Figuring out what is safe to rely on versus what needs to be guarded is an everyday part of programming. I guess one to look at it when developing a library is that by adding explicit guards they can sort of start to look like part of the API. Function X raises exception Y. When truly that's not the case because there should be no possible way to get there. And so an assert more clearly shows that it's not part of the API; doesn't represent a test case that needs to be supported after a refactor. Something like that.

jacobtylerwalls: ah, what a funny memory, ten minutes ago, that time I free-associated about "everyday programming" to someone with 10x my experience... 😉

mscuthbert: Will reply more soon. But my longer experience is more than exceeded by your talent and enthusiasm.

music21/style.py
jacobtylerwalls: Wow, good catch! I probably typed them out of order (this one first)

music21/voiceLeading.py
jacobtylerwalls: (Thanks.)

Raising exceptions didn't work here either without a lot of boilerplate (you apparently have to narrow in the same scope), so will just disable these ones.

music21/instrument.py
jacobtylerwalls: Will do (and for an entertaining read see https://github.com/python/mypy/issues/7511)

music21/style.py
jacobtylerwalls: Oh, it was to prevent an error in `class Comment`. I'll just add a disable over there. Some multiple inheritance problem I guess.

music21/harmony.py
```diff
@@ -191,7 +191,7 @@ def __init__(self,
                  figure: Optional[str] = None,
                  root: Optional[pitch.Pitch] = None,
                  bass: Optional[pitch.Pitch] = None,
-                 inversion: Optional[pitch.Pitch] = None,
+                 inversion: Optional[int] = None,
```
mscuthbert, line 194: great catch mypy!  :-)

mscuthbert (approved):
Thanks for the patient work on this!  :-)

I think that it's time to release a v7.3 with the changes so far and go straight to a making master a v8 branch for any future typings that allows for the `Literal[...] type` since we'll make 3.8 the lowest supported  system.  (a nice thing: we've now got SemVer aligned with minimum Python version).

music21/instrument.py
mscuthbert: If it was anyone but Guido, his "whatever" comment would be deleted for trolling.  I'm with others -- mypy is a type checker not a style enforcer.

PR #1202, opened 2022-01-22
Don't choke on conductor tracks lacking meters

jacobtylerwalls:
Fixes regression in v.7 (a56955ae50b87ee9f5145f8b6d63d629dc33e121). when making measures in MIDI import, if there is a non-empty conductor track, it was assumed to have a time signature. If it only had tempos, instrument names, etc., music21 raised:

```music21.exceptions21.StreamException: failed to find TimeSignature in meterStream; cannot process Measures```


mscuthbert (approved):
Looks great to me!

PR #1208, opened 2022-01-30
Avoid exporting superfluous meters and keys in staff groups

### LINKED ISSUE #1129: MusicXML importer: PartStaffs end up with too many (2) TimeSignatures in a non-first Measure

jacobtylerwalls:
Take 2 -- fixes #1129. Alternative to #1131.

@gregchapman-dev Another one here -- would you mind checking?

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -2471,7 +2471,7 @@ class PartExporter(XMLExporterBase):
             self.midiChannelList = parent.midiChannelList  # shared list
             self.makeNotation = parent.makeNotation
 
-        self.higher_sibling_in_group: Optional[stream.PartStaff] = None
+        self.previous_sibling_in_group: Optional[stream.PartStaff] = None
 
         self.instrumentStream = None
         self.firstInstrumentObject = None
@@ -5660,7 +5660,7 @@ class MeasureExporter(XMLExporterBase):
 
         return mxAttributes
 
-    def _matches_earlier_sibling_in_group(self,
+    def _matches_previous_sibling_in_group(self,
                                           obj: base.Music21Object,
                                           attr='keySignature',
                                           comparison='__eq__') -> bool:
@@ -5672,7 +5672,7 @@ class MeasureExporter(XMLExporterBase):
         '''
         if self.parent is None:  # pragma: no cover
             return False
-        if self.parent.higher_sibling_in_group is None:
+        if self.parent.previous_sibling_in_group is None:
             return False
         # Not a foolproof measure lookup: see more robust measure
         # matching algorithm in PartStaffExporterMixin.processSubsequentPartStaff().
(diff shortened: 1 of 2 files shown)


gregchapman-dev:
Code looks good by inspection... I'm not in a position to test this directly at the moment (everything is pulled apart), but your test case looks like it tests what I would have tested.  Thanks!

music21/musicxml/m21ToXml.py
```diff
@@ -2469,6 +2471,8 @@ def __init__(self, partObj: Union[stream.Part, stream.Score, None] = None, paren
             self.midiChannelList = parent.midiChannelList  # shared list
             self.makeNotation = parent.makeNotation
 
+        self.previous_sibling_in_group: Optional[stream.PartStaff] = None
```
mscuthbert, line 2474: since this is a public interface it should be camelCased.

mscuthbert: Explaining what a sibling is would be helpful.

music21/musicxml/m21ToXml.py
```diff
@@ -5654,6 +5660,32 @@ def wrapObjectInAttributes(self, objectToWrap, methodToMx):
 
         return mxAttributes
 
+    def _matches_previous_sibling_in_group(self,
```
mscuthbert, line 5663: this is okay as underscore case as private, but the name should match the attribute.

music21/musicxml/m21ToXml.py
```diff
@@ -5977,20 +6009,21 @@ def setMxAttributesObjectForStartOfMeasure(self):
...
-            smts = list(m.getElementsByClass('SenzaMisuraTimeSignature'))
-            if smts:
-                mxAttributes.append(self.timeSignatureToXml(smts[0]))
-
-            # For staves, see joinPartStaffs()
-            # TODO: part-symbol
-            # TODO: instruments
-            if m.clef is not None:
-                mxAttributes.append(self.clefToXml(m.clef))
+        if m.keySignature is not None and not self._matches_previous_sibling_in_group(
```
mscuthbert, line 6012: break two-part clauses between "None / and" -- wrap in parentheses.  (same below)

mscuthbert (changes requested):
LGTM -- just two style changes I'd prefer.  Thanks!

music21/musicxml/m21ToXml.py
jacobtylerwalls: Yeah, and it's not really any more keystrokes to call it `previousPartStaffInGroup` either...

PR #1211, opened 2022-02-06
instrumentLookup: new entries ++

MarkGotham:
Also consistency in:
- alphabetisation
- dict-closing `}` on new line

Note:
- no new classes added, only mappings;
- this commit = only `abbreviationsTo`, `englishTo`, and `bestNameTo`. Separate commit for languages involving transliteration (allowing that question to be handled separately).


music21/languageExcerpts/instrumentLookup.py
```diff
@@ -208,12 +216,15 @@
     'ocarina': 'Ocarina',
     'organ': 'Organ',
     'pan flute': 'PanFlute',
+    'percussion': 'Percussion',
     'piano': 'Piano',
     'piccolo': 'Piccolo',
     'pipe organ': 'PipeOrgan',
+    'pitched percussion': 'Pitched Percussion',
```
jacobtylerwalls, line 223: no space. here's a great example of why we should move over to class names as values, because this would have been caught.

jacobtylerwalls (commented):
Thanks, Mark, I would be in favor of merging this once the one bug below is fixed.

mscuthbert:
> Thanks, Mark, I would be in favor of merging this once the one bug below is fixed.

Can we make a test case that makes sure that all class names are real?

mscuthbert:
Thanks Mark.  Looks good.  Getting rid of Python 2 compatible strings and replacing with unicode is a big +.

PR #1212, opened 2022-02-07
Fix LocalCorpus.removePath() string/path confusion

jacobtylerwalls:
Removing paths using `LocalCorpus.removePath()` didn't work for corpora persisted to disk because the infrastructure expects strings, not `Path` objects. `Path` objects were being used since v5 when the `common.cleanpath()` default changed.

Added a test for the _in-memory_ corpus to at least add some coverage, but I didn't see a way to add coverage for the case that was failing without rewriting a lot of the system. (E.g. add more `verbose=` arguments to silence warnings, add a step to CI that creates a path that is not Desktop, etc.).

Adding that test was useful, though, since I discovered _that_ codepath _does_ expect `Path` objects.

### DIFF
--- a/music21/corpus/corpora.py
+++ b/music21/corpus/corpora.py
@@ -846,16 +846,30 @@ class LocalCorpus(Corpus):
 
         If that path is included in the list of persisted paths for the given
         corpus, it will be removed permanently.
+
+        >>> testCorpus = corpus.corpora.LocalCorpus(name='test')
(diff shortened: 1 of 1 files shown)


mscuthbert:
Thank you -- there are also still problems with saved Corpus objects having PosixPath objects rather than generic Path or plain strings in the metadata which makes searching impossible on Windows if the corpus is shared.  Haven't had time to investigate that.

Issue #1064, opened 2021-07-15
Instrument name lookup lists


MarkGotham:
Rather than adding ever more instruments to the `instrumentLookup.py` lists along with each possible small variant (e.g. singular vs plural), might a better approach be to add a `multiLingualLookUps` with basic regex that deal with most of the common instruments and variants at once?

Here's an example of how such a list might look for voices ...
```
('alt(o(s)?)?', 'alto'),  # ENG, FR, DE: alt, alto, altos
('bariton(e(s)?)?', 'baritone'),  # ENG, FR, DE (sing): bariton, baritone, baritones
('baritöne', 'baritone'),  # DE (pl)
('bass(e(s)?)?', 'bass'),  # ENG, FR: bass, basse, basses
('sopran(o(s)?)?', 'soprano'),  # ENG, FR, DE
('t(e|é)nors?', 'tenor'),  # ENG, FR (sing / pl); DE (sing)
('tenöre', 'tenor'),  # DE (pl)
('canto', 'voice'),  # IT
('chant(eu(r|se)?)?', 'voice'),  # FR
('gesang', 'voice'),  # DE
('(sing)?stimmen?', 'voice'),  # DE: stimme, stimmen, singstimme, singstimmen
('voc(a|e)', 'voice'),  # IT
('voi(x|ce)', 'voice'),  # ENG, FR
```
... paired with a simple function like ...
```
def multiLingualToBestName(query_string: str):
    for pattern, value in multiLingualLookUps:
        if re.search(pattern, query_string.lower()):
            return value
    return None
```
That would go in `instrumentLookup.py`, and need a couple of lines in `instrument.py` (before the others) ...
```
bestName = instrumentLookup.multiLingualToBestName(substring)
if not bestName:
    bestName = instrumentLookup.allToBestName[substring.lower()]
```

In the first instance, I'd suggest a simple form of this, replacing some (but perhaps not all) of the corresponding names from their respectively language lists. I.e. we can keep both initially to check things still work, and (assuming so) gradually transition over. As an additional check, the current, long lists could form the basis of more extensive test cases.

Then assuming that all works as expected, in a future PR, we can try to be more comprehensive:
- move all instruments up to the multilingual version where possible
- perhaps retain (much shorter) language-specific lists for special cases.

Ready to commit this for an initial sample of cases, if you approve the approach. If not, I'll just pile on some more instruments and variants!

MarkGotham:
[...]

PR #1215, opened 2022-02-08
Direct to className and add test

MarkGotham:


music21/languageExcerpts/instrumentLookup.py
```diff
@@ -1404,49 +1308,46 @@
...
-        self.assertEqual(allToBestName[russianTest], russianToBestName[russianTest])
-
-        spanishTest = 'acorde\xf3n'
-        self.assertEqual(allToBestName[spanishTest], spanishToBestName[spanishTest])
+    def testAllClassNames(self):
+        '''
+        Test that all class names are real.
+        '''
+        from music21 import instrument as instr
+        for v in allToClassName.values():
```
jacobtylerwalls, line 1349: Nitpick detail, but can these tests use self.subTest? Some examples in the codebase. So we don't have to re-run to get all the failures.

mscuthbert:
it looks really good.  It's not backwards compatible though.  I want to think for a bit if that matters so much though since these dictionaries have never been fully documented.  But they are exposed as part of the public interface, so we might need to merge this into a v.8 branch.  I know I've used the "abbreviationToBestName" in other code, though not the language lookups.

MarkGotham:
> it looks really good. It's not backwards compatible though. I want to think for a bit if that matters so much though since these dictionaries have never been fully documented. But they are exposed as part of the public interface, so we might need to merge this into a v.8 branch. I know I've used the "abbreviationToBestName" in other code, though not the language lookups.

Ok, noted and no problem. I'll leave it with you. I have a couple of final suggestions for this corner of the code base (or rather implementations of what we've discussed wrt this issue) but will hold until you decide on this.

mscuthbert:
I think that it's enough of an improvement that I'm okay with the lack of backwards compatibility in this shadowy part of the code.  Were there other things you wanted to add or shall I merge?

MarkGotham:
Thanks @mscuthbert.
This is complete in itself. 
I have a couple of final suggestions proposed on #1064 but they can be considered (and implemented or otherwise) separately.
So do go ahead and merge.

PR #1216, opened 2022-02-10
Represent inexpressible gaps as `<forward>` tags

jacobtylerwalls:
Solve an issue with musicxml export failing.

**Scenario**

Since eb1c953, musicxml import creates unlinked durations when there are discrepancies with the given types (e.g. due to tuplets and rounding errors). This can leave small gaps. If gaps are filled with hidden rests such as during voice creation, musicxml export chokes on "Inexpressible" durations.

**Solution**
Instead, for hidden inexpressible rests, just create `<forward>` tags.


mscuthbert:
Thanks!  LGTM.  Probably should just do this for all "inexpressible" durations, notes and rests.

Issue #254, opened 2017-10-05
Move warnings system to python warnings module


darthmall:
I'm parsing a pile of tunes in ABC from [thesession.org](https://thesession.org) and I'm encountering some invalid ABC (I think some of the forum members are making up their own shorthand for some things). Ideally, I'd like to just filter out these tunes, but music21 seems to be swallowing the errors here: https://github.com/cuthbertLab/music21/blob/13ec41980c8dec764d9d82b64588b787d29d6f8c/music21/abcFormat/__init__.py#L1547-L1549

An example can be found here: https://thesession.org/tunes/13690 (I've also attached a copy of the file: [shorelines.abc.txt](https://github.com/cuthbertLab/music21/files/1361051/shorelines.abc.txt).)

Could we let these exceptions bubble up and leave it up to the user to decide how they want to handle exceptions like this?

mscuthbert:
Could you wrap the parse command in a warnings.catch_filter and then discard anything that generates a warning?

darthmall:
This doesn't seem to work. Maybe I just set it up wrong, but as far as I can tell, `environLocal.warn` doesn't use the `warnings` module to log warnings, it just prints to `stderr`:

https://github.com/cuthbertLab/music21/blob/eeef3854f01976cdda2db7d31c5f670396cf5f3a/music21/environment.py#L1110

If I understand the `warnings` module correctly, `warnings.warn` does more than just print error messages, it uses classes derived from `Exception` to type warnings, which is how the filters work.

mscuthbert:
Ah, yes...we did that before the warnings module was really standardized.  I'll put it on a (long-range) to-do list to use warnings instead.  But you could switch sys.stderr to your own StringIO file and then check it to see if it's had anything written to it after parse and disregard anything like that.

For file formats that are hand coded (ABC, Humdrum, RomanText) music21 policy is to work around errors as best it can (like browsers do with HTML) -- human-encoded files tend to have lots of bugs and lax standards so we can't enforce perfect validity.  For file formats that are generated by other programs (musicxml, MIDI, etc.) then it's the responsibility of the creating software to generate correct code and except in a few cases we let errors from parsing those formats bubble up.

jacobtylerwalls:
- [ ] and in MIDI module, see: https://stackoverflow.com/questions/69378318/silence-music21-warnings

PR #1220, opened 2022-02-13
Make all import warnings MusicXMLWarning

jacobtylerwalls:
When this MusicXML import warning happens, it probably repeats hundreds of times (since it's just warning about less-than-ideal behavior by the software that produced the musicxml doc).
```
Cannot put in an element with a missing voice tag when no previous voice tag was given.  Assuming voice 1... 
```
To avoid duplicating this hundreds of times, just convert to a python warning instance (MusicXMLWarning) which by default emits once but can be filtered, shown always, raise exceptions, etc.

We were going to do this anyway as part of #254.


mscuthbert (approved):
Fine by me.  environLocal.warn should probably disappear or be replaced by something that uses the general Python warning system.  However, there are some warnings in there that are too important and should be emitted no matter what the settings are.

PR #1221, opened 2022-02-14
Distinguish which figures imply a bass note

### LINKED ISSUE #1218: romanText parsing of figures involving '4'
**music21 version**

`7.2.0`   

**Problem summary**

The romanText parser understands figures including '4' in unexpected ways (other than perhaps 6/4 chords).

**Steps to reproduce**

```
for fig in ("64", "54", "5/4", "4", "754", "74"):
    score = music21.converter.parse(
        f"m1 C: V{fig}", format="romanText"
    ).flatten()
    rn = next(score.getElementsByClass(music21.roman.RomanNumeral))
    print(fig, rn.pitchClasses)
```

```
64 [2, 7, 11] # OK, although some would of course say V64 should be [7, 0, 4]
54 [2, 7, 9] # expected [7, 0, 2]
5/4 [2, 7, 9] # expected [7, 0, 2]
4 [2, 7] # expected [7, 0]
754 [0, 5, 7, 11] # expected [7, 0, 2, 5]
74 [9, 2, 7] # expected [7, 0, 5]
```

All the chords seem to be transposed up 7 semitones. (7 -> 2, 0 -> 7, etc.)

jacobtylerwalls:
Fixes #1218 (Mark had additional proposals there that might benefit from a new issue)

@malcolmsailor @MarkGotham does this resolve your issue?

### DIFF
--- a/music21/roman.py
+++ b/music21/roman.py
@@ -61,6 +61,19 @@ _keyCache: Dict[str, key.Key] = {}
 _NOTATION_SINGLETON = fbNotation.Notation()
 
 
+# only some figures imply bass (e.g. "54" does not)
+FIGURES_IMPLYING_BASS: Tuple[Tuple[int, ...]] = (
+    # triads
+    (6,), (6, 3), (6, 4),
+    # seventh chords
+    (6, 5, 3), (6, 5), (6, 4, 3), (4, 3), (6, 4, 2), (4, 2), (2),
+    # ninth chords
+    (7, 6, 5, 3), (6, 5, 4, 3), (6, 4, 3, 2), (7, 5, 3, 2),
+    # eleventh chords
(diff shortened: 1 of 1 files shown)


malcolmsailor:
Thanks for the speedy work!

It almost resolves the issue, at least as far as I'm concerned. In C major, "V754" gives [7, 0, 2, 6] rather than [7, 0, 2, 5]. (I.e., F-sharp rather than F-natural as I would expect.) Not sure why, since "V74" gives [7,0,5].

jacobtylerwalls:
@malcolmsailor Thanks for finding the additional flaw. 853fe90 should take care of it.

mscuthbert:
Looks great to me.  Just to note though that roman numerals are not a complete descriptive system for all chords, so we're unlikely to make it so that every single chord round-trips perfectly nor that every chord symbol can be notated w/ Roman numerals.  But glad to get some more expansion as we can.

PR #1223, opened 2022-02-16
Make full scores containing piano parts exportable to braille

jacobtylerwalls:
Previously, full scores containing piano music could not be exported.

```python
>>> beach = corpus.parse('beach')
>>> beach.show('braille')
  File "/Users/.../music21/music21/stream/base.py", line 334, in show
  File "/Users/.../music21/music21/base.py", line 2758, in show
  File "/Users/.../music21/music21/converter/subConverters.py", line 502, in show
    super().show(obj, fmt, app=None, subformats=subformats, **keywords)
  File "/Users/.../music21/music21/converter/subConverters.py", line 212, in show
    returnedFilePath = self.write(obj, fmt, subformats=subformats, **keywords)
  File "/Users/.../music21/music21/converter/subConverters.py", line 510, in write
    dataStr = braille.translate.objectToBraille(obj, **keywords)
  File "/Users/.../music21/music21/braille/translate.py", line 186, in objectToBraille
    return streamToBraille(music21Obj,
  File "/Users/.../music21/music21/braille/translate.py", line 288, in streamToBraille
    return keyboardPartsToBraille(music21Stream,
  File "/Users/.../music21/music21/braille/translate.py", line 620, in keyboardPartsToBraille
    raise BrailleTranslateException('Can only translate two keyboard parts at a time')
music21.braille.translate.BrailleTranslateException: Can only translate two keyboard parts at a time
```

### DIFF
--- a/music21/braille/examples.py
+++ b/music21/braille/examples.py
@@ -265,7 +265,8 @@ Barline final ⠣⠅
         from music21.braille.translate import objectToBraille
         verdi = corpus.parse('verdi/laDonnaEMobile')
         x = objectToBraille(verdi, debug=True)
-        y = '''---begin grand segment---
+        y = '''⠠⠍⠕⠧⠑⠍⠑⠝⠞⠀⠠⠝⠁⠍⠑⠒⠀⠇⠁⠠⠙⠕⠝⠝⠁⠠⠑⠠⠍⠕⠃⠊⠇⠑⠲⠍⠭⠇
+---begin grand segment---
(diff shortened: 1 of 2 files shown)


mscuthbert (approved):
Looks great to me!  I think somewhere there is (was? should have been?) a general method that takes a Generic Stream and returns the best idea of what it might be (Score? Part? Measure?) -- something we might consider using here and getting rid of the limitation.  Then we can get rid of the special case for generic streams.

PR #1226, opened 2022-02-18
Improvements to single measure extraction to musicxml

jacobtylerwalls:
Improvements to `.measure(n).show()`:
- show the measure number (if no measure number style was set) so you can see what you exported
- show the part name and part abbreviation from the containing part
- fix a bug in musicxml export with a nonexistent member: `AttributeError: 'StreamStyle' object has no attribute 'measureNumberingStyle'`
- get the measure's clef from earlier in the context (previous measure)
- gather spanners

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -488,6 +488,7 @@ class GeneralObjectExporter:
         representation of a Measure, not for partial
         solutions in Part or Stream production.
         '''
+        m.coreGatherMissingSpanners()
         mCopy = m.makeNotation()
         if mCopy.style.measureNumbering is None:
             # Provide a default
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -492,8 +492,14 @@ class GeneralObjectExporter:
         if mCopy.style.measureNumbering is None:
             # Provide a default
             mCopy.style.measureNumbering = 'measure'
-        if not m.recurse().getElementsByClass('Clef').getElementsByOffset(0.0):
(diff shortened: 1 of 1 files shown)


music21/musicxml/m21ToXml.py
```diff
@@ -488,12 +488,26 @@ def fromMeasure(self, m):
         representation of a Measure, not for partial
         solutions in Part or Stream production.
         '''
+        m.coreGatherMissingSpanners()
```
mscuthbert, line 491: I wonder if this should happen by default in Stream.makeNotation()?  or will it lead to redundancies?

music21/musicxml/m21ToXml.py
```diff
@@ -488,12 +488,26 @@ def fromMeasure(self, m):
...
         mCopy = m.makeNotation()
-        if not m.recurse().getElementsByClass('Clef').getElementsByOffset(0.0):
+        if mCopy.style.measureNumbering is None:
+            # Provide a default
+            mCopy.style.measureNumbering = 'measure'
```
mscuthbert, line 495: I think that measureNumbering = 'start' should probably be applied to Parts created with the .measures(start, end) tag.

Btw -- I have a lot of measureNumbering off because it wasn't imported in musicxml readers and no number is better than a wrong number.  I haven't checked recently (just got Finale 27) to see if this is till the case.

mscuthbert (approved):
LGTM

music21/musicxml/m21ToXml.py
mscuthbert: Finale 27 properly imports measure numbers.

PR #1232, opened 2022-02-21
Import ottava objects from musicXML (as non-transposing)

jacobtylerwalls:
Import `spanner.Ottava` objects from `<octave-shift>` tags in MusicXML documents.

Set them as non-transposing, since MusicXML documents encode the sounding pitches, and that's what was imported before anyway. Backwards compatible in that respect.

Export was already supported.


mscuthbert (approved):
LGTM!

PR #1233, opened 2022-02-26
Surface `__eq__()` and `__getitem__()` in doc builds

jacobtylerwalls:
We weren't publishing any of the documentation for special ("dunder") methods like `__eq__()` and `__getitem__()`, some of which are very useful to read about, like the new v7 bracket syntax.

Presumably this was to avoid polluting every class with the docs for Python's base implementations on `list`, etc, but this PR works around that by loosening the check for `object` to any builtin (`list`, etc.)
***

Screen Shot 2022-02-26 at 12 25 44 PM


mscuthbert:
thanks!

Issue #904, opened 2021-03-08
Express durations with or without tuplets according to context


jacobtylerwalls:
**music21 version**
reproduced on current as well as 5.7.2

**Problem summary**
We have several use cases for automatic tuplet fixing. One of them is when a cross staff tuplet leaves a gap in one hand of piano music that has been separated. Using a quarter rest to fill the gap between two incomplete tuplets is a recipe for malformed musicxml output. The other use case for tuplet consolidation is if we end up creating more tuplets in #763. The humdrum parser calls `duration.TupletFixer`, but nobody else does. Maybe for v7, if we can swing it, we can look into getting this into the makeNotation/musicXML export pipeline. This might be a reason to put #763 on ice for now.

**Steps to reproduce**

```
s = stream.Stream()
s.append(note.Note(quarterLength=(1/6)))
s.insert((5/6), note.Note(quarterLength=(1/6)))
s.show()
```

**Expected vs. actual behavior**

Desired

![expected](https://user-images.githubusercontent.com/38668450/110344144-3523f580-7ffb-11eb-8eab-9d805617cd44.png)

Actual

![actual](https://user-images.githubusercontent.com/38668450/110344039-1aea1780-7ffb-11eb-920a-86edd902ca6d.png)

**More information**
@napulen noticed in https://github.com/MarkGotham/When-in-Rome/issues/17

jacobtylerwalls:
Ah, right, you already mentioned this was the very long term goal to handle these streams in https://github.com/cuthbertLab/music21/pull/763#issuecomment-757373425

napulen:
Just FYI, there may be other (real-world) pieces by Fanny Mendelssohn Hensel with very similar patterns, which also trigger this issue. 

Thanks for the minimal example presented here.

PR #1239, opened 2022-03-01
Fix next(activeSiteOnly=True) finding things in other streams

jacobtylerwalls:
Found while starting on #904


mscuthbert:
Thanks!

PR #1240, opened 2022-03-01
Apply tuplet to multiple components to express 5/6 or 7/3 QL

jacobtylerwalls:
Fixes #572
Same PR as #763, just with conflicts fixed and commits squashed

**Before**: Durations such as 5/6 or 7/3 QL were expressed as non-power of 2 tuplets, which caused display problems.

**Now**: Upon extracting the largest type (such as 0.5 QL from 5/6 QL or 2.0 QL from 7/3 QL), if the remainder can be expressed as a single tuplet, express the largest type and the remainder in smaller components over which a tuplet can be applied.

***
@gregchapman-dev would you be able see if it helps with your `makeRests()` issues?

This is not mergeable until we have a solution for #904. I have a diff for that I should be able to PR in a week or so--still testing it.

PR #1242, opened 2022-03-03
Add makeNotation routines for completing or consolidating tuplets

### LINKED ISSUE #904: Express durations with or without tuplets according to context

jacobtylerwalls:
Fixes #904

@gregchapman-dev , please share your findings if you give this a look -- you'll need makeNotation=True turned on during musicxml export, or else need to make your own calls to these new makeNotation-y routines. You might also want #1240, so a local branch merging all of that in might be the ticket. (#1240 avoids making the implausible tuplet, and then after running splitAtDurations() to get those 11 components isolated (again... another makeNotation-y routine to run yourself), this PR puts them back together in threes to get quarter rests again.)

This still doesn't get you a leading half rest -- for that, you'll want #992. Maybe @MarkGotham can reignite the effort there: I recall we were close, we solved the beat sequence extraction and just needed a PR to nail down the API, I think (?)


music21/stream/makeNotation.py
```diff
@@ -1857,6 +1855,203 @@ def setStemDirectionOneGroup(
...
+
+    New in v8.
+
+    >>> from music21.stream.makeNotation import splitElementsToCompleteTuplets
+    >>> s = stream.Stream(
+    ...    [note.Note(quarterLength=1/3), note.Note(quarterLength=1), note.Note(quarterLength=2/3)]
+    ... )
+    >>> splitElementsToCompleteTuplets(s)
+    >>> [el.quarterLength for el in s.notes]
+    [Fraction(1, 3), Fraction(2, 3), Fraction(1, 3), Fraction(2, 3)]
```
mscuthbert, line 1882: can we have a test that the second and third notes are tied?

mscuthbert:
If the tie adding test is added, this is ready to go and merge.  THANKS!

mscuthbert:
Congrats @jacobtylerwalls -- huge amount of work.

Issue #1123, opened 2021-09-11
Improving Noteworthy converter support


nvuaille:
**Motivation**
Improving the Noteworthy Converter, for more interoperability

**Feature summary**
Centered on version 1.75,  binary  and text format.

**Proposed implementation**
Improve supported NWC objects parsing  in https://github.com/cuthbertLab/music21/blob/master/music21/noteworthy/binaryTranslate.py
By design, this also require some improvements in https://github.com/cuthbertLab/music21/blob/master/music21/noteworthy/translate.py  (`binary translate` pass a text stream to 
the `translate` module) 

See my current branch [here](https://github.com/nvuaille/music21/tree/Noteworthy). I think I have most of the features I want, I still need update/fix current tests and add more =)

**Intent**

 * I plan on implementing this myself.

Is it something music21 would like to have ? Is there some recommendation about this kind of feature ?

In any case, thanks for all the work you did on this great tool !

mscuthbert:
most welcome contribution if you have it!  I'll need to fire up my PC to get some more files.

PR #1247, opened 2022-03-06
Work on noteworthy 1 75 binary reader

nvuaille:
* Improve support for many nwc objects:
    * alterations
    * grace notes
    * tempo mark
    * triplets
    * text position
    * chords / multiple pitches
    * repeat brackets

  * Readibility:
    * put constants in their own file
    * rename some methods (e.g. NWCStaff.parse() -> NWCStaff.parseStaff())

  * Use exception to print error infos on music21 objects creation
  * Update documentation
  * Update tests

refeers to https://github.com/cuthbertLab/music21/issues/1123


music21/noteworthy/translate.py
```diff
@@ -225,9 +257,7 @@ def setDurationForObject(self, generalNote, durationInfo):
 
         for kk in parts:
             if kk == 'Grace':
-                # Now it doesn't work, the function for grace notes have to be added here
-                warnings.warn('skipping grace note')
-                return
+                durationObject = durationObject.getGraceDuration()
```
jacobtylerwalls, line 260: Thank you for fixing this. I think the warnings filter in `testComplete` can now be removed. Would you be up for cleaning that up?

music21/noteworthy/translate.py
```diff
@@ -840,6 +972,61 @@ def createLyrics(self, attributes):
...
+    def createStaffInstrument(self, attributes):
+        patch = int(attributes.get('Patch', 0))
+        try:
+            staffInstru = instrument.instrumentFromMidiProgram(patch)
+            readTranspo = int(attributes.get('Trans', 1))
+            self.currentTransposition = 12 * readTranspo // 12
+            staffInstru.transposition = interval.Interval(readTranspo)
+            self.currentPart.append(staffInstru)
+        except InstrumentException:
+            for p in range(34, 81):
```
jacobtylerwalls, line 993: I don't think we should be creating 47 Instrument objects. Can we create one `UnpitchedPercussion` object?

music21/noteworthy/translate.py
```diff
@@ -840,6 +972,61 @@ def createLyrics(self, attributes):
...
+                self.currentPart.append(i)
+            self.currentPart.partName = 'Drumset'
+            self.currentPart.partAbbreviation = 'D.'
+
+    def createText(self, attributes):
+        text = attributes['Text']
+        # Rehearsal is associated to the score while TextExpression is staff-associated.
+        expr = expressions.TextExpression(text)
+        expr.positionPlacement = 'above' if int(attributes.get('Pos')) < 0 else 'below'
+        expr.style.relativey = attributes.get('Pos')
```
jacobtylerwalls, line 1006:
```suggestion
        expr.style.relativeY = attributes.get('Pos')
```

music21/noteworthy/translate.py
```diff
@@ -840,6 +972,61 @@ def createLyrics(self, attributes):
...
+        md.title = title
+        md.author = author
+        self.score.append(md)
+
+    def createEnding(self, attributes):
+        end = attributes['Endings']
+        if self.currentMeasure.leftBarline is None:
+            self.currentMeasure.leftBarline = bar.Barline()
+        self.repeatedMeasures = [self.currentMeasure]
+        endingType = end.split(',')[0]
```
jacobtylerwalls, line 1022: Reason we just need the first value? `RepeatBracket` can take a comma-separated list of endings.

jacobtylerwalls (commented):
Thank you for working on this improved converter! The adjusted doctest showing the better results is great.

Let me know if you have any questions or if I misunderstood something (since I don't know this format).

music21/noteworthy/binaryTranslate.py
```diff
@@ -309,7 +317,7 @@ def advanceToNotNUL(self, nul=b'\x00'):
             pp += 1
         self.parsePosition = pp
 
-    def parse(self):
+    def parseContent(self):
```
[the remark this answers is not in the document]
nvuaille: No strong arguments here. As far as I can remember it was a personnal readability helper, as several `parse` method are defined in this file (in different classes of course). Also, I thought it was more some kind of internal API than something people use from the other modules, but I can be wrong.

 I will revert if you prefer.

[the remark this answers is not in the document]
jacobtylerwalls: > I thought it was more some kind of internal API than something people use from the other modules, but I can be wrong.

You are probably right that no one is using these helper methods, but since they're not marked as private, I just think it's safer to not break compatibility if there's no upside beyond code readability. See below for a post where I suggested that someone override helper methods the MIDI parser rather than rewrite the whole converter:

https://groups.google.com/g/music21list/c/ByloLCLCtLA/m/Vy9kNw7hBgAJ

music21/noteworthy/translate.py
```diff
@@ -840,6 +972,61 @@ def createLyrics(self, attributes):
...
+        patch = int(attributes.get('Patch', 0))
+        try:
+            staffInstru = instrument.instrumentFromMidiProgram(patch)
+            readTranspo = int(attributes.get('Trans', 1))
+            self.currentTransposition = 12 * readTranspo // 12
+            staffInstru.transposition = interval.Interval(readTranspo)
+            self.currentPart.append(staffInstru)
+        except InstrumentException:
+            for p in range(34, 81):
+                i = instrument.unpitchedInstrumentFromPercMapPitch(p)
```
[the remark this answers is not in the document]
nvuaille: Oh thanks, good catch! It was on my branch but I did some extra-cleanup ...

Seeing other comment concerning this loop, I'm  going to rework that part.

music21/noteworthy/translate.py
```diff
@@ -453,20 +487,41 @@ def translateNote(self, attributes):
         durationInfo = attributes['Dur']
         pitchInfo = attributes['Pos']
 
-        n = note.Note()   # note!
+        p = self.getPitchFromPositionInfo(pitchInfo)
+
+        if self.currentClef == 'PERCUSSION':
+            n = note.Unpitched()
+            n.fromDisplayPitch(p)
```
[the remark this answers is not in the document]
nvuaille: Oh. I Will do a quick rework on my `Unpitched` instrument integration

music21/noteworthy/translate.py
```diff
@@ -840,6 +972,61 @@ def createLyrics(self, attributes):
...
+
+    def createStaffInstrument(self, attributes):
+        patch = int(attributes.get('Patch', 0))
+        try:
+            staffInstru = instrument.instrumentFromMidiProgram(patch)
+            readTranspo = int(attributes.get('Trans', 1))
+            self.currentTransposition = 12 * readTranspo // 12
+            staffInstru.transposition = interval.Interval(readTranspo)
+            self.currentPart.append(staffInstru)
+        except InstrumentException:
```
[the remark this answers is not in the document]
nvuaille: Actually unneeded, removed

music21/noteworthy/translate.py
nvuaille: As far as I understood, first value is the type and correspond to `RepearBracket.number` and (optional) second is the 'type', but I didn't find a correct doc about it, so I let it unused

nvuaille:
@jacobtylerwalls , thanks for your review, here is a new version improved with your comments

music21/noteworthy/binaryTranslate.py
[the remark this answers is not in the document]
nvuaille: Oh, miss your message before pushing. I'm going to revert that and push again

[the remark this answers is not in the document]
nvuaille: updated

jacobtylerwalls:
Great, thanks for the changes. I expect Myke will want to weigh in with his review. At that point please avoid force-pushing, as whoever merges has the option to squash the commits, and in the meantime it prevents people from assessing what changed between reviews. Thanks!

jacobtylerwalls (approved):

music21/noteworthy/translate.py
```diff
@@ -453,20 +486,39 @@ def translateNote(self, attributes):
...
 
-        n = note.Note()   # note!
+        p = self.getPitchFromPositionInfo(pitchInfo)
+
+        if self.currentClef == 'PERCUSSION':
+            n = note.Unpitched()
+            n.fromDisplayPitch(p)
+        else:
+            n = note.Note()   # note!
+            n.pitch = p
```
mscuthbert, line 496: this should be:

`n = note.Note(p)`

will speed up creation substantially.  (Otherwise a default C4 pitch is created and then dropped).

music21/noteworthy/translate.py
```diff
@@ -453,20 +486,39 @@ def translateNote(self, attributes):
...
 
         # if Lyrics
         if self.lyrics and self.lyricPosition < len(self.lyrics):
             n.addLyric(self.lyrics[self.lyricPosition])
 
-        self.currentMeasure.append(n)
+        self.getCurrentVoice().append(n)
+
+    # current voice has shorter duration. If no voices, return current measure
+    def getCurrentVoice(self):
```
mscuthbert, line 511: needs docs, and return typing.

music21/noteworthy/translate.py
```diff
@@ -453,20 +486,39 @@ def translateNote(self, attributes):
...
+        voice = None
```
mscuthbert, line 512: why not make `voice = self.currentMeasure` to make the end just `return voice`

music21/noteworthy/translate.py
mscuthbert: Especially since it doesn't necessarily return a stream.Voice object as the name implies.

music21/noteworthy/translate.py
```diff
@@ -453,20 +486,39 @@ def translateNote(self, attributes):
...
+        shorter = 999
```
mscuthbert, line 513: this should be a float so that typing works for it.  And typed as `OffsetQL` since it can also be a Fraction.  `float('inf')` would be the best way of implying that any voice is shorter than this.

music21/noteworthy/translate.py
```diff
@@ -453,20 +486,39 @@ def translateNote(self, attributes):
...
+        for item in self.currentMeasure:
+            if isinstance(item, stream.Voice):
```
mscuthbert, line 515: iterate as `for v in self.currentMeasure.getElementsByClass(stream.Voice):`

that way you're not setting activeSite etc. for non-matching elements.

music21/noteworthy/translate.py
mscuthbert: best would be renaming as `shortestStream`

music21/noteworthy/translate.py
mscuthbert: and rename as `shortestQL`

music21/noteworthy/translate.py
mscuthbert: as this is now called for every note, it needs to be optimized as much as possible.  I'm seeing most optimized version as:

```
if self.currentMeasure.isFlat:
     # O(1) in most cases.
     return self.currentMeasure

# O(n) only if voices already exist.
return min(
     self.currentMeasure.getElementsByClass(stream.Voice), 
     key=lambda v: v.quarterLength
)
```

music21/noteworthy/translate.py
```diff
@@ -476,33 +528,91 @@ def translateChord(self, attributes):
...
+        def getVoiceAtDuration(inner_self, voiceId, dur):
+            # first check if voice already exists in measure
+            voice = None
+            hasVoice = False
+            for item in inner_self.currentMeasure:
+                if isinstance(item, stream.Voice):
+                    hasVoice = True
+                    if item.id == voiceId:
+                        voice = item
+                        break
```
mscuthbert, line 556: as above, avoid iteration of measure if no voices yet, and in this case you can do

```
voice = None
voice = inner_self.currentMeasure.getElementsByClass(stream.Voice).getElementById(voiceId)
```

mscuthbert (changes requested):
I still need to review from l. 558 of nw/translate onwards, but looking very good.  Some changes requested, but nothing fundamental, and this will move to acceptance as soon as they're made and I'm able to verify that my collections of nwc and nwctxt files still parse.

nvuaille:
Hi @mscuthbert , thanks for the review!

I'm a bit busy in july but I should come back on this during august.

jacobtylerwalls (dismissed):
FYI still seeing `fromDisplayPitch` (nonexistent method) called in latest commit, so unresolving the relevant comment.

music21/noteworthy/binaryTranslate.py
```diff
@@ -466,14 +484,30 @@ def parse(self):
...
+
+        staffString = '|AddStaff|Name:' + label
+        dumpObjects.append(staffString)
+
+        staffInstruString = '|StaffInstrument|Name:' + instruName
+        if instruName in constants.MidiInstruments:
+            staffInstruString += '|Patch:'
+            staffInstruString += str(constants.MidiInstruments.index(self.instrumentName))
+
+        staffInstruString += '|Trans:' + str(self.transposition)
```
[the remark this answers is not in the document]
nvuaille: Indeed, I think falling back to the first midi instru ( `Acoustic Grand Piano`) is a better choice.

nvuaille (commented):
I just came back on it, sorry for the delay. I rebased on master but I created a new commit for the changes related to the review

music21/noteworthy/binaryTranslate.py
```diff
@@ -437,10 +449,16 @@ def toStream(self):
...
+    (see :meth:`parseHeader`). It may also contains some lyrics (see :meth:`parseLyrics`).
+    It defines a :meth:`dump` method that return a list of string containing the
+    nwctxt-formatted content of each object of the staff.
+    '''
     def __init__(self, parent=None):
-        self.parent = parent
+        self.parent: NWCConverter = parent
         self.lyrics = []
-        self.objects = []
+        self.objects: list[NWCObject] = []
```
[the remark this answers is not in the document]
mscuthbert: no longer relevant now, correct.

mscuthbert:
Sorry that I forgot to merge this.  Looks like it's good to go

mscuthbert (approved):

nvuaille:
Obviously I also forgot this one ... Thanks a lot for the integration !

PR #1252, opened 2022-03-18
Only set a default TempoText style if there is no style already set.

gregchapman-dev:
… And never set a default TempoText style when getting a property.

### DIFF
--- a/music21/tempo.py
+++ b/music21/tempo.py
@@ -183,29 +183,6 @@ class TempoText(TempoIndication):
     <music21.tempo.TempoText 'adagio'>
     >>> print(tm.text)
     adagio
-    >>> print(tm.style.absoluteY)
-    45
-    >>> tm.style.absoluteY = 33
-    >>> print(tm.style.absoluteY)
-    33
-    >>> print(tm.getTextExpression())
-    <music21.expressions.TextExpression 'adagio'>
-    >>> print(tm.style.absoluteY)
-    33
-    >>> te = music21.expressions.TextExpression('andante')
-    >>> te.style.absoluteY = 38
-    >>> print(te)
-    <music21.expressions.TextExpression 'andante'>
-    >>> print(te.style.absoluteY)
-    38
-    >>> tm.setTextExpression(te)
-    >>> print(tm.getTextExpression())
-    <music21.expressions.TextExpression 'andante'>
-    >>> print(tm.text)
-    andante
-    >>> print(tm.style.absoluteY)
-    38
-    >>>
     '''
 
     def __init__(self, text=None):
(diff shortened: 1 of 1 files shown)


gregchapman-dev:
Working on some tests...

gregchapman-dev:
Committed the tests. How do I rerun coveralls?

gregchapman-dev:
Figured it out (forgot to push).  I'll have to figure out why the tests failed another day.

gregchapman-dev:
I think this is (finally) ready to go.  Thanks!

mscuthbert:
Can you explain what the problem this PR is trying to solve and what existing cases might be affected by this change?

gregchapman-dev:
Yeah, I got a little ahead of myself…. Sorry!

There are three related problems fixed here, and the answers to your questions vary a bit for each one:

1. Problem being solved: TempoText.getTextExpression() overwrites the TempoText’s style (which is also the underlying TextExpression’s style, and the possibly-present-enclosing-MetronomeMark’s style) with special TempoText defaults (fontStyle=‘bold’, absoluteY=20 or 45).  The setting back to defaults is NOT reflected in the returned TextExpression, which is a deep copy taken before style is set back to defaults, but it does change the results of future calls.

Effect on existing cases: The only existing cases affected negatively by this fix would be code that sets MetronomeMark/TempoText style, and then makes a call that would change the style back to default, and _then_ has a bug where if the style _doesn't_ change back, bad things will happen.  This sounds silly, perhaps, but Murphy’s Law (and my experience) says it’s a real possibility.  However, I believe the benefit outweighs this potential downside.

2. Problem being solved: TempoText._setText (aka. tt.text = ‘andante’) was inconsistent in its efforts to link/set style properly.  If there already was a textExpression, that textExpression’s text would be set to ‘andante’, and style left alone (yay).  If there was not already a textExpression, a new TextExpression would be created, and then a decision was made as to which style to use:  If the TempoText did not already have style info, the TempoText would be set up to use the new TextExpression’s style (which would thus be created).  And then that style would be set to TempoText default.  This is all fine.  But here’s the inconsistency: if the TempoText _did_ already have style info, the new TextExpression would be set up to use the TempoText’s style (so far so good), and then that style would be whacked back to default, negating the effort to use the right style.

Effect on existing cases: Cases affected negatively by this fix would be code that is setting a TempoText’s text, then setting style fields on the TempoText, and _then_ expecting that a second set of the TempoText’s text would set the style back to default (which it would, of course, no longer do).  This is a real possibility.  I would be OK removing this particular fix if it’s really important to avoid that risk, especially since this bug can be worked around by getting/setting the TempoText style before/after setting TempoText.text (and probably around some other calls to MetronomeMark, too).

3. Problem being solved: TempoText.setTextExpression(te) was whacking TempoText.style back to default, but was _not_ linking the TextExpression and TempoText styles together.  I chose to fix this by doing something similar to what was being attempted in TempoText._setText (linking to the style that was already in place, and only whacking to default style if there were no previously set styles in place.

Effect on existing cases: Very similar to #2.  Potential workaround: get/set style around calls to TempoText.setTextExpression (and any MetronomeMark calls that call TempoText.setTextExpression).  Also, before setting the style after the call, relink the styles by hand.

Summary: (1) is the problem I initially set out to fix, since both musicdiff and my new Humdrum parser/writer are trying to get style exactly right, and having the text style whacked out from under me at (what appeared to be) random moments was distressing, to say the least.  (2) and (3) make things even more consistent and predictable, but I could be talked out of them if the risk to existing cases is judged to be too high.  The benefit of (3) is a bit higher than that of (2), in my opinion, since it fixes a failure to link styles, which could cause some really unpredictable behavior in score rendering/inspection.  If we reject (2), we could also modify (3), to re-link the styles, but go ahead and whack style back to default as before.

Thoughts?

Greg

mscuthbert:
Thanks for the great explanation.  All sounds good, and the improved cases far outweigh those that will break.  Merging.  Awesome work!

gregchapman-dev:
Thank you!

Greg

PR #1254, opened 2022-03-24
Pylint fixes to unblock CI

jacobtylerwalls:
updates for staying on latest pylint (2.14)

### DIFF
--- a/music21/configure.py
+++ b/music21/configure.py
@@ -13,6 +13,7 @@ import pathlib
 import re
 import time
 import sys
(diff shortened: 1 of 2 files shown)


mscuthbert:
looks good -- it'd probably save us headache in the future if RedHat didn't crash; even though we don't support.

mscuthbert:
But the whole notion of running configure.py should basically be obsolete.  It comes from the pre-pip days.

PR #1255, opened 2022-03-25
getAllNamesForInstrument (completes #1064)

MarkGotham:


mscuthbert:
closing and reopening because tests did not run.

music21/instrument.py
```diff
@@ -2485,6 +2485,80 @@ def fromString(instrumentString: str,
...
+
+    '''
+
+    language = language.lower()
+    instrumentNameDict = {}
+
+    instrumentClassName = instrumentClass.instrumentName
+
+    if language == 'all':
+        for lang in [str(x) for x in SearchLanguage][1:]:  # skip 'all' here
```
jacobtylerwalls, line 2552:
```suggestion
    if language == SearchLanguage.ALL:
        for lang in SearchLanguage:
            if lang is SearchLanguage.ALL:
                continue
```

jacobtylerwalls (commented):
a mypy/readability tip for you!

music21/instrument.py
MarkGotham: Thanks very much. Shame to check each time. There are only a few iterations here, but it might be useful to have iteration over enum in other contexts.

mscuthbert (approved):
LGTM.  In the future when concatenating a StrEnum to another string, I prefer `str(SearchLanguage.GERMAN) + 'toClassName'`  with the explicit casting to string to make it very clear what the output will be.

MarkGotham:
Thanks @mscuthbert. Noted!

PR #1261, opened 2022-04-03
Avoid `<forward>` tags from expressions out of measure bounds

jacobtylerwalls:
Avoid creating `<forward>` tags from expressions that might have an offset exceeding the measure.
Alternative to #706

Side benefit: creates a tendency toward positive `<offset>` values over negative ones.

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -3038,15 +3038,19 @@ class MeasureExporter(XMLExporterBase):
 
         self.currentVoiceId = voiceId
 
-        # group all objects by offsets and then do a different order than normal sort.
-        # that way chord symbols and other 0-width objects appear before notes as much as
-        # possible.
-        for objGroup in OffsetIterator(m):
-            groupOffset = m.elementOffset(objGroup[0])
+        # Group just notes and rests (GeneralNote) by their offsets so that
+        # <forward> tags (which become hidden rests) only bridge gaps between
+        # notes and rests, and not gaps formed with expressions that can
+        # be represented instead with <offset> tags.
+        # See m. 1 of Weber piece in core corpus
+        general_note_offsets = []
+        for objGroup in OffsetIterator(m, filterList=[lambda n: isinstance(n, note.GeneralNote)]):
+            general_note_offsets.append(m.elementOffset(objGroup[0]))
+        for i, groupOffset in enumerate(general_note_offsets):
             amountToMoveForward = int(round(divisions * (groupOffset
                                                              - self.offsetInMeasure)))
             if amountToMoveForward > 0:
-                # gap in stream!
+                # gap in stream between GeneralNote objects: create <forward>
                 mxForward = Element('forward')
                 mxDuration = SubElement(mxForward, 'duration')
                 mxDuration.text = str(amountToMoveForward)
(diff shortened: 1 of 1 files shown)


music21/musicxml/m21ToXml.py
```diff
@@ -5255,17 +5276,20 @@ def setOffsetOptional(self, m21Obj, mxObj=None, *, setSound=True):
...
         return mxOffset
 
-    def placeInDirection(self, mxObj, m21Obj=None):
+    def placeInDirection(self, mxObj, m21Obj=None, *, setSound=True):
         '''
-        places the mxObj <element> inside <direction><direction-type>
+        Places the mxObj <element> inside <direction><direction-type>
+        and sets <offset> if needed.
+
+        Changed in v7.3 -- added `setSound` keyword (see :meth:`setOffsetOptional`).
```
mscuthbert, line 5284: should be Changed in v8.

music21/musicxml/m21ToXml.py
```diff
@@ -3038,25 +3038,46 @@ def parseFlatElements(self, m, *, backupAfterwards=False):
...
+            else:
+                nextGroupOffset = general_note_offsets[i + 1]
+            for obj in m.getElementsByOffset(
+                offsetStart=groupOffset,
+                offsetEnd=nextGroupOffset,
+                includeEndBoundary=is_last_gn_offset,
+                mustFinishInSpan=False,
+                mustBeginInSpan=True,
+                includeElementsThatEndAtStart=True,
+            ):
```
mscuthbert, line 3076: getElementsByOffset is pretty slow -- and it ends up as a for-loop within a for-loop for what most of the time would be a no-op.

I wonder if there's a way to use the older code, getting everything in the OffsetIterator but only emitting a `<forward>` tag and setting offsetInMeasure if `any(isinstance(n, GeneralNote) for n in offsetGroup)`.  And making a separate list of offsetGroups with notes and without (or better, iterate over a cheap dataclass of `.offset`, `.elements`, `.hasSomeNotes`, `.isLast`, `.distanceToNextGroup`)  -- Does this make sense?

music21/musicxml/m21ToXml.py
```diff
@@ -5255,17 +5276,20 @@ def setOffsetOptional(self, m21Obj, mxObj=None, *, setSound=True):
...
         mxDirectionType = SubElement(mxDirection, 'direction-type')
         mxDirectionType.append(mxObj)
-        if (m21Obj is not None
-                and hasattr(m21Obj, 'placement')
-                and m21Obj.placement is not None):
-            mxDirection.set('placement', m21Obj.placement)
+        if m21Obj is not None:
+            if hasattr(m21Obj, 'placement') and m21Obj.placement is not None:
+                mxDirection.set('placement', m21Obj.placement)
+            self.setOffsetOptional(m21Obj, mxDirection, setSound=setSound)
```
mscuthbert, line 5292: thank you

music21/musicxml/m21ToXml.py
mscuthbert: er... `.distanceToNextGroupHavingNotes` I suppose.

mscuthbert (commented):
It looks really good -- I'm wondering if we can avoid doing an additional getElementsByOffset within the loop over OffsetIterator somehow.

music21/musicxml/m21ToXml.py
jacobtylerwalls: Smart, and it looks like just doing the `isinstance` check is enough.

PR #1265, opened 2022-04-07
[musicxml import] Catch BarException from presence of "times" attribute on start repeats

gregchapman-dev:
Just don't set it.

### DIFF
--- a/music21/musicxml/xmlToM21.py
+++ b/music21/musicxml/xmlToM21.py
@@ -4701,9 +4701,13 @@ class MeasureParser(XMLParserBase):
         else:
             raise bar.BarException('cannot handle mx direction format:', mxDirection)
 
-        if mxRepeat.get('times') is not None:
-            # make into a number
-            r.times = int(mxRepeat.get('times'))
+        try:
+            if mxRepeat.get('times') is not None:
+                # make into a number
+                r.times = int(mxRepeat.get('times'))
+        except bar.BarException:
+            # ignore BarException, just let the set of r.times fail silently
+            pass
 
         if inputM21 is None:
             return r
--- a/music21/musicxml/xmlToM21.py
+++ b/music21/musicxml/xmlToM21.py
@@ -4669,6 +4669,7 @@ class MeasureParser(XMLParserBase):
 
         Test that a forward repeat with times doesn't raise an exception, and
         that the resulting Repeat doesn't have times set.
+
         >>> mxStartBarline = ET.fromstring('<barline><bar-style>light-heavy</bar-style>' +
         ...       '<repeat direction="forward" times="2"/></barline>')
         >>> rs = MP.xmlToRepeat(mxStartBarline)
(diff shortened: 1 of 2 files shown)


gregchapman-dev:
The reason I want this fix is because this raise happens during parse of musicxml, and the musicxml isn't so bad that crashing during parse is warranted.  Example attached.  Here's the offending bit of musicxml:
'''
		<measure number="50">
			<print new-system="yes"/>
			<attributes>
				<divisions>1</divisions>
				<staves>2</staves>
			</attributes>
			<barline location="left">
				<bar-style>heavy-light</bar-style>
				<repeat direction="forward" times="2"/>
			</barline>
'''
[CRG001_MS.musicxml.zip](https://github.com/cuthbertLab/music21/files/8445774/CRG001_MS.musicxml.zip)

jacobtylerwalls:
Hi Greg. Did you consider just catching this failure in the musicxml importer? I agree that's a pretty minor detail not worth failing a whole file. (Finale loads the file without complaint.)

mscuthbert:
I prefer Jacob's way.  There's nothing more irritating that a property set that doesn't reflect:

```
>>> r.times = 5
>>> r.times
1
```

There's no exception on querying times on a start repeat (or there shouldn't be one) since reading a property should never raise an exception.

gregchapman-dev:
Agreed.  A raise during API usage makes sense, and the importer should just catch it silently.

gregchapman-dev:
This should be ready to merge.

music21/musicxml/xmlToM21.py
```diff
@@ -4666,6 +4666,14 @@ def xmlToRepeat(self, mxBarline, inputM21=None):
         'final'
         >>> r.direction
         'end'
+
+        Test that a forward repeat with times doesn't raise an exception, and
+        that the resulting Repeat doesn't have times set.
+        >>> mxStartBarline = ET.fromstring('<barline><bar-style>light-heavy</bar-style>' +
```
jacobtylerwalls, line 4672: A gotcha for future: if you build the documentation with `documentation/make.py` you'll notice this doesn't render as code without a blank line before the code block:
```suggestion

        >>> mxStartBarline = ET.fromstring('<barline><bar-style>light-heavy</bar-style>' +
```

music21/musicxml/xmlToM21.py
```diff
@@ -4701,9 +4709,13 @@ def xmlToRepeat(self, mxBarline, inputM21=None):
...
-        if mxRepeat.get('times') is not None:
-            # make into a number
-            r.times = int(mxRepeat.get('times'))
+        try:
+            if mxRepeat.get('times') is not None:
+                # make into a number
+                r.times = int(mxRepeat.get('times'))
+        except bar.BarException:
+            # ignore BarException, just let the set of r.times fail silently
+            pass
```
jacobtylerwalls, line 4718: I would just wrap the statement that can fail.
```suggestion
        if mxRepeat.get('times') is not None:
            try:
                # make into a number
                r.times = int(mxRepeat.get('times'))
            except bar.BarException:
                # ignore BarException, just let the set of r.times fail silently
                pass
```

jacobtylerwalls:
Thanks, Greg!

gregchapman-dev:
Thank you!

Issue #1214, opened 2022-02-08
romanText/tsvConverter.py update work in progress / coming soon


MarkGotham:
For information and to invite music21-related comments and suggestions:

The DCML standard for harmonic analysis has changed, so an update to the `music21/romanText/tsvConverter` is needed.

Here are the relevant links for discussion / work in progress:
- Discussion on When in Rome: https://github.com/MarkGotham/When-in-Rome/issues/32
- Draft implementation: https://github.com/malcolmsailor/music21/blob/master/music21/romanText/tsvConverter.py

mscuthbert:
I hope the new standard incorporates a version number. Older files will still be out there and need to be parsed in the old way.

MarkGotham:
Sure thing.

Probably the best approach is to set these numbers in relation the DCML version (previously 1.0; now 2.0)?

johentsch:
Thanks for the initative, @MarkGotham and @malcolmsailor! We'll be happy to provide support and assistance if required! This could be an important step for making the standards interoperable for everyone.

Yes, @mscuthbert, the DCML standard is [versioned](https://github.com/DCMLab/standards) and the employed version is indicated in the annotations' metadata (a promise yet to be filled for [ABC](https://github.com/DCMLab/ABC) and [MPS](https://github.com/DCMLab/mozart_piano_sonatas) in their soon-to-follow updates). Once ABC v2 will _finally_ be up and running in the near future, the "old way" will actually be deprecated. From that point, the latest regEx will be downward-compatible with all current and future `2.x.x` annotations.

malcolmsailor:
So would the best approach for the TSV converter be something like this then:

- allow user to specify DCML standard version
- if they don't, try to get it from the metadata
- if that fails, fall back to v2

malcolmsailor:
I implemented a flag allowing to switch between v1 and v2 parsing [here](https://github.com/malcolmsailor/music21/blob/master/music21/romanText/tsvConverter.py). Before I submit a pull request, one question:

I implemented a test where we convert to music21, then back to DCML, then back to music21 again, then compare the streams to ensure they are the same.

However, this test fails because vii (and possibly other roman numerals) seem to be treated differently going forward and going back: "#viio7" in a DCML file becomes "viio7" in music21, which is then written back out as "viio7" and then, when read in anew, becomes "bviio7". So I'm wondering if music21 provides any utilities that will allow us to write out "viio7" as "#viio7" and similar.

[...]

PR #1267, opened 2022-04-12
Adding DCML v2 parsing to tsvConverter.py

malcolmsailor:
See discussion at #1214 

There is an outstanding issue converting from music21 to the DCML format, as discussed in #1214. I do not believe that this pull request introduces this issue. Rather, I wrote a [test](https://github.com/malcolmsailor/music21/blob/46da6fbbd353378d5a4f78c5905fa38d29bdd063/music21/romanText/tsvConverter.py#L963) that uncovers the issue.


malcolmsailor:
Actually, I just noticed that the commented-out text in the linked code is out-of-date: the test no longer fails in quite the same way---I've fixed the issue where #vii became vii became bvii. But quality and inversion information seems not to propagate correctly, so 'viio7' becomes 'vii'. (This is all for v2; for v1, the test seems to fail more dramatically.)

malcolmsailor:
Whoops I missed pylint and flake in the contributing guidelines. I will address their objections.

music21/romanText/tsvConverter.py
```diff
@@ -155,18 +290,20 @@ def tabToM21(self):
...
+        >>> tabCd.representationType = 'm21'
         >>> tabCd.global_key = 'F'
         >>> tabCd.local_key = 'V'
-        >>> tabCd.representationType = 'm21'
         >>> m21Ch = tabCd.tabToM21()
 
         Now we can check it's a music21 RomanNumeral():
 
-        >>> m21Ch.figure
+        # >>> m21Ch.figure
```
mscuthbert, line 300: Commented out test is worse than no test.  Why does this not work and what is it instead?

music21/romanText/tsvConverter.py
```diff
@@ -155,18 +290,20 @@ def tabToM21(self):
...
         'vii'
         '''
 
-        if self.numeral:
+        if self.numeral == '@none':
+            thisEntry = harmony.NoChord()
+        else:
```
mscuthbert, line 306: what if self.numeral is None?  -- this seems to have taken away a possibility w/o comment

music21/romanText/tsvConverter.py
```diff
@@ -177,70 +314,19 @@ def tabToM21(self):
...
-            localKeyNonRoman = getLocalKey(self.local_key, self.global_key)
-
-            thisEntry = roman.RomanNumeral(combined, localKeyNonRoman)
-            thisEntry.quarterLength = self.length
-
-            thisEntry.op = self.op
-            thisEntry.no = self.no
-            thisEntry.mov = self.mov
-
+            if re.match(r'.*(i*v|v?i+).*', self.local_key, re.IGNORECASE):
```
mscuthbert, line 317: I can't follow.  needs explanation.

mscuthbert (changes requested):
I've gone through a bit of the PR -- there are a lot of changes that are not necessary for the issue.  Please limit changes to only what has to be changed to make the new version work.  I'm reviewing a part of the code base that I am not intimate with and lots of changes of spacing, etc. slow down the reviewing.  I stopped at line 200/300 of the tsvConverter.  Please contribute documentation and tests if you want this to be accepted.  Thank you!

malcolmsailor:
Regarding the formatting changes to parts of the code I did not otherwise edit, I believe what happened was that I have [black](https://github.com/psf/black) configured to run in my text editor on saving Python code and I neglected to turn that off before saving this file so it changed a great deal of the formatting without my noticing it. Sorry about that. I will try to restore the formatting and otherwise address your helpful comments. Thanks for your time!

malcolmsailor:
I made a new commit in response to your comments:

1. I restored the original formatting in the otherwise unchanged portions. Again, sorry about the inadvertent auto-format.
2. TabChord and TabChordV2 both inherit from a new abstract base class TabChordBase; attributes are defined explicitly rather than dynamically in the respective __init__ functions.
3. I tried to address your other comments.

I also made a few improvements. In particular, I addressed the bug discussed in #1214 by implementing writing to the `form` and `figbass` columns of tsv files, so that the [test](https://github.com/cuthbertLab/music21/blob/a1ad510356697f393bf6b636af8f45e81ad6ccc8/music21/romanText/tsvConverter.py#L1072) where we convert tsv -> m21 -> tsv -> m21 and then compare the m21 streams now works. (With one small issue concerning augmented sixth chords explained in a comment in the file.)

MarkGotham:
Hi all. Where are we with this? Ready to go? I'm happy to pitch in with dev. as necessary.

malcolmsailor:
> Hi all. Where are we with this? Ready to go? I'm happy to pitch in with dev. as necessary.

As far as I know it is ready to go; I did my best to address Michael's previous comments. If I missed anything or there are other outstanding issues I'm happy to address them too.

mscuthbert (changes requested):
Thanks for the contribution!  Many good things.  Be sure that you're familiar with music21 coding style before contributing a large quantity of code to the repo, as opposed to making an external tool.  Since I'll be taking responsibility for maintenance of this tool indefinitely, it needs to conform closely to existing style, docs, and testing.  Thanks!

music21/romanText/tsvConverter.py
```diff
@@ -166,81 +305,149 @@ def tabToM21(self):
...
-            thisEntry.op = self.op
-            thisEntry.no = self.no
-            thisEntry.mov = self.mov
+            if self.dcml_version == 1:
+                thisEntry.quarterLength = self.length
+                # following metadata attributes seem to be missing from
+                # dcml_version 2 tsv files
+                thisEntry.op = self.op
+                thisEntry.no = self.no
+                thisEntry.mov = self.mov
```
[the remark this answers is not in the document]
malcolmsailor: These lines (e.g., `thisEntry.op = self.op`) are in the existing codebase, not new with me. I can change them if you like though.

music21/romanText/tsvConverter.py
```diff
@@ -166,81 +305,149 @@ def tabToM21(self):
...
+        self.mn = None
+        self.mn_onset = None
+        self.timesig = None
+        self.globalkey = None
+        self.localkey = None
+        self.pedal = None
+        self.form = None
+        self.figbass = None
+        self.changes = None
+        self.phraseend = None
```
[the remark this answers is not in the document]
malcolmsailor: These attributes correspond to the headers of the columns of the DCML tsv files. The rationale for defining them separately in each subclass is that it makes it easier to compare the columns since they are all in one place. I can certainly change it if you prefer though.

music21/romanText/tsvConverter.py
```diff
@@ -166,81 +305,149 @@ def tabToM21(self):
...
+        'measure' has been removed from DCML v2 in favor of 'mn' and 'mc'. 'mn'
+        is equivalent to 'measure', so this property is provided as an alias.
+        '''
+        return int(self.mn)
 
-    >>> testTabChord1.numeral
-    'I'
-    '''
+    @property
+    def local_key(self):
```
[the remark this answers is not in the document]
malcolmsailor: By typing you mean type hints like (`def local_key(self) -> str:`)?

music21/romanText/tsvConverter.py
```diff
@@ -275,13 +482,25 @@ class TsvHandler:
...
+        self._extra_indices = None
+        self._tab_chord_cls = (TabChord, TabChordV2)[dcml_version - 1]
+        self.dcml_version = dcml_version
+        self.tsvData = self.importTsv()
+
+    def _get_heading_indices(self, header_row):
+        self._head_indices, self._extra_indices = {}, {}
+        for i, item in enumerate(header_row):
+            if item in self.heading_names:
+                self._head_indices[i] = item, self.heading_names[item]
```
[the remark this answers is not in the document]
malcolmsailor: the if condition checks if item is in heading_names and if not adds it to _extra_indices in the else clause.

music21/romanText/tsvConverter.py
```diff
@@ -291,14 +510,26 @@ def importTsv(self):
...
+
+    def _makeTabChord(self, row):
+        '''
+        Makes a TabChord out of a list imported from TSV data
+        (a row of the original tabular format -- see TsvHandler.importTsv()).
+        '''
+        # this method replaces the previously stand-alone makeTabChord function
+        thisEntry = self._tab_chord_cls()
+        for i, (name, type_) in self._head_indices.items():
+            setattr(thisEntry, name, type_(row[i]))
```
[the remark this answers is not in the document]
malcolmsailor: Here we are setting the attributes of TabChord or TabChordV2 (e.g., `measure`, `chord`, etc.) according to the values in the tsv row.
I use `setattr` here because TabChord and TabChordV2 have different attributes (corresponding to the different columns of the DCML tsv versions). The `name` values will only be those from the appropriate HEADERS because of lines 500-503, so we shouldn't  be setting any unanticipated attributes. We can add another layer of checks here though (e.g., `if name not in self.heading_names: continue`). Or I am open to suggestions for other ways of completing the same task.

music21/romanText/tsvConverter.py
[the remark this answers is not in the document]
malcolmsailor: @MarkGotham would you be willing to address this? It dates back to your v1 implementation.

music21/romanText/tsvConverter.py
```diff
@@ -166,81 +305,149 @@ def tabToM21(self):
...
-        if self.numeral:
-            if self.form:
-                if self.figbass:
-                    combined = ''.join([self.numeral, self.form, self.figbass])
-                else:
-                    combined = ''.join([self.numeral, self.form])
-            else:
-                combined = self.numeral
+        if self.numeral in ('@none', None):
+            thisEntry = harmony.NoChord()
```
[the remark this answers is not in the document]
malcolmsailor: In DCML v1, there is a "length" column from which we can infer quarterLength.
In DCML v2, there is no such column, so inferring quarterLength becomes nontrivial. If we need it for some reason, that would require some work---unless you know a fast way of calculating it.
I will add a check so that we add quarterLength in case DCML is v1.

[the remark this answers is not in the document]
mscuthbert: I think that you're determining the quarter length for normal RNs just below this -- I'm pretty sure that code just needs to be unindented.  If NoChord doesn't have a length assigned to it, then it will always last 1 QuarterNote which can't be intended.

malcolmsailor:
I did my best to address all your comments, Michael, and added a new PR.

> -1. Do not count on users being connected to the net to run tests. Create small, comprehensive test files that do not slow down testing.

Running on an actual ABC corpus file was giving me more confidence in the code, but I absolutely understand why that would be a bad idea. So today I did tsv -> music21 -> tsv -> music21 conversion on the entire ABC corpus, that comparing the music21 streams to make sure they were the same. There were quite a few harmonies that caused problems (mostly to do with vi in minor keys) so I added these to the test files `tsvEg_v2major.tsv` and `tsvEg_v2minor.tsv` as appropriate. Thanks to this I am confident that this PR makes the conversion quite a bit more accurate.

music21/romanText/tsvConverter.py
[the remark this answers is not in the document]
mscuthbert: Ah -- I see, you're getting the offsets properly but not the lengths.  What you might do then is in dcml v2 set all the quarterLengths to 0 -- that's better than showing them as quarter notes.  

Then when you're done with all parsing but before running makeMeasures, run Stream.extendDurations(harmony.Harmony, inPlace=True) if you'd like.  Or if measures are already made, it'll be a bit harder, but still possible.

mscuthbert (commented):
Some comments and notices.  Good work-- wanted to get the t.List etc. in for typing before you go too far.

music21/romanText/tsvConverter.py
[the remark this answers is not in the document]
malcolmsailor: The v1 TabChord has a length attribute that is taken from the "length" column in the DCML v1 files. But in the v2 files there is no such column, so the v2 TabChord doesn't have any such attribute. But when we put them into music21 streams, they do have measureNumber and beat attributes. Is there any way of populating the quarterLength attribute of the items of a stream from their measureNumber and beat attributes?

malcolmsailor:
> when you're done with all parsing but before running makeMeasures, run Stream.extendDurations(harmony.Harmony, inPlace=True) if you'd like. Or if measures are already made, it'll be a bit harder, but still possible.

The code doesn't call makeMeasures at all. The relevant methods (which I didn't write) seem to be TsvHandler.toM21Stream and TsvHandler.prepStream. The latter creates measures and inserts them into the stream, then the former populates these with the chords. I tried calling extendDuration on the whole stream but that didn't seem to work. I also tried

```
for m in self.m21stream.recurse().getElementsByClass(stream.Measure):
    m.extendDuration(harmony.Harmony, inPlace=True)
```

That didn't seem to work either. Any pointers would be appreciated!

music21/romanText/tsvConverter.py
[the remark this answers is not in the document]
johentsch: Currently, a column with lengths in quarters can be added to the TSVs optionally but I'm planning to make it a default in the upcoming weeks. If it helps the conversion, I can upgrade ABC v2 to that format?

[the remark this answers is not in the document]
malcolmsailor: Yes, that would be great!

music21/romanText/tsvConverter.py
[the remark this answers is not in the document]
mscuthbert: yes.  They're now mandatory in new additions to music21, and we've been spending a lot of time going back through existing code and adding them.  I've found that they reveal a lot of bugs that we didn't know were there.

music21/romanText/tsvConverter.py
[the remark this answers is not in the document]
mscuthbert: got it!

music21/romanText/tsvConverter.py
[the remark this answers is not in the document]
mscuthbert: If they have separate attributes, then you can call the same method name on the two different classes and they can work.  So move the logic to the TabChord/TabChordV2 itself.

mscuthbert:
> ```
> for m in self.m21stream.recurse().getElementsByClass(stream.Measure):
>     m.extendDuration(harmony.Harmony, inPlace=True)
> ```
> 
> That didn't seem to work either. Any pointers would be appreciated!

Try `self.m21stream.flatten().extendDuration(harmony.Harmony, inPlace=True)`

johentsch:
Using the current version @ [2f7db83](https://github.com/cuthbertLab/music21/pull/1267/commits/2f7db834091c25b660a04cb8076d7a7430126ac0), I wrote a small notebook to compare the chord tones expressed by the DCML labels with those of the m21 chords. Currently, there is divergence for all DCML labels containing chord tone replacement. 

Here a few examples from n01op18-1_01. Left DCML, right the converted `music21.roman.RomanNumeral` (dots stand for congruence):

```
V(64) <-> V
('C', 'F', 'A') != ('C', 'E', 'G')
.....
#viio7(6)/vi <-> #viio7/vi
('C#', 'E', 'A', 'Bb') != ('C##', 'E#', 'G#', 'B')

#viio7/vi <-> #viio7/vi
('C#', 'E', 'G', 'Bb') != ('C##', 'E#', 'G#', 'B')

#viio7(4)/ii <-> #viio7/ii
('F#', 'Bb', 'C', 'Eb') != ('F##', 'A#', 'C#', 'E')

#viio7/ii <-> #viio7/ii
('F#', 'A', 'C', 'Eb') != ('F##', 'A#', 'C#', 'E')

ii6(11#7b6) <-> ii6
('Bb', 'Eb', 'F#') != ('Bb', 'D', 'G')
```
Do the RomanNumeral objects provide a way to express chord tone replacement? If not, could it be an idea to create chord objects from the actual pitch class collections (left-hand) to maintain that information when creating the stream?

The handling of vi and vii in minor contexts is currently not working correctly by default (I know there must be a setting somewhere). Since the DCML labels express scale degrees of the natural minor scale it would probably make sense to fix the behaviour to 'flat' for both 6 and 7.

malcolmsailor:
Thanks, @johentsch! Would you be willing to share this notebook with me? That would be helpful!

music21/romanText/tsvConverter.py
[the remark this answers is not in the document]
malcolmsailor: So if I understand correctly, within these methods, you would like us to assign explicitly to each attribute, e.g.,
```
if "chord" in head_indices:
    i, type_to_coerce_to = head_indices["chord"]
    self.chord = type_to_coerce_to(row[i])
```
Repeated for each column name.

I can do this, but it is a lot of typing and will need to be altered any time we add or subtract columns. What about something like
```
for col_name, (i, type_to_coerce_to) in head_indices.items():
    if col_name in self.headers:
        setattr(self, col_name, type_to_coerce_to(row[i]))
```

MarkGotham:
Thanks all -- looking very promising!

Some quick notes, organised mostly around @johentsch 's comment/examples.

1. ^6 and ^7 in minor: As you suggest, many of the issues here relate to the notorious case of sixth and seventh degrees in minor. For that, `music21.roman.RomanNumeral` now has 4 options [documented here](https://web.mit.edu/music21/doc/moduleReference/moduleRoman.html?highlight=roman#music21.roman.Minor67Default). M21 default is `Minor67Default.QUALITY`. As @johentsch says, DCML use natural minor so this needs setting with m21's `Minor67Default.FLAT`. (Aside: These options were implemented in music21 after v1 of this converter -- hence the slightly hacky workarounds I wrote at the time).
2. Replacement: possible although a bit cumbersome. E.g., suspended 4th is `[add4][no3]`. We could consider a shorthand convention in romantext for this (e.g., `[sus4]`), but it's not currently there. @johentsch it looks like the general case maps `(N)` to `[add<N+1>][no<N>]`. Is that right?
3. `V(64)`: @johentsch do you distinguish cadential from passing at all? If not, it's probably wisest simply to map them all to `Cad64` then we runs checks for the rare cases in which its not cadential but a passing or otherwise an uncontroversial case of `V64` specifically (e.g., I-V64-I6). Unless someone sees a better approach, please implement this -- I hereby volunteer to handle the clean up.
4. I'll do something with the v1 metadata if necessary, though I think it better if we simply retire that and any other aspect of v1 that are both problematic and no longer relevant. 
6. Explicit quarter length column: Thanks @johentsch, yes please, +1!

Finally, checking whether the fantastic @napulen has eyes on this and anything to add. ;)

music21/romanText/tsvConverter.py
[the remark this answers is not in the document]
MarkGotham: Thanks. See below.

malcolmsailor:
I'm now working with the notebook @johentsch sent me. I believe I fixed the vii/vi stuff and I'm skipping the chords with chord-tone replacement for now.

I've found another category of error however involving augmented 6th chords. For example, on line 137 of n01op18-1_01.tsv, there is a chord whose label is "Ger6/vi". However, the "numeral" column is "vii", "form" is o, and "figbass" is 65. The music21 code uses these columns to infer the chord type. (I guess that DCML is indicating the augmented 6th through the "changes" column, which has b3. Thus a German sixth chord in C minor is like F-sharp diminished with a flat third.)

I believe the simplest thing to do may be to stop using the numeral, form, and figbass columns and instead parse these from the "label" directly.

malcolmsailor:
I pushed a new commit that does the following:
- adds type hints throughout
- fixes vi/vii handling using roman.Minor67Default.FLAT (I had to delete and/or adjust some of Mark's old tests as a result)
- populates quarterLength for v2 as well and adds a test that these add up to the correct value
- addresses various other comments from Michael and misc things
I also merged in current changes from the music21 repo since I believe that is why the automated tests failed on the last commit.

The main remaining issues seem to be:
1. addressing the chord tone replacement issue mentioned by Johannes
2. fixing conversion of certain other rare chords (besides the chords from 1., there are on the order of 10 chords in the ABC dataset that these seem to affect). I have been in touch by email with Johannes and Mark respectively concerning these.

malcolmsailor:
I should add that the commit also fixes the augmented sixth issue I mentioned in my previous comment, using the approach I suggested there. (It is possible that this issue remains in v1, however.)

johentsch:
> * Replacement: possible although a bit cumbersome. E.g., suspended 4th is `[add4][no3]`. We could consider a shorthand convention in romantext for this (e.g., `[sus4]`), but it's not currently there. @johentsch it looks like the general case maps `(N)` to `[add<N+1>][no<N>]`. Is that right?

This is right only for `2`, `4`, `6`, `9`, `11` etc. and for figures starting with `b` or `v` (~ downwards arrow). The exact details, including octave doubling, are explained [here in the tutorial](https://dcmlab.github.io/standards/build/html/tutorial/features.html#advanced-use-of-chord-tone-replacement).

> * `V(64)`: @johentsch do you distinguish cadential from passing at all? If not, it's probably wisest simply to map them all to `Cad64` then we runs checks for the rare cases in which its not cadential but a passing or otherwise an uncontroversial case of `V64` specifically (e.g., I-V64-I6). Unless someone sees a better approach, please implement this -- I hereby volunteer to handle the clean up.

Yes, the cadential one is always written as `V(64)`, but the passing chord as `I64`. E.g. `IV6 I64 ii65 V(64) V7 I`. 

> * Explicit quarter length column: Thanks @johentsch, yes please, +1!

Quarterbeats now available on the [v2_preview](https://github.com/DCMLab/ABC/tree/v2_preview) branch. Note that first endings receive `NA` values.

music21/romanText/tsvConverter.py
```diff
@@ -227,15 +228,9 @@ def _changeRepresentation(self):
...
-
         >>> tabCd._changeRepresentation()
         >>> tabCd.representationType
         'm21'
-
-        >>> tabCd.numeral
-        'vii'
         '''
 
         if self.representationType == 'm21':
```
johentsch, line 236: I noticed the style guide says that `enum` is to be preferred to string identifiers. Don't know how important this is to the main developers...

music21/romanText/tsvConverter.py
```diff
@@ -336,10 +335,14 @@ def tabToM21(self):
                 # otherwise, we assume self.local_key is already a pitch and
                 # pass it through unchanged
                 localKeyNonRoman = self.local_key
-            thisEntry = roman.RomanNumeral(combined, localKeyNonRoman)
+            thisEntry = roman.RomanNumeral(
+                combined,
+                localKeyNonRoman,
```
johentsch, line 340: Don't know how `getLocalKey()` operates but could it be that a `roman.Minor67Default` value could be needed here, as well, since the `.FLAT` behaviour also applies to DCML localkeys (if the globalkey is minor)?

malcolmsailor: Yes, that is now implemented---see lines 953--954. Thanks for your keen eye though :)

music21/romanText/tsvConverter.py
malcolmsailor: That could be changed, but that code is not introduced by this PR.

malcolmsailor:
Sorry I inadvertently pushed some incomplete code this morning. I'll patch it up and submit a new commit shortly.

malcolmsailor:
Added a new commit. I believe (hope) we are almost at the finish line now!

1. Population of TabChord attributes from tsv rows is now moved into the TabChord class per Michael's comments.
2. Handle conversion of dominant 7ths other than V7 (e.g., "IV7") in DCML -> m21
3. Added `handleAddedTones` to handle added tones (e.g., V(64), ii6(11#7b6)). There may still be a few rough edges there.

MarkGotham:
> > * Replacement: possible although a bit cumbersome. E.g., suspended 4th is `[add4][no3]`. We could consider a shorthand convention in romantext for this (e.g., `[sus4]`), but it's not currently there. @johentsch it looks like the general case maps `(N)` to `[add<N+1>][no<N>]`. Is that right?
> 
> This is right only for `2`, `4`, `6`, `9`, `11` etc. and for figures starting with `b` or `v` (~ downwards arrow). The exact details, including octave doubling, are explained [here in the tutorial](https://dcmlab.github.io/standards/build/html/tutorial/features.html#advanced-use-of-chord-tone-replacement).

Thanks! And also _mea cupla_ -- it seems that `V(4)` and `V(6)` work in roman text exactly as hypothetically discussed above. No need for `[addX][noY]`. Not sure why the 5th is cut out of all these cases, though that's a discussion for elsewhere.

> > * `V(64)`: @johentsch do you distinguish cadential from passing at all? If not, it's probably wisest simply to map them all to `Cad64` then we runs checks for the rare cases in which its not cadential but a passing or otherwise an uncontroversial case of `V64` specifically (e.g., I-V64-I6). Unless someone sees a better approach, please implement this -- I hereby volunteer to handle the clean up.
> 
> Yes, the cadential one is always written as `V(64)`, but the passing chord as `I64`. E.g. `IV6 I64 ii65 V(64) V7 I`.

That's great, thanks. So @malcolmsailor, definitely map `V(64)` to `Cad64`.

> > * Explicit quarter length column: Thanks @johentsch, yes please, +1!
> 
> Quarterbeats now available on the [v2_preview](https://github.com/DCMLab/ABC/tree/v2_preview) branch. Note that first endings receive `NA` values.

Thanks!

malcolmsailor:
> That's great, thanks. So @malcolmsailor, definitely map V(64) to Cad64.

All added tones handling is already implemented, including this. See my comment immediately above yours.

malcolmsailor:
> > > * Replacement: possible although a bit cumbersome. E.g., suspended 4th is `[add4][no3]`. We could consider a shorthand convention in romantext for this (e.g., `[sus4]`), but it's not currently there. @johentsch it looks like the general case maps `(N)` to `[add<N+1>][no<N>]`. Is that right?
> > 
> > 
> > This is right only for `2`, `4`, `6`, `9`, `11` etc. and for figures starting with `b` or `v` (~ downwards arrow). The exact details, including octave doubling, are explained [here in the tutorial](https://dcmlab.github.io/standards/build/html/tutorial/features.html#advanced-use-of-chord-tone-replacement).
> 
> Thanks! And also _mea cupla_ -- it seems that `V(4)` and `V(6)` work in roman text exactly as hypothetically discussed above. No need for `[addX][noY]`. Not sure why the 5th is cut out of all these cases, though that's a discussion for elsewhere.

`[addX][noY]` is already implemented now. Unless there's a compelling reason to change it I'm inclined to leave it. You're welcome to change it though.

MarkGotham:
Great to be approaching the finishing line! Interweaving here the discussion originating from https://github.com/MarkGotham/When-in-Rome/pull/42 ... it looks like:
- Phrase breaks are missing - can we reinstate?
- `Movement: N` line removed - can we reinstate?
- `#viio` should really be `viio` (even if `CAUTIONARY` handles them the same)
- Fast / slow 3-time. It looks like v1 defaulted to slow and v2 defaults to fast. See any 3-time movements, e.g., `/Quartets/Beethoven,_Ludwig_van/Op018_No1/2/analysis.txt`. 
- Actually beat assignments seem to be off for everything outside 4/4: e.g., see `Op018_No1/4/` in 2/4 (b.2 becomes 1.25), likewise in 2/2 and other oddities like b.1.67 > 1.17. Looks like some tests further are needed for the beat mapping.

malcolmsailor:
>

malcolmsailor:
[Sorry accidentally clicked on close just now]

> * Phrase breaks are missing - can we reinstate?

Sure, can you do that?

> * `Movement: N` line removed - can we reinstate?

I'm not sure I understand, but I think this is maybe an issue with the exported analyses in the When-in-Rome repository? So it just requires adding a line to the code there? (And isn't a music21 issue.)

> * `#viio` should really be `viio` (even if `CAUTIONARY` handles them the same)

Can you point to where this is a problem?

> * Actually beat assignments seem to be off for everything outside 4/4: e.g., see `Op018_No1/4/` in 2/4 (b.2 becomes 1.25), likewise in 2/2 and other oddities like b.1.67 > 1.17. Looks like some tests further are needed for the beat mapping.

Thanks for noticing this, I will have a look.

napulen:
Hello all, wonderful work. I see that this is all related to the `tsv` files. I have mostly worked with the DCML `mscx` directly. Something I have noticed, maybe irrelevant here (but I'll note it just in case), is that translating one of these `mscx` files into `mxl` results in annotations that are parsable by `music21` as `NoHarmony` objects (I think). This is great because the translation of the syntax can also be done on the `mxl` files directly. However, in those `mscx` files, the exact location (offset) of the annotation is not written into the `mxl`. Instead, they are attached to the closest note/rest in the staff. If there is, say, a whole note in a measure with two annotations (on `b1` and `b3`), those two annotations would be collapsed as part of `b1` (the location of the whole note). If somehow that was addressed, I think the usability of the parsers would just double because one could use the `mxl` files and/or `tsv`s.

Kind regards to all.

malcolmsailor:
>> Actually beat assignments seem to be off for everything outside 4/4: e.g., see Op018_No1/4/ in 2/4 (b.2 becomes 1.25), likewise in 2/2 and other oddities like b.1.67 > 1.17. Looks like some tests further are needed for the beat mapping.

> Thanks for noticing this, I will have a look.

OK should be fixed now. The problem was in all time signatures include 4/4, it's simply that in v2 mn_onset is notated in fractions of a whole note rather than quarter notes.

malcolmsailor:
I just noticed that when writing v2 to romantext with `converter.subConverters.ConverterRomanText().write`, there don't seem to be any time signatures in the output. I'm unsure of why this is; the time signatures are inserted in the stream in `TsvHandler.prepStream` and I don't believe the relevant code in that method has changed between v1 and v2. Moreover, I've inspected the streams and they contain time signatures. E.g.,

>>> stream[meter.TimeSignature].first()
<music21.meter.TimeSignature 3/4>

But after

>>> converter.subConverters.ConverterRomanText().write(stream, "romanText")
PosixPath('/var/folders/9j/7wpbgd7x53l4p8fy_9bf8sbm0000gn/T/music21/tmp6tii72qc.rntxt')

The file begins

```
[metadata snipped]

m1 F: I
m2 V
m3 I
m4 IV6
```

The time signature's offset seems to be 0:

```
>>> stream[meter.TimeSignature].first().activeSite.offset
0.0
>>> stream[meter.TimeSignature].first().offset
0.0
```

I'll investigate further but if anyone else has any suggestions I'd appreciate hearing them.

malcolmsailor:
I worked out the problem: in `writeRoman.py`, `RnWriter.prepSequentialListOfLines` only recognizes time signatures that are inside measures, and the first time signature was being inserted at the beginning of the part (and not in the first measure). This seems likely to be an oversight in `RnWriter` to me but I went ahead and changed the TsvHandler to insert the time signature (and key signature too, for good measure) in the first measure of the part.

mscuthbert:
That's generally where they're expected in music21.  Though now it's pretty easy to patch RNWriter to use the getTimeSignatures() function to get all of them regardless of where they are.

music21/romanText/tsvConverter.py
```diff
@@ -186,8 +187,15 @@ def __init__(self):
         self.changes = None
         self.phraseend = None
 
+        # the following attributes are overwritten by properties in TabChordV2
+        # because of changed column names in DCML v2
+        self.local_key = None
+        self.global_key = None
+        self.beat = None
+        self.measure = None
```
mscuthbert, line 195: can these be typed?  Basically anything that's being changed should be typed.

music21/romanText/tsvConverter.py
```diff
@@ -227,15 +235,9 @@ def _changeRepresentation(self):
...
 
-        >>> tabCd.numeral
-        '#vii'
-
         >>> tabCd._changeRepresentation()
         >>> tabCd.representationType
         'm21'
-
-        >>> tabCd.numeral
-        'vii'
```
mscuthbert, line 95: without these tests, it's hard to see what representationType does -- is there a demonstration that shows what DCML vs m21 type does?

music21/romanText/tsvConverter.py
```diff
@@ -312,20 +323,22 @@ def tabToM21(self):
...
+                combined = self.chord
+            else:
+                # previously this code only included figbass in combined if form
+                # was not falsy, which seems incorrect
+                combined = ''.join(
+                    [attr for attr in (self.numeral, self.form, self.figbass) if attr]
+                )
+
+                if self.relativeroot:  # special case requiring '/'.
+                    combined = ''.join([combined, '/', self.relativeroot])
```
mscuthbert, line 341: this can be `combined += '/' + self.relativeroot` and be cleaner

music21/romanText/tsvConverter.py
```diff
@@ -312,20 +323,22 @@ def tabToM21(self):
...
-
-            if self.relativeroot:  # special case requiring '/'.
-                combined = ''.join([combined, '/', self.relativeroot])
+            if self.dcml_version == 2 and self.chord:
+                combined = self.chord
+            else:
+                # previously this code only included figbass in combined if form
+                # was not falsy, which seems incorrect
+                combined = ''.join(
+                    [attr for attr in (self.numeral, self.form, self.figbass) if attr]
```
mscuthbert, line 337: no need for outside [] -- a generator expression can be passed to join.

music21/romanText/tsvConverter.py
```diff
@@ -336,22 +349,49 @@ def tabToM21(self):
...
-                thisEntry.no = self.no
-                thisEntry.mov = self.mov
-
-            thisEntry.pedal = self.pedal
-
-            thisEntry.phraseend = None
-
+                thisEntry.editorial.op = self.extra.get("op", "")
+                thisEntry.editorial.no = self.extra.get("no", "")
+                thisEntry.editorial.mov = self.extra.get("mov", "")
```
mscuthbert, line 364: music21 style is single quotes.

music21/romanText/tsvConverter.py
```diff
@@ -336,22 +349,49 @@ def tabToM21(self):
...
+        # if dcml_version == 2, we need to calculate the quarterLength
+        #   later
+        thisEntry.quarterLength = 0.0
         return thisEntry
 
+    def populate_from_row(
+        self,
+        row: t.List[str],
+        head_indices: t.Dict[str, t.Tuple[int, t.Type]],
+        extra_indices: t.Dict[int, str]
```
mscuthbert, line 377: music21 public attributes are in camelCase .  Same with methodNames.

music21/romanText/tsvConverter.py
```diff
@@ -166,81 +305,149 @@ def tabToM21(self):
...
-    ...                                        'C', 'I', '', 'I', '', '', '', '', 'false']
-    >>> testTabChord1 = romanText.tsvConverter.makeTabChord(tabRowAsString1)
+    @property
+    def beat(self):
+        '''
+        'beat' has been removed from DCML v2 in favor of 'mn_onset' and
+        'mc_onset'. 'mn_onset' is equivalent to 'beat', except that 'mn_onset'
+        is zero-indexed where 'beat' was 1-indexed. This property reproduces
+        the former 'beat' by adding 1 to 'mn_onset'.
+        >>> tabCd = romanText.tsvConverter.TabChordV2()
```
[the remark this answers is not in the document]
mscuthbert: Still needs space (a line break) between the text and the documentation.

music21/romanText/tsvConverter.py
```diff
@@ -388,30 +420,45 @@ def __init__(self):
...
-        is zero-indexed where 'beat' was 1-indexed. This property reproduces
-        the former 'beat' by adding 1 to 'mn_onset'.
+        is zero-indexed where 'beat' was 1-indexed, and 'mn_onset' is in
+        fractions of a whole-note rather than in quarter notes.
         >>> tabCd = romanText.tsvConverter.TabChordV2()
-        >>> tabCd.mn_onset = 0
+        >>> tabCd.mn_onset = 0.0
+        >>> tabCd.beat
+        1.0
+        >>> tabCd.mn_onset = 1/2
```
mscuthbert, line 433: write as 0.5 -- otherwise it will be mistaken as a Fraction class and loses the parallel with beat.

music21/romanText/tsvConverter.py
```diff
@@ -388,30 +420,45 @@ def __init__(self):
...
         '''
         return int(self.mn)
 
+    @measure.setter
+    def measure(self, measure: int):
+        self.mn = int(measure) if measure is not None else None
+
     @property
-    def local_key(self):
+    def local_key(self) -> str:
```
mscuthbert, line 461: these underscore cases are fine because they represent functionality with the same name in DCML v1.

music21/romanText/tsvConverter.py
```diff
@@ -516,24 +562,25 @@ def importTsv(self):
...
         # this method replaces the previously stand-alone makeTabChord function
         thisEntry = self._tab_chord_cls()
-        for i, (col_name, type_to_coerce_to) in self._head_indices.items():
-            # set attributes of thisEntry according to values in row
-            setattr(thisEntry, col_name, type_to_coerce_to(row[i]))
-        thisEntry.extra = {
-            col_name: row[i] for i, col_name in self._extra_indices.items() if row[i]
-        }
-        thisEntry.representationType = 'DCML'  # Added
+        thisEntry.populate_from_row(row, self._head_indices, self._extra_indices)
```
mscuthbert, line 572: populateFromRow

music21/romanText/tsvConverter.py
```diff
@@ -560,30 +607,43 @@ def toM21Stream(self):
...
-
-            if thisChord.representationType == 'DCML':
-                thisChord._changeRepresentation()
+            if m21Measure is None:
+                # TODO: m21Measure should never be None if prepStream is
+                #   correctly implemented. We need to handle None to satisfy
+                #   mypy. If it *is* None, then there is a bug in the
+                #   implementation. What is correct behavior in this instance?
+                #   Raise a bug?
+                raise ValueError
```
mscuthbert, line 628: do not raise a class -- raise an instance with a message:  `raise ValueError('m21Measure should not be None')`  All code paths should be tested.

music21/romanText/tsvConverter.py
```diff
@@ -599,24 +659,24 @@ def prepStream(self):
             s.insert(0, metadata.Metadata())
 
             firstEntry = self.chordList[0]  # Any entry will do
-            s.metadata.opusNumber = firstEntry.op
-            s.metadata.number = firstEntry.no
-            s.metadata.movementNumber = firstEntry.mov
+            s.metadata.opusNumber = firstEntry.extra.get('op', '')
+            s.metadata.number = firstEntry.extra.get('no', '')
+            s.metadata.movementNumber = firstEntry.extra.get('mov', '')
```
mscuthbert, line 664: rewrite as checks:

```
if 'op' in firstEntry.extra:
    s.metadata.opusNumber = firstExtry.extra['op']
```

The reason is that there's a difference between an opusNumber of '' and no opusNumber defined.  The former will print to musicxml or other places that the opusNumber is ''; the latter will not give an opus number.

music21/romanText/tsvConverter.py
```diff
@@ -692,14 +751,17 @@ def m21ToTsv(self):
...
         tsvData = []
         # take the global_key from the first item
         global_key = next(
             self.m21Stream.recurse().getElementsByClass('RomanNumeral')
         ).key.tonicPitchNameWithCase
 
         for thisRN in self.m21Stream[roman.RomanNumeral]:
+            if thisRN is None:
+                # shouldn't occur, but to satisfy mypy
+                continue
```
mscuthbert, line 764: that is strange -- mypy should know that it will be a roman.RomanNumeral -- have you merged master from music21 recently?

music21/romanText/tsvConverter.py
```diff
@@ -717,18 +779,26 @@ def _m21ToTsv_v1(self):
...
-            thisEntry.no = self.m21Stream.metadata.number
-            thisEntry.mov = self.m21Stream.metadata.movementNumber
+            ts = thisRN.getContextByClass(meter.TimeSignature)
+            if ts is None:
+                thisEntry.timesig = ''
+            else:
+                thisEntry.timesig = ts.ratioString
+            thisEntry.extra["op"] = self.m21Stream.metadata.opusNumber
+            thisEntry.extra["no"] = self.m21Stream.metadata.number
+            thisEntry.extra["mov"] = self.m21Stream.metadata.movementNumber
```
mscuthbert, line 789: single quotes.  and note that these might be None -- so better to add ` or ''` afterwards.  mypy did not catch this because `extra` is not typed.

music21/romanText/tsvConverter.py
```diff
@@ -173,9 +175,8 @@ def __init__(self):
         self.numeral = None
         self.relativeroot = None
         self.representationType = None  # Added (not in DCML)
-        self.extra = None
+        self.extra = {}
```
mscuthbert, line 178: type as `t.Dict[str, str]` -- which would catch a bug below.

music21/romanText/tsvConverter.py
```diff
@@ -717,18 +779,26 @@ def _m21ToTsv_v1(self):
...
             thisEntry.global_key = global_key
             thisEntry.local_key = thisRN.key.tonicPitchNameWithCase
             thisEntry.pedal = None
             thisEntry.numeral = thisRN.romanNumeral
             thisEntry.form = getForm(thisRN)
             # Strip any leading non-digits from figbass (e.g., M43 -> 43)
-            thisEntry.figbass = re.match(r'^\D*(\d.*|)', thisRN.figuresWritten).group(1)
+            figbassm = re.match(r'^\D*(\d.*|)', thisRN.figuresWritten)
+            if figbassm is not None:
+                thisEntry.figbass = figbassm.group(1)
```
mscuthbert, line 799: write as figbassMatch or something like that.  It looks like it means small figured bass.

music21/romanText/tsvConverter.py
```diff
@@ -850,6 +928,66 @@ def getForm(rn: roman.RomanNumeral) -> str:
     return ''
 
 
+def handleAddedTones(dcml_chord: str) -> str:
```
mscuthbert, line 931: public attributes need to be CamelCased

music21/romanText/tsvConverter.py
```diff
@@ -850,6 +928,66 @@ def getForm(rn: roman.RomanNumeral) -> str:
...
+        return dcml_chord
+    primary = m.group('primary')
+    added_tones = m.group('added_tones')
+    secondary = m.group('secondary') if m.group('secondary') is not None else ''
+    figure = m.group('figure')
+    if primary == 'V' and added_tones == '64':
+        return 'Cad64' + secondary
+    added_tone_tuples: t.List[t.Tuple[str, str, str, str, str]] = list(
+        # after https://github.com/johentsch/ms3/blob/main/src/ms3/utils.py
+        re.findall(r"((\+|-)?(\^|v)?(#+|b+)?(1\d|\d))", added_tones)
```
mscuthbert, line 968: single quotes.  

What is the license on the ms3/utils.py file?  If not BSD, need permission to reuse -- it's probably within the realm of fair use, but I don't want any claims from the GPL community that would make it a problem.

music21/romanText/tsvConverter.py
```diff
@@ -850,6 +928,66 @@ def getForm(rn: roman.RomanNumeral) -> str:
...
+            continue
+        if added_or_removed != '+' and int(factor) < threshold:
+            if above_or_below == 'v' or alteration in ('b', ''):
+                increment = -1
+            else:
+                increment = 1
+            replaced_factor = str(int(factor) + increment)
+            omissions.append(f'[no{replaced_factor}]')
+        additions.append(f'[add{alteration}{factor}]')
+    return primary + "".join(omissions) + "".join(additions) + secondary
```
mscuthbert, line 988: quotes

music21/romanText/tsvConverter.py
```diff
@@ -929,33 +1053,10 @@ def characterSwaps(preString, minor=True, direction='m21-DCML'):
...
-                postString = preString[:position] + insert + preString[position:]
-        else:
-            postString = preString
+    return preString
 
-    return postString
 
-
-def getLocalKey(local_key, global_key, convertDCMLToM21=False):
+def getLocalKey(local_key: str, global_key: str, convertDCMLToM21: bool = False):
```
mscuthbert, line 1059: type return value, especially important since this does not return a key.Key object as implied.

music21/romanText/tsvConverter.py
```diff
@@ -1091,8 +1206,8 @@ def testTsvHandler(self):
                 # Ensure that both m21 streams are the same
                 self.assertEqual(len(stream1.recurse()), len(stream2.recurse()))
                 for i, (item1, item2) in enumerate(zip(
-                    stream1.recurse().getElementsByClass('RomanNumeral'),
-                    stream2.recurse().getElementsByClass('RomanNumeral')
+                    stream1.recurse().getElementsByClass(harmony.Harmony),
+                    stream2.recurse().getElementsByClass(harmony.Harmony)
```
mscuthbert, line 1210: better as simply: `stream1[harmony.Harmony]`

mscuthbert (changes requested):
Looking good.  still needs more lint/style-guide/typing .  Thanks!

mscuthbert:
Coverage is very good, but look at https://coveralls.io/builds/51509206/source?filename=music21%2FromanText%2FtsvConverter.py#L269 and see that there are test lines that are not being run, and the regexp for Mm7 isn't being tested.  Most of the other uncovered lines are pretty trivial.

music21/romanText/tsvConverter.py
malcolmsailor: It's GNU v3. The only thing taken from @johentsch's code is the regex, which I then altered and will in fact alter further since we are not using the group that captures the entire match. Perhaps simplest solution to your concerns is just to remove the attribution? Or clarify "regex modeled on URL"?

music21/romanText/tsvConverter.py
malcolmsailor: I removed them because after using `roman.Minor67Default.FLAT`, tabCd.numeral is now '#vii' in both cases, so it seemed trivial. I have replaced it with a more up-to-date test.

Also, I just realized that this means some lines in the docstring text above are now out of date so I revised them.

music21/romanText/tsvConverter.py
malcolmsailor: Yes, and I'm not getting that warning again after deleting these lines.

music21/romanText/tsvConverter.py
malcolmsailor: It turns out that typing these causes mypy to object because they are overridden by properties in TabChordV2.
E.g., `music21/romanText/tsvConverter.py:425: error: Signature of "beat" incompatible with supertype "TabChordBase"`

This seems to be a bug in mypy, which there is an open issue about [here](https://github.com/python/mypy/issues/4125). (I like [this comment](https://github.com/python/mypy/issues/4125#issuecomment-870105336) on why overriding attributes by properties might be a good idea.) 

These aren't really "new" attributes, I just moved them from TabChordV1 to TabChordBase per your request. Is it alright if we just leave them untyped for now?

music21/romanText/tsvConverter.py
malcolmsailor: I typed this as you request and mypy still doesn't object to the code below, not sure why.

malcolmsailor:
I'm sorry about all the double-quotes, it is pretty deeply ingrained habit to use them. I've mostly gotten the hang of going through and replacing them after making changes for this project but yesterday I forgot. Have you thought about using something like [https://github.com/zheller/flake8-quotes/](https://github.com/zheller/flake8-quotes/) for the sake of people like me?

> Coverage is very good, but look at https://coveralls.io/builds/51509206/source?filename=music21%2FromanText%2FtsvConverter.py#L269 and see that there are test lines that are not being run, and the regexp for Mm7 isn't being tested. Most of the other uncovered lines are pretty trivial.

The test lines that were not being run turn out to be obsolete, so I removed them. And I believe the revised doctest for _changeRepresentation should now cover the regexp for Mm7.

music21/romanText/tsvConverter.py
mscuthbert: yeah, I might do a big switch to our StrEnum for a lot of parts of the system before V8, so that can be done then.

music21/romanText/tsvConverter.py
mscuthbert: ah -- if they're blocked by a bug in mypy, then we can leave them untyped.  If we need another round of review, maybe put a comment in the source code about why they're untyped.  Otherwise I'll forget and just ask again (or try to do it myself) next time the code is touched.

malcolmsailor:
As you may observe the workflow tests seem to be failing on the following test:

```
FAIL: testPickleMidi (freezeThaw.Test)
  File "/home/runner/work/music21/music21/music21/freezeThaw.py", line 1226, in testPickleMidi
    self.assertEqual(d.parts[1].flatten().notes[20].volume._client.__class__.__name__,
AssertionError: 'ReferenceType' != 'weakref'
- ReferenceType
+ weakref
```

When i run multiprocessTest.py in my local environment I don't get any such error. And I just merged the latest master in this morning. So I'm not sure why the error is happening, or how to reproduce it for further debugging. It doesn't seem to have anything to do with this PR but of course I can't be sure. Any suggestions?

music21/romanText/tsvConverter.py
mscuthbert: With GNU I want to be extra cautious.  How about rewriting it with re.VERBOSE which would be a lot easier anyhow, and then sufficiently derived as not to have license issues:

```
re.findall(
    r'''(
          (\+|-)?   # front alterations, or whatever this is.
          (\^|v)?   # what is this?
          (#+|b+)?    # sharps or flats
          (1\d|\d)   # numbers 0-19, in practice, 1-13
    )''', added_tones, re.VERBOSE)
```

That would be REALLY helpful for reading later.
re.VERBOSE is the best way to make a complex
regular expression understandable and debuggable.

I should use this more.

mscuthbert:
> Have you thought about using something like https://github.com/zheller/flake8-quotes/ for the sake of people like me?

I did not know about this!  Adding it!  Thanks!

mscuthbert:
The error seems to be a change in Python 3.10 version naming for weakreferences.  Pushing a fix on another branch.

mscuthbert:
Apparently this happened https://github.com/python/cpython/issues/79512

mscuthbert:
Closing and reopening because of bugs in Github right now.

MarkGotham:
Hey @malcolmsailor, all. Picking up a few points here.

1. Ger6: M21 has a mapping method for getting aug 6ths from the literal roman numeral figures for which they stand as shorthand. So one approach would be to create the Roman numerals and then run `chord.isAugmentedSixth()`.
2. Write Roman: thanks for spotting that. I'm happy to deal with it.
3. Single/double quotes. Double does seem to be more standard. Is m21 firmly committed to single or is there any prospect of moving over? A switch would needs careful handling for cases like single within double (or vice versa) but otherwise doesn't affect functionality, so not a breaking change, right?

Please let me know if I've missed anything else – thanks!

mscuthbert:
> 3. Single/double quotes. Double does seem to be more standard. Is m21 firmly committed to single or is there any prospect of moving over? A switch would needs careful handling for cases like single within double (or vice versa) but otherwise doesn't affect functionality, so not a breaking change, right?

Music21 is firmly committed to single quotes and camelCase.  

[String Quotes](https://peps.python.org/pep-0008/#string-quotes)
```
In Python, single-quoted strings and double-quoted strings are the same. This PEP does not make a recommendation for this. Pick a rule and stick to it. When a string contains single or double quote characters, however, use the other one to avoid backslashes in the string. It improves readability.
```

Music21 does not follow PEP 257 because (a) I didn't know about it in 2005 when we started, and (b) the first major dev and I both find single quotes more easily counted to make sure there are three of them.

The nice part of starting a project is that you get to make the rules.  I don't like following others' unless they have a very good reason for them.  Music21 uses CamelCase because the first version of m21 was written in Perl where that was the standard.  See https://github.com/cuthbertLab/music21/blob/master/CONTRIBUTING.md

music21/romanText/tsvConverter.py
mscuthbert: This is an AWESOME Example!  Now I get it.

music21/romanText/tsvConverter.py
```diff
@@ -659,14 +660,18 @@ def prepStream(self) -> stream.Score:
...
-                + '_No' + firstEntry.extra.get('no', '')
-                + '_Mov' + firstEntry.extra.get('mov', '')
-            )
+            title = []
+            if 'op' in firstEntry.extra:
+                s.metadata.opusNumber = firstEntry.extra['op']
+                title.append('Op' + s.metadata.opusNumber)
+            if 'no' in firstEntry.extra:
+                s.metadata.number = firstEntry.extra['no']
+                title.append('No' + s.metadata.number)
```
mscuthbert, line 669: These optional title additions are very good.

music21/romanText/tsvConverter.py
```diff
@@ -898,7 +900,7 @@ def write(self, filePathAndName: str):
 def getForm(rn: roman.RomanNumeral) -> str:
     '''
     Takes a music21.roman.RomanNumeral object and returns the string indicating
-    "form" expected by the DCML standard.
+    'form' expected by the DCML standard.
```
mscuthbert, line 903: hehe, this was actually okay before -- I've updated CONTRIBUTING.md to explain that in docs we generally use American English, which is double quotes.  But no worries!

mscuthbert (commented):
Awesome -- I believe that everything I wanted except the GNU regex rewrite has been done.  I would put the rewrite in myself, except I don't know what two of the capture groups are doing:

```
re.findall(
    r'''(
          (\+|-)?   # front alterations, or whatever this is.
          (\^|v)?   # what is this?
          (#+|b+)?    # sharps or flats
          (1\d|\d)   # numbers 0-19, in practice, 1-13
    )''', added_tones, re.VERBOSE)
```

It seems like members of the DCML, TSV, Annotated Beethoven Corpus, are still weighing in with suggestions and potential improvements.  My thought, unless there are objections, it to merge this great contribution as soon as that fix is in, and then to continue the discussion in an issue and future pull requests.  We're over 150 conversation points on this PR as it is, and it's time to let people try this out with real data and see what we all can do with it.

(But if there are objections let them be heard).

johentsch:
> ...except I don't know what two of the capture groups are doing

A suggestion how the comments in the verbose regEx could look like:

```
re.findall(
    r'''(
          (\+|-)?   # added non-chord or removed chord tone
          (\^|v)?   # (up/down) inverting the default direction of chord tone replacement
          (#+|b+)?  # sharps or flats altering the local key's scale degree
          (1\d|\d)  # numbers 0-19, in practice, 1-14
    )''', added_tones, re.VERBOSE)
```

malcolmsailor:
> Awesome -- I believe that everything I wanted except the GNU regex rewrite has been done.

I've done this, but as I did I noticed there also a few edge cases where that function is not matching what DCML expects so I'm fixing that up. I'll have a commit shortly.

music21/romanText/tsvConverter.py
malcolmsailor: I just steamrolled through with find and replace!😅

MarkGotham:
Thanks @malcolmsailor , @johentsch , @mscuthbert.

+1 for merge, further testing, and new smaller issue-discussions as required.

malcolmsailor:
`handleAddedTones` now gives the same result as @johentsch's ms3 library parser in the vast majority of cases in the ABC corpus. When it doesn't, it is often due to #1369. In any case the remaining disparities are very rare and all concern highly unusual chords. I have a script (based on a notebook from Johannes) to compare the ms3 and music21 output that I can share with anyone who wants to continue working on that.

As far as I can tell, the flake8 and mypy failures are not due to code merged into this PR.

mscuthbert:
> As far as I can tell, the flake8 and mypy failures are not due to code introduced by me.

yeah.  I did a wrong branch push yesterday.  I've now introduced branch protection to avoid that happening again.  Will merge after required tests pass.

mscuthbert (approved):

mscuthbert:
Merged!  Congrats!

PR #1275, opened 2022-04-15
Deprecate `StreamIterator.__getattr__()`

### LINKED ISSUE #1137: Deprecate `StreamIterator.__getattr__()`
The following code silently fails (unless running in `-Wall`, which will merely say it's inefficient):

```python
p = stream.Part()
p[stream.Measure].insert(0, stuff)
```

The user forgot to access the iterator and isn't told this didn't work.

The existing code has warnings for attributes, but silently dropping arguments to bound methods seems bad.

I suggest in v.8 it's time to finally deprecate `StreamIterator.__getattr__()` and remove in v.9 (with flat and the rest).

jacobtylerwalls:
The upgrade path is to call `.stream()` before attribute access.

Resolves #1137

### DIFF
--- a/music21/stream/iterator.py
+++ b/music21/stream/iterator.py
@@ -211,6 +211,7 @@ class StreamIterator(prebase.ProtoM21Object):
         self.cleanup()
         raise StopIteration
 
+    @common.deprecated('v8', 'v9', 'Call .stream() before attribute access for efficiency.')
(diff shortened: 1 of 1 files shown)


mscuthbert:
agreed -- thanks.  v9 is a good time to remove the functionality -- maybe later -- but this version is definitely time to make it more explicit.

PR #1288, opened 2022-05-01
v8: makeRests(inPlace=True) returns None for scores

jacobtylerwalls:
Let `makeRests(inPlace=True)` return None for consistency with other methods (and with itself).


mscuthbert:
@jacobtylerwalls -- i've debated making inPlace just return the same thing as not inPlace in v8 because it'd help the typing a lot.  BUT I don't like what it does to the doctests.

PR #1299, opened 2022-05-09
Make cautionaryPitchClass=True work with chords

### LINKED ISSUE #1227: Make cautionaryPitchClass work inside a single chord
**Motivation**
`cautionaryPitchClass=True` could make the following example print an F-natural at the bottom.

```python
>>> c = chord.Chord('F4 F#5')
>>> c.show()
```
Current:
Screen Shot 2022-02-18 at 2 02 42 PM

**Feature summary**
I think this could be on by default, even in a diatonic context. Or we can just do this in keyless contexts.

**Proposed implementation**
Similar to #1190 , maybe handle in tandem

**Intent**

[x] I plan on implementing this myself
[ ] I am starting a discussion with the hope that community members will volunteer their time to create this. I understand that individuals work on features of interest to them and that this feature may never be implemented.

jacobtylerwalls:
Fixes #1227

### DIFF
--- a/music21/pitch.py
+++ b/music21/pitch.py
@@ -5039,6 +5039,14 @@ class Pitch(prebase.ProtoM21Object):
                       and cautionaryNotImmediateRepeat is False
                       and pPastInMeasure is False):
                     set_displayStatus(True)
+
+                # Avoid making early, incorrect assumptions if this pitch is part of a chord.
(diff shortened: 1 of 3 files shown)


mscuthbert (approved):
Looks great to me. Thanks!

PR #1300, opened 2022-05-13
import typing as t

mscuthbert:
Because we're importing so much from typing, it makes more sense to standardize importing the entire module as `t` and then using `t.Union` etc.  When 3.9 is the minimum version, this will be a lot less line noise.

Removing existing uses of `t` as a variable makes everything cleaner with better names.


mscuthbert:
@jacobtylerwalls -- sanity check me if this is the direction we want to go.

PR #1310, opened 2022-05-17
Docs improvements.  Fix Pylint RecursiveIt typing

mscuthbert:
also fix shadowing Typing


mscuthbert:
Did not succeed in fixing typing for both pycharm and mypy

Issue #1284, opened 2022-04-27
7.3.3 tag is a v8 commit


jacobtylerwalls:
Downloading 7.3.3 from pip/PyPi works as intended, but downloading the source of 7.3.3 from GitHub Releases will download 333f2a2, which is Py3.8+.

Maybe we can force-push the 7.3.3 tag to the correct commit? I'm suggestible.

mscuthbert:
weird ummm...not sure how that could have happened.  Probably need to do a 7.3.4 release to fix.  but I'm not sure if this might happen again.

mscuthbert:
oh... it might have been tagged from master -- it's my first non-master release I've done in a long time.

jacobtylerwalls:
I can probably get to it at some point. I'm hoping one can just force-push the tags. No use doing a new version, since it's already correct on pypi!

jacobtylerwalls:
On second thought, I agree with the entertaining section about this in `git tag --help` that calls my suggestion to force-push the tags "insane".

So, I guess, yes--maybe a 7.3.4 is worth doing just so that pypi is in sync with github releases and tags.

mscuthbert:
Ah well... I'll look to see if there is anything to back port to justify a 7.3.4.

Probably they are:
* scale.nextPitch() as an alternative to scale.next()
* un-deprecating tree/toStream methods chordify() etc.  they're likely to be retained now that I'm using them more.

jacobtylerwalls:
Closing as moot.

mscuthbert:
thanks!

PR #1314, opened 2022-05-21
More specific instructions for tagging commits

jacobtylerwalls:
I'm a big fan of checklists. So I just added a suggestion to boost confidence when doing tags in the future. (Ref #1284)


mscuthbert:
Thanks!

PR #1323, opened 2022-06-06
Fix a regression with chord accidentals

jacobtylerwalls:
Better solution than #1299. That PR caused a regression with this case:

```python
from music21 import chord
c = chord.Chord("D# F# A#")
c.show(app='Finale')
```
Screen Shot 2022-06-06 at 4 01 46 PM

***
A better solution is to actually distinguish "past" from "simultaneous" rather than mixing the two notions.

### DIFF
--- a/music21/pitch.py
+++ b/music21/pitch.py
@@ -4863,7 +4863,8 @@ class Pitch(prebase.ProtoM21Object):
             else:
                 return  # exit: nothing more to do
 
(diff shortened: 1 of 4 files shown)


mscuthbert:
Thamks for the great catch -- feel free to merge after the `test_pitch.py` failure is updated to use a keyword arg.

PR #1326, opened 2022-06-18
Fix shiftElements docs

jacobtylerwalls:
`startOffset` and `endOffset` in this method work like excluding filters -- so this should read that elements outside the bounds are _not_ shifted.


mscuthbert:
thanks!

PR #1331, opened 2022-06-29
[Musicxml export] Preserve whitespace in TextBox

jacobtylerwalls:
I found that `xml:space="preserve"` is the way to preserve the display of newlines in a `<credit-words>`.
music21 was splitting text content on newlines and creating distinct `<credit-word>` elements, but without changing the position attributes, they were superimposed:

**Before**
Screen Shot 2022-06-29 at 11 24 24 AM

**After**
Screen Shot 2022-06-29 at 11 24 31 AM


jacobtylerwalls:
It's the Amy Beach choral work in the corpus.

Issue #1335, opened 2022-07-02
MusicXML voice numbers should be unique within a multi-staff part (reuse of voice numbers confuses Dorico)


gregchapman-dev:
**music21 version**

`7.3`   

**Problem summary**

Dorico believes the notes are in the same voice, so they end up rendered as a cross-staff two-note chord.  Musescore doesn't mind.

Here's an example MusicXML file (produced by music21) that re-uses voice numbers within a multi-staff part.
```
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE score-partwise  PUBLIC "-//Recordare//DTD MusicXML 3.1 Partwise//EN" "http://www.musicxml.org/dtds/partwise.dtd">
<score-partwise version="3.1">
  <movement-title></movement-title>
  <identification>
    <creator type="composer"></creator>
    <encoding>
      <encoding-date>2022-06-18</encoding-date>
    </encoding>
  </identification>
  <defaults>
    <scaling>
      <millimeters>7</millimeters>
      <tenths>40</tenths>
    </scaling>
  </defaults>
  <part-list>
    <score-part id="P1">
      <part-name />
    </score-part>
  </part-list>
  <!--=========================== Part 1 ===========================-->
  <part id="P1">
    <!--========================= Measure 1 ==========================-->
    <measure number="1">
      <attributes>
        <divisions>1</divisions>
        <staves>2</staves>
        <clef number="1">
          <sign>G</sign>
          <line>2</line>
        </clef>
        <clef number="2">
          <sign>F</sign>
          <line>4</line>
        </clef>
      </attributes>
      <note>
        <pitch>
          <step>C</step>
          <octave>5</octave>
        </pitch>
        <duration>1</duration>
        <voice>1</voice>
        <type>quarter</type>
        <staff>1</staff>
      </note>
      <note>
        <pitch>
          <step>D</step>
          <octave>5</octave>
        </pitch>
        <duration>1</duration>
        <voice>1</voice>
        <type>quarter</type>
        <staff>1</staff>
      </note>
      <note>
        <pitch>
          <step>E</step>
          <octave>5</octave>
        </pitch>
        <duration>1</duration>
        <voice>1</voice>
        <type>quarter</type>
        <staff>1</staff>
      </note>
      <note>
        <pitch>
          <step>F</step>
          <octave>5</octave>
        </pitch>
        <duration>1</duration>
        <voice>1</voice>
        <type>quarter</type>
        <staff>1</staff>
      </note>
[...]

PR #1336, opened 2022-07-02
Fix issue #1335: voice numbers written in MusicXML must be unique…

gregchapman-dev:
…within the StaffGroup.


music21/musicxml/partStaffExporter.py
```diff
@@ -790,6 +792,23 @@ def moveMeasureContents(measure: Element, otherMeasure: Element, staffNumber: in
                             'stem', 'notehead', 'notehead-text', 'staff'
                         ]
                     )
+#                 if voice is not None:
```
mscuthbert, line 795: rem unused new comment

mscuthbert:
I think that this looks good. @jacobtylerwalls did a lot of work on the PartStaffExporter, so I'd like him to weigh in before merging.  Please remove newly added code that is commented out.  thanks!

music21/musicxml/partStaffExporter.py
```diff
@@ -773,13 +771,17 @@ def moveMeasureContents(measure: Element, otherMeasure: Element, staffNumber: in
...
                 voice = elem.find('voice')
+                # otherMeasure is already using voice numbers 1 through maxVoices.
+                # Adjust/set this note's voice number to avoid that range.
                 if voice is not None:
-                    if otherMeasureLackedVoice and voice.text:
-                        # otherMeasure assigned voice 1; Bump voice number here
-                        voice.text = str(int(voice.text) + 1)
+                    if voice.text:
+                        # increment current voice number by maxVoices
+                        voice.text = str(maxVoices + int(voice.text))
```
jacobtylerwalls, line 779: Won't this destroy intentional cross-staff beams?

jacobtylerwalls (changes requested):
Thanks for picking this up. Can you add a test showing we don't destroy intentional cross-staff beaming? If you parse in a musicxml document with cross-staff beams (voice 1 on both hands) then I think we still want the same voice number on export, not voices n and n+1. No?

I have a feeling you might reply that your module uses voice numbers in a different way than musicxml, and you'd prefer not to delete voices in measures that don't need display voices and prefer not to make your voice numbers distinct since you wish to signal analytical unity. But until we create a way to divorce "display" voices from "analytical" voices in #890, I don't see how. I probably need to close #890 and create a more approachable summary after the discussion on #915, but I just didn't get to it yet.
***
From #915:

> Cross-staff music: PartStaffExporter currently bumps all voice numbers in the subsequent part staff to ensure distinct voice numbers, i.e. to ensure the 99% use case of no cross staff. When we WANT cross-staff, PartStaffExporter needs to get that knowledge from somewhere. Something like, per measure, if two music21 voices share the same .id/.group and .retain = True, if they've not already been given the same .sequence, give them the same .sequence and renumber everything else around it. This is a bit like the implementation in flattenUnnecessaryVoices, above, but different insofar as it would need to be calculated per measure over all voices from all PartStaffs in a joinable group.

Nasty problem. See also https://github.com/w3c/mnx/issues/183.

music21/musicxml/partStaffExporter.py
gregchapman-dev: I think it will.  Damn.

I don't believe my Humdrum parser is assigning voice numbers (but I'll check more thoroughly).  I think that voices are currently being blindly assigned voice numbers 1..N (per PartStaff) _somewhere_, and that means that partStaffExporter (unfixed) always thinks there are cross-staff voices.  My fix makes it so partStaffExporter _never_ thinks there are cross-staff voices, which is also incorrect.  I need to find who is assigning these bad voice numbers and fix it there, so I can put partStaffExporter back the way it was.

gregchapman-dev: Musescore/Finale don't care about voice numbers, so maybe we never noticed this bug when it was introduced?  Dorico is very unhappy, though.

jacobtylerwalls: I think it's this:
https://github.com/cuthbertLab/music21/blob/2dccb5b34d181eb2038b37ea8d722fa1fe848b4a/music21/musicxml/m21ToXml.py#L3106

Your test case has memory locations for the id number (default), and that's nonsense in a musicxml document, so this is the part we could make part-staff-aware.

gregchapman-dev: I think you're right.  Not sure how to do that, though.  I could (thinking out loud) mark these ids as "arbitrary" somehow, so partStaffExporter knows it can overwrite them?

jacobtylerwalls: You have access to the PartExporter as self.parent, and from there the part object, and `PartExporter.groupsToJoin`, so you could walk the groups, find out if the current part is not-the-first, and then advance the voice numbers. Something like that. Then no changes (hopefully) in partStaffExporter, or even some code we could remove, possibly (haven't looked).

jacobtylerwalls: A model:
https://github.com/cuthbertLab/music21/blob/2dccb5b34d181eb2038b37ea8d722fa1fe848b4a/music21/musicxml/m21ToXml.py#L2728-L2729

gregchapman-dev: I'll give it a shot.  Learning, learning, learning... :-)

music21/musicxml/m21ToXml.py
```diff
@@ -3076,6 +3083,77 @@ def mainElementsParse(self):
...
+                partStaff: stream.Part,
+                measure: stream.Measure) -> int:
+            # find the partExporter for partStaff
+            partStaffExp: t.Optional[PartExporter] = None
+            for partx in scoreExp.partExporterList:
+                if partx.stream is partStaff:
+                    partStaffExp = partx
+                    break
+            if partStaffExp is None:
+                return 0
```
jacobtylerwalls, line 3108: Realistically this is not so bad, but we could do better than executing a PartStaff search in every single measure.

See the `mergeInstrumentStreamPartStaffAware` model: the search is done once at the part level. Maybe the PartExporter needs a `.joinableGroup` attribute to hold the `StaffGroup` it participates in.

music21/musicxml/m21ToXml.py
```diff
@@ -3076,6 +3083,77 @@ def mainElementsParse(self):
...
+            for partStaff in group:
+                if partStaff is partExp.stream:
+                    # skip our own PartStaff
+                    continue
+
+                otherMeasure: stream.Measure = partStaff.measure(
+                    myMeasure.measureNumberWithSuffix()
+                )
+                if otherMeasure is None:
+                    # if measures don't have numbers, we can't do this
```
jacobtylerwalls, line 3147: I think we can do better than this, since many music21 operations can leave measures unnumbered or poorly numbered, especially if you're doing anything algorithmic, or you're appending streams (see recent list posts). Use your IDE to search for "Measure 0" and you'll find plenty of documented examples of streams all having no measure numbers.

I had a go at an algorithm for matching measure numbers in `processSubsequentPartStaff`. I see two options: shoehorn all of your voice numbering changes into that after-process inside `processSubsequentPartStaff`, which is a little bit like your original approach (sorry) -- OR -- factor out my logic into a generator function kind of like `zip()` that will spit out tuples of measures, eg...
```python
>>> list(zip_measures(p))
[(m, m), (m, None), (m, None), (m, m), ...
```
Let me know if this makes sense. I'm just feeling that if we can't support the use case of no measure numbers, then we're trading the constraint of "hey Greg, make sure your humdrum parser creates voice numbers" for "hey Greg, make sure your humdrum parser creates measure numbers."

jacobtylerwalls (commented):
Thanks for pushing forward -- let me know what you think about my questions below.

mscuthbert:
I'm a little concerned that we're going further in the direction of manipulating XML that we have generated (which I didn't like in the initial PartStaffExporter proposal) and away from changing Python objects such as Voices etc. Can we do more preprocessing before XML is written (since this is already a deepcopy) to keep the complexity down?

gregchapman-dev:
OK, I'm back to looking at this "voice numbers" PR.

I will definitely add support for unnumbered measures, and I'll try to manipulate the voice.id rather than the generated XML.

gregchapman-dev:
I like this a lot better.  I scan the score to compute the voiceNumberOffset for each PartExporter, _then_ do the parse.  Since I'm not doing measure-by-measure offsets (I use the same offset in every Measure in a Part), the measure numbers don't matter at all.  And the partExp.voiceNumberOffset is used to offset the voice.id, avoiding further manipulation of generated XML.

music21/musicxml/m21ToXml.py
```diff
@@ -1735,6 +1736,66 @@ def parseFlatScore(self):
...
+
+            # part is in a staffGroup, so we need to find the maximum number
+            # of voices used in any measure in this part.  We also find the
+            # maximum "low enough to not be considered a memory location"
+            # voice id.  The maxVoiceNum for a part is the higher of
+            # maxNumVoices-1 and maxNonMemLocationVoiceId.
+            maxNumVoices = 0
+            maxNonMemLocationVoiceId = 0
+            for measure in partExp.stream[stream.Measure]:
+                voices = tuple(measure[stream.Voice])
```
jacobtylerwalls, line 1772:
```suggestion
                voices = measure[stream.Voice]
```

music21/musicxml/m21ToXml.py
```diff
@@ -1735,6 +1736,66 @@ def parseFlatScore(self):
...
+                maxNumVoices = max(maxNumVoices, len(voices))
+                for voice in voices:
+                    if (isinstance(voice.id, int)
+                            and voice.id < defaults.minIdNumberToConsiderMemoryLocation):
+                        maxNonMemLocationVoiceId = max(maxNonMemLocationVoiceId, voice.id)
+            partExp.maxVoiceNum = max(maxNumVoices - 1, maxNonMemLocationVoiceId)
```
jacobtylerwalls, line 1778: Could you explain why the "minus 1" comes in? If there are two voices, shouldn't the maxVoiceNum be 2? (A 1 gets added back, below, which kind of adds to the understanding barrier for me.)

jacobtylerwalls (commented):
Thanks, this is looking good. I agree with Myke that in 2022 if I were doing the partStaffExporter again I would want to do it all without post-processing any xml.

My main question is just why we are tracking the voice number offsets at the part level? I would have thought they need to be tracked at the measure level.

TREBLE
m1.           m2.          
v1, v2       --no-voices--      
BASS
m1.           m2.          
v1, v2.       v1, v2

Wouldn't this PR now unnecessarily turn the bass clef in measure 2 into voices 3 and 4 (instead of leaving it as 1 and 2?) This would be worth testing in some musicxml readers -- especially when the voice numbers end up above 4. And a unit test with multiple measures would be good.

gregchapman-dev:
> My main question is just why we are tracking the voice number offsets at the part level? I would have thought they need to be tracked at the measure level.

For simplicity, because I found I didn’t care so much about the case below (using higher voice numbers than necessary in some measures).  But since you do care, I will use your cool idea about zip_measures, and track voice number offsets at the measure level...

> Wouldn't this PR now unnecessarily turn the bass clef in measure 2 into voices 3 and 4 (instead of leaving it as 1 and 2?) This would be worth testing in some musicxml readers -- especially when the voice numbers end up above 4. And a unit test with multiple measures would be good.

I will add such a test.

music21/musicxml/m21ToXml.py
gregchapman-dev: Yeah, I was thinking in 0-based indices (as one does), and then converting to 1-based at the last minute, but you're right, if I keep voice numbers 1-based throughout (as they are in MusicXML) it's much clearer.

music21/musicxml/m21ToXml.py
gregchapman-dev: I'll see if this works; I was assuming that measure[stream.Voice] would return an iterator, so I would need to make an iterable of it in order to call len().

gregchapman-dev: Well, isn't that lovely!  `StreamIterator.__len__()` for the win.

jacobtylerwalls:
Great. Yeah it's not so much that I care at a philosophical level, just that of all the musicxml readers out there, some of them (MuseScore?) I would expect to do funny things when the voice numbers start getting high.

gregchapman-dev:
OK, ready for another review.

music21/musicxml/m21ToXml.py
```diff
@@ -1758,43 +1758,105 @@ def prePartProcess(self):
...
+                partExp.voiceNumberOffsetForMeasure[measure] = voiceNumberOffset
+
+    @staticmethod
+    def zip_measures(streams: t.List[stream.Stream]):
+        numStreams = len(streams)
+        if numStreams == 0:
+            return
+
+        def getSimultaneousMeasureIndices(
+                measures: t.List[t.Optional[stream.Measure]]) -> t.List[int]:
```
jacobtylerwalls, line 1810:
```suggestion
                measures: t.Sequence[t.Optional[stream.Measure]]) -> t.List[int]:
```

music21/musicxml/m21ToXml.py
```diff
@@ -2548,6 +2671,19 @@ def __init__(self,
...
         self.lastDivisions = None
 
+        # voiceNumberOffsetForMeasure is the offset that must be applied to any voice
+        # numbers automatically generated for the key measure in this part, in order
+        # to avoid voice numbers (either explicitly set, or automatically generated)
+        # in other parts in this staffGroup.  This is computed in a pre-pass over all
+        # the voices in all the measures in all the parts (in ScoreExporter).
+        self.voiceNumberOffsetForMeasure: t.Dict[stream.Measure, int] = {}
+
+        # The maximum voiceNum found in the voices in all the measures in this part.
```
jacobtylerwalls, line 2681: Isn't this no longer referring to "all the measures"?

music21/musicxml/m21ToXml.py
```diff
@@ -1735,6 +1736,128 @@ def parseFlatScore(self):
...
+                for i in range(0, partIndexInGroup):
+                    m = simultaneousMeasures[i]
+                    if m is None:
+                        continue
+                    pex = partExporterForPart[staffGroupStreamList[i]]
+                    voiceNumberOffset += pex.maxVoiceNumForMeasure[m]
+                partExp.voiceNumberOffsetForMeasure[measure] = voiceNumberOffset
+
+    @staticmethod
+    def zip_measures(streams: t.List[stream.Stream]):
```
jacobtylerwalls, line 1804: This is public, so it should be `zipMeasures`, and have a docstring.

jacobtylerwalls, line 1804:
```suggestion
    def zip_measures(streams: t.Iterable[stream.PartStaff]):
```

jacobtylerwalls (commented):
Thanks for the changes, this is looking great.

music21/musicxml/m21ToXml.py
```diff
@@ -1758,43 +1758,105 @@ def prePartProcess(self):
...
```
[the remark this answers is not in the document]
gregchapman-dev: Yes.

music21/musicxml/m21ToXml.py
gregchapman-dev: Yes.

gregchapman-dev: This doesn't work, because t.Iterable is not Sized, so you can't call len() on it.  (That's weird, why not?)

music21/musicxml/m21ToXml.py
gregchapman-dev: Yes.

music21/musicxml/m21ToXml.py
gregchapman-dev: Yes.

music21/musicxml/m21ToXml.py
jacobtylerwalls: Ah, so `Sequence`?

gregchapman-dev: 🤦🏻‍♂️ Yes.  Pushed again.

gregchapman-dev:
Wow, flake8 got a lot pickier throughout music21.  Hope that doesn't block me from getting this in.

jacobtylerwalls (approved):
Thanks, Greg!

jacobtylerwalls:
> Wow, flake8 got a lot pickier throughout music21. Hope that doesn't block me from getting this in.

See #1347, you could merge master now.

music21/musicxml/m21ToXml.py
```diff
@@ -1735,6 +1736,139 @@ def parseFlatScore(self):
         pp.parse()
         self.partExporterList.append(pp)
 
+    def prePartProcess(self):
```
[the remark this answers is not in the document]
mscuthbert: And then we can just renumber voices here and not make the PartExporter have to keep track of them.

mscuthbert (changes requested):
Loving the huge amounts of tests and attention to detail!

My comments are mostly on trying to simplify the logic and remove the zipMeasures and other helper functions that seem to largely duplicate functionality that's already core to music21, and make it so that it's much easier to see still what ScoreExporter does in the case where there aren't staffgroups, PartStaffs etc.  Very close!

gregchapman-dev:
I'm doing the following:

`for measureStack in OffsetIterator(staffGroupScore).getElementsByClass(stream.Measure):`

and I never get any measureStacks.  Does OffsetIterator not recurse?  The staffGroupScore I'm testing with contains 2 PartStaff objects, each of which has a Measure.

gregchapman-dev:
I believe this is complete, and ready for review now.

mscuthbert:
> I'm doing the following:
> 
> `for measureStack in OffsetIterator(staffGroupScore).getElementsByClass(stream.Measure):`
> 
> and I never get any measureStacks. Does OffsetIterator not recurse? The staffGroupScore I'm testing with contains 2 PartStaff objects, each of which has a Measure.

AH.  it might not.  SHOOT.

I'm working on new Iterator types right now, and I think I've gotten down how to make them work more simply.  So I should be able to add a recurse=True to OffsetIterator instead.  Okay, no requirement to implement it here.

mscuthbert (approved):

mscuthbert:
Everything looks great!  Congrats!

PR #1337, opened 2022-07-03
Fix for #547 : Initial proposal for ArpeggioMark and ArpeggioMarkSpanner classes.

### LINKED ISSUE #547: Support <arpeggiate/> for xmlToM21.py
Is this planned?
If I could make a pull request to support this, what should I do, to be exact: In musicXML arpeggio lives in the `<notations/>` tag, but `music21` puts extra elements in `articulations`. So for the pull request I should make an `Arpeggiate` class inside `articulations`?
There are working parsers in the file so I guess it's worth a shot. I just want to know your opinion before making it for extra cover :D

gregchapman-dev:
Fixes #547

### DIFF
--- a/music21/expressions.py
+++ b/music21/expressions.py
@@ -1465,7 +1465,7 @@ class ArpeggioMarkSpanner(spanner.Spanner):
     bracket instead of a squiggly line, used to indicate a non-arpeggiated multi-chord
     intervening in a sequence of arpeggiated ones).
 
-    >>> ams = expressions.ArpeggioMarkSpanner('non-arpeggio')
+    >>> ams = expressions.ArpeggioMarkSpanner(arpeggioType='non-arpeggio')
     >>> c1 = chord.Chord('C3 E3 G3')
     >>> c2 = chord.Chord('C4 E4 G4')
     >>> ams.addSpannedElements([c1, c2])
@@ -1475,8 +1475,9 @@ class ArpeggioMarkSpanner(spanner.Spanner):
     <music21.expressions.ArpeggioMarkSpanner
      <music21.chord.Chord C3 E3 G3><music21.chord.Chord C4 E4 G4>>
     '''
-    def __init__(self, arpeggioType: t.Optional[str] = None, *arguments, **keywords):
+    def __init__(self, *arguments, **keywords):
         super().__init__(*arguments, **keywords)
+        arpeggioType: t.Optional[str] = keywords.get('arpeggioType', None)
         if arpeggioType is None:
             arpeggioType = 'normal'
         if arpeggioType not in ('normal', 'up', 'down', 'non-arpeggio'):
--- a/music21/expressions.py
+++ b/music21/expressions.py
@@ -1457,8 +1457,12 @@ class ArpeggioMark(Expression):
 class ArpeggioMarkSpanner(spanner.Spanner):
     '''
     ArpeggioMarkSpanner is a multi-staff or multi-voice (i.e. multi-chord) arpeggio.
-    The spanner should contain all the simultaneous Chords/Notes that are to be
-    arpeggiated together.
+    The spanner should contain all the simultaneous Chords that are to be
+    arpeggiated together.  If there is only one arpeggiated note in a particular staff
+    or voice (i.e. the rest are in other staves/voices), then in that case only the
+    spanner can contain a Note.  Do not ever put a Note that is within a Chord into a
+    spanner; put the Chord in instead.  And do not ever put an ArpeggioMark in a note
+    or chord's .expressions.
 
     The parameter arpeggioType can be 'normal' (a squiggly line), 'up' (a squiggly line
     with an up arrow), 'down' (a squiggly line with a down arrow), or 'non-arpeggio' (a
--- a/music21/expressions.py
+++ b/music21/expressions.py
@@ -1490,6 +1490,29 @@ class ArpeggioMarkSpanner(spanner.Spanner):
             )
         self.type = arpeggioType
 
+    def noteExtremes(self) -> t.Tuple[t.Optional['music21.note.Note'],
+                                      t.Optional['music21.note.Note']]:
+        '''
+        Return the lowest and highest note spanned by the element,
+        extracting them from Chords if need be.
+
+        >>> ch = chord.Chord(['C4', 'E4', 'G4'])
+        >>> n = note.Note('C#3')
+        >>> nonArp = expressions.ArpeggioMarkSpanner([ch, n])
+        >>> nonArp.noteExtremes()
+        (<music21.note.Note C#>, <music21.note.Note G>)
+        '''
+        from music21 import chord
+        from music21 import note
+        notes = []
+        for n_or_ch in self:
+            if isinstance(n_or_ch, note.Note):
+                notes.append(n_or_ch)
+            elif isinstance(n_or_ch, chord.Chord):
+                notes.extend(n_or_ch.notes)
+        return (min(notes), max(notes))
+
+
 # ------------------------------------------------------------------------------
 class Test(unittest.TestCase):
 
--- a/music21/expressions.py
+++ b/music21/expressions.py
@@ -1426,10 +1426,6 @@ class TremoloSpanner(spanner.Spanner):
             raise TremoloException('Number of marks must be a number from 0 to 8') from ve
 
 
-class ArpeggioException(exceptions21.Music21Exception):
-    pass
-
-
 class ArpeggioMark(Expression):
     '''
     ArpeggioMark must be applied to a Chord (not to a single Note).
@@ -1447,10 +1443,12 @@ class ArpeggioMark(Expression):
     >>> am.type
     'down'
     '''
-    def __init__(self, arpeggioType: str = 'normal'):
+    def __init__(self, arpeggioType: t.Optional[str] = None):
         super().__init__()
+        if arpeggioType is None:
+            arpeggioType = 'normal'
         if arpeggioType not in ('normal', 'up', 'down', 'non-arpeggio'):
-            raise ArpeggioException(
+            raise ValueError(
                 'Arpeggio type must be "normal", "up", "down", or "non-arpeggio"'
             )
         self.type = arpeggioType
@@ -1459,8 +1457,8 @@ class ArpeggioMark(Expression):
 class ArpeggioMarkSpanner(spanner.Spanner):
     '''
     ArpeggioMarkSpanner is a multi-staff or multi-voice (i.e. multi-chord) arpeggio.
-    The spanner should contain all the simultaneous Chords that are to be arpeggiated
-    together.  It should not contain individual notes.
+    The spanner should contain all the simultaneous Chords/Notes that are to be
+    arpeggiated together.
 
     The parameter arpeggioType can be 'normal' (a squiggly line), 'up' (a squiggly line
     with an up arrow), 'down' (a squiggly line with a down arrow), or 'non-arpeggio' (a
@@ -1477,10 +1475,12 @@ class ArpeggioMarkSpanner(spanner.Spanner):
     <music21.expressions.ArpeggioMarkSpanner
      <music21.chord.Chord C3 E3 G3><music21.chord.Chord C4 E4 G4>>
     '''
-    def __init__(self, arpeggioType: str = 'normal', *arguments, **keywords):
+    def __init__(self, arpeggioType: t.Optional[str] = None, *arguments, **keywords):
         super().__init__(*arguments, **keywords)
+        if arpeggioType is None:
+            arpeggioType = 'normal'
         if arpeggioType not in ('normal', 'up', 'down', 'non-arpeggio'):
-            raise ArpeggioException(
+            raise ValueError(
                 'Arpeggio type must be "normal", "up", "down", or "non-arpeggio"'
             )
         self.type = arpeggioType
--- a/music21/expressions.py
+++ b/music21/expressions.py
@@ -1447,7 +1447,7 @@ class ArpeggioMark(Expression):
     >>> am.type
     'down'
     '''
-    def __init__(self, arpeggioType: str):
+    def __init__(self, arpeggioType: str = 'normal'):
         super().__init__()
         if arpeggioType not in ('normal', 'up', 'down', 'non-arpeggio'):
             raise ArpeggioException(
@@ -1477,7 +1477,7 @@ class ArpeggioMarkSpanner(spanner.Spanner):
     <music21.expressions.ArpeggioMarkSpanner
      <music21.chord.Chord C3 E3 G3><music21.chord.Chord C4 E4 G4>>
     '''
-    def __init__(self, arpeggioType: str, *arguments, **keywords):
+    def __init__(self, arpeggioType: str = 'normal', *arguments, **keywords):
         super().__init__(*arguments, **keywords)
         if arpeggioType not in ('normal', 'up', 'down', 'non-arpeggio'):
             raise ArpeggioException(
--- a/music21/expressions.py
+++ b/music21/expressions.py
@@ -1426,6 +1426,65 @@ class TremoloSpanner(spanner.Spanner):
             raise TremoloException('Number of marks must be a number from 0 to 8') from ve
 
 
+class ArpeggioException(exceptions21.Music21Exception):
+    pass
+
+
+class ArpeggioMark(Expression):
+    '''
+    ArpeggioMark must be applied to a Chord (not to a single Note).
+
+    The parameter arpeggioType can be 'normal' (a squiggly line), 'up' (a squiggly line
+    with an up arrow), 'down' (a squiggly line with a down arrow), or 'non-arpeggio' (a
+    bracket instead of a squiggly line, used to indicate a non-arpeggiated chord
+    intervening in a sequence of arpeggiated ones).
+
+    >>> am = expressions.ArpeggioMark('normal')
+    >>> am.type
+    'normal'
+
+    >>> am = expressions.ArpeggioMark('down')
+    >>> am.type
+    'down'
+    '''
+    def __init__(self, arpeggioType: str):
+        super().__init__()
+        if arpeggioType not in ('normal', 'up', 'down', 'non-arpeggio'):
+            raise ArpeggioException(
+                'Arpeggio type must be "normal", "up", "down", or "non-arpeggio"'
+            )
+        self.type = arpeggioType
+
+
+class ArpeggioMarkSpanner(spanner.Spanner):
+    '''
+    ArpeggioMarkSpanner is a multi-staff or multi-voice (i.e. multi-chord) arpeggio.
+    The spanner should contain all the simultaneous Chords that are to be arpeggiated
+    together.  It should not contain individual notes.
+
+    The parameter arpeggioType can be 'normal' (a squiggly line), 'up' (a squiggly line
+    with an up arrow), 'down' (a squiggly line with a down arrow), or 'non-arpeggio' (a
+    bracket instead of a squiggly line, used to indicate a non-arpeggiated multi-chord
+    intervening in a sequence of arpeggiated ones).
+
+    >>> ams = expressions.ArpeggioMarkSpanner('non-arpeggio')
+    >>> c1 = chord.Chord('C3 E3 G3')
+    >>> c2 = chord.Chord('C4 E4 G4')
+    >>> ams.addSpannedElements([c1, c2])
+    >>> ams.type
+    'non-arpeggio'
+    >>> ams
+    <music21.expressions.ArpeggioMarkSpanner
+     <music21.chord.Chord C3 E3 G3><music21.chord.Chord C4 E4 G4>>
+    '''
+    def __init__(self, arpeggioType: str, *arguments, **keywords):
+        super().__init__(*arguments, **keywords)
+        if arpeggioType not in ('normal', 'up', 'down', 'non-arpeggio'):
+            raise ArpeggioException(
+                'Arpeggio type must be "normal", "up", "down", or "non-arpeggio"'
+            )
+        self.type = arpeggioType
+
 # ------------------------------------------------------------------------------
 class Test(unittest.TestCase):
 
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -1440,7 +1440,7 @@ class ScoreExporter(XMLExporterBase, PartStaffExporterMixin):
 
         self.scoreMetadata = None
 
-        self.spannerBundle = None
+        self.spannerBundle: t.Optional[spanner.SpannerBundle] = None
         self.meterStream: t.Optional[stream.Stream[meter.TimeSignatureBase]] = None
         self.scoreLayouts = None
         self.firstScoreLayout = None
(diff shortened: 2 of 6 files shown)


gregchapman-dev:
MusicXML parser/writer are next (of course).

gregchapman-dev:
OK, support for ArpeggioMark (including m21ToXml.py and xmlToM21.py) is there.  I have made no attempt to support ArpeggioMarkSpanner for cross-staff arpeggios, since I have no MusicXML examples to parse.

Can I get this reviewed, please?  Maybe we can merge it as is (minus the declaration of ArpeggioMarkSpanner) and write a new issue for ArpeggioMarkSpanner?

music21/expressions.py
```diff
@@ -1426,6 +1426,65 @@ def numberOfMarks(self, num):
...
+    >>> ams.type
+    'non-arpeggio'
+    >>> ams
+    <music21.expressions.ArpeggioMarkSpanner
+     <music21.chord.Chord C3 E3 G3><music21.chord.Chord C4 E4 G4>>
+    '''
+    def __init__(self, arpeggioType: str = 'normal', *arguments, **keywords):
+        super().__init__(*arguments, **keywords)
+        if arpeggioType not in ('normal', 'up', 'down', 'non-arpeggio'):
+            raise ArpeggioException(
```
jacobtylerwalls, line 1483: Just a style thing, but I prefer standard Python exceptions for user errors, to allow defining and catching "Library failures" separately. So just ValueError would be fine here, and remove the custom exception. I know that's kind of a departure from the way music21 was written, but I've already coerced Myke into merging #995 :D

music21/expressions.py
```diff
@@ -1426,6 +1426,65 @@ def numberOfMarks(self, num):
...
+                'Arpeggio type must be "normal", "up", "down", or "non-arpeggio"'
+            )
+        self.type = arpeggioType
+
+
+class ArpeggioMarkSpanner(spanner.Spanner):
+    '''
+    ArpeggioMarkSpanner is a multi-staff or multi-voice (i.e. multi-chord) arpeggio.
+    The spanner should contain all the simultaneous Chords that are to be arpeggiated
+    together.  It should not contain individual notes.
```
jacobtylerwalls, line 1463: Does this limitation come from musicxml? A three-staff StaffGroup with a single note on each staff could still yield an arpeggiated trichord? Or dyad on two staves? 🤔 

It's worth just a bit of thought to make sure we can make music21 as liberal as possible even if musicxml is more conservative.

music21/expressions.py
```diff
@@ -1426,6 +1426,65 @@ def numberOfMarks(self, num):
...
+    'normal'
+
+    >>> am = expressions.ArpeggioMark('down')
+    >>> am.type
+    'down'
+    '''
+    def __init__(self, arpeggioType: str = 'normal'):
+        super().__init__()
+        if arpeggioType not in ('normal', 'up', 'down', 'non-arpeggio'):
+            raise ArpeggioException(
```
jacobtylerwalls, line 1453: ValueError

music21/musicxml/m21ToXml.py
```diff
@@ -4774,6 +4828,15 @@ def expressionToXml(self, expression):
             if k in classes:
                 mx = Element(v)
                 break
+        if mx is None:
+            # 'ArpeggioMark' maps to two different elements
+            if 'ArpeggioMark' in classes:
```
jacobtylerwalls, line 4833: See Myke's posts about v8 on the list, we're moving to more standard `isinstance()` instead of .classes

jacobtylerwalls (commented):
Forgive me for doing the trivial bits of the review first, but I have takeout to go pick up :-)

music21/musicxml/m21ToXml.py
```diff
@@ -4421,21 +4441,39 @@ def noteToNotations(self, n, noteIndexInChord=0, chordParent=None):
...
+                # no expression is applicable, or expressionToXml() didn't recognize it.
+                continue
+
+            if 'Ornament' in expObj.classes:
+                if mxOrnaments is None:
+                    mxOrnaments = Element('ornaments')
+                mxOrnaments.append(mxExpression)
+                # print(mxExpression)
+            else:
+                notations.append(mxExpression)
```
jacobtylerwalls, line 4476: Seems like this creates a behavior change: now exporting ornaments stored on notes that are not first in the chord. We should revert or discuss it separately if you have a use case (see #909)

music21/musicxml/test_m21ToXml.py
```diff
@@ -458,6 +458,63 @@ def testOutOfBoundsExpressionDoesNotCreateForward(self):
...
+            'arpeggiate',
+            'arpeggiate',
+            'arpeggiate',
+        )
+        s = converter.parse(testPrimitive.arpeggio32d)
+        x = self.getET(s)
+        # helpers.dump(x)
+        mxPart = x.find('part')
+        mxMeasure = mxPart.find('measure')
+        for i, mxNote in enumerate(mxMeasure.findall('note')):
```
jacobtylerwalls, line 490: Thanks for the test. I'm just a little worried about debugging it if it fails, say with musicxml 5.0 someday. Can you indent one more time under `with self.subTest(note_index=i):` so that your variable `i` is in the failure message if there's a failure?

music21/musicxml/test_xmlToM21.py
```diff
@@ -1134,6 +1134,37 @@ def testFretIndication(self):
...
 
+    def testArpeggioMarks(self):
+        from music21 import converter
+        from music21.musicxml import testPrimitive
+
+        s = converter.parse(testPrimitive.arpeggio32d)
+        p = s.parts[0]
+        m = p.measure(1)
+        for i, el in enumerate(m):
+            if 2 <= i <= 8:
```
jacobtylerwalls, line 1145: This is okay for now (unless Myke says otherwise), but since the tests are another form of documentation I wouldn't want to model too much depending on an element being at a certain index.

When we changed MIDI parsing to start creating time signatures or measures, people who were indexing into `myPart[0]` and expecting to find voices or notes were unexpectedly getting something else. There's a Coursera course that tells people to do it the bad way so about every four months there's a Stack Overflow post that goes "how come my Pat Metheny MIDI file crashes when I run this music21 code?"

For instance, if there were a clef and a key and a meter, the notes would probably start at index 3. And so if musicxml 5.0 starts implying 4/4 by default... you get the idea how this could break. But like I said, just something for future unless Myke feels strongly.

music21/musicxml/xmlToM21.py
```diff
@@ -2813,12 +2813,12 @@ def xmlToChord(self, mxNoteList: t.List[ET.Element]) -> chord.ChordBase:
                 sp.replaceSpannedElement(n, c)
             for art in n.articulations:
                 if type(art) in seenArticulations:  # pylint: disable=unidiomatic-typecheck
-                    pass
+                    continue
```
jacobtylerwalls, line 2816: Thank you! For Myke's benefit: this was identified as a TODO in https://github.com/cuthbertLab/music21/issues/909#issuecomment-796903557.

music21/musicxml/xmlToM21.py
```diff
@@ -3645,6 +3644,19 @@ def flatten(mx, name):
                 fermata.shape = mxObj.text.strip()
             n.expressions.append(fermata)
 
+        # get any arpeggios, store in expressions.
+        for mxObj in mxNotations.findall('arpeggiate'):
+            arpeggioType: str = mxObj.get('direction')
+            if not arpeggioType:
+                arpeggioType = 'normal'
```
jacobtylerwalls, line 3651: The constructor defaults to 'normal', so I'm not certain you need this check here. For consistency you could just use `textStripValid` like the surrounding cases. (I'm guessing that some musicxml writers at some point in time were adding extra whitespace, but I don't have direct experience with that.)

jacobtylerwalls (commented):
Gave it a more substantive look. Looking good! Noticed one potential behavior change.

music21/expressions.py
jacobtylerwalls: If there's implementation work to do, then we can totally punt that to a future PR. I'm just asking if there's any reason we have to reaffirm the musicxml limitation right out of the gate.

gregchapman-dev: I've removed this comment.  It's still true of ArpeggioMark, since an arpeggio on a single note is meaningless, but ArpeggioMarkSpanner could span a bunch of single notes easily.

music21/expressions.py
gregchapman-dev: Agreed.

music21/expressions.py
gregchapman-dev: Yep

music21/musicxml/m21ToXml.py
gregchapman-dev: Agreed.  I'll leave the older 'if k in classes' alone, though (for now).

music21/musicxml/m21ToXml.py
gregchapman-dev: Ooh!  Good eye.  I have a fix.

music21/musicxml/test_m21ToXml.py
gregchapman-dev: Looked around a bit to see how the heck _that_ works.  Got it.

music21/musicxml/test_xmlToM21.py
gregchapman-dev: Agreed.  I hope you like the new, improved test.

music21/musicxml/xmlToM21.py
gregchapman-dev: 👍🏼

music21/musicxml/xmlToM21.py
gregchapman-dev: I've just removed the check.  I don't use mxObj.text here, so I'm not sure how textStripValid will apply.

gregchapman-dev: And I've changed the constructors to take None (and map it to 'normal').

gregchapman-dev:
OK, I think I've addressed all of @jacobtylerwalls' review.

jacobtylerwalls (approved):
Great, thanks for the updates!

gregchapman-dev:
I'm still working on MusicXML parser/writer support for ArpeggioMarkSpanner (multi-chord/note arpeggios).  I have the writer working (I'll push it for your perusal shortly), but I can't test it very well without the parser (well, _I_ can, since I have a version of my Humdrum parser that can parse multi-chord arpeggios, but no-one else can, at least not very easily).  On to the parser!

gregchapman-dev:
Here is the parser support for multi-chord arpeggios.

gregchapman-dev:
Hmmm. Time to write more tests.

gregchapman-dev:
Once a week ping :-) 
Arpeggio support is ready for final review.

music21/musicxml/m21ToXml.py
```diff
@@ -3364,7 +3366,18 @@ def _spannerEndParameters(spannerClass, sp):
...
+        if arpeggioNumber == -1:
+            arpeggioNumber = self.nextArpeggioNumber
+            self.nextArpeggioNumber += 1
+            if self.nextArpeggioNumber > 16:
+                self.nextArpeggioNumber = 1
+            self.arpeggioNumbers[arp] = arpeggioNumber
+
+        return arpeggioNumber
+
+    def objectAttachedSpannersToNotations(self, obj, noteIndexInChord, objectSpannerBundle=None):
```
jacobtylerwalls, line 3380: We have permission to break the signature in a major version change, but I'm wondering if we can't just make this new param keyword-only and add a default value like 0?

music21/expressions.py
```diff
@@ -1426,6 +1426,65 @@ def numberOfMarks(self, num):
...
+    >>> c1 = chord.Chord('C3 E3 G3')
+    >>> c2 = chord.Chord('C4 E4 G4')
+    >>> ams.addSpannedElements([c1, c2])
+    >>> ams.type
+    'non-arpeggio'
+    >>> ams
+    <music21.expressions.ArpeggioMarkSpanner
+     <music21.chord.Chord C3 E3 G3><music21.chord.Chord C4 E4 G4>>
+    '''
+    def __init__(self, arpeggioType: t.Optional[str] = None, *arguments, **keywords):
```
jacobtylerwalls, line 1478: I'm not aware of spanners in music21 with positional arguments, since it breaks usual positional usage:

```python
>>> s.insert(expressions.ArpeggioMarkSpanner(s.first(), s.last()))
  File "/Users/jwalls/music21/music21/expressions.py", line 1483, in __init__
    raise ValueError(
ValueError: Arpeggio type must be "normal", "up", "down", or "non-arpeggio"
```

So I would look for a solution that made arpeggioType a keyword argument.

music21/musicxml/m21ToXml.py
```diff
@@ -4403,12 +4450,32 @@ def noteheadToXml(self, n: note.NotRest) -> Element:
...
         return mxNotehead
 
+    @staticmethod
+    def topAndBottomIndices(chordOrNote) -> t.Tuple[int, int]:
+        # Use sortAscending if you can, assume ascending if you can't.
+        # If chordOrNote is a note (not expected, but possible), (0, 0) will be returned.
+        # It would be nice if PercussionChord could sortAscending (ascending on the
+        # staff, that is)...
+        notes = list(chordOrNote.notes)
+        top: int = len(notes)
```
jacobtylerwalls, line 4460: Chord.notes is a tuple, so I don't think you need to listify it.
```suggestion
        top: int = len(chordOrNote.notes)
```

music21/musicxml/m21ToXml.py
```diff
@@ -4403,12 +4450,32 @@ def noteheadToXml(self, n: note.NotRest) -> Element:
...
+        bottom: int = 0
+
+        if isinstance(chordOrNote, chord.Chord):
+            ascendingChord = chordOrNote.sortAscending(inPlace=False)
+            topNote = ascendingChord.notes[-1]
+            bottomNote = ascendingChord.notes[0]
+            top = notes.index(topNote)
+            bottom = notes.index(bottomNote)
```
jacobtylerwalls, line 4468:
```suggestion
            top = chordOrNote.notes.index(topNote)
            bottom = chordOrNote.notes.index(bottomNote)
```

music21/musicxml/m21ToXml.py
gregchapman-dev: That seems awkward (it would be the only noteIndexInChord treated this way in the file). How about giving it a default?  The only thing broken then is calls with positional spannerBundle params (which now must be called with a name).

music21/expressions.py
gregchapman-dev: Yes.

music21/musicxml/m21ToXml.py
gregchapman-dev: Yes.

music21/musicxml/m21ToXml.py
gregchapman-dev: Yes.

music21/musicxml/m21ToXml.py
gregchapman-dev: I've pushed what this would look like, so you can review it.

gregchapman-dev:
Wow, flake8 got a lot pickier throughout music21. Hope that doesn't block me from getting this in.

mscuthbert:
ArpeggioMark is totally approved.  That can be merged.  Thank you!

I think from the discussion in #1142 (HammerOn, HammerOff) -- I'm still not willing to go with spanners being inside music21 objects because of the way that they're gathered -- what happens if we only show Staff 2 but the spanner is the .expressions on an object in Staff 1.  If the spanner is in Staff 1, music21 is smart enough to do the right thing (or if it's not, there's one place we need to fix it).  I'm also still scared by the circular reference -- note-holds-spanner-holds-note-holds-spanner...etc.  And it's even harder for cases like chords (what if a chord has notes 0,1,2 in staff 1 and notes 3,4 in staff 2, but only notes 2-3 are arpeggiated?) Then we have chord holds spanner which holds notes which aren't in the Stream.  I think that the general solution to the problem raised in #1142 needs to be solved before going with ArpeggioMarks inside of expressions.

Maybe there is something like a spanner that references an ordered list of other m21 objects that could be made, since the entire infrastructure of spanner doesn't seem to be needed here.

gregchapman-dev:
ArpeggioMarkSpanner is never put in obj.expressions (and shouldn't be, ever).  I will push another update to make the ArpeggioMarkSpanner docs more clear on this subject.

mscuthbert:
> ArpeggioMarkSpanner is never put in obj.expressions, and shouldn't ever contain notes that are within chords. I have pushed another update to make the ArpeggioMarkSpanner docs more clear on this subject.

Ah... I totally misread this.  Let me check further.

music21/musicxml/m21ToXml.py
```diff
@@ -4403,12 +4450,31 @@ def noteheadToXml(self, n: note.NotRest) -> Element:
             mxNotehead.set('color', color)
         return mxNotehead
 
+    @staticmethod
+    def topAndBottomIndices(chordOrNote) -> t.Tuple[int, int]:
+        # Use sortAscending if you can, assume ascending if you can't.
+        # If chordOrNote is a note (not expected, but possible), (0, 0) will be returned.
+        # It would be nice if PercussionChord could sortAscending (ascending on the
```
mscuthbert, line 4457: yes -- that would be nice.

music21/musicxml/m21ToXml.py
```diff
@@ -4403,12 +4450,31 @@ def noteheadToXml(self, n: note.NotRest) -> Element:
...
+        # staff, that is)...
+        top: int = len(chordOrNote.notes)
+        bottom: int = 0
+
+        if isinstance(chordOrNote, chord.Chord):
+            ascendingChord = chordOrNote.sortAscending(inPlace=False)
```
mscuthbert, line 4463: This can be done inPlace and then no need for the extra indexes, etc.

By the time we're exporting to musicxml there's already been a deepcopy done once, so everything is fair game to manipulate.

music21/musicxml/m21ToXml.py
```diff
@@ -4421,9 +4487,33 @@ def noteToNotations(self, n, noteIndexInChord=0, chordParent=None):
             # get expressions from first note of chord
             chordOrNote = chordParent
 
-        # only apply expressions to notes or the first note of a chord...
-        if isSingleNoteOrFirstInChord:
-            for expObj in chordOrNote.expressions:
+        # apply all expressions apart from arpeggios only to the first note of a chord.
+        topNoteIndex: int
+        bottomNoteIndex: int
+        foundTopAndBottom = False
```
mscuthbert, line 4493: see next comment

music21/musicxml/m21ToXml.py
```diff
@@ -4421,9 +4487,33 @@ def noteToNotations(self, n, noteIndexInChord=0, chordParent=None):
...
+                        mxExpression.set('type', 'bottom')
+                    elif topNoteIndex == noteIndexInChord:
+                        mxExpression = self.expressionToXml(expObj)
+                        mxExpression.set('type', 'top')
+                else:
+                    # <arpeggiate> goes on every note in the chord
+                    mxExpression = self.expressionToXml(expObj)
+                if mxExpression is None:
+                    # the ArpeggioMark is not applicable on this note.
+                    continue
```
mscuthbert, line 4513: Let's move all this (including the cached setup code) to its own arpeggioMarkToMxExpression method -- that way it can be tested separately and it doesn't interfere with reading this very important method.  It's a lot of extra lines to read through for the rather uncommon case of a ArpeggioMark let alone a Non-Arpeggiate mark.  Then it'll be easier to add future musicxml features as well.  (it's quite likely that this module is going to be split into smaller ones, like I started to do with `<sound>` and the refactor will be that easier too).

mscuthbert (changes requested):
Got it!  This is great!  A few small changes/refactors for readability, etc. then let's merge this today.  (@jacobtylerwalls feel free also to merge if you see that the changes are made and I don't check in later today).

music21/musicxml/m21ToXml.py
gregchapman-dev: done.

music21/musicxml/m21ToXml.py
gregchapman-dev: done.

gregchapman-dev:
I believe this is ready to go (I hope!)

music21/musicxml/m21ToXml.py
```diff
@@ -3425,13 +3425,10 @@ def objectAttachedSpannersToNotations(self, obj, noteIndexInChord=0, objectSpann
...
                 # <non-arpeggiate> goes only on top and bottom note in chord
-                if not foundTopAndBottom:
-                    topNoteIndex, bottomNoteIndex = self.topAndBottomIndices(obj)
-                    foundTopAndBottom = True
-                if bottomNoteIndex == noteIndexInChord:
+                if noteIndexInChord == 0:
                     mxArpeggio = Element('non-arpeggiate')
                     mxArpeggio.set('type', 'bottom')
-                elif topNoteIndex == noteIndexInChord:
+                elif noteIndexInChord == len(obj.notes):
```
mscuthbert, line 3431: this will never happen, off by one error.  `len(obj.notes) - 1`  can we have a test to make sure these are generated correctly.

music21/musicxml/m21ToXml.py
```diff
@@ -4450,23 +4450,21 @@ def noteheadToXml(self, n: note.NotRest) -> Element:
...
-
-        return top, bottom
+    def arpeggioMarkToMxExpression(self, arpeggioMark, chordOrNote, noteIndexInChord):
+        mxExpression = None
+        if arpeggioMark.type == 'non-arpeggio':
+            # <non-arpeggiate> goes on top and bottom note in chord
+            if noteIndexInChord == 0:
+                mxExpression = self.expressionToXml(arpeggioMark)
+                mxExpression.set('type', 'bottom')
+            elif noteIndexInChord == len(chordOrNote.notes) - 1:
```
mscuthbert, line 4460: here done right.  :-)

music21/musicxml/m21ToXml.py
gregchapman-dev: Yeah, my multiStaffArpeggio test doesn't have any non-arpeggiates in it.  I'll fix that (and then this bug).

music21/musicxml/m21ToXml.py
gregchapman-dev: sigh...  This one was tested, too.

mscuthbert:
Unfortunately, after adding some tests, the ArpeggioMarkSpanner for non-arpeggiate is not working, because the marks don't go on the lowest and highest notes of a single chord, but the lowest and highest notes of the entire ArpeggioMarkSpanner, so we'll need to extract that information instead.

music21/musicxml/m21ToXml.py
mscuthbert: it's in my newest PR

mscuthbert:
It's working now.  Just need to lint

music21/musicxml/m21ToXml.py
mscuthbert: looks good to me.

music21/expressions.py
mscuthbert: Fixed.

PR #1340, opened 2022-07-06
Validate figures provided to `RomanNumeral()`

### LINKED ISSUE #1280: Raise an exception when `RomanNumeral` gets non-numeric non-accidental figures
**Motivation**
In v8, I suggest raising an exception when something non-numeric or non-accidental-ish is passed to the RomanNumeral constructor:

```python
>>> roman.RomanNumeral('V¯\_ (ツ)_/¯')
<music21.roman.RomanNumeral V¯\_ (ツ)_/¯>
```

The more realistic case:
```
>>> roman.RomanNumeral('viiº').commonName  # because that symbol is subtly different from °
'minor triad'
```

**Feature summary**
see above. Need a way to delimit allowed figures (numeric, b, #, n? hyphen?)

**Intent**

jacobtylerwalls:
Fixes #1280

### DIFF
--- a/music21/roman.py
+++ b/music21/roman.py
@@ -2145,10 +2145,10 @@ class RomanNumeral(harmony.Harmony):
         # immediately fix low-preference figures
         if isinstance(figure, str):
             figure = figure.replace('0', 'o')  # viio7
-            # /o is just a shorthand for ø -- so it should not be stored.
-            figure = figure.replace('/o', 'ø')
             figure = figure.replace('º', 'o')
             figure = figure.replace('°', 'o')
+            # /o is just a shorthand for ø -- so it should not be stored.
+            figure = figure.replace('/o', 'ø')
         else:
             raise TypeError(f"Expected str or int: got {type(figure)}")
         # end immediate fixes
--- a/music21/roman.py
+++ b/music21/roman.py
@@ -2136,7 +2136,8 @@ class RomanNumeral(harmony.Harmony):
             figure = figure.replace('0', 'o')  # viio7
             # /o is just a shorthand for ø -- so it should not be stored.
             figure = figure.replace('/o', 'ø')
-            figure = figure.replace('º', '°')
+            figure = figure.replace('º', 'o')
+            figure = figure.replace('°', 'o')
         else:
             raise TypeError(f"Expected str or int: got {type(figure)}")
         # end immediate fixes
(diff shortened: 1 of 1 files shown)


music21/roman.py
```diff
@@ -2122,13 +2134,16 @@ def __init__(
         # immediately fix low-preference figures
         if isinstance(figure, str):
             figure = figure.replace('0', 'o')  # viio7
-
-        if isinstance(figure, str):
             # /o is just a shorthand for ø -- so it should not be stored.
             figure = figure.replace('/o', 'ø')
-
+            figure = figure.replace('º', '°')
```
mscuthbert, line 2139: I'm not sure that I'm liking the use of degree character for diminished when seen in the context of half diminished. The two characters are supposed to be identical except for the presence of the slash, and this change moves them to be farther apart not closer. 

All else approved.

mscuthbert (commented):
See comment.

music21/roman.py
jacobtylerwalls: Happy to take suggestions for what to do with the less common symbol `º` (instead of `°`). Should it raise an exception? If someone wrote `ii/º7` it would convert to `ii°7`. Is that a problem?

jacobtylerwalls: OH -- is it that both `viiº` and `vii°` should go to `viio` so that it's more like `iiø7'?

mscuthbert: Ah, no -- my phone font was really bad at the time, and it looked like the last one was even smaller than it was.  Still I wish that the o and ø looked better.

mscuthbert: Yeah, thinking more and doing some edits on RomanNumeral, ° is just too small and hard to type to make it a preferred character the way ø is, so I think probably better to make viio7 the default.

jacobtylerwalls: > ° is just too small and hard to type

Lol. When I need it I usually just hold down the option key and then mash the keyboard until it comes up.

jacobtylerwalls:
@malcolmsailor @MarkGotham do either of you have objections to normalizing the various characters that either are or look like degree symbols to `o`?

The context is that Mark noticed in #1280 the following bug when using the "wrong" symbol:
```
>>> roman.RomanNumeral('viiº').commonName  # because that symbol is subtly different from °
'minor triad'
```

Will normalizing to "o" break any of your tools? (E.g., will you have a bumpy upgrade path if you have comparisons to literals like `'°' in rn.figure`?)

malcolmsailor:
I have no objection!

PR #1359, opened 2022-08-06
Allow iterables of qualified class names in `Stream.__getitem__` searches

### LINKED ISSUE #1360: Stream.__getitem__ is stricter about subclassing `Music21Object` than `ClassFilter`
**music21 version**
de1be4b76af0ce2bae9bdfb2a7f02eaa787370ec

**Problem summary**
Filtering streams with the square-bracket syntax enforces stricter conditions on types (must subclass `Music21Object`) than long-hand searches with `.getElementsByClass()`.

**Steps to reproduce**

```python
>>> from music21 import *
>>> s = stream.Stream()
>>> s.append(bar.Repeat())
>>> [x for x in s.getElementsByClass(repeat.RepeatMark)]
[<music21.bar.Repeat direction=start>]
>>> [x for x in s[repeat.RepeatMark]]
  File "/Users/jacobwalls/music21/music21/stream/base.py", line 619, in __getitem__
    raise TypeError(
TypeError: Streams can get items by int, slice, class, or string query; got <class 'type'>
```

The two forms should behave the same: we should either enforce Music21Object inheritance for class queries in both forms, or we should loosen it to ProtoM21Object to ease the problem of searching for mixins like RepeatMark.

There are some documented places that say to search on the string "RepeatMark" instead of `repeat.RepeatMark` for this very reason, despite that it works with the longhand form.

**More information**
Found while doing #1359, and could potentially be fixed in the same PR.

jacobtylerwalls:
- Since v7, square brackets on Streams are syntactic sugar for `.recurse().getElementsByClass()`, but unlike that long form, the short form didn't cope with iterables of qualified class names such as `[note.Note, note.Rest]`:

```
TypeError: Streams can get items by int, slice, class, or string query; got <class 'tuple'>
```

Inspired by mailing list message.

Also,
- Remove the constraint that the types provided must be subclasses of `Music21Object` to match longhand syntax (Fixes #1360)


music21/stream/base.py
```diff
@@ -494,6 +494,8 @@ def __getitem__(self,
         2
         >>> len(s[note.Note])
         6
+        >>> len(s[[note.Note, note.Rest]])
```
jacobtylerwalls, line 497: This didn't work before. Now it works with or without the inner list brackets (i.e. it works with an implicit tuple from the multiple arguments to __getitem__)

music21/stream/base.py
```diff
@@ -605,7 +609,12 @@ def __getitem__(self,
 
             return t.cast(M21ObjType, searchElements[k])
 
-        elif isinstance(k, type) and issubclass(k, base.Music21Object):
+        elif all(
+            isinstance(maybeType, type) and issubclass(maybeType, base.Music21Object)
```
jacobtylerwalls, line 613: This condition is actually stricter than the long-form, see #issue number

music21/stream/base.py
mscuthbert: Better not to demo it with the additional inner list brackets (put in non-doc test) -- that looks really awkward, and the point of the syntax is to make m21 friendlier.

music21/stream/base.py
```diff
@@ -434,13 +434,17 @@ def __getitem__(self, k: slice) -> t.List[M21ObjType]:
     @overload
     def __getitem__(
         self,
-        k: t.Type[ChangedM21ObjType]
+        k: t.Union[t.Type[ChangedM21ObjType], t.Collection[t.Type[ChangedM21ObjType]]]
     ) -> iterator.RecursiveIterator[ChangedM21ObjType]:
```
mscuthbert, line 438: does this properly type the output for mypy?  (just want to be sure that that was checked)

music21/stream/base.py
```diff
@@ -496,6 +500,8 @@ def __getitem__(self,
         2
         >>> len(s[note.Note])
         6
+        >>> len(s[[note.Note, note.Rest]])
```
jacobtylerwalls, line 503:
```suggestion
        >>> len(s[note.Note, note.Rest])
```

music21/stream/base.py
```diff
@@ -486,7 +490,7 @@ def __getitem__(self,
         3.0
 
 
-        If a class is given then an iterator of elements
+        If a class (or iterable of classes) is given, then an iterator of elements
```
mscuthbert, line 493: better to demo these two things separately.  `If a class is given, then a RecursiveIterator of elements matching the class is returned, similar...`

then later, "Multiple classes can be passed in separated by commas.  Any element matching any of the classes will be returned."    with the demo of that.

music21/stream/base.py
```diff
@@ -607,7 +615,12 @@ def __getitem__(self,
 
             return t.cast(M21ObjType, searchElements[k])
 
-        elif isinstance(k, type) and issubclass(k, base.Music21Object):
+        elif all(
+            isinstance(maybeType, type) and issubclass(maybeType, base.Music21Object)
+            for maybeType in (
+                k if common.isIterable(k) else [k]  # type: ignore
+            )
+        ):
```
mscuthbert, line 623: not such a fan of this type of generator predicate.  Clearer to make two separate elif statements, one for the single class and second one for the isIterable case.  Students read this code and emulate it.

mscuthbert (changes requested):
Looks good.  One question and some doc updates.

music21/stream/base.py
jacobtylerwalls: I think I'm not quite following -- can you explain what you mean? (`mypy` passes, and I didn't touch the return type)

mscuthbert: What I mean is something like, say you start typing:

```
for el in s[note.Note, chord.Chord]:
    print(el.pitches)
    print(el.recurse())
```
will the `el.pitches` line start to auto-suggest .pitches when you've typed .pi... (like does it know that all possible members have .pitches in it), and will mypy choke on the `el.recurse()` -- knowing that the union of note.Note and chord.Chord does not have `.recurse()` defined on all members -- or even one member.

And if you do:

```
for el in s[note.Note, chord.Chord]:
    if not isinstance(el, note.Note):
         print(el.closedPosition())
```
and know that this is fine?  I'm just wondering if the typing is inferred from the ChangedM21Type.

And most importantly, does this still raise a mypy error:

```
for el in s[note.Note]:
     print(el.isMajorTriad())
```

while this passes?

```
for el in s[note.Note]:
     print(el.pitch)
```

I'll just grab the PR and play with it.  But I'm wondering if putting the two different types of things in the same `@overload` might cause problems.

music21/stream/base.py
```diff
@@ -516,6 +514,18 @@ def __getitem__(self,
...
+        any of the requested classes will be matched.
+
+        >>> len(s[note.Note, note.Rest])
+        9
+
+        >>> for note_or_rest in s[note.Note, note.Rest]:
+        ...     if isinstance(note_or_rest, note.Note):
+        ...         print(note_or_rest.name, end=' ')
+        ...     else:
+        ...         print("Rest", end = ' ')
```
mscuthbert, line 527: quotes and spacing.  (Sorry that the new linter doesn't do doctests.

mscuthbert (commented):
Looks good -- I'll play with this and make the tiny quote/space change myself.

mscuthbert:
No, in fact the new typing doesn't even know that el is a Music21Object.  So the overload needs to be written separately.

There was actually a reason for the stricter [RecurseType] needing to be a Music21Object -- that's how the operator knew that everything coming out was still a Music21Object.  There's a way to make it still work.

mscuthbert:
For typing I think we're going to need to leave in the stricter version and probably make .getElementsByClass() stricter too.  People will need to do the .getElementsByClass('RepeatMark') way, or `.addFilter(lambda x: isinstance(x, RepeatMark))` ... or we'll make RepeatMark a Music21Object.   That's probably the best way to do things.

Adding general t.Type as one of the things that could filter meant that we were losing the amazing type-inference that was one of the main points of doing this refactor.  Unfortunately Python does not yet have a `t.Type[t.Not[M21ObjType]]`

mscuthbert (approved):

PR #1361, opened 2022-08-06
MusicXML export: Make rests before making notation

### LINKED ISSUE #904: Express durations with or without tuplets according to context

jacobtylerwalls:
Follow up to #1242. Didn't quite fix #904 because we need to make rests _before_ making notation (if we are to have the opportunity to re-express the rests inside tuplets).

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -469,6 +469,14 @@ class GeneralObjectExporter:
 
         Changed in v8 -- fills gaps with rests before calling makeNotation
         to avoid duplicating effort with :meth:`PartExporter.fixupNotationMeasured`.
+
+        >>> v = stream.Voice(note.Note())
+        >>> m = stream.Measure([meter.TimeSignature(), v])
+        >>> GEX = musicxml.m21ToXml.GeneralObjectExporter(m)
+        >>> out = GEX.parse()  # out is bytes
+        >>> outStr = out.decode('utf-8')  # now is string
+        >>> '<note print-object="no" print-spacing="yes">' in outStr
+        True
         '''
         classes = obj.classes
         outObj = None
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -541,6 +541,17 @@ class GeneralObjectExporter:
         return self.fromMeasure(m)
 
     def fromStream(self, st):
+        '''
+        Changed in v8 -- fills gaps with rests before calling makeNotation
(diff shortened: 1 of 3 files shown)


music21/musicxml/m21ToXml.py
```diff
@@ -1676,16 +1696,6 @@ def parsePartlikeScore(self):
...
-
-        Hide rests created at this late stage.
-
-        >>> v = stream.Voice(note.Note())
-        >>> m = stream.Measure([meter.TimeSignature(), v])
-        >>> GEX = musicxml.m21ToXml.GeneralObjectExporter(m)
-        >>> out = GEX.parse()  # out is bytes
-        >>> outStr = out.decode('utf-8')  # now is string
-        >>> '<note print-object="no" print-spacing="yes">' in outStr
-        True
```
jacobtylerwalls, line 1688: This test is moved to fromGeneralObject() doctest

music21/musicxml/test_m21ToXml.py
```diff
@@ -419,9 +419,13 @@ def testMeasurePadding(self):
         s = stream.Score([converter.parse('tinyNotation: 4/4 c4')])
         s[stream.Measure].first().paddingLeft = 2.0
         s[stream.Measure].first().paddingRight = 1.0
+        # workaround until getET() helper starts calling fromGeneralObject
+        s = GeneralObjectExporter().fromGeneralObject(s)
```
jacobtylerwalls, line 423: We really need to get all the musicxml tests to declare whether or not they depend on notation being made. This was a problem _before_ we added makeNotation=False as an option -- essentially, getET() bypasses part of the normal musicxml export path.

mscuthbert:
Thanks!

Issue #1358, opened 2022-08-06
Other inversions of dominant 7ths


malcolmsailor:
**music21 version**

`8.0.0a9`

**Problem summary**

According to the [docs](https://web.mit.edu/music21/doc/moduleReference/moduleRoman.html?highlight=roman), "Dominant 7ths can be specified by putting d7 at end". Naively, I expected "d43" to then be a dominant 7th in 43 position. It isn't, but we can get that with "43d7". Perhaps a note of clarification in the docs could be useful? Or just changing the behavior for "d43"?

**Steps to reproduce**

```
>>> from music21 import roman
>>> RN = roman.RomanNumeral
>>> RN("IV7").pitches
(<music21.pitch.Pitch F4>, <music21.pitch.Pitch A4>, <music21.pitch.Pitch C5>, <music21.pitch.Pitch E5>)
>>> RN("IVd7").pitches
(<music21.pitch.Pitch F4>, <music21.pitch.Pitch A4>, <music21.pitch.Pitch C5>, <music21.pitch.Pitch E-5>)
>>> RN("IV43").pitches
(<music21.pitch.Pitch C4>, <music21.pitch.Pitch E4>, <music21.pitch.Pitch F4>, <music21.pitch.Pitch A4>)
>>> RN("IVd43").pitches # I naively expected C-Eb-F-A
(<music21.pitch.Pitch C4>, <music21.pitch.Pitch E4>, <music21.pitch.Pitch F4>, <music21.pitch.Pitch A4>)
>>> RN("IV43d7").pitches # OK
(<music21.pitch.Pitch F4>, <music21.pitch.Pitch E-5>, <music21.pitch.Pitch A5>, <music21.pitch.Pitch B5>)
```

malcolmsailor:
Actually I just noticed that the last pitch of `RN("IV43d7").pitches` is B, which is not what I expected. So maybe there is no way to use "d" to change an inversion to a dominant 7th?

malcolmsailor:
It looks to me like this could be solved by matching patterns beyond "d7" here:

https://github.com/cuthbertLab/music21/blob/de1be4b76af0ce2bae9bdfb2a7f02eaa787370ec/music21/roman.py#L2337

A regex that matches d followed by the various 7th chord patterns (7, 65, 43, 42, 2... others?) should do it. I could implement this if you like.

PR #1363, opened 2022-08-07
match d43, d65, etc.

malcolmsailor:
I went ahead and implemented a regex to make other inversions prefaced by "d" also dominant seventh chords as I suggested in #1358. It was just a couple lines of code and it will allow me to clean up some of the stray chords still not matching with the tsvConverter.

NB pylint objects `music21/roman.py:2378: [R0201(no-self-use), RomanNumeral._findSemitoneSizeForQuality] Method could be a function` but this is not code introduced by me.

My remaining question is: perhaps 'd' should indicate dominant seventh quality no matter what digits follow? I.e., if a user indicates something weird like "d479', we should just pass the digits straight through but set impliedQuality to 'dominant-seventh'?


mscuthbert:
> NB pylint objects music21/roman.py:2378: [R0201(no-self-use), RomanNumeral._findSemitoneSizeForQuality] Method could be a function but this is not code introduced by me.

We have no-self-use disabled in the music21 pylint configuration.  Make sure that pylint is reading the file.

music21/roman.py
```diff
@@ -1606,13 +1606,20 @@ class RomanNumeral(harmony.Harmony):
     >>> cp(r)
     ['A-5', 'C6', 'E-6', 'G-6']
 
-    >>> r = roman.RomanNumeral('VId7')
+    >>> r = roman.RomanNumeral('VId2')
     >>> r.figure
-    'VId7'
+    'VId2'
```
mscuthbert, line 1611: this should be VId42 -- music21 prefers and emphasizes the 42 rather than 2 for 3rd inversion.

This should also have text such as "inversions of dominant 7ths can be created by specifying 'd' followed by the figures"

music21/roman.py
```diff
@@ -1785,6 +1792,11 @@ class RomanNumeral(harmony.Harmony):
     >>> cp(r)
     ['G5', 'B5', 'D6', 'F6']
 
+    >>> r = roman.RomanNumeral('IVd6/5')
+    >>> r.key = key.Key('Eb')
+    >>> cp(r)
+    ['C5', 'E-5', 'G-5', 'A-5']
+
```
mscuthbert, line 1799: put this in a test below.  It uses a form that we need to parse but that we're not encouraging.

music21/roman.py
```diff
@@ -2334,10 +2346,9 @@ def _setImpliedQualityFromString(self, workingFigure):
             workingFigure = workingFigure[1:]
             impliedQuality = 'augmented'
             # impliedQualitySymbol = '+'
-        elif workingFigure.endswith('d7'):
+        elif m := re.match(r'(?P<leading>.*)d(?P<figure>7|6/?5|4/?3|4/?2|2)$', workingFigure):
```
mscuthbert, line 2349: rewrite as:

`elif 'd' in workingFigure and (m := ...what_you_currently_have...):`

because regular expression matching is slow, so we only want to do a regular expression match if there is already a 'd' there.  It will make it slightly slower for d65 etc. chords, but much faster for all other chords.

mscuthbert: In fact, it'd probably be better to throw exception on chords that can't be made dominant 7th chords:

```
if 'd' in workingFigure:
    m = re.match(r'(?P<leading>.*)d(?P<figure>7|6/?5|4/?3|4/?2|2)$', workingFigure)
    if m is None:
        raise RomanException(
              f'Cannot make a dominant-seventh chord out of {workingFigure}.'
        )
    ...rest of your code...
```

So, no, I don't want implied quality for things that aren't dominant sevenths.  If someone wants that they can use the "add" --  RomanNumeral('IVd65[add2]') or set the `implied_quality` later.

mscuthbert (changes requested):
Great!  Some small and fast changes then ready to go.  Thanks!

music21/roman.py
malcolmsailor: This is in the "OMIT_FROM_DOCS" section so I thought it was an appropriate place for such a thing.

malcolmsailor:
> We have no-self-use disabled in the music21 pylint configuration. Make sure that pylint is reading the file.

I don't see it in the .pylintrc in the repository root directory. Is there more pylint configuration elsewhere?

music21/roman.py
mscuthbert: AH!!! IT IS!  MY BAD!

mscuthbert:
> > We have no-self-use disabled in the music21 pylint configuration. Make sure that pylint is reading the file.
> 
> I don't see it in the .pylintrc in the repository root directory. Is there more pylint configuration elsewhere?

Ah!  For various reasons, we have to put the configuration in twice -- once in music21/test/testLint.py and once in the .pylintrc -- they're supposed to be kept perfectly in sync, but obviously they weren't for this point.

mscuthbert:
Looking good.  Letting tests run (I should've had you do a one-byte PR to get you approved as not "first-time-contributor" long ago), and if all pass will merge.  THANKS!

malcolmsailor:
> Looking good. Letting tests run (I should've had you do a one-byte PR to get you approved as not "first-time-contributor" long ago), and if all pass will merge. THANKS!

As you can see the tests are failing with the same error I referenced in #1267; any suggestions for what might be going on with that are welcome.

mscuthbert:
It's the 3.10.6 change that was causing the failure.  Fix pushed and running tests again.

mscuthbert:
> > We have no-self-use disabled in the music21 pylint configuration. Make sure that pylint is reading the file.
> 
> I don't see it in the .pylintrc in the repository root directory. Is there more pylint configuration elsewhere?

Ah -- no-self-use was moved to an optional extension in recent versions of pylint, and is no longer checked.  So the solution is to upgrade your pylint.

mscuthbert (approved):

mscuthbert:
Merged.  Congrats!

PR #1364, opened 2022-08-09
Add setters back for metadata-typing

mscuthbert:
Adding setters back to metadata.py for properties, each with a note saying that they don't actually run.  My type checker in PyCharm was going crazy saying that title could not be set, etc. and will help with typing of the value of the setter.

Ping @gregchapman-dev @ahankinson -- good to get a set of eyes that I didn't break anything

### DIFF
--- a/music21/metadata/__init__.py
+++ b/music21/metadata/__init__.py
@@ -620,7 +620,7 @@ class Metadata(base.Music21Object):
     @copyright.setter
     def copyright(self, value: str) -> None:
         '''For type checking only. Does not run.'''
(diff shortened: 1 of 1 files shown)


gregchapman-dev:
Looks good. Should the date setter have a deprecation comment?

Greg

mscuthbert:
> Looks good. Should the date setter have a deprecation comment? Greg
> […](#)

I'll add one in the next PR, but it doesn't really matter since setter docs are always ignored and only the getter docs are displayed.

PR #1368, opened 2022-08-11
Accidental commit to master -- damage?

mscuthbert:
Accidentally committed to the main branch.  Checking if any damage was done and if I need to revert or keep going.  (Was working on expressions and thought I was on a branch)

### DIFF
--- a/music21/base.py
+++ b/music21/base.py
@@ -4058,11 +4058,10 @@ class ElementWrapper(Music21Object):
         return object.__getattribute__(storedObj, name)
 
 
-
-
 # ------------------------------------------------------------------------------
 # define presented order in documentation
 _DOC_ORDER = [Music21Object, ElementWrapper]
+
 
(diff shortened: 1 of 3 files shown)


ahankinson:
If it's helpful, you can set up branch protection rules in GitHub so that some branches can only accept merges and not direct pushes. 

https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches

mscuthbert:
> If it's helpful, you can set up branch protection rules in GitHub so that some branches can only accept merges and not direct pushes.

I tried that a few years ago, but it wasn't possible to do before.  I'll check to see if it is possible now.

mscuthbert:
Thanks @ahankinson -- it's apparently been possible now since sometime in 2021.

PR #1371, opened 2022-08-11
Test protected branch workflow

mscuthbert:
Add small note to expressions


mscuthbert:
Coverage run has been moved to Python 3.9, and is explicitly NOT a required test so that merges after a complete test pass can take place in 5 minutes rather than 10-12 (Coverage slows down Python runs significantly)

PR #1375, opened 2022-08-14
Prevent too large `<backup>` values in PartStaffExporter

jacobtylerwalls:
The PartStaffExporter was producing `<duration>` values under `<backup>` tags that were too large by summing the durations of all chord members (instead of just taking one of the durations).

Adds a regression test failing on master:
```
FAIL: testBackupAmount (__main__.Test.testBackupAmount)
Regression test for chord members causing too-large backup amounts.
  File "/Users/jacobwalls/music21/music21/musicxml/partStaffExporter.py", line 1079, in testBackupAmount
    self.assertEqual(
AssertionError: '30240' != '10080'
- 30240
+ 10080


Ran 21 tests in 2.173s

FAILED (failures=1)
```

### DIFF
--- a/music21/musicxml/partStaffExporter.py
+++ b/music21/musicxml/partStaffExporter.py
@@ -743,7 +743,12 @@ class PartStaffExporterMixin:
 
         # Create <backup>
         amountToBackup: int = 0
-        for dur in otherMeasure.findall('note/duration'):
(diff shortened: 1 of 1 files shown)


mscuthbert (approved):
Thanks for the fix!

PR #1377, opened 2022-08-14
Explicit Keywords on all music21 objects

### LINKED ISSUE #839: Keywords need to be part of the function signature
It's time to stop creating methods and classes like this:

```
def __init__(self, **keywords):
     if 'hello' in keywords:
         hello = keywords['hello']
     else:
         hello = 'hi'
```

All the keywords expected by a method should be in the function signature:

```
def __init__(self, *, hello='hi'):
```

It's still okay to allow a generic `**keywords` argument, if the class is expected to be subclassed by something with more keywords or if it calls a subfunction that might have more keywords:

```
def __init__(self, *, hello='hi', **keywords):
      super().__init__(**keywords, hello=hello)
```

But it's not at all apparent to a user what is possible with `**keywords` in the current state.

Do this before (or start it) as part of #763 show implementation.

mscuthbert:
Fixes #839

Each push will include one or more modules.

Starting with Base.

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -61,6 +61,7 @@
         <list>
           <option value="type" />
           <option value="format" />
+          <option value="id" />
         </list>
       </option>
     </inspection_tool>
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,7 +40,8 @@ or Stack Overflow.
 ## Submitting Pull Requests ##
 
 Open an issue to propose a feature or report a bug before raising a pull request.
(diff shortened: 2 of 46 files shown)


mscuthbert:
Misc fixes:
* Metadata date formats derive from DatePrimitive not DateSingle (attn: @gregchapman-dev)
* remove unused note.linkage
* kwargs -> keywords throughout
* fix metadata tests pinned to 8.0.0a9

Modules/Classes affected include:

* base
* Tuplet
* note
* Stream
* chordBase
* discrete analysis
* Variant (now takes only one arg)
* midi realtime, etc.
* Chorales, Date primitives, RepeatBracket
* layout

Issue #858, opened 2021-02-15
Repeats in Roman text


MarkGotham:
`romanText.rtObjects` includes a provision for start/end repeat tokens in Roman text (e.g. `:||`) but `romanText.translate` leaves those tokens unprocessed.

**Minimal Proposal:**
Process the token to create a `bar.Repeat` object for these tokens (with correct direction for start versus end repeats).

**Additional Proposal:**
As this is an in-line token, could we accept beat definitions like `m4 V :|| b3 vi` (i.e. mid-measure repeats) and split the measure into two partial measures at that point, adding the repeat marks at the start/end of those partial measures?

(That's assuming I haven't missed a smarter way of doing mid-measure repeats that's consistent with musicxml etc.)

MarkGotham:
p.s.
You can get these repeat markings into a rntxt-generated score by using first/second time bars (`mXa` / `mXb`) but apparently not by typing them in directly (i.e. for cases without first/second time).

mscuthbert:
Repeat signs can go anywhere in music21, but they gain much more functionality if they're put into two separate measures -- including export to systems that can only handle repeats at ends/beginnings of measures (I believe that MusicXML supports them mid-measure also, but no software will respect them.)

My sense for direction to go.  First to parse end of bar repeats / beginning of bar repeats.  make sure they work.  Then make a second PR for mid-bar repeats/etc.  -- ensure that they appear where they should in the bar.  Then create a generalized "splitOnMidMeasureBarlines" function (not in Stream.base.py, but maybe in stream.makeNotation?) that given a measure either returns itself intact or two new measures with the elements put into the proper sub-measure.  Then another function called "splitMeasuresOnMidMeasureBarlines" that works on a Part and returns a new Part with the measures split up. (or, optionally, the same Part if nothing needs to be done).  Then have that function run as part of the RNText process.  If well-tested enough, it's something that we can consider doing in makeNotation on musicxml output.

napulen:
One comment on this: the notation `||` in RomanText means *phrase boundary*, which might be confusing if at some point `:||` means *repeat barline*. In that case, `||` would make more sense as a *double barline* or maybe *end barline*. In fact, I originally interpreted `||` as a double barline, and then I had to look into the documentation to see what it really meant.

MarkGotham:
Thanks @mscuthbert and @napulen! 

Responding to @napulen's comment re: `||`, you're completely right, though:
[...]

PR #1378, opened 2022-08-14
fixed up romanText repeats

malcolmsailor:
This PR addresses much of #858, though not (I think) the case of mid-measure repeats.

It turns out that romanText repeats were already largely implemented (by whoever wrote the romanText/translate.py, presumably), but there were a couple bugs that were preventing it from working as expected and there was no test.

I wrote some test cases and debugged the existing code. The bugs are explained in comments in the code. As far as I can tell, the repeats now work smoothly.

I also have a few (non-urgent, I think) questions in the comments. I'll repeat them here:

1. repeat.Expander does not work on stream.Stream, only on stream.Part. Is this expected behavior?
2. in the case of 1st and 2nd endings, spanner.RepeatBracket.getNumberList does not return the measure numbers I expect. What is the correct approach here?

### DIFF
--- a/music21/romanText/translate.py
+++ b/music21/romanText/translate.py
@@ -1635,7 +1635,6 @@ m2: I :||
         _repeat_tester(end_repeat, 'end', 3.0, 2)
         _test_expanded(s, 12.0)
 
-        # TODO flake quotes
         single_bar_repeats = '''Time Signature: 2/4
 m1: ||: I :||
 '''
--- a/music21/romanText/translate.py
+++ b/music21/romanText/translate.py
@@ -130,13 +130,16 @@ import unittest
 from music21 import bar
 from music21 import base
 from music21 import common
+from music21 import converter
 from music21 import exceptions21
 from music21 import harmony
 from music21 import key
 from music21 import metadata
 from music21 import meter
 from music21 import note
+from music21 import repeat
 from music21 import roman
+from music21 import spanner
 from music21 import stream
 from music21 import tie
 from music21.romanText import rtObjects
@@ -674,7 +677,6 @@ class PartTranslator:
         self.numberOfAtomsInCurrentMeasure = len(measureToken.atoms)
         # first RomanNumeral object after a key change should have this set to True
         self.setKeyChangeToken = False
-
         for i, a in enumerate(measureToken.atoms):
             isLastAtomInMeasure = (i == self.numberOfAtomsInCurrentMeasure - 1)
             self.translateSingleMeasureAtom(a, m, isLastAtomInMeasure=isLastAtomInMeasure)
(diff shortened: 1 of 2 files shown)


malcolmsailor:
pylint objects to a bunch of stuff in translate.py but I don't believe any of it is introduced by this PR

mscuthbert:
yes, the problems were introduced in this PR when additional modules were imported at the top of the file.  romanText.translate cannot import converter, because romanText is itself a converter, so we'd get a circular import.

malcolmsailor:
Oops, sorry about that, right you are. It slipped my mind that I had imported converter. I should have been more careful and linted the code without my changes.

music21/romanText/translate.py
```diff
@@ -781,15 +782,21 @@ def translateSingleMeasureAtom(self, a, m, *, isLastAtomInMeasure=False):
         elif isinstance(a, rtObjects.RTChord):
             self.processRTChord(a, m, self.currentOffsetInMeasure)
         elif isinstance(a, rtObjects.RTRepeat):
-            if self.currentOffsetInMeasure == 0:
+            if isinstance(a, rtObjects.RTRepeatStop) and isLastAtomInMeasure:
+                # This condition needs to be moved here because it is possible
+                #   and indeed likely that a end repeat will have an offset of
```
mscuthbert, line 787: "an end repeat"

music21/romanText/translate.py
```diff
@@ -1605,6 +1606,94 @@ def testTuplets(self):
...
+            e = repeat.Expander(p)
+            self.assertTrue(e.repeatBarsAreCoherent())
+            p2 = e.process()
+            self.assertEqual(p2.quarterLength, quarterLength)
+
+        # Test simple repeats
+        simple_repeats = '''Time Signature: 3/4
+m1: ||: V
+m2: I :||
+'''
```
mscuthbert, line 1632: note that these can be indented to look nice and then use `textwrap.dedent('''...''')` on it.  No Big Deal

music21/romanText/translate.py
```diff
@@ -1605,6 +1606,94 @@ def testTuplets(self):
...
+        s = converter.parse(simple_repeats, format='romanText')
+        start_repeat, end_repeat = s[bar.Repeat]
```
mscuthbert, line 1634: better to assign to an iterator and test that `self.assertEqual(len(iterator), 2)` so that if something fails later we can see what the problem is rather than a cryptic "to few/many elements to unpack".  Again, NBD.

music21/romanText/translate.py
```diff
@@ -1605,6 +1606,94 @@ def testTuplets(self):
...
+        _repeat_tester(end_repeat2, 'end', 3.0, 2)
+        first_ending, second_ending, third_ending = s[spanner.RepeatBracket]  # pylint: disable=unused-variable
+        # TODO the following tests don't work; repr(first_ending) gives
+        #   '<music21.spanner.RepeatBracket 1 <music21.stream.Measure 2a offset=3.0>>'
+        #   but first_ending.getNumberList() returns [1]
+        #   What is the correct approach to checking that the measures are as we
+        #   expect? Or should we just forget about it since _test_expanded passes?
+        # self.assertEqual(first_ending.getNumberList(), ['2a'])
+        # self.assertEqual(second_ending.getNumberList(), ['2b'])
+        _test_expanded(s, 18.0)
```
mscuthbert, line 1679: I'm not sure that I follow, but can you make an issue if it's something beyond this PR?  It seems like we're not setting numberList properly.

mscuthbert (approved):
I'm going to merge this, since three of the issues are small (style/grammar) changes in test code, and the one substantive one (the TODO) is described carefully and can be a separate issue.  I'm concerned that getNumberList() is returning false information, but that seems like either a misunderstanding of all of us, or a separate bug.

music21/romanText/translate.py
malcolmsailor: Wow how have I not known about that module all these years. Thanks!

music21/romanText/translate.py
malcolmsailor: I investigated this some more and I realized that RepeatBracket.number or getNumberList indicates not the measure numbers that the bracket spans, but instead the "ending number" (e.g., 1 for the first ending, 2 for the second ending, and so on). I was misunderstanding.

I'll submit a new commit to clean up this misleading comment and implement this test correctly.

PR #1379, opened 2022-08-14
Add TypeGuard from typing_extensions

mscuthbert:
Should help mypy with places where we use isNum, isListLike, and isIterable.


mscuthbert:
Not a help -- making things worse.

PR #1385, opened 2022-08-15
Contributing

MarkGotham:
Bringing over a few topics discussed in other issues/PRs.
Feel free to take some / all / none of these (all separate commits in order to support that).
Likewise, feel free to re-order as needed (I didn't discern an order to follow).
Thanks!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -115,6 +115,9 @@ Conventions:
   - dependencies:
     - Music21 supports all core python modules ... but nothing more, 
       so avoid external dependencies such as numpy.
+  - assignments:
+    - use 2+ lines instead of multiple assignments like `(a, b = c, d)`.
+    - use explicit parentheses when left and right side are not balanced in number of assignments.
   - methods:
     - no more than three positional arguments (in addition to `self`)
     - keyword arguments should be keyword-only by using `*`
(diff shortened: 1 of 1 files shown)


mscuthbert:
Some of this is obsolete -- music21 has required external modules in requirements.txt since version 7 at least. 

Make sure the examples are in the right place/indentation level -- the getElementsByClass() example has nothing to do with underscore/camelCase. -- a live markdown editor like one in JetBrains will help.  And be sure the examples are actual python code:  (a, b = c,d) does not compile so of course I don't like that.  The actual rule is don't make multiple assignments where the user has to read across commas to see what b is assigned to.  This is fine:  `a, b = myTuple` or `x, y = getBoundingBox()`;  I'd probably even be okay with `(a, b) = (c, d)` if parenthesized like that, since it's easy to see what things are.  What I hate hate hate is: 

`a, b, c = 'hello', self.getOffsetBySite(previousActiveSite) + 3, 2`

mscuthbert:
Dependencies -- it's all about discussion before adding any new ones, and the general answer is probably no.  AND they should only be imported within a method that will actually use them, unless the usage is ubiquitous throughout music21.  So I definitely don't load scipy when `import music21` is called, since that takes forever.

mscuthbert:
In fact, I just wrote:

```
        for chunk in dateStr.split('/'):
            if chunk == '--':
                value, error = None, None
            else:
                value, error = self._stripError(chunk)
```
where to me the `value, error = None, None` seemed a nice parallel to `value, error = self._stripError(chunk)`

mscuthbert:
I think that the great ideas from this have been incorporated or we've figured out reasons not to.  Closing for now.  Reopen if there are parts missed.

PR #1394, opened 2022-08-17
Remove *arguments where not used 

mscuthbert:
The use of *arguments just masks errors and pushes debugging down the road.  I could never remember which Classes use them and which don't.

Leaving out interval.py for now, since that needs a substantially bigger rewrite.


mscuthbert:
Lots of extra commits because this was based on the **keywords branch, and that was squashed.  This needs to be squashed not merged as well.

PR #1405, opened 2022-09-02
Parse double flats in RomanText

### LINKED ISSUE #1390: Parsing double-flat keys in RomanText
**music21 version**

`8.0.0a9`   

**Problem summary**

The RomanText parser can't handle keys with double-flats, which occasionally occur in highly chromatic music (e.g., in [this Reger example](https://github.com/MarkGotham/When-in-Rome/blob/master/Corpus/Textbooks/Reger%2C_Max/Modulation/21/analysis.rntxt)).

**Steps to reproduce**

```
>>> music21.converter.parse("m1 Bbb: I", format="romanText")[music21.roman.RomanNumeral].first()
<music21.roman.RomanNumeral I in B- major> # should be B-flat-flat, not B-flat
```

Double-sharps are ok:
```
>>> music21.converter.parse("m1 F##: I", format="romanText")[music21.roman.RomanNumeral].first()
<music21.roman.RomanNumeral I in F## major>
```

And music21 can handle double-flat keys in general:
```
>>> music21.key.Key("B--")
<music21.key.Key of B-- major>
```

I can have a look into this.

malcolmsailor:
Fixes #1390.

For the vast majority of cases (where there is just a single flat symbol) the regex on line 83 is overkill so we only run that if the preceding conditions don't match.

### DIFF
--- a/music21/key.py
+++ b/music21/key.py
@@ -75,7 +75,7 @@ def convertKeyStringToMusic21KeyString(textString):
     >>> key.convertKeyStringToMusic21KeyString('Ebb')
     'E--'
     '''
-    if (not textString[-1] == 'b') or textString == 'b':
+    if (not textString.endswith('b')) or textString == 'b':
         pass
     elif textString == 'bb':
         textString = 'b-'
--- a/music21/key.py
+++ b/music21/key.py
@@ -72,22 +72,19 @@ def convertKeyStringToMusic21KeyString(textString):
     'B--'
     >>> key.convertKeyStringToMusic21KeyString('bbb')
     'b--'
+    >>> key.convertKeyStringToMusic21KeyString('Ebb')
+    'E--'
     '''
-    if textString == 'bb':
+    if (not textString[-1] == 'b') or textString == 'b':
+        pass
+    elif textString == 'bb':
         textString = 'b-'
     elif textString == 'Bb':
         textString = 'B-'
     elif len(textString) == 2 and textString[-1] == 'b':
-        textString = textString[:1] + '-'
-    elif textString.endswith('b'):
-        m = re.match(r'''
-                (?P<leading_chars>.[^b]*) # matches at least one leading character
-                (?P<flats>b+)
-            ''',
-            textString,
-            flags=re.VERBOSE)
-        if m:
-            textString = m.group('leading_chars') + '-' * len(m.group('flats'))
+        textString = textString[0] + '-'
+    elif set(textString[1:]) == {'b'}:
+        textString = textString[0] + '-' * (len(textString) - 1)
     return textString
 
 
--- a/music21/key.py
+++ b/music21/key.py
@@ -68,13 +68,26 @@ def convertKeyStringToMusic21KeyString(textString):
     'b#'
     >>> key.convertKeyStringToMusic21KeyString('c')
     'c'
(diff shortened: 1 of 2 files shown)


music21/key.py
```diff
@@ -68,13 +68,26 @@ def convertKeyStringToMusic21KeyString(textString):
...
+    'b--'
     '''
     if textString == 'bb':
         textString = 'b-'
     elif textString == 'Bb':
         textString = 'B-'
-    elif textString.endswith('b') and not textString.startswith('b'):
-        textString = textString.rstrip('b') + '-'
+    elif len(textString) == 2 and textString[-1] == 'b':
+        textString = textString[:1] + '-'
```
mscuthbert, line 81: this can just be `textString[0]`

music21/key.py
```diff
@@ -68,13 +68,26 @@ def convertKeyStringToMusic21KeyString(textString):
...
+    elif textString.endswith('b'):
```
mscuthbert, line 82: why not an additional check for `elif textString in ('b', 'B'): pass` so that the keys of 'b' or 'B' don't need to go through the regex?

but I wonder if the regex is overkill.  Can't it be:

```
elif textString.endswith('b'):
     textString = textString[0] + '-' * (len(textString) - 1)
```

and assume all characters in between are 'b's.  Or if we don't want to make the assumption and let things like "E-b" or "C#b" fail, something like:

```
elif set(textString[1:]) == {'b'}:
     textString = textString[0] + '-' * (len(textString) - 1)
```

music21/key.py
```diff
@@ -68,13 +68,26 @@ def convertKeyStringToMusic21KeyString(textString):
     'b#'
     >>> key.convertKeyStringToMusic21KeyString('c')
     'c'
+    >>> key.convertKeyStringToMusic21KeyString('Bbb')
+    'B--'
+    >>> key.convertKeyStringToMusic21KeyString('bbb')
+    'b--'
```
mscuthbert, line 74: Also a test for a different double-flat key than b, say 'Ebb' since that follows a completely different code path potentially than 'bbb'

mscuthbert (changes requested):
A few small changes to make the common case of Key of 'b' or 'B' work more quickly.

music21/key.py
malcolmsailor: I think both cases follow the same path since the checks for B/b/Bb/bb only apply when the key has two or fewer characters. But I added such a test.

I wasn't sure what the right balance between comprehensiveness and readability was here since these are actually doctests (albeit ones that people who aren't actually developing music21 are probably unlikely to consult).

malcolmsailor:
OK, done. Rather than checking for 'B' explicitly I just pass if the last character is anything other than 'b'; then we don't check all those other redundant conditions if e.g., the key is "F#" or "E", etc.

malcolmsailor:
It just occurred to me that this function will throw an exception if textString is the empty string. Should we handle that case or is the exception ok?

mscuthbert:
Looks great!  Thanks @malcolmsailor and for the last patch @jacobtylerwalls  We can make empty string throw an exception later, but if so we should do it explicitly and not just with an `IndexError` which would be strange.

mscuthbert (approved):

Issue #1402, opened 2022-08-28
I need a way to have a Spanner (e.g. Crescendo) start or end with a timestamp instead of a note/chord/rest


gregchapman-dev:
**Motivation**
Several file formats (MEI, Humdrum) often describe Crescendos/Diminuendos as objects with duration, rather than associating them directly with notes in the same staff.  If there is no note in the same staff with the appropriate start/end time, a parser will need to create an invisible Rest (in a separate Voice) instead, complicating the Score unnecessarily.

**Feature summary**
Add a new object (or perhaps extend the current Dynamic object) to represent a crescendo/diminuendo that is not a Spanner, but instead has duration.

**Proposed implementation**
Simplest bit is defining the new data structure.  Medium bit is coming up with a name.  The trickiest bit will be support for the new object in the MusicXML writer.

**Intent**

[x ] I plan on implementing this myself.
[ ] I am willing to pay to have this feature added.
[ ] I am starting a discussion with the hope that community members will volunteer their time to create this. I understand that individuals work on features of interest to them and that this feature may never be implemented.

gregchapman-dev:
I'm not sure of the v8 schedule; I would love to get this into v8 if at all possible.

gregchapman-dev:
I'll try it out and see what works and doesn't work.  Can a Spanner have no notes?  Just an offset and duration?

gregchapman-dev:
Ah... the first question that needs to be answered, I think, is "how would a crescendo with duration be represented in a MusicXML file"?  And I fear that the duration would advance the time pointer, causing the immediately following note (in the MusicXML) to be offset to the right by the duration of the crescendo.  I think we would actually have to use <backup/> to get the time pointer back to the beginning of the crescendo.  Thoughts?

jacobtylerwalls:
The `<offset>` tag is supposed to take care of all of that, so that the musicxml "time point" can be offset as much as you like. Have you had a look at `setOffsetOptional`? You could also export a 1-measure document straight out of Finale.

mscuthbert:
I believe that the system (using a duration, etc.) would work if at least one note is in the spanner.  I don't know if it works with zero (but can you really crescendo over nothing?)

Definitely hairpins are really under tested in music21.

gregchapman-dev:
[...]

PR #1407, opened 2022-09-05
Allow DynamicWedge to have no spanned elements and be inserted with offset and duration instead

gregchapman-dev:


gregchapman-dev:
This is completely untested as yet.  It's a straw-man that I am now attempting to use in my Humdrum parser, to see if it does what I want.

gregchapman-dev:
Obviously, xmlToM21.py will need to be enhanced to parse `<wedge/>` elements with offsets, or we won't be able to parse what we write.

gregchapman-dev:
I'd love some review of this straw-man implementation.

gregchapman-dev:
I think this implementation is a fail.  See issue #1402 for further discussion.

Issue #1413, opened 2022-09-08
`Roman.Numeral.transpose()` partially doesn't work.


malcolmsailor:
**music21 version**

`8.0.0a12`   

**Problem summary**

```
In [29]: RN = music21.roman.RomanNumeral

In [30]: RN("I", "C").transpose(4).pitches # Looks good!
Out[30]: (<music21.pitch.Pitch E4>, <music21.pitch.Pitch G#4>, <music21.pitch.Pitch B4>)

In [31]: RN("I", "C").transpose(4) # oops, key should be E major
Out[31]: <music21.roman.RomanNumeral I in C major>

In [32]: RN("I", "C").transpose(4).key # oops again
Out[32]: <music21.key.Key of C major>
```

Note that if the roman numeral is embedded in a part, the keys in the part are transposed but not the keys of the roman numeral:

```
In [34]: rntxt = """Time Signature: 4/4
    ...: m1 C: I"""

In [37]: converter.parse(rntxt, format="romanText").transpose(4).show("text")
{0.0} <music21.metadata.Metadata object at 0x1076f61f0>
{0.0} <music21.stream.Part 0x10429e400>
    {0.0} <music21.stream.Measure 1 offset=0.0>
        {0.0} <music21.key.Key of E major>
        {0.0} <music21.meter.TimeSignature 4/4>
        {0.0} <music21.roman.RomanNumeral I in C major>
```

The romanText writer seems to rely on the RomanNumeral keys, so it then ignores the transposition. I.e., after running `converter.parse(rntxt, format="romanText").transpose(4).write("romanText", "rntxttemp")` the contents of "rntxttemp" are:

```
Composer: Composer unknown
Title: Title unknown
Analyst: 
Proofreader: 

Time Signature: 4/4
m1 C: I
```

Note that the key is C and not E!

I can have a stab at fixing this. It seems to me that all that should be necessary is to call `transpose()` on the keys within the RomanNumerals.

Incidentally, in making this issue I noticed that `.show("romanText")` doesn't seem to work (I would have expected it to just print out the romanText representation, analogous to `.show("text")`).

mscuthbert:
If you transpose I in C major up a P5, does it become I in G major or V in C major?

It all depends on the context what the correct answer should be.  If you're transposing the whole piece, then just the key should be transposed.  But then there's another problem: I frequently share Key objects between roman numerals:

```
e = key.Key('E')
i1 = roman.RomanNumeral('I', e)
[...]

PR #1414, opened 2022-09-08
transpose key of RomanNumeral

malcolmsailor:
Fixes the issue in #1413. I closely modeled RomanNumeral.transpose() on Harmony.transpose().

### DIFF
--- a/music21/roman.py
+++ b/music21/roman.py
@@ -42,6 +42,8 @@ environLocal = environment.Environment('roman')
 
 # TODO: setting inversion should change the figure
 
+T = t.TypeVar('T', bound='RomanNumeral')
+
 # -----------------------------------------------------------------------------
 
 
@@ -3088,6 +3090,28 @@ class RomanNumeral(harmony.Harmony):
             raise RomanNumeralException(
                 f'_updatePitches() was unable to derive pitches from the figure: {self.figure!r}'
             )  # pragma: no cover
+
+    def transpose(self, value, *, inPlace=False) -> t.Optional[T]:
+        '''
+        Overrides :meth:`~music21.harmony.Harmony.transpose` so that `key`
+        attribute is transposed as well.
+
+        >>> rn = roman.RomanNumeral('I', 'C')
+        >>> rn
+        <music21.roman.RomanNumeral I in C major>
+        >>> rn.transpose(4)
+        <music21.roman.RomanNumeral I in E major>
+        >>> rn.transpose(-4, inPlace=True)
+        >>> rn
+        <music21.roman.RomanNumeral I in A- major>
+        '''
+        post = super().transpose(value, inPlace=inPlace)
+        if not inPlace:
+            post.key = self.key.transpose(value, inPlace=False)
+            return post
+        else:
+            self.key.transpose(value, inPlace=True)
+            return None
 
 
     # PUBLIC PROPERTIES #
--- a/music21/roman.py
+++ b/music21/roman.py
@@ -3110,7 +3110,7 @@ class RomanNumeral(harmony.Harmony):
             post.key = self.key.transpose(value, inPlace=False)
             return post
         else:
-            self.key.transpose(value, inPlace=True)
+            self.key = self.key.transpose(value, inPlace=False)
             return None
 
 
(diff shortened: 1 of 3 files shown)


music21/roman.py
```diff
@@ -3089,6 +3091,28 @@ def _updatePitches(self):
...
+        >>> rn.transpose(-4, inPlace=True)
+        >>> rn
+        <music21.roman.RomanNumeral I in A- major>
+        '''
+        post = super().transpose(value, inPlace=inPlace)
+        if not inPlace:
+            post.key = self.key.transpose(value, inPlace=False)
+            return post
+        else:
+            self.key = self.key.transpose(value, inPlace=False)
```
mscuthbert, line 3113: why couldn't this be `self.key.transpose(value, inPlace=True)` ?

music21/roman.py
```diff
@@ -3089,6 +3091,28 @@ def _updatePitches(self):
                 f'_updatePitches() was unable to derive pitches from the figure: {self.figure!r}'
             )  # pragma: no cover
 
+    def transpose(self, value, *, inPlace=False) -> t.Optional[T]:
```
mscuthbert, line 3094: for this to work, self has to be annotated with T as well.

mscuthbert: see Harmony L 2338

mscuthbert:
I didn't see that this was ready because it was marked with an "x" -- reviewed now -- one fix needed to not get incorrect typing, one nice to have.

Consider looking at the paradigm in key.Key transpose where the two inPlace options are typed differently so that it's obvious to the typing system whether to expect a RN or None to be returned.

music21/roman.py
malcolmsailor: That's what I had in my initial commit (a40e41a7f04b829ef02e9d8cb66a7dea1a78af70) but I changed it based on the discussion in #1413 and your feedback at https://github.com/cuthbertLab/music21/issues/1413#issuecomment-1247205169.

music21/roman.py
malcolmsailor: OK, fixed, but mypy didn't seem to be complaining. The "x" seems to be because coverage decreased slightly. (Not quite sure why since both `inPlace=True` and `inplace=False` are tested in the `RomanNumeral.transpose()` doctest.)

malcolmsailor:
mypy is failing now:

```
music21/audioSearch/__init__.py:529: error: Incompatible return value type (implicitly returns "None", expected "List[int]")
music21/interval.py:363: error: Incompatible return value type (implicitly returns "None", expected "Tuple[Literal['C', 'D', 'E', 'F', 'G', 'A', 'B'], int]")
music21/stream/iterator.py:796: error: A function returning TypeVar should receive at least one argument containing the same Typevar
Found 3 errors in 3 files (checked 259 source files)
```

I don't get these errors on my local machine; I just tried merging the current master into the branch and I still don't get them. So I'm not sure whether this PR is somehow introducing them?

music21/roman.py
mscuthbert: ah okay -- yes, we solved the problem in RomanText but not elsewhere, my bad.

mscuthbert:
> mypy is failing now:
> 
> ```
> music21/audioSearch/__init__.py:529: error: Incompatible return value type (implicitly returns "None", expected "List[int]")
> music21/interval.py:363: error: Incompatible return value type (implicitly returns "None", expected "Tuple[Literal['C', 'D', 'E', 'F', 'G', 'A', 'B'], int]")
> music21/stream/iterator.py:796: error: A function returning TypeVar should receive at least one argument containing the same Typevar
> Found 3 errors in 3 files (checked 259 source files)
> ```
> 
> I don't get these errors on my local machine; I just tried merging the current master into the branch and I still don't get them. So I'm not sure whether this PR is somehow introducing them?

They must have upgraded their version of mypy today and it now finds errors that it didn't before.  Those aren't from your work, so we can go and merge and I'll squash them separately.  THANKS!

mscuthbert:
oh, wait I turned on merge protection to prevent me from accidentally committing to master and I can't overrule this without some major tweaking (2-factor, etc.) so I'll get a separate fix in and then merge this.  Sorry for the delay.

mscuthbert:
Closing and reopening to get a clean board.

Issue #1415, opened 2022-09-08
Allow writing to file-like objects with romanText writer


malcolmsailor:
**Motivation**

I would like to be able to pass a file-like object to the RomanText writer. Judging from [subConverters.write](https://github.com/cuthbertLab/music21/blob/00971f4a8e8d44e326e35bd922fd02311bb8f467/music21/converter/subConverters.py#L262), which delegates to [subConverters.writeDataStream](https://github.com/cuthbertLab/music21/blob/00971f4a8e8d44e326e35bd922fd02311bb8f467/music21/converter/subConverters.py#L275), this capability is expected to be supported, but in `ConverterRomanText` it [is not](https://github.com/cuthbertLab/music21/blob/00971f4a8e8d44e326e35bd922fd02311bb8f467/music21/converter/subConverters.py#L1317). I can implement this modeling the new code on writeDataStream in the parent class `subConverter`.

**Intent**

[x] I plan on implementing this myself.
[ ] I am willing to pay to have this feature added.
[ ] I am starting a discussion with the hope that community members will volunteer their time to create this. I understand that individuals work on features of interest to them and that this feature may never be implemented.

mscuthbert:
Thanks -- that'd be nice to have.

PR #1416, opened 2022-09-08
romanText write to file-like objects

malcolmsailor:
Realizes #1415.

One difference between my new code and subConverter.writeDataStream() is that the latter calls fp.close() when fp is a file-like object. However, I don't think this is really ideal---if the caller is passing an open file, they probably expect it to remain open. (They may, for example, wish to continue writing to it later.) And in the case of writing to StringIO objects, (which is my use case) StringIO buffer contents are discarded when calling StringIO.close().

### DIFF
--- a/music21/converter/subConverters.py
+++ b/music21/converter/subConverters.py
@@ -1323,9 +1323,14 @@ class ConverterRomanText(SubConverter):
         if fp is None:
             fp = self.getTemporaryFile()
 
-        with open(fp, 'w', encoding='utf-8') as text_file:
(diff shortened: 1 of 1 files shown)


mscuthbert:
does this work with BytesIO ?   Actually, I'm thinking that we should be getting rid of the encoding='utf-8' above anyhow -- that's old Python 2 code still living in here.  If we did that then we wouldn't need to worry about BytesIO -- people will be expected to pass in an object that can have strings written to it.

mscuthbert:
though maybe the encoding doesn't harm anything,

PR #1418, opened 2022-09-09
write 'slow 3/8' etc.

### LINKED ISSUE #1417: RomanText writer doesn't write "slow 3/8"
**music21 version**

`8.0.0a12`   

**Problem summary**

If music21 reads a RomanText file in "slow 3/8", then writes it out again, it writes it in "3/8" (which defaults to "fast 3/8"), omitting the "slow". The output file will likely be unreadable since it may contain beats that are then illegal, e.g., "b2".

**Steps to reproduce**

```
>>> from io import StringIO
>>> rntxt = """Time Signature: slow 3/8
... m1 C: I b2 V"""
>>> s = music21.converter.parse(rntxt, format="romanText")
>>> text_stream = StringIO()
>>> s.write("romanText", text_stream)
>>> newrntxt = text_stream.getvalue()
>>> print(newrntxt) # NB not "Slow 3/8"
Composer: Composer unknown
Title: Title unknown
Analyst: 
Proofreader: 

Time Signature: 3/8
m1 C: I b2 V
>>> s2 = music21.converter.parse(newrntxt, format="romanText")
[rest of TraceBack clipped]
    raise RomanTextTranslateException(
music21.romanText.translate.RomanTextTranslateException: too many notes in this measure: m1 C: I b2 V
```

I will see if I can't fix this myself.

malcolmsailor:
Fixes #1417.

One question might be: do we want the writer to write "fast 3/8" in the case of fast 3/8, etc? I didn't implement this since my approach is to change the existing behavior as little as possible. But I'm inclined to think it might be better to be maximally explicit and write "fast 3/8".

Another question could be: is there such thing as "slow 9/8" or "slow 12/8", etc? And what about "slow 3/16" etc.? I only check for 3/8 and 6/8 since these are the only such meters mentioned in the docs as far as I could tell. And in fact I believe I have never even seen "slow 6/8", only "slow 3/8".

### DIFF
--- a/music21/romanText/writeRoman.py
+++ b/music21/romanText/writeRoman.py
@@ -249,7 +249,14 @@ class RnWriter(prebase.ProtoM21Object):
             tsThisMeasure = thisMeasure.getElementsByClass(meter.TimeSignature)
             if tsThisMeasure:
                 firstTS = tsThisMeasure[0]
-                self.combinedList.append(f'Time Signature: {firstTS.ratioString}')
+                if (firstTS.ratioString in ('3/8', '6/8')
+                        and firstTS.beatDivisionCountName == 'Simple'):
+                    tsPrefix = 'slow '
+                else:
+                    tsPrefix = ''
+                self.combinedList.append(
+                    f'Time Signature: {tsPrefix}{firstTS.ratioString}'
+                )
                 if len(tsThisMeasure) > 1:
                     unprocessedTSs = [x.ratioString for x in tsThisMeasure[1:]]
                     msg = f'further time signature change(s) unprocessed: {unprocessedTSs}'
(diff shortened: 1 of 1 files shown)


mscuthbert:
Let's have further discussion on #1417 first.

mscuthbert:
Converting to draft until the issues in #1417 are resolved.

mscuthbert:
#1417 seems far enough along that I'm ready to take this as a merge --

malcolmsailor:
> #1417 seems far enough along that I'm ready to take this as a merge --

Yes, I've been occupied with other things but I will revise this based on the discussion there shortly.

mscuthbert:
Hi @malcolmsailor -- do you have time to catch this up? it was very close to a merge.  THANKS!

malcolmsailor:
I'll try to look into it this weekend

mscuthbert:

Issue #1412, opened 2022-09-07
First/Second endings in DCML conversion


malcolmsailor:
**music21 version**

`8.0.0a12`

**Problem summary**

I've noticed that the TSV converter implemented by @MarkGotham and updated by myself doesn't take any account of first and second endings. When there are first and second endings in the DCML files, it simply puts the contents of *both* endings into the numbered measures. (Whereas, according to my understanding of the RomanText standard, it should label first endings "a" and second endings "b".)

**Steps to reproduce**

Here's a simple TSV that allows us to reproduce this behavior (it's the 'volta' column that indicates first/second endings):

```
mc 	mn 	mc_onset 	mn_onset 	timesig 	staff 	voice 	volta 	label 	globalkey 	localkey 	pedal 	chord 	special 	numeral 	form 	figbass 	changes 	relativeroot 	cadence 	phraseend 	chord_type 	globalkey_is_minor 	localkey_is_minor 	chord_tones 	added_tones 	root 	bass_note
 1 	 1 	       0 	       0 	2/4     	      	      	      	.C.I  	C         	I        	      	I     	        	I       	     	        	        	             	        	FALSE     	           	                   	                  	            	            	     	
 2 	 2 	       0 	       0 	2/4     	      	      	    1 	V     	C         	I        	      	V     	        	V       	     	        	        	             	        	FALSE     	           	                   	                  	            	            	     	
 3 	 2 	       0 	       0 	2/4     	      	      	    2 	I     	C         	I        	      	I     	        	I       	     	        	        	             	        	FALSE     	           	                   	                  	            	            	     	
						
```

If we put this into a file and run the following commands:

```
>>> from music21 import romanText
>>> f = "music21/romanText/tsvEg_v2_repeats.tsv"
>>> stream = romanText.tsvConverter.TsvHandler(f, dcml_version=2).toM21Stream()
>>> converter.subConverters.ConverterRomanText().write(stream, 'romanText', fp="temp.txt")
```

The contents of "temp.txt" are then:

```
Composer: Composer unknown
Title: Title unknown
Analyst: 
Proofreader: 

Time Signature: 2/4
m1 C: I
m2 V I
```
Note that measure 2 contains both V (from ending 1) and I (from ending 2) at beat 1.

I am happy to implement a fix myself but I'm wondering if anybody has any pointers to where I can read about the best way of creating/manipulating endings in music21. Thanks!

gregchapman-dev:
[...]

PR #1434, opened 2022-09-22
Fixes bug in repeat tests in romanText/translate.py; handles empty measures

malcolmsailor:
While working on addressing #1412 I noticed that there was a syntax error in the romanText in `testRepeats` (my bad!). Each measure number was followed by a colon (i.e., "m1:" rather than "m1"). The first thing this PR fixes is removing those colons.

The colons caused the parser to insert a note in each measure. When I removed them, I discovered that the test was actually failing on measures that are empty except for forward/backward repeat bars. To fix this required two things:
1. explicitly set the offset of backward repeat bars according to the time-signature length
2. if a measure has 0 duration, explicitly set its duration according to the current time-signature

I also added a new test for the copying of empty measures.

### DIFF
--- a/music21/romanText/translate.py
+++ b/music21/romanText/translate.py
@@ -575,7 +575,7 @@ class PartTranslator:
 
         else:
             m = self.translateSingleMeasure(measureLineToken)
-            if m.barDuration.quarterLength == 0:
+            if self.tsCurrent is not None and m.barDuration.quarterLength == 0:
                 m.duration = duration.Duration(
                     self.tsCurrent.barDuration.quarterLength
                 )
--- a/music21/romanText/translate.py
+++ b/music21/romanText/translate.py
@@ -130,7 +130,6 @@ import unittest
 from music21 import bar
 from music21 import base
 from music21 import common
-from music21 import duration
 from music21 import exceptions21
 from music21 import harmony
 from music21 import key
@@ -575,10 +574,8 @@ class PartTranslator:
 
         else:
             m = self.translateSingleMeasure(measureLineToken)
-            if self.tsCurrent is not None and m.duration.quarterLength == 0:
-                m.duration = duration.Duration(
-                    self.tsCurrent.barDuration.quarterLength
-                )
+            if m.duration.quarterLength == 0:
+                self.fillMeasureFromPreviousRn(m)
             p.coreAppend(m)
 
     def fillToMeasureToken(self, measureToken: rtObjects.RTMeasure):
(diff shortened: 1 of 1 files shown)


mscuthbert:
Thanks!

music21/romanText/translate.py
```diff
@@ -574,6 +575,10 @@ def translateMeasureLineToken(self, measureLineToken: rtObjects.RTMeasure):
 
         else:
             m = self.translateSingleMeasure(measureLineToken)
+            if self.tsCurrent is not None and m.duration.quarterLength == 0:
+                m.duration = duration.Duration(
+                    self.tsCurrent.barDuration.quarterLength
+                )
```
mscuthbert, line 581: this isn't right, unfortunately.  Duration on a stream is special property that automatically manages its quarterLength based on the duration of elements inside of it.  You'd be repeating an empty measure, not a measure of RN.

What should happen, I think is that the code from 593-605 (was 588-600) from "fillToMeasureToken" should be abstracted out into a new method (`fillMeasureFromPreviousRn(mFill: stream.Measure) -> None`) and then that method should be called instead.  Something needs to go into the repeat measure.  We can't repeat an empty span of time.

music21/romanText/translate.py
```diff
@@ -1661,10 +1682,10 @@ def _test_ending_contents(
...
         empty_bars_with_repeats = textwrap.dedent('''
             Time Signature: 2/4
-            m1: I
-            m2: ||:
-            m3: :||
-            m4: ||: :||
+            m1 I
+            m2 ||:
+            m3 :||
+            m4 ||: :||
```
mscuthbert, line 1688: Add tests that the repeated measures do have an RN.figure == 'I' token in it after doing above.

mscuthbert (changes requested):
Realizing that the empty measures can't end empty (and duration on a stream should not be manipulated) -- proposed solution in comments.

music21/romanText/translate.py
malcolmsailor: OK, I'll do this.

In this connection, it might be worth clarifying the docs on when (if ever) explicitly setting duration is possible:

https://web.mit.edu/music21/doc/moduleReference/moduleStreamBase.html#music21.stream.base.Stream.duration

They describe explicitly setting duration in the way I did above as an "advanced usage". I assumed an "advanced usage" would be ok inside the music21 source which is why I thought it would be ok to do.

mscuthbert: Yeah, I'll try to improve the docs for that.

It basically means that no matter how long the duration of the measure is, it'll always report some other number, which is something that an end user might want to do, but which our parsers shouldn't do.

music21/romanText/translate.py
```diff
@@ -590,26 +587,29 @@ def fillToMeasureToken(self, measureToken: rtObjects.RTMeasure):
...
+            self.fillMeasureFromPreviousRn(mFill)
             appendMeasureToRepeatEndingsDict(self.lastMeasureToken,
                                              mFill,
                                              self.repeatEndings, i)
             p.coreAppend(mFill)
         self.lastMeasureNumber = measureToken.number[0] - 1
         self.lastMeasureToken = measureToken
 
+    def fillMeasureFromPreviousRn(self, mFill: stream.Measure) -> None:
+        if self.previousRn is not None:
```
mscuthbert, line 599: this is fine. but there's an easy refactor to avoid nesting.

```
if self.previousRn is None:
    return
```

music21/romanText/translate.py
```diff
@@ -1695,6 +1695,14 @@ def _test_ending_contents(
         _repeat_tester(end_repeat1, 'end', 2.0, 3)
         _repeat_tester(start_repeat2, 'start', 0.0, 4)
         _repeat_tester(end_repeat2, 'end', 2.0, 4)
+        for measure in s[stream.Measure]:
+            rn_iter = measure[roman.RomanNumeral]
+            self.assertEqual(len(rn_iter), 1)
+            # mypy complains about the next line because
+            #   RecursiveIterator.first() has t.Optional type, but we know
+            #   it will not be None because we have just asserted that rn_iter
+            #   has length 1
```
mscuthbert, line 1704: that's fine.  :-)

mscuthbert (approved):

PR #1393, opened 2022-08-17
Write roman ts and repeats

MarkGotham:
Write complement to #1378 along with a bit of clean up

PR #1435, opened 2022-09-22
Romantext writer handle repeat bars and measure suffixes like m1a

malcolmsailor:
While working on addressing #1412 I noticed that the RomanText writer is not writing out repeat bars or measure suffixes like "m1a".

I.e., previously if we read in

```
Time Signature: 2/4
m1 ||: C: I
m2 V :||
m3 ||: I :||
m4 ||: I
m5a V :||
m5b I
```

Then wrote it out, we would get

```
Time Signature: 2/4
m1 C: I
m2 V
m3 I
m4 I
m5 V
m5 I
```

This PR fixes this issue by implementing writing of repeat bars and measure suffixes.

### DIFF
--- a/music21/romanText/writeRoman.py
+++ b/music21/romanText/writeRoman.py
@@ -13,10 +13,12 @@ Writer for the 'RomanText' format (Tymoczko, Gotham, Cuthbert, & Ariza ISMIR 201
 '''
 
 import fractions
+import textwrap
 import unittest
 
 import typing as t
 
+from music21 import bar
 from music21 import base
 from music21 import metadata
 from music21 import meter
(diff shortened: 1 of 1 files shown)


music21/romanText/writeRoman.py
```diff
@@ -332,11 +360,14 @@ def rnString(measureNumber: int,
     '''
 
     if inString:
```
mscuthbert, line 362: In definition of rnString, change type to t.Union[int, str]

mscuthbert: But it looks like you're concating int(measure.number) + measure.suffix only to separate them later.  why not pass in the suffix separately?

I didn't look carefully enough at `rnString()` when this went in.  It's quite inefficient.  It seems to me that it really should be an object whose __str__ gives the correct result so you don't need to keep passing in a the same string to concat to and read from over and over.  But that's a later/larger refactor.

mscuthbert (changes requested):
The only thing that must change is the typing on the `def rnString()` -- mypy should be picking that up as an error.  But there's the opportunity to make this much better than constantly reading from the same string.

music21/romanText/writeRoman.py
```diff
@@ -332,11 +360,14 @@ def rnString(measureNumber: int,
...
-        inStringMeasureNumber = int(inString.split(' ')[0][1:])
-        if inStringMeasureNumber != measureNumber:
+        inStringMeasureNumber = inString.split(' ')[0][1:]
```
malcolmsailor, line 363: > But it looks like you're concating int(measure.number) + measure.suffix only to separate them later. why not pass in the suffix separately?

Do you mean on this line? The only thing I changed was removing the cast to int. It doesn't split the suffix, it splits the inString on space and then slices the "m" from the beginning. As far as I can tell the suffix stays put once it is added.

I agree this function is clearly inefficient. It seems the obvious thing to do would be accumulate each "word" to a list and then do `' '.join` when finished.

malcolmsailor:
I fixed the type annotation. I notice that ironically the mypy run on github now fails, but it succeeds on my local machine, and it seems to be objecting to something in numpy.

mscuthbert:
> I fixed the type annotation. I notice that ironically the mypy run on github now fails, but it succeeds on my local machine, and it seems to be objecting to something in numpy.

Seems like a known issue in numpy -- I'll do a patch to the github actions workflow for now. 
https://github.com/python/mypy/issues/13627

mscuthbert:
Patch is in -- rerunning tests

mscuthbert (approved):

music21/romanText/writeRoman.py
MarkGotham: Thanks all.
+1 for sorting out `rnString`.
Mea culpa for how it looks as it stands though I suspect it ended up that way as one of those solutions that seeks to please everyone ... but ends up half way between different solutions and ends up pleasing no one ;)
See #1393 for an alternative that removes it completely.
I'm happy to advance this sometime.

PR #1443, opened 2022-09-25
Music21 py3.10/11 only

mscuthbert:
To make music21 able to be developed faster and more of a joy for the dev team to work on, music21 will now support only Python 3.10+ for now, and drop its prior policy of supporting previous 3 versions.

Major bug fixes might appear in patches to older versions.

Because we want one fast build that is required to pass and one slower build to measure code coverage, but Py 3.11 is not ready to run dependencies yet, we're running coverage on 3.10.6 and fast tests on 3.10-latest (currently 3.10.7)

### DIFF
--- a/.github/workflows/maincheck.yml
+++ b/.github/workflows/maincheck.yml
@@ -14,7 +14,7 @@ jobs:
         runs-on: ubuntu-latest
         strategy:
             matrix:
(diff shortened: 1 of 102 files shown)


mscuthbert:
Fixes #1444

PR #1451, opened 2022-10-06
Add converter.toData, improve braille docs

mscuthbert:
Adds converter.toData -- a drop in for getting the raw data from a format.

Improve documentation for Braille


mscuthbert:
Keep getting coverage drops because I keep deleting multiple blank lines in doc tests.  :-)

PR #1457, opened 2022-10-10
`__eq__` comparison on time signatures

MarkGotham:


MarkGotham:
... I've reviewed the attrs for ratioString and beatSequence and am pretty sure that:
- that's a good basis for this comparison
- these tests should work ...

music21/meter/base.py
```diff
@@ -41,6 +41,7 @@
 environLocal = environment.Environment('meter')
 
 if TYPE_CHECKING:
+    from music21 import meter
```
mscuthbert, line 44: Unused

music21/meter/base.py
```diff
@@ -489,6 +490,64 @@ def __init__(self, value: str = '4/4', divisions=None):
     def _reprInternal(self):
         return self.ratioString
 
+    def __eq__(self, other) -> bool:
+        '''
+        Establishes equality between two time signatures
```
mscuthbert, line 495: Dunder methods `__xxx__` don't display docs.  So put the basic information in the class documentation at the top, and just put corner case doctests here.

music21/meter/base.py
```diff
@@ -489,6 +490,64 @@ def __init__(self, value: str = '4/4', divisions=None):
...
+
+        >>> oneKindOf44 = meter.TimeSignature('4/4')
+        >>> sameKindOf44 = meter.TimeSignature()
+        >>> oneKindOf44 == sameKindOf44
+        True
+
+        >>> otherKindOf44 = meter.TimeSignature('Cut')
+        >>> oneKindOf44 == otherKindOf44
+        False
+
```
mscuthbert, line 540: Here are some things that weren't tested:

Is TS('Cut') == TS('2/2')  (I think no)

Is TS('slow 6/8') == TS('fast 6/8') ?  (I think no, and beatSequence will take care of it?)

BIG ONE:

Is TS('4/4') == note.Note() and reversed: is note.Note == TS('4/4')

you're find that your test will raise an attribute error.

Look at the __eq__ for other music21.note -- and others to fix this.  The first line should always be something like:

```
if not isinstance(other, type(self)):
    return False  # or return NotImplemented -- we do both in the current code, should unify
```

music21/meter/base.py
```diff
@@ -489,6 +490,64 @@ def __init__(self, value: str = '4/4', divisions=None):
...
+        '''
+
+        if self.ratioString == other.ratioString:  # test 1
```
mscuthbert, line 543: why not make this 
```
if self.ratioString != other.ratioString:
    return False
```
and then you don't need nested comparisons

mscuthbert (changes requested):
Some changes requested -- the part about checking class is very important.

music21/meter/base.py
MarkGotham: Thanks @mscuthbert , I'll implement these tests to keep the code up-to-date with the discussion ... though I stopped adding tests before because even the few existing tests I already don't pass. Apologies if that wasn't clear.

There's something spooky going on here with `meter.TimeSignature` and ... and we're bit early for Halloween.

Shall we try and fix-finish this off as a placeholder to be superseded by the higher level #1458 at some point? I.e., as a kind of local test case before the more ambitious project. Or do you prefer to implement that higher level solution first?

music21/meter/base.py
MarkGotham: Why not indeed!

music21/meter/base.py
MarkGotham: Thanks @mscuthbert.

I understand that they don't display. Copying here from the practice seen elsewhere. E.g., `note.GeneralNote.__eq__` contains:

```
        '''
        General Note objects are equal if their durations are equal, and
        they have the same articulation and expression classes (in any order),
        and their ties are equal.
        '''
```

Should that comment (and any others like it) then move up to the class docs? Happy to implement (not in this PR, obv.)

music21/meter/base.py
MarkGotham: True. This was in for a different kind of type checking now removed.

MarkGotham:
Latest:
- implements new `__eq__` syntax #1466
- fixes (uncomments) the failing tests highlighted in #1466.
- replaces `beatSequence` check (does not catch fast vs slow 6/8) with `beatCount` and `ratioString`

Possible remaining questions:
- Do we want `assertRaises` on the comparison of different classes or is `False` good (as here)?
- Basic `__eq__` method in `core` seems to have been the missing link. Now fixed. Q: all and only the right checks there or something else needed?

music21/meter/base.py
jacobtylerwalls: Actually in #1233 I made an exception for `__eq__`.

MarkGotham: Thanks @jacobtylerwalls.
This PR emulates @mscuthbert 's practice (discussed here and seen in #1466) of explaining equality in the class docs.
I'm obv. happy to implement either (i.e., as here or move back to `__eq__`).

jacobtylerwalls: Sounds good. Just adding context to advertise we won't need a second refactor of the docstrings that survived #1466.

MarkGotham: Great. How does the rest of this PR look to you?

music21/meter/base.py
```diff
@@ -489,6 +541,16 @@ def __init__(self, value: str = '4/4', divisions=None, **keywords):
     def _reprInternal(self):
         return self.ratioString
 
+    def __eq__(self, other) -> bool:
+        '''
+        See notes at class doc and in meter.tests
+        '''
+
+        if not super().__eq__(other):
+            return False
```
jacobtylerwalls, line 550: I think Myke mentioned some inconsistencies exist in music21 with this, but I'd like to move toward `NotImplemented` rather than False, which is what Python defaults to: https://docs.python.org/3/reference/datamodel.html#object.__eq__

Also, I think I first learned of this from mypy, so I'm wondering why mypy doesn't complain (maybe it's an info-level issue)?

mscuthbert: I'm settling on False, because the base assumption in the new `__eq__` system is that two objects have to be the same class to be equal, so there's no point in returning NotImplemented since it should never find a True value later on.

For `__lt__` or `__gt__` there is still a good reason to return NotImplemented.

Note that bool(NotImplemented) returns True with a deprecation warning for Py3.9 and newer, so including it in a hierarchy of `__eq__` runs into major problems.

I'm doing a major rewrite of MeterSequence/MeterTerminal now, but it's getting a bit out of control.  So maybe we should set up something along the way, but definitely I think slow 6/8 and fast 6/8 should be not equal, same was 4/4 and 2/2.

music21/meter/base.py
```diff
@@ -448,10 +448,62 @@ class TimeSignature(TimeSignatureBase):
     '2/4+3/16' or '11/16 (2/4+3/16)'.  Or you might want the written
     TimeSignature to contradict what the notes imply.  All this can be done
     with .displaySequence.
+
+    Equality
+    --------
```
mscuthbert, line 453: sadly -- it turns out this doesn't work -- Sphinx crashes.  Need to just label with bold: `**Equality**` and lose the heading marker `---------`.

music21/meter/base.py
mscuthbert: This `__eq__` routine isn't necessary.  The `equalityAttributes` are enough to do the trick.  So delete this routine.

music21/meter/core.py
```diff
@@ -418,6 +418,13 @@ def __init__(self, value=None, partitionRequest=None):
 
     # SPECIAL METHODS #
 
+    def __eq__(self, other) -> bool:
+
+        if not super().__eq__(other):
+            return False
+
+        return True
+
```
mscuthbert, line 427: This is also not needed.   If you're just returning the value of a `super()` call, then there's no point in subclassing.

music21/meter/base.py
jacobtylerwalls: Makes sense about the new __eq__ system!

MarkGotham: Thanks. True. The equalityAttributes are enough here. Deleted.
Further notes on other parts of this discussion below for ease of reference.

music21/meter/core.py
MarkGotham: Ok thanks. True in so far as the actual test is not needed ... but the method is not expendable. Something is needed here, perhaps just to give MeterTerminal some notion of equality. FWIW empty equalityAttributes tuple (`equalityAttributes = ()`) doesn't work.

MarkGotham:
Latest: the 3x equality attributes seem to cover all and only the right comparisons:
- `beatCount`: slow 6/8 and fast 6/8 successfully treated as unequal
- `symbol`: Cut vs 2/2
- `ratioString` (which actually stands in for much more than just a string): 2+3 != 3+2. 

See tests at the end of `meter/tests` for coverage and proof of this particular pudding ;)

music21/meter/core.py
```diff
@@ -419,10 +419,6 @@ def __init__(self, value=None, partitionRequest=None):
     # SPECIAL METHODS #
 
     def __eq__(self, other) -> bool:
-
-        if not super().__eq__(other):
-            return False
-
         return True
```
mscuthbert, line 422: I still don't understand this -- what does this help in the system?  I don't know of any object that always compares True to every other object.

MarkGotham: Hi @mscuthbert .

Thanks for this. As I say, this odd little method is just a placeholder - not the final solution, but a demo/reminder that something is needed on this class and none of the following are acceptable alternatives:
- `def __eq__` with a simple `pass` 
- no `def __eq__` but with `equalityAttributes = ()` on the class instead
- neither of those two.

Excuse the exploratory approach but I don't know why this behaves as it does - whether it's to do the new wider handling of eq you've started implementing, or something on meter.

In short, I'm just reporting on the behaviour I see for this case and what 'works' (i.e., what apparently is needed to get the musical tests returning the correct bools).

My hope is that having this oddity and discussion point helps us diagnose what's going on and replace it with something more plausible. Likely it will feed into your wider re-working of meter.

The rest seems to work well ... and without oddities.

Thanks for any ideas for a fix. Equally, just say if you want to drop this PR, e.g., if it's not a good time for it / doesn't fit with your root-and-branch reform of the meter trees ... pun intended ;)

mscuthbert: I think it's best to comment it out.  I've never heard of an object where `if OBJ == True` is the same as `if OBJ == False` and `if OBJ == 5` and `if OBJ == 2` are all True.  It leads to huge issues with using the obj in a hash situation.

mscuthbert:
Hi @MarkGotham -- just a ping on this.

MarkGotham:
Hi @mscuthbert , @jacobtylerwalls , all

Thanks for the nudge.

_Eureka!_

Coming back to this fresh, I realise that the much-discussed and not resolved placeholder `__eq__` method was a red herring and the issue was with a single test case that a pre-existing "Work in Progress" test that's out of scope. Specifically:
- test case pre-existing this PR (was commented out as marked as WIP)
- I thought it could now be included, so removed the comment marks
- but actually I can't because it tests on a different class (`MeterSequence`).

BTW this is one of two such test case (pre-existing this PR, commented out, marked as WIP) -- the other one is on `TimeSignature` objects, is now included, and works fine.

So, in summary:
- This PR delivers on its promise: `__eq__ comparison on time signatures`
- We could consider a separate PR for `__eq__ comparison on MeterSequence`
- If you're content with this then I can probably also finalise (restore time signatures to) #1454.

mscuthbert (approved):
Looks good now.  Thanks for groking the equityAttributes -- this PR really inspired making all that finally work.  Will merge when the little edit passes

PR #1461, opened 2022-10-12
try some performance improvements

mscuthbert:
right now it's actually slower on the computer, but my CPU is going crazy.  This should prevent the creating of two durations for note objects on copying.


mscuthbert:
Was about 1 minute faster (15%) than recent builds.  Merging.

mscuthbert:
Wrong PR -- I thought that this was merged already.

PR #1463, opened 2022-10-12
WIP: Instantiation w/ **keywords tests

mscuthbert:
TEMPORARILY Restoring an old requirement for Music21Objects that require them to be instantiable without ().

I think that this is the wrong approach now that I'm working on it, and will probably close in favor of a `__new__` based deepcopy rather than `__init__` based.


mscuthbert:
Closing in favor of #1464

PR #1464, opened 2022-10-14
Faster deepcopy from reduce_ex

mscuthbert:
Use reduce_ex and the private copy._reconstruct() to take advantage of C-language state for custom deepcopies.

PurgeOrphans after every deepcopy.

Change to derived contextSites -- the derivation origin is included in the context sites.  The former behavior looks like a bug.

volume keeps a strong reference to its client.

StreamStatus keeps a strong reference to its client.

Stream.toSoundingPitch checks derivations.  This was intended behavior originally but was relying on orphans not being purged to do so.

Stream.measure()  [not .measures() nor Score.measure()] gets a substantial speedup by not creating a new Part.

TODO: remove "collect" from .measure() and see if gatherSpanners can also be removed from Score.measure()

Remove weakrefs from getstate several places.


mscuthbert:
Realtime improvement: about 10% on small scores.

However, not seeing a speedup on Beethoven opus132.  6% slower.  Might be some sort of hickup with extra copies of a site?  30000 notes.  Or the removal of the duration deepcopy speedups?  Will try to fix that.

mscuthbert:
Removed slowdowns.  Good performance now, and it's a simplification to the system in any case.

Issue #1458, opened 2022-10-10
v9 -- equality hierarchy


mscuthbert:
As we've done with Keywords (#1377) make an equality hierarchy for `__eq__` within music21.  Comes out of discussion at #1454 and #1457.

All Music21Objects should define a list of attributes that must compare to `True` in order for the two M21 objects to be equal.  In addition to being the same class.  And then base.Music21Object can define `__eq__` something like:

```
SENTINEL_SELF = object()
SENTINEL_OTHER = object()

def __eq__(self, other):
    if not isinstance(other, type(self)):
        return False  # or NotImplemented?

    for attr in type(self)._getEqualityAttributes():
        if getattr(self, attr, SENTINEL_SELF) != getattr(other, attr, SENTINEL_OTHER):
            return False
    return True

equalityAttributes = ('id', 'groups', 'duration')

@cache
def _getEqualityAttributes(self):
     cls = type(self)
     equalityAttributes = set()
     for klass in [cls, *cls.mro()]:
          equalityAttributes |= set(klass.equalityAttributes)
    return equalityAttributes
```

this way it cannot be the case that two GeneralNotes would be not equal but two Notes with the same GeneralNote parameters would be equal.

Subclasses can define other things that need to be equal beyond just attributes, but I think this would be clearer.

Should also document that Streams do contains based on is not ==, which leads to things like:

>>> n = note.Note('C#4')
>>> m = note.Note('C#4')
>>> n == m
True
>>> s = stream.Stream([n])
>>> m in s
False
>>> m in list(s)
True

A new major release is a good time to do this.

**Intent**

[x] I plan on implementing this myself.
[ ] I am willing to pay to have this feature added.
[ ] I am starting a discussion with the hope that community members will volunteer their time to create this. I understand that individuals work on features of interest to them and that this feature may never be implemented.

mscuthbert:
Fixed in #1466

PR #1466, opened 2022-10-17
Equality changes -- hierarchy

mscuthbert:
* Add an equality hierarchy to music21 to enable easier development (needed for @MarkGotham PRs)
* Add documentation for how equality and hashing work in music21
* Add notes in dev documents about equality and hashing in music21 and why there are mistakes that can't be fixed.
* Before there was a mishmash of which Music21Objects could be hashed.  Aim for all to be able to -- with consequences on devs if they mutate them later.
* Use deque in a few places where pop(0) was being used.  Remove pop(False)


mscuthbert:
Fixes #1458

PR #1467, opened 2022-10-18
All Music21Objects must be hashable and default instantiate

mscuthbert:
All Music21Objects must not crash on `m = MyM21Object()` and `{m: 1}` must be valid.


mscuthbert:
Coverage keeps decreasing as I delete extraneous blank lines.

PR #1477, opened 2022-11-01
horns typically hunt in pairs

MarkGotham:
This small change is needed for a lot of orchestral scores with part names like "Horns 1 and 2 in Eb"


mscuthbert:
i corni sono felici ancora!

PR #1481, opened 2022-11-19
Lowercase some RomanNumeral figures created with integers

### LINKED ISSUE #1450: Major/minor distinction in `roman.RomanNumeral` from degree
**music21 version**

Github, dev.

**Problem summary**

RomanNumeral from degree always return major triads.

**Steps to reproduce**

```
>>> majorKeyObj = key.Key("C")
>>> roman.RomanNumeral(1, majorKeyObj)
<music21.roman.RomanNumeral I in C major>

>>> minorKeyObj = key.Key("c")
>>> roman.RomanNumeral(1, minorKeyObj)
<music21.roman.RomanNumeral I in c minor>
```

The above should return
- `I` in major (upper case)
- `i` in minor (lower case)

Instead, they both return `I` (upper case).

jacobtylerwalls:
Closes #1450

Previously, RomanNumerals created with integer figures were always uppercase, even when `caseMatters` was True.

Now, certain figures are lowercased according to the mode.

***

cc/ @MarkGotham would you be willing to review?

I didn't use structural pattern matching per this comment:
https://github.com/cuthbertLab/music21/blob/d77f4059a6943aa43b0dea388faa2bf2b8257c6c/music21/midi/translate.py#L1281-L1283

### DIFF
--- a/music21/roman.py
+++ b/music21/roman.py
@@ -2268,8 +2268,20 @@ class RomanNumeral(harmony.Harmony):
         self.scaleCardinality: int = 7
 
         if isinstance(figure, int):
-            self.caseMatters = False
             figure = common.toRoman(figure)
+            if self.caseMatters:
+                mode = ''
+                if isinstance(keyOrScale, key.Key):
+                    mode = keyOrScale.mode
+                elif isinstance(keyOrScale, scale.DiatonicScale):
+                    mode = keyOrScale.type
+                elif isinstance(keyOrScale, str) and keyOrScale:
+                    mode = 'major' if keyOrScale[0].isupper() else 'minor'
+                if (
+                    (mode == 'major' and figure in ('II', 'III', 'VI', 'VII'))
+                    or (mode == 'minor' and figure in ('I', 'II', 'IV', 'V'))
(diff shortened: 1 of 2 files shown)


music21/roman.py
```diff
@@ -4389,6 +4401,24 @@ def test_sevenths_on_alteration(self):
         rn = RomanNumeral('bVII7', 'C', seventhMinor=Minor67Default.CAUTIONARY)
         self.assertEqual(rn.seventh.name, 'A')
 
+    def test_int_figure_case_matters(self):
+        '''https://github.com/cuthbertLab/music21/issues/1450'''
+        minorKeyObj = key.Key("c")
```
mscuthbert, line 4406: trying to standardize docstrings as always like:

```
'''
adfahdfkladf
'''
```

even if a one-liner.

The linter should be catching the double quote "c" -- not sure why not here.

mscuthbert (approved):
Looks great to me, good approach.

jacobtylerwalls:
Thanks, Myke. Feel free to unpin `mypy` later, of course!

PR #1482, opened 2022-11-23
Add `__eq__` for ChordStepModification

jacobtylerwalls:
Add a convenience `__eq__` comparison for `ChordStepModification` to avoid having to compare each property manually.


mscuthbert:
thanks!

PR #1485, opened 2022-11-24
Pin flake8 to v5

mscuthbert:
* Flake8 does not work with flake-quotes

* Consolidate checks into one run (why two -- I'll never know)

* install from requirements_dev.txt


jacobtylerwalls:
> Consolidate checks into one run (why two -- I'll never know)

I think the separation allowed having the pylint badge on the readme, but that's not of much value, so I'll just open a PR to strip it.

PR #1479, opened 2022-11-19
Implement new spanner.SpannerAnchor class

gregchapman-dev:
Fixes #1402

Issue #1419, opened 2022-09-10
Start and end notes under `Ottava` spanner exported incorrectly due to `toWrittenPitch()`


NeroBlackstone:
**music21 version**

`7.3.3`   

**Problem summary**

Pitch of some notes incorrect after .musicxml file parsed by  music21.

**Steps to reproduce**

```
from music21 import *
s=converter.parse('1.musicxml')
s.show()
```

**Expected vs. actual behavior**

Original file:
![original xml](https://user-images.githubusercontent.com/17470335/189474510-0684b8ed-b0d8-4852-b0eb-5c19d07b0fa0.png)

After parse:
![after parse](https://user-images.githubusercontent.com/17470335/189474519-8fcacfa2-0342-4a17-b9d2-2df612387c9b.png)

So C5 is incorrectly parsed as C4.

**More information**

musicxml file:
[1.zip](https://github.com/cuthbertLab/music21/files/9539801/1.zip)

OS: linux-lts 5.15.64-1

mscuthbert:
Hi Nero -- what does the MusicXML file output from MuseScore look like in m1 (beginning) and m8 (end)?  There are four possible problems:

1) MuseScore is not outputting MusicXML the way you intend the piece to be interpreted (either user error or MuseScore error; not music21).
2) Music21 is not reading the MusicXML properly (music21 problem -- we would fix); the Stream is incorrect.
3) Music21 is not outputting the MusicXML properly (music21 problem -- we would fix)
4) MuseScore is not reading properly formatted MusicXML properly (MuseScore error; not music21)

Can you isolate what is happening w/ the MusicXML specs and determine which of these 4 errors is happening here?  Otherwise, if you need someone to do this for you you may need to look into paid consulting support.

NeroBlackstone:
Thank you, mscuthbert.

I have read .musicxml file before and after music21 parser.
And I think it's possible music21 bug:

For your question 1, I think muse score generate correct .musicxml file.
For example in measure 1: (Original pitch in .mscz is **B-5 F5 E-6 F6**)
After musescore .musicxml export:

``` xml
<measure number="1" width="207.21">
...
<pitch>
          <step>B</step>
          <alter>-1</alter>
          <octave>5</octave>
</pitch>
....
<pitch>
          <step>F</step>
          <octave>5</octave>
</pitch>
...
<pitch>
          <step>E</step>
          <alter>-1</alter>
          <octave>6</octave>
</pitch>
...
<pitch>
[...]

PR #1486, opened 2022-11-25
Ottava transposition fixes

gregchapman-dev:


gregchapman-dev:
This is just the first few things mentioned in issue #1419.  Lots more work to do, but at this point the notes that are actually in the Ottava are transposed correctly.  I'm now off to make sure _all_ the notes are in the Ottava (and in other Spanners as well).

gregchapman-dev:
Oh, you might notice that I have the SpannerAnchor implementation from PR #1479 included here.  I will continue that work in this PR as well (I think it's done, but I might find some bugs as I work on Ottavas).

mscuthbert:
> Oh, you might notice that I have the SpannerAnchor implementation from PR #1479 included here. I will continue that work in this PR as well (I think it's done, but I might find some bugs as I work on Ottavas).

It'd be better if we can get that one merged first so it's easier to see the separation of the two.  I'll look at it now.

mscuthbert:
Or.......(thinking out loud too later) .... what happens if instead of using the spanner anchor, we check for non-zero Spanner duration in the case of one element-spanners?

gregchapman-dev:
I actually tried that first (on a different branch), but a Spanner with a duration can't reach into subsequent Measures.  You gotta have an anchor (GeneralNote or SpannerAnchor) over in that Measure.

gregchapman-dev:
Interesting test failure: Since I'm no longer checking `s.atSoundingPitch is True` (in m21ToXml.py's ScoreExporter.parse() and PartExporter.parse()) before calling s.toWrittenPitch (because the ottava transposition code in toWrittenPitch needs to run in all cases), _treatAtSoundingPitch is now raising an exception because s.atSoundingPitch is 'unknown' in all contextSites.  This is the case in many tests.  @mscuthbert and @jacobtylerwalls, I'm wondering if _treatAtSoundingPitch should interpret `s.atSoundingPitch == 'unknown'` as "at written pitch" instead of raising an exception, since that's what m21ToXml.py was effectively doing.  I will do that for now, but please let me know if there's something else I should do here.

gregchapman-dev:
Ah, even better, just return 'unknown' from _treatAsSoundingPitch instead of raising an exception, skipping the whole transposition side effect.

gregchapman-dev:
Huh.  I think I worked around the circular imports (but I had to leave the searchStream parameter un-type-hinted).  Hopefully there's a better way.

gregchapman-dev:
@jacobtylerwalls, please take a look at my changes to xmlToM21.py, which fill out any Ottavas during MusicXML parse.  I got stymied a bit by separateOutPartStaves, since I need to search the correct partStaff for the intermediate elements.  I ended up finding the correct partStaff using a potentially very expensive search for ottava.getFirst() in each partStaff.  This is crude, but it works.

Better, I think, would be to modify separateOutPartStaves to (1) put the Ottava in the correct partStaff (it currently puts all spanners in the first partStaff, which is fine for many spanners, but is definitely wrong for Ottavas), and then (2) fill out the Ottava right then while we have our hands on the right partStaff to search for intermediate elements.  I'll take a shot at doing that, but any pointers would be helpful, since separateOutPartStaves is fairly complex.

gregchapman-dev:
From the last checkin comment:

> In xmlToM21.py, let's try filling in all spanners.  Result: mozartTrioK581Excerpt
has two slurs whose startNote isn't in the searchStream.  I'll figure that out shortly.

I figured it out.  mozartTrioK581Excerpt has a slur that starts just before the end of the excerpt in both parts 1 (Clarinet) and 2 (violin 1).  Our MusicXML parser is apparently seeing the next slur start in the document as a stop for that trailing slur.  So we end up with two bad slurs: one that starts at the end of part1 and ends in measure 6 of part 2, and one that starts at the end of part 2 and ends in measure 6 of part 3 (violin 2).  And since the Slur spanner ends up associated with the Part where it ends, we have a start note in the slur that is not in that Part.  I no longer crash (yay) when this happens, but... I think there's a bug in the parser that gets tripped up by this not-that-well-formed-because-it's-an-excerpt MusicXML file. (If I remove the trailing slur starts, all is well.)

gregchapman-dev:
I wrote up a new issue for the xmlToM21.py bug with slurs that start but don't stop (issue #1489).

gregchapman-dev:
Hmmm... After a bunch of testing, I am slowly coming to the conclusion that while the utility "fillIntermediateSpannedElements" might be useful, I really only want to call it on Ottavas in xmlToM21.py:PartParser (and similarly in my own parsers).  I keep running into problems with other spanners.  For example, a Diminuendo that lasts exactly one quarter note (there's one quarter note in the spanner) gets stretched to a dotted-quarter because there's a dotted-quarter note at the same offset in a different Voice in the Measure.  For an Ottava, that's appropriate, because that dotted-quarter falls under the Ottava as well, and may need to be transposed.  But when a spanner is simply about duration, I think filling it is wrong.  Thoughts, @jacobtylerwalls and @mscuthbert?

gregchapman-dev:
Oh, filling an ArpeggioMarkSpanner should also be done (if the parser's file format doesn't already carefully describe all the chords/notes in the arpeggio, and instead only describes the top and bottom chords/notes).  But that's a custom "vertical" fill algorithm that I need to work on now.

gregchapman-dev:
Another commit with another big question for @mscuthbert and @jacobtylerwalls:  Inheriting accidental display options during transposition is something that appears to be handled somewhat inconsistently, and I needed an option to force inheritance (1) at end of Humdrum parse, and (2) when comparing scores in musicdiff, where the visibility of an accidental in one score vs another is considered a difference worth noting (but of course, musicdiff can't compare values of notes in Ottavas until both versions of the Ottava have been transposed into a similar state).

So, in my most recent commit, I added an `inheritAccidentalStatus: bool = False` parameter to most transpose routines, so that I could request this behavior by passing `inheritAccidentalStatus=True`.

I think there is more work to do here, to make the behavior more consistent when the parameter is False, but that is riskier work than belongs here.  Some routines don't inherit, some always do, and there's one that inherits all but `Accidental.displayStatus`.

mscuthbert:
> Hmmm... After a bunch of testing, I am slowly coming to the conclusion that while the utility "fillIntermediateSpannedElements" might be useful, I really only want to call it on Ottavas in xmlToM21.py:PartParser (and similarly in my own parsers). I keep running into problems with other spanners. For example, a Diminuendo that lasts exactly one quarter note (there's one quarter note in the spanner) gets stretched to a dotted-quarter because there's a dotted-quarter note at the same offset in a different Voice in the Measure. For an Ottava, that's appropriate, because that dotted-quarter falls under the Ottava as well, and may need to be transposed. But when a spanner is simply about duration, I think filling it is wrong. Thoughts, @jacobtylerwalls and @mscuthbert?

To avoid this problem, I think that this sounds like something that should be on the individual spanner (and I'm trying not to put more things on the Stream object -- it's too complex) so you might use it in a loop like:

```
for ottava in sc[spanner.Ottava]:
    ottava.fillIntermediateSpannedElements()
```
so it only functions on one type of spanner.

I do think though that the ways to avoid the diminuendo problem are two-fold:  (1) if there is one element in a spanner, fillIntermediateSpannedElements shouldn't do anything.  If we want a routine to do that, it should be called something else.  (2) fillIntermediateSpannedElements should start with the first element and call `.next()` [possibly filtered by the superset of classes of first and last] until it reaches the last element (skipping adding anything that's already in the spanner, in case the stream has been edited.   .next() is smart enough (or should be) to know that for a note in a given voice, notes in other voices will not be found by .next().  

Note that this has the potential of including the entire stream if, say, the last element is not accessible by .next() but that reveals a conceptual problem in the setup of the spanner.

Because elements can be in multiple different hierarchies, starting with activeSite is key.

(I think I need a test that .next() doesn't get side-tracked into different hierarchies...somehow I don't remember ever testing for this).

Sorry that as mentioned earlier, I'm swamped until end of semester (two more weeks) so unlikely to do substantive reviews.  Any help from others is appreciated.

jacobtylerwalls:
>(I think I need a test that .next() doesn't get side-tracked into different hierarchies...somehow I don't remember ever testing for this).

Sounds a bit like #1239!

> Sorry that as mentioned earlier, I'm swamped until end of semester (two more weeks) so unlikely to do substantive reviews. Any help from others is appreciated.

Same here...

gregchapman-dev:
This is ready to go (well, I do need to improve test coverage).  Can I get an initial review, @jacobtylerwalls and @mscuthbert?

gregchapman-dev:
Merged in my SpannerAnchor changes, to try to stay up-to-date.

jacobtylerwalls:
If it helps anyone, here is a view of the changes without the spanner anchor PR:

https://github.com/gregchapman-dev/music21/compare/gregc/spannerAnchor...gregchapman-dev:music21:gregc/ottavaTransposeFix

jacobtylerwalls:
@gregchapman-dev, do you mind presenting the case that doesn't behave as you expect with regard to accidentals?

The two main things to be aware of are that diatonic intervals will preserve displayType; chromatic intervals won't--and that pitches created with ints or otherwise have `spellingIsInferred=True` will reset their displayType. I think with those tools in hand, you should be able to get to the output you want with the existing APIs. If not, I'd be very eager to see a failing test case!

gregchapman-dev:
OK, I got over my "don't EVER change default behavior in code you don't really know" reflexes, and removed all the inheritAccidentalDisplay params, leaving a single routine (Interval._diatonicTransposePitch) that needed fixes to no longer lose accidental display options.  This is a much smaller PR now, and should be far easier to review.

Unfortunately, a single figured bass test is now failing (I believe, due to the changes in Interval._diatonicTransposePitch), and I can't really figure out why.  Some helpful pointers would be appreciated.

gregchapman-dev:
@jacobtylerwalls @mscuthbert Unfortunately, a single figured bass test is now failing (I believe, due to the changes in Interval._diatonicTransposePitch), and I can't really figure out why (I've stepped through the code, but figured bass realization isn't something I understand.  Some helpful pointers would be appreciated!

mscuthbert:
I'm not sure -- it uses figuredBass/examples.py - exampleB which has some odd notation.  Does

```
    s = converter.parse('tinynotation: 4/4 D4 A4_7,5,#3 B-4 F4_6 G4_6 AA4_7,5,#3 D2',
                        makeNotation=False)
    s.insert(0, key.Key('d'))
    fb = realizer.figuredBassFromStream(s)
    fb.generateBassLine().show()
```

does this resemble the output in the docs for figuredBass/examples?

gregchapman-dev:
@mscuthbert Thanks for the pointer.  I can see in the docs what it's supposed to look like.  The failure is that fbRealization1.getNumSolutions() returns 0.  Which seems really wrong.  The only thing I can think of is that my changes to Interval._diatonicTransposePitch are causing it to transpose to an incorrect pitch (this routine gets called a LOT during figured bass realization).  But my fixes should only have changed accidental display options in the results, not changed pitch.

music21/interval.py
```diff
@@ -3443,6 +3443,8 @@ def _diatonicTransposePitch(self,
         PRIVATE METHOD: Return p even if inPlace is True
         '''
         # NOTE: this is a performance critical method
+        from music21.pitch import Accidental
```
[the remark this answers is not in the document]
mscuthbert: Actually, no need to import accidental.  Just set pitch.accidental = 0 to get a natural.  (A little internal easter egg to avoid imports.)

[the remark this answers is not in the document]
mscuthbert: and add `type: ignore` to the line you do it on, since mypy doesn't know about Variant getters/setters.

music21/interval.py
```diff
@@ -3490,10 +3493,40 @@ def _diatonicTransposePitch(self,
...
+                # Make a real Accidental that represents "non-displayed natural"
+                accid2 = Accidental('natural')
+                accid2.displayStatus = False  # implied by pitch2.accidental is None
+            else:
+                accid2 = pitch2.accidental
+
+            accid2.inheritDisplay(accid1)
+
+            if pitch2.accidental is None:
+                # accid2 was a made-up Accidental that represents "non-displayed natural",
```
mscuthbert, line 3516: here's where accid2 might be created only if needed

```
if pitch1.accidental is not None:
    pitch2.accidental = 0
    pitch2.accidental.inheritDisplay(pitch1.accidental)
```

music21/interval.py
```diff
@@ -3490,10 +3493,40 @@ def _diatonicTransposePitch(self,
                 # cannot import Accidental here.
                 pitch2.accidental = halfStepsToFix  # type:ignore
 
-            # inherit accidental display type etc. but not current status
-            if pitch2.accidental is not None and pitch1.accidental is not None:
-                pitch2.accidental.inheritDisplay(pitch1.accidental)
-                pitch2.accidental.displayStatus = None  # set accidental display to None
+            # inherit accidental display options
```
mscuthbert, line 3496: I'm not sure what is happening here that wasn't covered in the previous lines -- this whole set of branches is confusing to me, and there are several possibilities that we're creating accidentals but possibly not using them (which can lead up to a 20% performance slow-down).  What's the particular case that was broken before but works now?

music21/interval.py
```diff
@@ -3490,10 +3493,40 @@ def _diatonicTransposePitch(self,
...
+                # and then it inherited some display options.  If those display options
+                # still say "non-displayed", we can leave pitch2.accidental as None. But
+                # otherwise we need to set accid2 into pitch2.accidental.
+                if accid2.displayStatus is not False:
+                    pitch2.accidental = accid2
+        else:
+            # no halfStepsToFix, so pitch2 is fine as is, but...
+            # We set pitch2.accidental to None, so we might have lost some
+            # display options. So we restore oldPitch2Accidental if that makes sense.
+            if (pitch2.accidental is None
```
mscuthbert, line 3526: pitch2.accidental will always be None here, so this line can be dropped.

But is it true that the displayStatus of the new natural-accidental will always be the same as the old natural-accidental?  That seems like a dangerous assumption.

music21/layout.py
```diff
@@ -442,6 +442,19 @@ def __init__(self,
...
+        searchStream,  # yikes
+        *,
+        includeEndBoundary: bool = False,
+        mustFinishInSpan: bool = False,
+        mustBeginInSpan: bool = True,
+        includeElementsThatEndAtStart: bool = False
+    ):
+        # There is no fill of intermediate elements in a StaffGroup.  All elements must be
+        # chosen carefully by the client.
+        return
```
mscuthbert, line 456: why?  Why can't it fill in [stream.Part] elements?

music21/musicxml/m21ToXml.py
```diff
@@ -3228,6 +3227,24 @@ def mainElementsParse(self):
             # Assumes voices are flat...
             self.parseFlatElements(v, backupAfterwards=backupAfterwards)
 
+    def moveForward(self, byOffset: OffsetQLIn):
```
mscuthbert, line 3230: public method.  Needs docs and tests.

music21/musicxml/m21ToXml.py
```diff
@@ -3228,6 +3227,24 @@ def mainElementsParse(self):
...
+        amountToMoveForward: int = int(round(byOffset * self.currentDivisions))
+        if amountToMoveForward:
+            mxForward = Element('forward')
+            mxDuration = SubElement(mxForward, 'duration')
+            mxDuration.text = str(amountToMoveForward)
+            self.xmlRoot.append(mxForward)
+            self.offsetInMeasure += byOffset
+
+    def moveBackward(self, byOffset: OffsetQLIn):
```
mscuthbert, line 3239: public method.  Needs docs and tests.

music21/musicxml/m21ToXml.py
mscuthbert: But this is NICE!  Thank you

music21/musicxml/m21ToXml.py
```diff
@@ -37,6 +37,7 @@
 from music21 import clef
 from music21 import chord
 from music21 import common
+from music21.common.types import OffsetQLIn
```
mscuthbert, line 40: Should be OffsetQL (not OffsetQLIn)  -- OffsetQLIn can be int, float, Fraction.  OffsetQL is just float, Fraction (the "In" means we're taking In external data and someone might provide an int as a setter, but we always use OffsetQL internally.).

music21/musicxml/m21ToXml.py
```diff
@@ -3265,55 +3280,88 @@ def parseFlatElements(
...
-        # group all objects by offsets and then do a different order than normal sort.
-        # that way chord symbols and other 0-width objects appear before notes as much as
+        # We order things differently if there are any SpannerAnchors in the flat stream.
+        # If there are SpannerAnchors, we do two passes. The first pass skips all
+        # prePostObjectSpanners processing, and then the second pass goes back to the
+        # beginning and does all the prePostObjectSpanners processing.
+        # If there are no SpannerAnchors, we do one pass, doing any prePostObjectSpanners
+        # processing as we process each element.
+
+        hasSpannerAnchors: bool = bool(m[spanner.SpannerAnchor])
```
mscuthbert, line 3290: since this is already a flat stream, `bool(m.getElementsByClass(spanner.SpannerAnchor))` is more efficient.

music21/musicxml/m21ToXml.py
```diff
@@ -3265,55 +3280,88 @@ def parseFlatElements(
 
         self.currentVoiceId = voiceId
 
-        # group all objects by offsets and then do a different order than normal sort.
-        # that way chord symbols and other 0-width objects appear before notes as much as
+        # We order things differently if there are any SpannerAnchors in the flat stream.
+        # If there are SpannerAnchors, we do two passes. The first pass skips all
+        # prePostObjectSpanners processing, and then the second pass goes back to the
```
mscuthbert, line 3285: Ugh.  This phrasing is my fault, I think.  Is there a better phrase than `prePostObjectSpanners` it was okay when it only appeared twice in the code but now that it's all over the place i don't have a clue what that means (same in `skipPrePostObjectSpanners`).  relatedSpanners might be better?  skipRelatedSpanners, etc.

music21/musicxml/m21ToXml.py
```diff
@@ -3265,55 +3280,88 @@ def parseFlatElements(
...
-                self.parseOneElement(n)
+                self.parseOneElement(n, skipPrePostObjectSpanners=hasSpannerAnchors)
+
+        # Second pass: If this flat stream (measure/voice) contains any SpannerAnchors, we
+        # will perform a second pass, emitting any pre/postList for any object that is
+        # functioning as a spanner anchor (the SpannerAnchors, as well as any GeneralNotes
+        # that start or end spanners).
+
+        if hasSpannerAnchors:
+            # return to the beginning of the measure, to emit any SpannerAnchors.
```
mscuthbert, line 3335: at the end of this, should we jump forward to where we were before backing up?  In case there is more to do -- because backupAfterwards setting might suppose that we're at the end of the measure.

music21/musicxml/m21ToXml.py
```diff
@@ -3265,55 +3280,88 @@ def parseFlatElements(
...
+                self.moveBackward(self.offsetInMeasure)
 
         self.currentVoiceId = None
 
-    def parseOneElement(self, obj):
+    def parseOneElement(
+        self,
+        obj,
+        prePostObjectSpannersOnly: bool = False,
+        skipPrePostObjectSpanners: bool = False
```
mscuthbert, line 3363: needs docs about this (public method).  It seems like there's just one attribute here.  Something like  `appendSpanners: AppendSpanners = AppendSpanners.RELATED`  where AppendSpanners is an Enum of RELATED (normal usage), NONE (when appending objects in the context of spannerAnchors), RELATED_ONLY (don't append the object.

music21/musicxml/m21ToXml.py
```diff
@@ -3322,8 +3370,26 @@ def parseOneElement(self, obj):
...
+            if not preList and not postList:
+                return
+
+            for sp in preList:  # directions that precede the element
+                root.append(sp)
+
+            if postList and obj.quarterLength > 0:
+                # gap in stream before spanner stop: create <forward>
+                # Ignore Harmony objects having writeAsChord = False
+                if 'GeneralNote' in obj.classes and getattr(obj, 'writeAsChord', True):
```
mscuthbert, line 3383: why GeneralNote and not Harmony?

music21/musicxml/m21ToXml.py
```diff
@@ -3322,8 +3370,26 @@ def parseOneElement(self, obj):
         self.objectSpannerBundle = self.spannerBundle.getBySpannedElement(obj)
         preList, postList = self.prePostObjectSpanners(obj)
 
-        for sp in preList:  # directions that precede the element
-            root.append(sp)
+        if prePostObjectSpannersOnly:
+            if not preList and not postList:
```
mscuthbert, line 3374: I think I'd like to see this as three stages -- 

(1) write the preList elements unless AppendSpanners.NONE.  
(2) write the object...unless AppendSpanners.RELATED_ONLY in which case move forward.
(3) write the postList elements unless AppendSpanners.NONE

The logic here is now much more confusing to me and requires updating multiple places for any later developments.

music21/musicxml/m21ToXml.py
```diff
@@ -3369,8 +3435,28 @@ def parseOneElement(self, obj):
             if parsedObject is False:
                 environLocal.printDebug(['did not convert object', obj])
 
-        for sp in postList:  # directions that follow the element
-            root.append(sp)
+        if not skipPrePostObjectSpanners:
+            for sp in postList:  # directions that follow the element
+                root.append(sp)
+
+    def _hasPrePostObjectSpanners(self, obj) -> bool:
```
mscuthbert, line 3442: rename to hasRelatedSpanners

music21/musicxml/m21ToXml.py
```diff
@@ -3369,8 +3435,28 @@ def parseOneElement(self, obj):
...
+
+        # this list of spanner classes must match the paramsSet keys in prePostObjectSpanners()
+        for m21spannerClass in ('Ottava', 'DynamicWedge', 'Line'):
+            for thisSpanner in spannerBundle.getByClass(m21spannerClass):
+                if thisSpanner.isFirst(obj) or thisSpanner.isLast(obj):
+                    return True
+
+        return False
 
     def prePostObjectSpanners(self, target):
```
mscuthbert, line 3461: rename to `relatedSpanners(self, target: base.Music21Object) -> tuple[Sequence[Element], Sequence[Element]`  THANKS!

music21/musicxml/xmlToM21.py
```diff
@@ -1571,23 +1571,57 @@ def parse(self):
...
+                self.stream.coreInsert(0, sp)
+            completedSpanners.append(sp)
         # remove from original spanner bundle
-        for sp in rm:
+        for sp in completedSpanners:
             self.spannerBundle.remove(sp)
         # s is the score; adding the part to the score
         self.stream.coreElementsChanged()
 
+        partStaffs: list[stream.PartStaff] = []
```
mscuthbert, line 1589: partStaves -- just for similarity with existing pluralization.

music21/musicxml/xmlToM21.py
```diff
@@ -1571,23 +1571,57 @@ def parse(self):
...
+            spannerPart: stream.Part | None = None
+            if partStaffs:
+                spannerPart = self.findFirstPartContaining(sp.getFirst(), partStaffs)
+            else:
+                spannerPart = self.stream
+
+            if spannerPart:
+                spannerPart.coreInsert(0, sp)
+                spannerPart.coreElementsChanged()
+                sp.fillIntermediateSpannedElements(spannerPart)
```
mscuthbert, line 1610: move to separate method.  `self.replaceOttavas(completedSpanners, partStaves)` -- when a large part of a generalized function deals with one type of element, it makes reading much easier to see what to skip (and lets me skip over a while method call when debugging something unrelated to Ottavas.

music21/musicxml/xmlToM21.py
```diff
@@ -1571,23 +1571,57 @@ def parse(self):
...
+        obj: base.Music21Object,
+        parts: list[stream.Part]
+    ) -> stream.Part | None:
+        # This can be quite slow if the object is a long way into the first partstaff
+        # or in a second (or third?) partstaff.  I hope for a faster alternative.
+        for part in parts:
+            for mObj in part.recurse().getElementsByClass(obj.classes[0]):
+                if mObj is obj:
+                    return part
+        return None
```
mscuthbert, line 1623: This should use `containerInHierarchy()` repeatedly until a Part is found.

mscuthbert (commented):
Reviewed up through the musicxml directory with comments there.  No bugs found, but some ineffiicencies.  Huge amount of work!

music21/musicxml/m21ToXml.py
jacobtylerwalls: I was the last one to touch this; Greg is just duplicating it.

GeneralNote is the 99.9% case here. We're trying to _ignore_ any Harmony objects that have writeAsChord = False, hence the three-argument getattr() to ensure the attribute is either missing or True.

(I get that three-argument getattr can be a "buh?" though...)

music21/stream/base.py
```diff
@@ -5371,14 +5371,13 @@ def toSoundingPitch(self, *, inPlace=False):
         'C3'
 
         If 'atSoundingPitch' is unknown for this Stream and all of its parent Streams
-        then will raise a StreamException:
+        then no transposition will take place (this used to raise an exception):
 
         >>> s = stream.Score()
         >>> p = stream.Part(id='partEmpty')
         >>> s.append(p)
```
mscuthbert, line 5378: Oh, while we're here, can we change this to `s.insert(0.0, p)` -- trying to get people not to get in the habit of appending Part objects to Scores (makes no difference here, but just good form)

music21/stream/base.py
```diff
@@ -5371,14 +5371,13 @@ def toSoundingPitch(self, *, inPlace=False):
...
         >>> p.toSoundingPitch()
-        Traceback (most recent call last):
-        music21.exceptions21.StreamException: atSoundingPitch is unknown: cannot transpose
+        <music21.stream.Part partEmpty>
```
mscuthbert, line 5380: unfortunately, now it's unclear what this call is demonstrating.  

need to see what p's atSoundingPitch was before ('unknown') since that's never displayed, and what happens when 'unknown' stream calls toSoundingPitch() -- and that can no longer be demonstrated with an empty stream.

I'm -1 on the change to the status of 'unknown' streams because of this.

music21/stream/base.py
```diff
@@ -5448,6 +5458,7 @@ def toWrittenPitch(self, *, inPlace=False):
         * Changed in v5 returns `None` if `inPlace=True`
         '''
         from music21 import spanner
+        from music21 import stream
```
mscuthbert, line 5461: not needed.  You can just do `assert isinstance(partLike, Stream)` within stream/base.py.

music21/stream/tests.py
```diff
@@ -4575,18 +4575,17 @@ def collectAccidentalDisplayStatus(s_inner):
 
         # mStream.show()
 
-        # after transposition all accidentals are reset
-        # note: last d# is not showing in Finale, but this seems to be a
-        # finale error, as the musicxml is the same in all D# cases
+        # after diatonic transposition no accidentals are reset
```
mscuthbert, line 4578: I'm still not convinced that we can keep displayStatus after a diatonic transposition:

```
>>> n = note.Note('F')
>>> n.pitch.accidental = pitch.Accidental('natural')
>>> m = stream.Measure([n, note.Note('F#')])
>>> m.makeNotation(inPlace=True)
>>> n.pitch.accidental.displayStatus
False
>>> m2 = m.transpose('A1')
>>> m2[note.Note][0]
<music21.note.Note F#>
>>> print(m2[note.Note][0].pitch.accidental.displayStatus)
None
```

it would be a major mistake to keep that as False after the transposition.

music21/stream/tests.py
```diff
@@ -4575,18 +4575,17 @@ def collectAccidentalDisplayStatus(s_inner):
...
-                          ('G#', None), ('C#', None), ('C#', None), ('C#', None), ('C#', None),
-                          ('D#', None), ('D#', None), ('C#', None), ('D#', None), 'x', 'x',
-                          ('F#', None), ('F#', None), ('G#', None), ('G#', None), ('C#', None),
-                          ('C#', None), ('D#', None), ('D#', None), 'x', 'x', 'x', 'x', 'x', 'x',
-                          ('F#', None), ('F#', None), ('B#', None), ('B#', None),
-                          ('C#', None), ('C#', None)]
+                         ['x', ('G#', False), ('G#', False), 'x', 'x', 'x', 'x',
+                          ('B#', True), ('B#', False), ('C#', False), ('F#', False), ('G#', False),
+                          ('G#', False), ('C#', False), ('C#', False), ('C#', False), ('C#', False),
+                          ('D#', False), ('D#', False), ('C#', False), ('D#', False), 'x', 'x',
```
mscuthbert, line 4584: If, for instance, we chose to transpose just the measure (and not the original key signature of 3 sharps also), the D# False would be incorrect.  There is no D# in the key signature.

mscuthbert (changes requested):
I've finished the review.  (I really wish we had broken this into 4 PRs: one for SpannerAnchor, one for Fill Spanners, one for ottava, and one for changes in accidental making -- but that's water under the bridge now, I suppose).  Making as request changes because there are some potential bugs with inheriting displayStatus that I don't think have been resolved.  

Also lacks additional tests for some of the most delicate things that have been changed.

(One thing I didn't mark: why do Spanners that don't fill anything need to override the fill method?  Can't they just mark fillElementTypes = [] and the base spanner fillIntermediateSpannedElements (or fill() ) check that if fillElementTypes is empty it should just return immediately.

music21/stream/tests.py
gregchapman-dev: Hmmm... This is distressing.  I was imagining that diatonic transpositions were all about changing _key_.  But as you say, the key doesn't necessarily get transposed.  But I really need to be able to transpose to written or sounding pitch (for transposing instruments as well as for ottavas) while maintaining displayStatus, and I believe that should be valid for those cases.  Maybe I need to re-introduce inheritAccidentalDisplay: bool as a parameter, so that those who know they are changing key (or transposing by octaves) can set it True and get the benefit.

jacobtylerwalls: Are you sure you don't want to be manipulating `displayType` rather than `displayStatus`? Status is quasi-private.

mscuthbert: What @jacobtylerwalls is exactly what I came to say.  `displayType` determines the rules about when the accidental should be displayed.  'always' (or 'even-tied') is the equivalent of don't hide this accidental ever.  'never' is don't show this accidental ever.  If `displayStatus` is None, then it will recompute it using `displayType` and context.  For `always` and `never` that recomputation is very fast.

mscuthbert:
Sorry that I didn't see the previous discussion of `unknown` and `atSoundingPitch` before commenting here.

I think perhaps--thinking more--we should leave unknown as it was before but set a default that `atSoundingPitch` = True -- streams by default store pitches at sounding transposition (unrelated to ottava), and if people want to store streams at writtenPitch they will need to set it manually (or if it's unknown, manually setting it to `unknown` is still possible -- damn musicologists.  :-D  )

gregchapman-dev:
I do understand this, I think.

music21’s MusicXML parser (and my Humdrum and MEI parsers) computes displayStatus (not displayType) from the parsed file itself.  This is appropriate, since the parsed file knows exactly  what accidentals are printed.

Transposition for Parts with transposing instruments, and for Ottavas, is important to file writers (that might need to encode as sounding or written pitch) and to tools like musicdiff, that are trying to compare written notation.

So, doing those key-change and octave-change transpositions must not lose that parsed info about whether the accidental is printed (that’s displayStatus, not displayType).

mscuthbert:
> I do understand this, I think. music21’s MusicXML parser (and my Humdrum and MEI parsers) computes displayStatus (not displayType) from the parsed file itself. This is appropriate, since the parsed file knows exactly what accidentals are printed. Transposition for Parts with transposing instruments, and for Ottavas, is important to file writers (that might need to encode as sounding or written pitch) and to tools like musicdiff, that are trying to compare written notation. So, doing those key-change and octave-change transpositions must not lose that parsed info about whether the accidental is printed (that’s displayStatus, not displayType).

It sounds like something that a particular routine might do -- like toWrittenPitch or something, but not the general transpose method.

gregchapman-dev:
Proposal for `accid.displayStatus` (which I am working on now): add a new `treatAsKeyChange: bool = False` parameter to various transpose routines (to be determined, but I will try to minimize), which if set True (for example, by `s.toWrittenPitch`) causes all `displayStatus`es to be maintained during diatonic transpositions.  If set False, the behavior would be like it was before this PR, with the exception that if the diatonic interval is an octave (or multiple octaves), `displayStatus` will be maintained (it's not a key change, but the `displayStatus` values are known to be good after the transposition).

music21/interval.py
```diff
@@ -3490,10 +3493,40 @@ def _diatonicTransposePitch(self,
...
+                accid1 = Accidental('natural')
+                accid1.displayStatus = False  # implied by pitch1.accidental is None
+            else:
+                accid1 = pitch1.accidental
+
+            accid2: Accidental
+            if pitch2.accidental is None:
+                # Make a real Accidental that represents "non-displayed natural"
+                accid2 = Accidental('natural')
+                accid2.displayStatus = False  # implied by pitch2.accidental is None
```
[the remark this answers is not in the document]
gregchapman-dev: done.

music21/interval.py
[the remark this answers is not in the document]
gregchapman-dev: done.

music21/interval.py
```diff
@@ -3490,10 +3493,40 @@ def _diatonicTransposePitch(self,
                 # cannot import Accidental here.
```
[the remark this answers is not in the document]
gregchapman-dev: done.

music21/interval.py
gregchapman-dev: I rewrote this code to be cleaner (I'll do the "make it optional" thing in a bit).

music21/interval.py
gregchapman-dev: Rewritten.

music21/interval.py
gregchapman-dev: displayStatus inheritance now only happens if `treatAsKeyChange` is True or the Interval is a unison or octave shift.  No-one yet sets `treatAsKeyChange` to True.

gregchapman-dev: And now both s.toWrittenPitch and s.toSoundingPitch will end up with treatAsKeyChange True, due to changes in s._transposeByInstrument.  Tests have been modified to test both cases (treatAsKeyChange set and not set).

music21/stream/tests.py
gregchapman-dev: Latest commit answers this (with `treatAsKeyChange` parameter, set appropriately in `s._transposeByInstrument`).

gregchapman-dev:
Proposal above (`treatAsKeyChange: bool = False`) has been pushed.  Enjoy!  I'm off to handle all the rest of these review comments.

music21/stream/tests.py
gregchapman-dev: Handled by `treatAsKeyChange: bool = False` proposal, and the test has been updated to test both values of that parameter.

music21/layout.py
gregchapman-dev: I guess it could, going by order of Parts in the Score.  But in any case, I have no implementation for this yet, so I'm leaving StaffGroup in the (new) default state, which is that `fill()` does nothing because `self.fillElementTypes` is left empty.

music21/spanner.py
```diff
@@ -209,6 +212,7 @@ def __init__(self,
                  *spannedElements: t.Union[base.Music21Object,
                                            Sequence[base.Music21Object]],
                  **keywords):
+        from music21 import note
```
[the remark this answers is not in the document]
gregchapman-dev: Done (both module import and a better/faster implementation of findFirstPartContaining).

music21/spanner.py
```diff
@@ -528,6 +547,82 @@ def replaceSpannedElement(self, old, new) -> None:
         # environLocal.printDebug(['replaceSpannedElement()', 'id(old)', id(old),
         #    'id(new)', id(new)])
 
+    def fillIntermediateSpannedElements(
+        self,
+        searchStream,  # yikes
```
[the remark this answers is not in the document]
gregchapman-dev: Replaced "yikes" comment, and renamed fillIntermediateSpannedElements -> fill.

music21/spanner.py
```diff
@@ -528,6 +547,82 @@ def replaceSpannedElement(self, old, new) -> None:
...
+        *,
+        includeEndBoundary: bool = False,
+        mustFinishInSpan: bool = False,
+        mustBeginInSpan: bool = True,
+        includeElementsThatEndAtStart: bool = False
+    ):
+        if t.TYPE_CHECKING:
+            from music21 import stream
+            assert isinstance(searchStream, stream.Stream)
```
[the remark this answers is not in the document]
gregchapman-dev: I can't import stream at the top level; the following happens:
    
class DynamicWedge(spanner.Spanner):
AttributeError: partially initialized module 'music21.spanner' has no attribute 'Spanner' (most likely due to a circular import)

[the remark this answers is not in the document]
gregchapman-dev: I allow searchStream to be None now, using self.getFirst().activeSite if available, raising an exception if no activeSite.

music21/spanner.py
```diff
@@ -528,6 +547,82 @@ def replaceSpannedElement(self, old, new) -> None:
...
+            endOffsetInHierarchy = (
+                self.getLast().getOffsetInHierarchy(searchStream) + self.getLast().quarterLength
+            )
+
+        for foundElement in (searchStream
+                .recurse()
+                .getElementsByOffsetInHierarchy(
+                    startOffsetInHierarchy,
+                    endOffsetInHierarchy,
+                    includeEndBoundary=includeEndBoundary,
```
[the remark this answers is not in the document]
gregchapman-dev: I decided to leave all the options here, so as not to be different from getElementsByOffsetInHierarchy (so people can think about them the same way).  I did default them differently, though, for my own needs.

If you really want me to remove some, let me know.

music21/spanner.py
```diff
@@ -1128,9 +1301,11 @@ class Slur(Spanner):
     '''
 
     def __init__(self, *spannedElements, **keywords):
+        from music21 import note
```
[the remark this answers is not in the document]
gregchapman-dev: In the latest push, Ottavas are the only Spanner that actually can be filled, because it's the only Spanner that we need to fill inside of music21 (and in my music21 parsers) to get Ottava transposition working.  My goal here is to get all this in before v9 is released, so I'm trying to lop off unused stuff that might take too long to get right.  Please let me know if there are particular other-than-Ottava Spanners you want fillable in v9.

music21/spanner.py
```diff
@@ -1166,7 +1341,9 @@ class MultiMeasureRest(Spanner):
     }
 
     def __init__(self, *spannedElements, **keywords):
+        from music21 import note
         super().__init__(*spannedElements, **keywords)
+        self.fillElementTypes = [note.Rest]
```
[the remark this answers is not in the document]
gregchapman-dev: Yeah, if only it actually worked... In the latest push, Ottavas are the only Spanner that actually can be filled, because it's the only Spanner that we need to fill inside of music21 (and in my music21 parsers) to get Ottava transposition working. My goal here is to get all this in before v9 is released, so I'm trying to lop off unused stuff that might take too long to get right. Please let me know if there are particular other-than-Ottava Spanners you want fillable in v9.

music21/musicxml/m21ToXml.py
gregchapman-dev: I'm not sure.  In the emitted MusicXML we _are_ at the end of the measure (although, self.offsetInMeasure might be at an offset before the furthest offset in the measure, if the spanners didn't reach the end of the measure).  Does that matter?  Is it worth adding yet another `<forward>` tag to the emitted MusicXML for that?  Hmmm... looking at mainElementsParse, it looks like the only thing that might end up at the wrong offset is the rightBarline.  Do you need a `<forward>` before a barline to get the barline to be at the right offset?

gregchapman-dev: But that rightBarline error might also happen if a second voice is shorter than a first voice.  So we're not making anything worse????

music21/musicxml/m21ToXml.py
gregchapman-dev: @jacobtylerwalls I'm still actually not sure I understand what you were trying to accomplish with the Harmony check in the "emit the note" case.  It looks like you skip updating self.offsetInMeasure for certain Harmony objects, but you still emit them?

music21/musicxml/m21ToXml.py
gregchapman-dev: Added docs.  Tests still to come.

music21/musicxml/m21ToXml.py
gregchapman-dev: Added docs.  Tests still to come.

music21/musicxml/m21ToXml.py
gregchapman-dev: Done.

music21/musicxml/m21ToXml.py
gregchapman-dev: done

music21/musicxml/m21ToXml.py
gregchapman-dev: done

music21/musicxml/m21ToXml.py
gregchapman-dev: done.  I called them NORMAL, RELATED_ONLY, and NONE.

music21/musicxml/m21ToXml.py
gregchapman-dev: Yes!  Much better.

music21/musicxml/m21ToXml.py
gregchapman-dev: done

music21/musicxml/m21ToXml.py
gregchapman-dev: done

music21/musicxml/xmlToM21.py
gregchapman-dev: done.

music21/musicxml/xmlToM21.py

music21/musicxml/xmlToM21.py
gregchapman-dev: Done, but please review the code, since I ended up taking a short-cut that I _think_ will always work.

music21/spanner.py
```diff
@@ -528,6 +547,82 @@ def replaceSpannedElement(self, old, new) -> None:
...
+            return
+
+        if self.getFirst() is None:
+            # no spanned elements?  Nothing to fill.
+            return
+
+        endElement: base.Music21Object | None = None
+        if len(self) > 1:
+            # Start and end elements are different, we can't just append everything, we need
+            # to save off the end element, remove it, add everything, then add the end element
```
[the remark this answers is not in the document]
gregchapman-dev: done.

music21/spanner.py
```diff
@@ -528,6 +547,82 @@ def replaceSpannedElement(self, old, new) -> None:
...
+            # filling, the new intermediate elements will come after the existing ones,
+            # regardless of offset.  But first and last will still be the same two elements
+            # as before, which is the most important thing.
+            endElement = self.getLast()
+            if t.TYPE_CHECKING:
+                assert endElement is not None
+            self.spannerStorage.remove(endElement)
+
+        try:
+            startOffsetInHierarchy: OffsetQL = self.getFirst().getOffsetInHierarchy(searchStream)
```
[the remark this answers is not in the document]
gregchapman-dev: done.

music21/spanner.py
```diff
@@ -528,6 +547,82 @@ def replaceSpannedElement(self, old, new) -> None:
...
+                endOffsetInHierarchy = (
+                    endElement.getOffsetInHierarchy(searchStream) + endElement.quarterLength
+                )
+            except sites.SitesException:
+                # print('end element not in searchStream')
+                self.addSpannedElements(endElement)
+                return
+        else:
+            endOffsetInHierarchy = (
+                self.getLast().getOffsetInHierarchy(searchStream) + self.getLast().quarterLength
```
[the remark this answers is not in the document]
gregchapman-dev: done

music21/spanner.py
```diff
@@ -528,6 +547,82 @@ def replaceSpannedElement(self, old, new) -> None:
...
+            )
+
+        for foundElement in (searchStream
```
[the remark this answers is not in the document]
gregchapman-dev: All fixed.

gregchapman-dev:
Current status: nearly all implementation changes are made (a few review comments are still not resolved above, mostly with questions from me).  Lots of doc and test still to do.

music21/musicxml/m21ToXml.py
gregchapman-dev: I'm currently thinking we should jump forward only if (1) the spanner pass ended up short of the notes pass AND (2) backupAfterwards is False.  We should never jump backward (even if it could happen that the spanner pass was longer duration than the notes pass, which I think it couldn't).

music21/musicxml/test_m21ToXml.py
```diff
@@ -58,6 +58,13 @@ def getET(self, obj, makeNotation=True):
         helpers.indent(mxScore)
         return mxScore
 
+    def stripInnerSpaces(self, txt):
```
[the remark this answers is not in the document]
gregchapman-dev: done

music21/spanner.py
```diff
@@ -610,6 +705,84 @@ class _SpannerRef(t.TypedDict):
...
+        {0.0} <music21.stream.Measure 0 offset=0.0>
+            {0.0} <music21.stream.Voice 0x...>
+                {0.0} <music21.note.Note C>
+                {2.0} <music21.spanner.SpannerAnchor object at 0x...>
+                {4.0} <music21.spanner.SpannerAnchor object at 0x...>
+    {4.0} <music21.dynamics.Crescendo <music21.note.Note C><music21.spanner.SpannerAnchor...>>
+    {4.0} <music21.dynamics.Diminuendo <...SpannerAnchor...><...SpannerAnchor...>>
+
+    SpannerAnchors aways have a duration of 0, and if any attempt is made to
+    change this, TypeError will be raised.
```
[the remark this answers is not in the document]
gregchapman-dev: Hmmm... I think of a SpannerAnchor as anchoring one end of the spanner, never both.  However, your point about things like augmentOrDiminish is compelling.  I will allow SpannerAnchors to have duration, and if the SpannerAnchor is the last (or only) element in a Spanner, then m21ToXml.py will need to compute the spanner end offset as anchor.offset + anchor.duration.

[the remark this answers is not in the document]
gregchapman-dev: Looks like m21ToXML.py already gets it right by moving forward when it skips the element (because that element might be a SpannerAnchor, or it might be a Note already emitted in a previous pass):

```
        else:  # appendSpanners == AppendSpanners.RELATED_ONLY
            # skip the element itself, just move forward by element duration
            if postList and obj.quarterLength > 0:
```

music21/spanner.py
```diff
@@ -610,6 +705,84 @@ class _SpannerRef(t.TypedDict):
     spanner: 'Spanner'
     className: str
 
+class SpannerAnchor(base.Music21Object):
+    '''
+    A simple Music21Object that can be used to define the beginning or end
+    of a Spanner, in the place of a GeneralNote.
+
+    This is useful for (e.g.) a Crescendo that ends partway through a
+    note (e.g. in a violin part).
```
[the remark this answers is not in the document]
gregchapman-dev: done.

music21/spanner.py
```diff
@@ -610,6 +705,84 @@ class _SpannerRef(t.TypedDict):
...
+    Here's an example of a whole note that has a Crescendo for the first
+    half of the note, and a Diminuendo for the second half of the note.
+
+    >>> score = stream.Score()
+    >>> part = stream.Part()
+    >>> score.insert(0, part)
+    >>> measure = stream.Measure()
+    >>> part.insert(0, measure)
+    >>> voice = stream.Voice()
+    >>> measure.insert(0, voice)
```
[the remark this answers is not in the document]
gregchapman-dev: done.

music21/spanner.py
```diff
@@ -610,6 +705,84 @@ class _SpannerRef(t.TypedDict):
...
+
+    >>> n = note.Note('C', quarterLength = 4)
```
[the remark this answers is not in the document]
gregchapman-dev: done

music21/spanner.py
```diff
@@ -610,6 +705,84 @@ class _SpannerRef(t.TypedDict):
...
+    >>> sa1 = spanner.SpannerAnchor()
+    >>> sa2 = spanner.SpannerAnchor()
+    >>> voice.insert(0, n)
+    >>> voice.insert(2, sa1)
+    >>> voice.insert(4, sa2)
```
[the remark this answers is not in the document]
gregchapman-dev: done

music21/spanner.py
```diff
@@ -610,6 +705,84 @@ class _SpannerRef(t.TypedDict):
...
+    >>> cresc = dynamics.Crescendo(n, sa1)   # cresc from n to sa1
+    >>> dim = dynamics.Diminuendo(sa1, sa2)  # dim from sa1 to sa2
```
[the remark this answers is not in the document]
gregchapman-dev: done.

music21/spanner.py
```diff
@@ -610,6 +705,84 @@ class _SpannerRef(t.TypedDict):
...
+    >>> part = stream.Part()
+    >>> score.insert(0, part)
+    >>> measure = stream.Measure()
+    >>> part.insert(0, measure)
+    >>> voice = stream.Voice()
+    >>> measure.insert(0, voice)
+
+    >>> n = note.Note('C', quarterLength = 4)
+    >>> sa1 = spanner.SpannerAnchor()
+    >>> sa2 = spanner.SpannerAnchor()
```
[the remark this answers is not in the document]
gregchapman-dev: done.

music21/spanner.py
```diff
@@ -610,6 +705,84 @@ class _SpannerRef(t.TypedDict):
...
+    >>> cresc = dynamics.Crescendo(n, sa1)   # cresc from n to sa1
+    >>> dim = dynamics.Diminuendo(sa1, sa2)  # dim from sa1 to sa2
+    >>> score.append((cresc, dim))
+    >>> score.show('text')
+    {0.0} <music21.stream.Part 0x...>
+        {0.0} <music21.stream.Measure 0 offset=0.0>
+            {0.0} <music21.stream.Voice 0x...>
+                {0.0} <music21.note.Note C>
+                {2.0} <music21.spanner.SpannerAnchor object at 0x...>
+                {4.0} <music21.spanner.SpannerAnchor object at 0x...>
```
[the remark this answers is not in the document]
gregchapman-dev: done.

music21/spanner.py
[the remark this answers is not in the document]
gregchapman-dev: done.

gregchapman-dev:
> Sorry that I didn't see the previous discussion of unknown and atSoundingPitch before commenting here. 
I think perhaps--thinking more--we should leave unknown as it was before but set a default that atSoundingPitch = True -- streams by default store pitches at sounding transposition (unrelated to ottava), and if people want to store streams at writtenPitch they will need to set it manually (or if it's unknown, manually setting it to unknown is still possible -- damn musicologists. :-D )

I will attempt this right now.  In `Stream.__init__` I will set `self._atSoundingPitch` to True, and in `Stream._treatAtSoundingPitch` I'll `raise StreamException` if you try to transpose a stream whose `self._atSoundingPitch` has been set to 'unknown' (presumably by a damn musicologist. :-D )

gregchapman-dev:
Crap, for the `self._atSoundingPitch = True` change, I'm going to have to regenerate the corpus (because all those default 'unknown's are stashed there).  Sigh...

gregchapman-dev:
Oh dear, this isn't going to work.  By defaulting atSoundingPitch to True, we've lost the ability to set atSoundingPitch to False on the Score (only), and have every substream also be implied to be at written pitch (because they are, every one of them, explicitly set to be at sounding pitch).  This causes several tests to fail (notably the tests for toSoundingPitch).  I can fix the tests by setting atSoundingPitch to False on every stream (not just the Score), but that's not reasonable for other users.

I think I cannot make this change right now.

music21/stream/base.py
```diff
@@ -5371,14 +5371,13 @@ def toSoundingPitch(self, *, inPlace=False):
         'C3'
 
         If 'atSoundingPitch' is unknown for this Stream and all of its parent Streams
-        then will raise a StreamException:
+        then no transposition will take place (this used to raise an exception):
```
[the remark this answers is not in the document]
gregchapman-dev: done.

music21/stream/base.py
gregchapman-dev: I have modified the test a bit to show that p.atSoundingPitch is still 'unknown' after the call to toSoundingPitch.

music21/stream/base.py
gregchapman-dev: done.

music21/stream/base.py
gregchapman-dev: done.

music21/musicxml/m21ToXml.py
gregchapman-dev: done.

music21/musicxml/m21ToXml.py
gregchapman-dev: Tests added.

music21/musicxml/m21ToXml.py
gregchapman-dev: Tests added.

gregchapman-dev:
OK, I'm all done except:
1. There's still an open question about Harmony object handling.
2. Ottava is the only Spanner that is fillable for now.  OK?  I'll be happy to add a bunch more later,  but I don't want this PR to miss v9.
3. I need final approval of the new handling of `s.atSoundingPitch == 'unknown'`, since I couldn't make the requested change to set atSoundingPitch to True in `Stream.__init__`.  I removed the exception because I unconditionally call toWrittenPitch during MusicXML export now (to get the Ottavas right, even if atSoundingPitch is 'unknown').
4. I still need to get better test coverage.

music21/musicxml/m21ToXml.py
jacobtylerwalls: Yes, if writeAsChord is False, they need to be emitted (I don't know offhand which routine emits the proper tags for them), but they shouldn't take up any time (since they're not being written-as-chord, i.e. as `<note>` tags.)

mscuthbert:
Merging master (including SpannerAnchor PR merged)

mscuthbert:
> 2. Ottava is the only Spanner that is fillable for now.  OK?  I'll be happy to add a bunch more later,  but I don't want this PR to miss v9.

That's okay -- we can start with one and move forward later.

> 3. I need final approval of the new handling of `s.atSoundingPitch == 'unknown'`, since I couldn't make the requested change to set atSoundingPitch to True in `Stream.__init__`.  I removed the exception because I unconditionally call toWrittenPitch during MusicXML export now (to get the Ottavas right, even if atSoundingPitch is 'unknown').

Ah -- this is something I didn't think of.  We could possibly have something like:

`Stream.atSoundingPitch = SoundingPitch.INHERIT_SOUNDING`

where we make an Enum like:

```
class SoundingPitch(Enum):
     SOUNDING: True
     TRANSPOSED: False
     UNKNOWN: 'unknown'
     INHERIT_SOUNDING: 'inherit_sounding'  # inherit from activeSite but default to SOUNDING
     INHERIT_WRITTEN: 'inherit_written'  # inherit from activeSite but default to WRITTEN
```

But this sounds like another PR -- so we could make this PR make no changes to SoundingPitch right now -- this PR only works where atSoundingPitch() has been set to a particular setting (False?) and then quickly get a second PR for that one in. 


> 4. I still need to get better test coverage.

Thanks -- Appreciate that.

gregchapman-dev:
> so we could make this PR make no changes to SoundingPitch right now -- this PR only works where atSoundingPitch() has been set to a particular setting (False?) and then quickly get a second PR for that one in.

This PR has to remove the exception for transpose of atSoundingPitch == 'unknown'.  m21ToXml.py calls toWrittenPitch(ottavasToSounding=True) without first checking `atSoundingPitch is True` in this PR, because the ottava transposition still needs to happen.  So it crashes on a lot of scores if I don't remove that exception.

gregchapman-dev:
Merge from master has exposed/introduced a test failure.  Investigating (along with some necessary linting).

gregchapman-dev:
@mscuthbert, the testSieveScaleA failure seems to be caused by this PR specifically, but I can't figure it out.  Help?  I don't know what "sieve" means in this context.

mscuthbert:
> Proposal above (`treatAsKeyChange: bool = False`) has been pushed. Enjoy! I'm off to handle all the rest of these review comments.

Great work, but I'm concerned about how "treatAsKeyChange" is appearing on every transposition method throughout the system -- I thought that something like this would be very closely constrained.  I'm still not sure why displayType cannot be used for the same purpose without needing to be put throughout the code:

```
def transposeAndRetainDisplayStatus(s_orig, intv):
     s = copy.deepcopy(s_orig)
     s.makeNotation(inPlace=True)  # or just makeAccidentals
     displays = {}
     for p in s.pitches:
           if not p.accidental:
               p.accidental = pitch.Accidental(0)
           displays[id(p)] = p.accidental.displayType
           if p.accidental.displayStatus:
               p.accidental.displayType = 'always'
           else:
               p.accidental.displayType = 'never'
     s.transpose(intv, inPlace=True)
     s.makeNotation(inPlace=True)  # or just makeAccidentals
     for p in s.pitches:
           p.accidental.displayType = displays[id(p)]
     return s
```

or something like that.

gregchapman-dev:
Parsers (including the MusicXML parser, see xmlToPitch and both of mine) set displayStatus, not displayType.  This is as it should be: many file formats (MEI in particular) are very clear about the visibility of accidentals, so the parser shouldn't also ask makeAccidentals' opinion.  That's also why my OMR clients always set makeNotation=False during write/show.  It's better to crash during write (and figure out what went wrong) than to unknowingly write out a score that has a visible accidental (or other notation) that wasn't in the score the OMR system scanned and interpreted.

So I really think we need `treatAsKeyChange: bool`. I tried my best to minimize where it would get added (I started with the routine that actually used that parameter to behave differently, and then started at toWrittenPitch and toSoundingPitch, found the common call below them, and only added treatAsKeyChange when needed, all the way down.

gregchapman-dev:
Looked at your `transposeAndRetainDisplayStatus` routine a bit closer.  There is something there that might help:

If _transposeByInstrument can save off all the displayStatuses for all the pitches in the (recursive) Score, and then restore them all at the end of _transposeByInstrument, then we wouldn't have to pass treatAsKeyChange around.  _transposeByInstrument knows if it's a key change already, and does all the work.  Can this really work?

Interval._diatonicTransposePitch (where the rubber meets the road) would still need to maintain displayStatus if the interval doesn't change pitch class (Ottavas, Unisons, etc), but not for any other case.

gregchapman-dev:
I think that will work!  I've written the code changes for _transposeByInstrument (which is called by both toWrittenPitch and toSoundingPitch), and it's pretty simple.  I'm going to finish that up (removing treatAsKeyChange everywhere) and push it for your approval.

gregchapman-dev:
There ya go!  I like this a lot better.

gregchapman-dev:
I fixed the failing SieveScale test by making sure a Unison/Octave doesn't have fractional semitones.  This only affected displayStatus, but I still don't know why that caused the test to fail.  Whatever, I fixed a real bug, and the test stopped failing.

gregchapman-dev:
Back to more tests.

gregchapman-dev:
I think I've got this all done now.  Test coverage is good, too.

mscuthbert:
> I think that will work! I've written the code changes for _transposeByInstrument (which is called by both toWrittenPitch and toSoundingPitch), and it's pretty simple. I'm going to finish that up (removing treatAsKeyChange everywhere) and push it for your approval.

I'm still not sure that this takes every potential place where displayStatus could be wrong after a diatonic transposition.  I mention in the review transposition on an atonal score, but also transposition of a mid-measure instrument or key change.  It seems like too much to trust when there's really only one place (makeAccidentals) which has been fire-tested against hiding an accidental that needs to be made.

I'm happy to have a keyword to `toSoundingPitch()` and `toWrittenPitch()` of `preserveAccidentalDisplay: bool = False` that gets passed to `_transposeByInstrument()` and if it's True then we run the routines.

Though this also looks like it should be a `contextlib.contextmanager` somewhere so it can be tested separately (see common.classTools.saveAttributes):

```
run_transpose = lambda: focus.transpose(trans, inPlace=True, classFilterList=classFilterList)
if preserveAccidentalDisplay and self.haveAccidentalsBeenMade():
     with saveAccidentalDisplayStatus(focus):
           run_transpose()
else:
     run_transpose()
````

Then it can be documented and tested separately and used elsewhere in the system or by others (your OMR programs?), and its usage in _transposeByInstrument becomes basically self-documenting.

File: music21/stream/base.py [hunk 47]
```diff
@@ -5287,11 +5289,33 @@ def _transposeByInstrument(
             trans = inst.transposition
             if reverse:
                 trans = trans.reverse()
+
+            displayStatuses = {}
+            if transposeKeySignature and displayStatusesAreSet:
```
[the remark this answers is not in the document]
mscuthbert: See notes in "conversation" about this.

mscuthbert (changes requested):
Couple of small things regarding `transpositionChangesPitchClass` (probably not needed) and making preserveDisplayStatus optional in toSounding, etc. otherwise really close.

Thanks for the added tests.

gregchapman-dev:
I'll have some time Friday to work through this latest review.  Thanks!

music21/interval.py
```diff
@@ -3261,7 +3243,9 @@ def isSkip(self) -> bool:
 
     @property
     def transpositionChangesPitchClass(self) -> bool:
-        return self.simpleName != 'P1'
+        if self.simpleName == 'P1' and float(self.semitones) == float(int(self.semitones)):
+            return False
+        return True
```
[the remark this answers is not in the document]
gregchapman-dev: @mscuthbert I thought `.pitchClass` was the same thing as `.name`, so yeah, transpositionChangesPitchName is in fact what I'm looking for.  But how do I check for that?  `self.simpleName == 'P1'` is True for a DiatonicInterval of 0.5 semitones (Oblique Perfect Unison).  That's why I added "and self.semitones is not fractional" to the condition.

music21/interval.py
```diff
@@ -3484,7 +3460,8 @@ def _diatonicTransposePitch(self,
         # NOTE: this is a performance critical method
 
         inheritAccidentalDisplayStatus: bool = False
-        if treatAsKeyChange or not self.transpositionChangesPitchClass:
+        if not self.transpositionChangesPitchClass:
```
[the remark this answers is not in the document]
gregchapman-dev: See above (Oblique Perfect Unison problem).

music21/chord/__init__.py
```diff
@@ -4667,7 +4667,7 @@ def sortFrequencyAscending(self):
         newChord._notes.sort(key=lambda x: x.pitch.frequency)
         return newChord
 
-    def transpose(self, value, *, inPlace=False):
+    def transpose(self, value, *, treatAsKeyChange=False, inPlace=False):
```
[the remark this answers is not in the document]
gregchapman-dev: treatAsKeyChange is gone (in a later commit).

music21/interval.py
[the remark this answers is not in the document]
gregchapman-dev: I've changed the property name to `transpositionChangesPitchName`, but left the "self.semitones is not fractional" check in place for now, since otherwise it gives wrong answers for P1 +/- half-a-semitone.

music21/stream/base.py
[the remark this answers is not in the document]
gregchapman-dev: I have followed your suggestions in the "conversation": makeNotation.py now has a `saveAccidentalDisplayStatus` generator/contextmanager that `_transposeByInstrument` uses if it was passed `preserveAccidentalDisplay=True` (`toWrittenPitch` and `toSoundingPitch` are also passed `preserveAccidentalDisplay`, which they pass to `_transposeByInstrument`).  I couldn't use the `lambda`, though, because pylint complained about using `focus` in the `lambda` inside a loop where `focus` was defined (or something).

I had not heard of contextmanager before; very cool, and fun to learn something new.

music21/interval.py
```diff
@@ -3232,6 +3259,10 @@ def isStep(self) -> bool:
     def isSkip(self) -> bool:
         return self.diatonic.isSkip
 
+    @property
+    def transpositionChangesPitchClass(self) -> bool:
+        return self.simpleName != 'P1'
```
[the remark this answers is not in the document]
gregchapman-dev: I've changed the property name to transpositionChangesPitchName, but left the "self.semitones is not fractional" check in place for now, since otherwise it gives wrong answers for P1 +/- half-a-semitone.

music21/spanner.py
```diff
@@ -2187,6 +2186,67 @@ def testSpannerRepr(self):
...
+        self.assertEqual(len(ott3), 1)
+        self.assertIs(ott3.getFirst(), expectedElements[0])
+
+        # Ottava with start element not in searchStream, end element is
+        expectedElements = [note.Note('F'), m.notes[0]]
+        ott4 = spanner.Ottava(expectedElements)
+        ott4.fill(m)
+        self.assertEqual(len(ott4), 2)
+        for i, el in enumerate(ott4.getSpannedElements()):
+            self.assertIs(el, expectedElements[i])
```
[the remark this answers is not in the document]
gregchapman-dev: Done.

music21/stream/base.py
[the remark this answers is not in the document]
gregchapman-dev: Done.

gregchapman-dev:
I believe I have responded to all review comments now.  The only remaining issue is whether there is a better implementation of `Interval.transpositionChangesPitchName`, currently implemented as:

```
    @property
    def transpositionChangesPitchName(self) -> bool:
        if self.simpleName == 'P1' and float(self.semitones) == float(int(self.semitones)):
            return False
        return True
```

`self.simpleName == 'P1'` on its own does not return the correct `True` value for an Oblique Perfect Unison.

gregchapman-dev:
pylint issues are not mine.  Yet Another Lint-Gets-Pickier-Whenever-Greg-Pushes-A-Big-PR.  🙄

mscuthbert:
Oh, geez -- this looks like it's https://github.com/PyCQA/pylint/issues/7494 -- I wish they wouldn't change something like this on minor releases which are supposed to be backwards compatible.  Will look in morning.  (I don;t think v9 will come out by first day of class.  So we'll get this in before it is and then I'll have the students upgrade...)

mscuthbert:
fixed -- closed and reopened to run again.

mscuthbert:
Ah, I suppose transposing microtones _can_ change the name of the pitch, but it doesn't have to:

```
In [20]: i = interval.Interval('P1')

In [21]: i.chromatic.semitones = 0.01

In [22]: p = pitch.Pitch('C4')

In [23]: for j in range(20):
    ...:     print(p.name)
    ...:     p.transpose(i, inPlace=True)
    ...: 
C
C
C
C
C
C
C
C
C
C
C
C
C
C
C
C
```

though there is something strange happening here:

```
In [24]: i = interval.Interval('P1')

In [25]: i.chromatic.semitones = 0.5

In [27]: p = pitch.Pitch('C4')

In [28]: for j in range(9):
    ...:     print(p.name, p.ps)
    ...:     p.transpose(i, inPlace=True)
    ...: 
C 60.0
C~ 60.5
C# 61.0
C# 61.0
C# 61.0
C# 61.0
C# 61.0
C# 61.0
C# 61.0
```

(why does it stop?

gregchapman-dev:
This is getting deeper than I am happy addressing in this PR.  Perhaps I should get rid of `transpositionChangesPitchName`, since it is so problematic, and simply perform that same check inline when I want to know if a transposition can safely maintain the existing accidental display status (i.e. if the transposition ends up leaving the note as it is, or transposing it by some number of (exact) octaves).

gregchapman-dev:
Latest push does exactly that: inlines the check.  I'm checking whether or not I can safely keep displayStatus.  So any "truly perfect" unison or octave is fine.

gregchapman-dev:
@mscuthbert Any chance of getting this merged soon?

mscuthbert (approved):
Thanks!  All looks good -- now that we have pylint working again.

PR #1490, opened 2022-11-30
Synchronize Measure IDs on Musicxml out

rigaux:
Adding IDs to measures in MusicXML files. I added a call to _synchronizeIds(), just as it works already for other elements (eg, notes).

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -3143,6 +3143,9 @@ class MeasureExporter(XMLExporterBase):
         self.xmlRoot = Element('measure')
         self.currentDivisions = defaults.divisionsPerQuarter
 
+        # Adding ids to measure in the MusicXML export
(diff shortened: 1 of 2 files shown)


mscuthbert:
Can you explain what this change does and add a test to show what the changed output is.  THanks!

(Also a small flake issue)

mscuthbert:
Pinging @rigaux

rigaux:
Sorry for the late update. So: the goal is to export measure IDs to MusicXML. Our group annotates music score, and we need to refer to measures.

At the end, a measure element features an id attribute, e.g., <measure id="m1" number="1">

I also added a test 'testExportIdOfMeasure ' in musicxml/test_m21ToXxml.py. It checks the existence of the 'id' attribute.

mscuthbert:
just fixed the flakes myself.  Will merge if passes.

PR #1225, opened 2022-02-18
Fix MIDI export of meter changes

jacobtylerwalls:
Fixes #1224

Thanks @saintmatthieu for the report.

PR #1492, opened 2022-12-03
Remove unnecessary coreElementsChanged() call

jacobtylerwalls:
The associated _coreInsert call was removed in #1225 without also removing coreElementsChanged() afterward.


mscuthbert:
Closing and reopening to get the mypy fixed.

PR #1493, opened 2022-12-03
Translate "implicit" attribute of MusicXML measures

jacobtylerwalls:
Requested at https://groups.google.com/g/music21list/c/vMaz9wJVPyc.

Used `numberImplicit` to rhyme with attributes on MetronomeMark.

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -402,7 +402,7 @@ class GeneralObjectExporter:
           <!--=========================== Part 1 ===========================-->
           <part id="...">
             <!--========================= Measure 1 ==========================-->
-            <measure number="1">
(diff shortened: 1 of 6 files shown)


mscuthbert:
I don't like numberImplicit. I think it should be like showMeasureNumber="default/always/never" (maybe with an enum?)

numberImplicit sounds like it has something to do with "get measure number from context" but that's not what the MusicXML attribute means. 

Thanks!

mscuthbert:
Super amazing!  Thank you for the hard work on the fix.  It looks great to me.  Will figure out what new mypy is complaining about later.  Merging!

mscuthbert:
looking further, I should probably have put this under `.style.measureNumbering` -- I forgot we had that.  Might refactor before v9 is released.

PR #1496, opened 2022-12-07
adds doc for RN from int, following #1481

MarkGotham:
Also a de facto doc test for the upper/lower case issue discussed there.


MarkGotham:
#1450

MarkGotham:
@jacobtylerwalls 
Uncontroversial, I think ;)

mscuthbert:
Thanks!

PR #1497, opened 2022-12-10
Correct diminished and half-diminished 11th chord types

jacobtylerwalls:
The following 11th chord types were inconsistent with 9th and 13th chords and with common usage AFAICT:
- half-diminished 11th: was using a diminished 9th
- diminished 11ths: was using a diminished 9ths and diminished 11th

My understanding is that diminshed/half-diminished refers to the sevenths--everything else needs to be given with "flat/sharp-11" if wanted, see `half-diminished-minor-ninth` earlier in this file.

### DIFF
--- a/music21/harmony.py
+++ b/music21/harmony.py
@@ -94,8 +94,8 @@ CHORD_TYPES = collections.OrderedDict([
     ('minor-11th', ['1,-3,5,-7,9,11', ['m11', 'min11']]),  # Y
     ('augmented-major-11th', ['1,3,#5,7,9,11', ['+M11', 'augmaj11']]),  # N
     ('augmented-11th', ['1,3,#5,-7,9,11', ['+11', 'aug11']]),  # N
-    ('half-diminished-11th', ['1,-3,-5,-7,-9,11', ['ø11']]),  # N
-    ('diminished-11th', ['1,-3,-5,--7,-9,-11', ['o11', 'dim11']]),  # N
+    ('half-diminished-11th', ['1,-3,-5,-7,9,11', ['ø11']]),  # N
+    ('diminished-11th', ['1,-3,-5,--7,9,11', ['o11', 'dim11']]),  # N
 
     # thirteenths
     ('major-13th', ['1,3,5,7,9,11,13', ['M13', 'Maj13']]),  # Y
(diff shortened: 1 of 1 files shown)


mscuthbert:
I had no idea we had half-diminished or fully diminished 11ths (or 9ths or 13ths).  Your proposal does seem consistent (the only thing I'd change is it's not all about the 7th; it's the 7th and the fifth).

But I can't seem to find anything authoritative that backs up your view, and I know that the student who entered all of these didn't have this level of music theory, so she must have gotten the notes from somewhere -- I don't know if it's somewhere authoritative or just another site.  I want to merge, but given that it's a backwards-incompatible change, I'd like to see some reference that this interpretation is correct before doing so.  It might be that our 9th and 13th chords are incorrectly encoded instead.
(I suppose there's no fully diminished 13th because that would give a halfstep between adjacent notes? but it also seems strange.)

jacobtylerwalls:
Right. I think part of the issue is that 11th and 13ths don't entail the presence of 9ths at all, so we're going to have trouble finding an authoritative source saying that 1-b3-b5-b7-9-11 is a thing at all. My guess is that if 11ths entail 9ths, you're thinking "in scales" as opposed to in "chord symbols", which musicians fluently switch between anyway.

What I think we can find sources for is that:
- [extensions](https://www.google.com/books/edition/The_Jazz_Theory_Book/iyNQpJ4oaMcC?hl=en&gbpv=1&bsq=extensions) are usually diatonic steps above the ionian root, third, and fifth
- alterations must be specified with flats and sharps, and diminished ninths and elevenths in particular would be "[avoid notes](https://www.google.com/books/edition/The_Berklee_Book_of_Jazz_Harmony/ufxSDwAAQBAJ?hl=en&gbpv=1&bsq=harmonic%20avoid%20notes)" forming ic1's with chord tones--so you have to specify them if wanted eg. V7(b9)
- [diminished 11ths don't make sense in this context](https://books.google.com/books?id=iyNQpJ4oaMcC&q=11th+chords#v=snippet&q=the%2011th%20can't%20be%20lowered&f=false) (where spelling is not relevant to soloing): just become major tenths
- ø9 implies the "natural" (i.e. major) 9th. (See paragraph 16 of [MTO review of Levine](https://www.mtosmt.org/issues/mto.00.6.1/mto.00.6.1.rawlins.html) or Aebersold's [Scale Syllabus](http://dsapresents.org/guitar/wp-content/uploads/2011/08/Shell-Chords-Scales-Arrpeggios-Scale-Syllabus-Aebersold.pdf) for Cø9).

My guess is that once you've made the choice of treating Cø11 as referring to Fb, you're inclined to flatten the 9ths as well to not cause the [01246] dissonance of Bb-C-D-Eb-Fb, but that's a subsequent domino that's falling.

jacobtylerwalls:
> (I suppose there's no fully diminished 13th because that would give a halfstep between adjacent notes? but it also seems strange.)

Actually, I think that confirms my view -- a natural 13th is enharmonically equivalent to a diminished seventh, so it would be superfluous to specify it. Instead if a "diminished 13th" is wanted, give Co7(b13), or just give up and use the octatonic scale.

mscuthbert:
(Thanks!)

PR #1508, opened 2023-01-01
Remove `forceSource=True` from user's guide examples

jacobtylerwalls:
This keyword is mentioned back in ch. 8, but it's described as ~_you don't need it_, so I was surprised to see it pop up here.


mscuthbert:
Oh, yeah.  those aren't done yet and bugs still emerging, but right, I shouldn't have left them in.

PR #1516, opened 2023-01-15
Give GeneralMordent placement like Turn and Trill have.

gregchapman-dev:
m21ToXml.py already checks if hasattr(blah, 'placement'), so all we need is this one line change.


mscuthbert:
thanks!

PR #1527, opened 2023-02-18
Observe updatePitches=False when bass or root provided

### LINKED ISSUE #1350: Possible bug/slowness in Harmony constructor
**music21 version**

`8.0.9a`   

**Problem summary**

In harmony.py, l. 246 currently reads:

    if (updatePitches
            and self._figure  # == '' or is not None
            or 'root' in self._overrides
            or 'bass' in self._overrides):
        self._updatePitches()

before a formatting change in 2020 it was:

    if (updatePitches and self._figure is not None
            or 'root' in self._overrides or 'bass' in self._overrides):

before that it was:

    if self._figure is not None or self._root or self._bass:            
        self._updatePitches()

The way `W and X or Y or Z` is parsed is `(W and X) or Y or Z`.  
I think we mean `W and (X or Y or Z)`

I'm wondering if it is supposed to be something like this:

    if (updatePitches 
        and (self._figure
                or 'root' in self._overrides 
                or 'bass' in self._overrides)):

Does that seem right to you, @jacobtylerwalls ?

jacobtylerwalls:
Closes #1350

### DIFF
--- a/music21/graph/primitives.py
+++ b/music21/graph/primitives.py
@@ -851,7 +851,7 @@ class GraphColorGridLegend(Graph):
         >>> fig = matplotlib.pyplot.figure()
         >>> subplot = colorLegend.makeOneRowOfGraph(fig, 0, 'Scriabin Mapping', rowData)
         >>> subplot
-        <AxesSubplot: >
+        <Axes: >
         '''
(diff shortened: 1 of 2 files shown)


mscuthbert (approved):
thanks!

PR #1533, opened 2023-02-24
Add support for delayed turns (and delayed inverted turns).

gregchapman-dev:


gregchapman-dev:
Hoping this can make v9 as well.

music21/expressions.py
```diff
@@ -1027,42 +1030,97 @@ class Turn(Ornament):
     A turn or Gruppetto.
 
     * Changed in v7: size is a Generic second.  removed unused nachschlag component.
+    * Changed in v9: Added support for delayed vs non-delayed Turn.
     '''
-    def __init__(self, **keywords):
+    def __init__(self, delay: OrnamentDelay | OffsetQL = OrnamentDelay.NO_DELAY, **keywords):
```
mscuthbert, line 1035: If anything needs to be changed, I'd prefer this to be keyword only:  `self, *, delay: ...` -- if no other changes needed, then it's good.

music21/expressions.py
```diff
@@ -1082,16 +1140,16 @@ def realize(self, srcObj: 'music21.note.Note', *, inPlace=False):
...
-        <music21.note.Note C> 128th
-        <music21.note.Note B> 128th
-        <music21.note.Note C> 128th
+        <music21.note.Note D> 256th
+        <music21.note.Note C> 256th
+        <music21.note.Note B> 256th
+        <music21.note.Note C> 256th
         >>> n2.duration.type
-        'zero'
+        '64th'
```
mscuthbert, line 1152: Can we show that `newOrigNote is n2`

music21/expressions.py
```diff
@@ -1170,8 +1244,8 @@ def realize(self, srcObj: 'music21.note.Note', *, inPlace=False):
 
 
 class InvertedTurn(Turn):
-    def __init__(self, **keywords):
-        super().__init__(**keywords)
+    def __init__(self, delay: OrnamentDelay | OffsetQL = OrnamentDelay.NO_DELAY, **keywords):
```
mscuthbert, line 1247: as above

music21/musicxml/m21ToXml.py
```diff
@@ -5216,6 +5216,13 @@ def expressionToXml(self, expression):
         >>> MEX.dump(mxExpression)
         <inverted-turn placement="above" />
 
+        >>> invDelayedTurn = expressions.InvertedTurn(delay=1.)
```
mscuthbert, line 5219: style: 1.0 -- never a trailing period to create a float.

music21/test/test_expressions.py
```diff
@@ -66,14 +66,16 @@ def testGetRepeatExpression(self):
...
+        # n1 is a dotted-quarter-note with a non-delayed inverted turn
         p1 = stream.Part()
         m1 = stream.Measure()
         m2 = stream.Measure()
         p1.append(clef.TrebleClef())
         p1.append(key.Key('F', 'major'))
         p1.append(meter.TimeSignature('2/4'))
         n1 = note.Note('C5', type='half')
-        turn0 = expressions.Turn()
+        turn0 = expressions.Turn(delay=1.)  # delay is a quarter-note
```
mscuthbert, line 78: as above no trailing dot.

music21/test/test_expressions.py
```diff
@@ -88,16 +90,33 @@ def testExpandTurns(self):
...
+        # because turn0.quarterLength defaults to 0.25 (16th-note), each of the four turn
+        # notes is ql=0.25 instead of 0.375 (realize doesn't stretch to fill).
+        self.assertEqual(realized2[0].quarterLength, 0.25)
         self.assertEqual(realized2[1].quarterLength, 0.25)
+        self.assertEqual(realized2[2].quarterLength, 0.25)
+        self.assertEqual(realized2[3].quarterLength, 0.25)
 
+        # Here, delay is still a quarter-note (delay=1.0), and each of the four turn notes
+        # is ql=0.125 (because turn0.quarterLength has been set to 0.125, and realize
+        # doesn't stretch to fit).
```
mscuthbert, line 112: ah -- this is the only substantive issue I see and disagree with -- I think it should be an error to have an ornament where the sum of the realized durations doesn't equal the length of the original.  In the case where there's an explicit delay float set, and an explicit quarterLength for the turn, the last note should be held to the length of the remaining unused duration.

This is common where there's a dotted-eighth + 16th pattern with a delayed turn on the eighth, where the turn is usually executed as 3 notes on the 2nd 16th and the last 16th holds the original note:

C[8.] D[16]  =>  C[16] D[32trip] C[32trip] B[32trip] C[16] D[16]

(which would make a great demo).  It's seen all over the music of Weber.

What I don't want happening is that there's a score with lots of ornaments and someone calls "realizeOrnaments" and now there's gaps in the stream.  That should be an error.

mscuthbert (changes requested):
This is GREAT!  There is one substantive issue (last comment) that needs to be addressed about float delays, and a few style things, but hope I can get this in today.  Thanks!

music21/test/test_expressions.py
gregchapman-dev: OK, I think what you mean is that we _should_ stretch to fit, but only stretch the last note of the turn, leaving the first three at the prescribed `turn.quarterLength` duration.  I'll do that, and take a shot at putting that Weber turn in the doctest.

music21/expressions.py

music21/expressions.py

music21/musicxml/m21ToXml.py
gregchapman-dev: Done.

music21/expressions.py
gregchapman-dev: Done.

music21/test/test_expressions.py
gregchapman-dev: Done.

music21/test/test_expressions.py
gregchapman-dev: Done.

gregchapman-dev:
Assuming all the checks pass, this should be ready to go.

gregchapman-dev:
@mscuthbert Checks have passed, looks ready to go.

mscuthbert:
Great. Will be in the office later today and looks like it's a ready to merge.

mscuthbert (approved):
Thanks!  Looks perfect!

mscuthbert:
Merging.

Thinking in the future we might need to have something like "minQuarterLength" and "maxQuarterLength" instead of a bare quarterLength idea -- but that's an idea for later.

gregchapman-dev:
Thanks!

PR #1539, opened 2023-03-22
Import and export of figured bass elements from MEI and musicxml file formats

mxordn:
(1) Added the possibility to import musicxml and MEI representations of figured bass notations.

(2) Proposes also a class to represent Figured Bass in a music21 stream. The class is called FiguredBassIndication and inherits from the harmony object using the already existing objects from the figuredbass module.

Any comments are highly appreciated. A discussion in the google list is here available: https://groups.google.com/g/music21list/c/-Z_UQfZ9SPk/m/gyE9DciUCgAJ


gregchapman-dev (commented):
Quick comment/worry about mei/base.py: I think you started from an older version; you are missing a bunch of stuff that happened in the last year or so (for example the metadata changes I made in v8, and all the typing changes Myke made as well).

mxordn:
> Quick comment/worry about mei/base.py: I think you started from an older version; you are missing a bunch of stuff that happened in the last year or so (for example the metadata changes I made in v8, and all the typing changes Myke made as well).

This is interesting. I synced with the main repo some weeks ago and thought I used the mei/base.py that is in the current master branch. I will have a look at it.

music21/figuredBass/notation.py
```diff
@@ -371,6 +402,15 @@ def _getFigures(self):
 
         self.figures = figures
 
+        figuresFromNotaCol = []
+
+        for i in range(len(self.origNumbers)):
```
ahankinson, line 407: Better to iterate over the actual list like this:

```suggestion
        for i, number in enumerate(self.origNumbers):
```

Then you can remove:

`number = self.orgNumbers[i]`

music21/figuredBass/notation.py
```diff
@@ -412,14 +452,18 @@ class Figure(prebase.ProtoM21Object):
             ''',
     }
 
-    def __init__(self, number=1, modifierString=None):
+    def __init__(self, number=1, modifierString=None, isExtender=None):
         self.number = number
         self.modifierString = modifierString
         self.modifier = Modifier(modifierString)
+        if self.number == '_':
```
ahankinson, line 459: This whole block could be replaced with:

`self.isExtender = (self.number == "_")`

music21/harmony.py
```diff
@@ -2492,6 +2493,32 @@ def transpose(self: NCT, _value, *, inPlace=False) -> NCT | None:
 
 # ------------------------------------------------------------------------------
 
+class FiguredBassIndication(Harmony):
+    isFigure = True
+    tstamp = 1
+    def __init__(self, figs=None, **keywords):
```
ahankinson, line 2499: It would be good to have type annotations for the `__init__` here. `figs: ?? | None = None`

music21/harmony.py
```diff
@@ -2492,6 +2493,32 @@ def transpose(self: NCT, _value, *, inPlace=False) -> NCT | None:
...
+        super().__init__(**keywords)
+        if figs:
+            if isinstance(figs, list):
+                fig_string = str(figs[0])
```
ahankinson, line 2503: I suspect this whole block could be just:

`",".join(figs)`

You're also duplicating the first element in your string by starting your `fig_string` with the first element. Is that intentional?

music21/mei/base.py
```diff
@@ -2493,6 +2504,52 @@ def chordFromElement(elem, slurBundle=None):
...
+
+    return fb_harmony_tag
+
+def figuredbassFromElement(elem, slurBundle=None):
+    if elem.get(_XMLID):
+        fb_id = elem.get(_XMLID)
+    fb_notation = ''
+    dauer: float = 0
+    # loop through all child elements and collect <f> tags
+    for subElement in elem.findall('*'):
```
ahankinson, line 2531: You don't need to loop through all of the elements -- you can just do an xquery to find just the `<f>` elements and return a list of those.

music21/mei/base.py
```diff
@@ -2493,6 +2504,52 @@ def chordFromElement(elem, slurBundle=None):
...
+        if subElement.tag == f'{MEI_NS}f':
+            if subElement.text is not None:
+                if fb_notation != '':
```
ahankinson, line 2534: Rather than doing some complicated looping, you can just gather all of the text values into a list, and then `join` them: `",".join(["list", "of", "text", "values"])`

music21/mei/base.py
mxordn: Do you mean something like `elem.findall('f'{MEI_NS}f'')`? Then the first if statement is not necessary. I think I did this to be able to grab other harmony tags here in the future.

music21/harmony.py
mxordn: type annotation is added.

music21/figuredBass/notation.py
mxordn: Right.

music21/figuredBass/notation.py
mxordn: Done. Thanks for the optimazation.

music21/harmony.py
mxordn: Thanks for this. Didn't see this error in the result. Seems the notation object sorts out duplicates.

music21/mei/base.py
mxordn: This simplifies a lot! I changed it like you suggested.

mxordn:
Finally I have an overview over the tests in music21 and flake8. Sorry for so many corrections and retests. I managed now to run all the tests successfully on my machine. The linter hint to iterate over the dictionary rather than calling .keys() to Me looks not like a real improvement that simplifies or speeds up the process (```music21/musicxml/m21ToXml.py:1707```).

music21/harmony.py
```diff
@@ -2492,6 +2493,33 @@ def transpose(self: NCT, _value, *, inPlace=False) -> NCT | None:
 
 # ------------------------------------------------------------------------------
 
+class FiguredBassIndication(Harmony):
```
jacobtylerwalls, line 2496: I'm wondering what you think we should do about the existing `figure` property, which as of now will give:
```python
>>> fbi = harmony.FiguredBassIndication(root='D', figure='643')
>>> fbi
<FiguredBassIndication figures: >
>>> fbi.figure
'643'
```

music21/harmony.py
```diff
@@ -2492,6 +2493,33 @@ def transpose(self: NCT, _value, *, inPlace=False) -> NCT | None:
...
+    isFigure = True
+    tstamp = 1
```
jacobtylerwalls, line 2498: If these are class attributes, won't each instance contaminate the the others when attempting to set them? Also, can you explain what tstamp is needed for? (Can we just use `.duration` and `writeAsChord=False`?)

music21/musicxml/m21ToXml.py
```diff
@@ -3358,6 +3405,7 @@ def parseFlatElements(
                     if hasSpannerAnchors:
                         self.parseOneElement(obj, AppendSpanners.NONE)
                     else:
+                        # ENTRY FOR Figured Bass Indications
```
jacobtylerwalls, line 3408: If you'd like to add a comment clarifying why harmony objects get handled here, I'd suggest improving the comment above ("including ChordSymbols or FiguredBassIndications not written as chord").
```suggestion
```

music21/musicxml/xmlToM21.py
```diff
@@ -5148,6 +5172,62 @@ def xmlToChordSymbol(
...
+        if d:
+            fbi.quarterLength = d.quarterLength
+        # call function in parent to add found objects.
+        self.parent.appendFbis(fbi, offsetFbi)
+
+    def _getFigurePrefixOrSuffix(self, figure, presuf: str = 'prefix') -> str:
+        if figure.findall(presuf):
+            for fix in figure.findall(presuf):
+                if fix.text:
+                    print(fix.text)
```
jacobtylerwalls, line 5227:
```suggestion
```

music21/musicxml/m21ToXml.py
```diff
@@ -3488,6 +3536,29 @@ def parseOneElement(
...
+        if self.stream:
+            measureRange = (self.stream.offset, self.stream.offset + self.stream.highestTime)
+
+            # Look if there are figures in the current measure and insert them
+            # to add them later.
+            if self.parent and self.parent.fbis:
+                for o, f in self.parent.fbis.items():
+                    if o >= measureRange[0] and o < measureRange[1]:
+                        for fbi in f:
+                            self.stream.insert(o - self.stream.offset, fbi)
```
jacobtylerwalls, line 3558: Why are we inserting the figured bass objects into the score instead of into the measure? From the test file parsed in the previous comment:

```
>>> s.show('text')
{0.0} <music21.metadata.Metadata object at 0x10846cc50>
{0.0} <music21.stream.Part MusicXML Part>
    {0.0} <music21.instrument.Instrument 'P1: MusicXML Part: '>
        {0.0} <music21.key.Key of C major>
        {0.0} <music21.meter.TimeSignature 4/4>
        {0.0} <music21.note.Note G>
        {0.5} <music21.note.Note G>
        {1.0} <music21.note.Note G>
        {1.75} <music21.note.Note G>
        {2.0} <music21.note.Note G>
        {3.0} <music21.note.Note G>
{0.0} <FiguredBassIndication figures: 3>
{1.0} <FiguredBassIndication figures: #1,b3,♮5>
{1.75} <FiguredBassIndication figures: 6>
{2.0} <FiguredBassIndication figures: 5/,b127/>
{3.0} <FiguredBassIndication figures: >
```

music21/musicxml/m21ToXml.py
```diff
@@ -3332,7 +3366,20 @@ def parseFlatElements(
...
+        # first group them in one list together with their note, instead of handling them
+        # seperately. O
+        groupedObjList: list = []
+        for els in objIterator:
+            if len(groupedObjList) > 0:
+                if len(els) == 1 and isinstance(els[0], harmony.FiguredBassIndication):
+                    # print('**multiple fb found**')
+                    groupedObjList[-1].append(els[0])
+                    continue
+            groupedObjList.append(els)
```
jacobtylerwalls, line 3380: Did you explore solutions that did not involve another iteration of the entire stream?

music21/musicxml/xmlToM21.py
```diff
@@ -2850,6 +2865,15 @@ def xmlToNote(self, mxNote: ET.Element) -> None:
         self.offsetMeasureNote += offsetIncrement
         self.endedWithForwardTag = None
 
+        # reset offset for figures. This is needed to put in
+        # multiple FiguredBassIndications at one note at the right offset.
+        # Musicxml puts <figured-bass> tags immediately before a <note> tag,
+        # which means that we have to reset a given offset duration of some
+        # <figured-bass> tags after inserting the coressponding note and
+        # before going to a new note.
+        self.lastFigureDuration = 0
```
jacobtylerwalls, line 2874: I would appreciate if you could share more of your findings around duration/offset/tstamp. It's not yet clear to me why we can't just treat the `<figured-bass>` tags like all the other musicxml tags and just insert them into the measure as we go, like almost all the other musicxml methods in this class. It's true, if there are more than one at the same time point, we need to adjust the last one instead of creating a new one, but that's just a couple lines of code (e.g. `if self.stream.getElementsByOffset(self.offsetInMeasure, classList=[Figured...])`

jacobtylerwalls (commented):
Thanks for starting on a PR for figured bass parsing.

The next most important thing to work on is adding test coverage. There is already a sample file in the repo. I've parsed it in one of the review comments. Here is an [example](https://github.com/cuthbertLab/music21/blob/37763c432629f06a3a32340e60e60b2f382b685b/music21/musicxml/test_xmlToM21.py#L1400) test that uses the same set of lilypond test files. For the musicxml side, we'll need a test in the import test file, the export test file, and then doctests on your new public methods. There's also the MEI side to think about.

Finally, I'm hoping you can add some more details around how you arrived at the current solution involving another iteration pass and storing off a map of offsets. I would suggest letting the MeasureParser do most of the work for you. I can sketch out this idea later if helpful, but wanted to hear if you had already explored this.

mscuthbert:
What Jacob said: tests, docs, and stubs of implementation are more important at this stage than a fully fleshed out working version that isn't tested or documented.  Code is easy -- understanding is hard.

music21/musicxml/m21ToXml.py
mxordn: I think the measure object might be a good place as well. It will then be a similar solution to the way figures are represented in MEI. I had the figured bass handling of lilypond in mind when writing the importer. I will think it over, but I think both are valid. Maybe the measure is a better place to keep the score cleaned up.

mxordn:
Thanks for the review and the information. I will come back to this. But I'm quite busy during the next three weeks. So please be patient with me not being fast in going through the points. Originally the implementation was an experiment for me and an exploration of how one could process a corpus of baroque music and make some analysis.

mscuthbert:
Hi. This PR currently changes 200+ files and has hundreds of commits. This won't be possible to review. Can you close and open a new one just with the changes from the "master" branch?

mxordn:
As I came back to this topic today, I thought similar. Rebasing my fork to the new version triggered the checks here I think. I'll close this PR and open a new one when I have the stuff ready.

PR #1540, opened 2023-03-25
Minimize gaps produced by quantization algorithm

### LINKED ISSUE #1536: Trailing rests created that don't agree with the quantization scheme
**music21 version**

`8.1.0`   

**Problem summary**
I want to flatten a score for reading by another program which needs a single track.
When I read a score with converter.parse, flatten it and re-write it with score.write, the notes' timings are wonky (it sounds wrong).
Using other quarterLengthDivisors doesn't change the results.
This happens on many, but not all files.
What is the correct way to do this please?

**Steps to reproduce**

```
from music21 import *
import subprocess
import os

def makexml(source, target):
    # Convert a midi to MusicXML with MuseScore
    musescore_path = r'C:\Program Files\MuseScore 3\bin\MuseScore3.exe'
    cmd_str = f'"{musescore_path}" "{source}" -o "{target}" -w'
    subprocess.run(cmd_str)

piece = 'mazurka'  # or 'rondo'
# Convert with MuseScore, as recommended here
# https://github.com/cuthbertLab/music21/issues/1530#issuecomment-1440759880
xml_path = 'temp.mxl'
makexml(piece + '.mid', xml_path)
# Changing from default quarterLengthDivisors=((4, 3)) to quarterLengthDivisors=((8, 6))
# makes no difference, output still wonky
score = converter.parse(xml_path)
os.remove(xml_path)
score = score.flatten()  # Flattened or not, still wonky
score.write(fmt='midi', fp=piece + '.flat.mid')
```

See attached mazurka and rondo files.
[samples.zip](https://github.com/cuthbertLab/music21/files/10981106/samples.zip)

jacobtylerwalls:
Fixes #1536 by improving the algorithm that "looks ahead" to minimize gaps when quantizing durations from its first draft in #1062, as long as the stream is sorted first.

Looking ahead was the right idea, but using the same divisor was not right, because in a passage of triplets, on beats, you have a tie between divisors (for calculating offsets). For the screenshotted example below, that means the last triplet-eighths of a beat were quantized as sixteenths (because there was a tie for the "best divisor" for the quantization of the offset at the X.0 beat).

This refactor solves the reported issue and also includes a simpler way of achieving #641's requirement of nonzero durations on non-grace notes.

- This also uncovered some infelicities with Stream.insert(ignoreSort=True), adding an explicit test.
- Now streams parsed from MIDI are explicitly sorted.

**Before**
Screenshot 2023-03-25 at 5 53 43 PM

**After**
Screenshot 2023-03-25 at 5 53 19 PM

### DIFF
--- a/music21/stream/base.py
+++ b/music21/stream/base.py
@@ -86,7 +86,7 @@ RecursiveLyricList = note.Lyric | None | list['RecursiveLyricList']
 
 BestQuantizationMatch = namedtuple(
     'BestQuantizationMatch',
-    ['error', 'tick', 'match', 'signedError', 'divisor']
+    ['remainingGap', 'error', 'tick', 'match', 'signedError', 'divisor']
 )
 
 class StreamDeprecationWarning(UserWarning):
@@ -9361,18 +9361,34 @@ class Stream(core.StreamCore, t.Generic[M21ObjType]):
         # this presently is not trying to avoid overlaps that
         # result from quantization; this may be necessary
 
-        def bestMatch(target, divisors):
+        def bestMatch(target, divisors, gapToFill=0.0):
             found = []
             for div in divisors:
-                match, error, signedErrorInner = common.nearestMultiple(target, (1 / div))
-                # Sort by unsigned error, then "tick" (divisor expressed as QL, e.g. 0.25)
-                found.append(BestQuantizationMatch(error, 1 / div, match, signedErrorInner, div))
+                tick = 1 / div
+                match, error, signedErrorInner = common.nearestMultiple(target, tick)
+                if match % tick:
+                    remainingGap = max(gapToFill - match, 0.0)
+                else:
+                    remainingGap = 0.0
+                # Sort by remainingGap, then unsigned error, then "tick"
+                # (divisor expressed as QL, e.g. 0.25)
+                found.append(
+                    BestQuantizationMatch(
+                        remainingGap, error, tick, match, signedErrorInner, div))
             # get first, and leave out the error
             bestMatchTuple = sorted(found)[0]
             return bestMatchTuple
 
-        # if we have a min of 0.25 (sixteenth)
-        # quarterLengthMin = quarterLengthDivisors[0]
+        def findNextElementNotCoincident(
+            useStream: Stream,
+            startIndex: int,
+        ) -> tuple[base.Music21Object | None, BestQuantizationMatch | None]:
+            for next_el in useStream._elements[startIndex:]:
+                next_offset = useStream.elementOffset(next_el)
+                look_ahead_result = bestMatch(float(next_offset), quarterLengthDivisors)
+                if look_ahead_result.match > o:
+                    return next_el, look_ahead_result
+            return None, None
 
         if inPlace is False:
             returnStream = self.coreCopyAsDerivation('quantize')
@@ -9385,6 +9401,11 @@ class Stream(core.StreamCore, t.Generic[M21ObjType]):
 
         rests_lacking_durations: list[note.Rest] = []
         for useStream in useStreams:
+            # coreSetElementOffset() will immediately set isSorted = False,
+            # but we need to know if the stream was originally sorted to know
+            # if it's worth "looking ahead" to the next offset. If a stream
+            # is unsorted originally, this "looking ahead" could become O(n^2).
+            originallySorted = useStream.isSorted
             for i, e in enumerate(useStream._elements):
                 if processOffsets:
                     o = useStream.elementOffset(e)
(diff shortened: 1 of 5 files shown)


music21/stream/base.py
```diff
@@ -86,7 +86,7 @@
 
 BestQuantizationMatch = namedtuple(
     'BestQuantizationMatch',
-    ['error', 'tick', 'match', 'signedError', 'divisor']
+    ['remainingGap', 'error', 'tick', 'match', 'signedError', 'divisor']
```
mscuthbert, line 89: okay, but generally add new attributes to a namedtuple at the end so anyone using `[0]` still gets the same result.

music21/stream/tests.py
```diff
@@ -4726,6 +4738,13 @@ def testSortAndAutoSort(self):
         self.assertEqual([(0.0, 2), (0.0, 30), (5.0, 25), (8.0, 10), (10.0, 2),
                           (15.0, 10), (20.0, 2), (22.0, 1.0)], match)
 
+    def testInsertIgnoreSort(self):
+        '''A sorted stream does not become unsorted when ignoreSort=True.'''
+        s = Stream()
+        s.repeatAppend(note.Note(), 4)
+        s.insert(1, tempo.MetronomeMark(), ignoreSort=True)
+        self.assertTrue(s.isSorted)
```
mscuthbert, line 4746: this seems strange to me -- can you explain this?  The Stream is no longer sorted -- why is it reporting True for isSorted?  I don't think I like this.  `isSorted` should be something that should be able to be trusted unless private methods are being used.

jacobtylerwalls: Ah, I think I just botched the setup of the test case. Inserting something at the end would make a more clear test case, I'll fix it.

What motivated this: the docs for insert say:
> If `ignoreSort` is True then the inserting does not
        change whether the Stream is sorted or not (much faster if you're
        going to be inserting dozens
        of items that don't change the sort status)

But this test fails on master, which is to say, providing ignoreSort=True _does_ change the sort status, even if the stream stays sorted (as I'll update the test to actually demonstrate!) This is because insert() was unconditionally calling coreElementsChanged with clearIsSorted=True.

> isSorted should be something that should be able to be trusted unless private methods are being used.

Yeah, maybe ignoreSort needs to become exposed on coreInsert only?

music21/stream/base.py
jacobtylerwalls: True, but then we could no longer use it for sorting (unless there's something clever I'm not thinking of).

music21/stream/tests.py
jacobtylerwalls: How I ended up here was:

1. I noticed that MIDI files have a tendency to produce simultaneities that don't collapse into chords, because of differing durations. So iterating elements and checking the element at i + 1 might still give you the same offset; I need the next distinct offset. I didn't want to have to look backwards in the elements array, so I realized I wanted sorted streams.
2. So I made MIDI parsing produce a stream with isSorted reported as True.
3. Then I went to add a test case to testQuantize, which builds streams with insert(). I then had to debug why the stream was reporting itself as unsorted. I chose between using ignoreSort=True versus doing an explicit sort.
4. ignoreSort=True wasn't preserving the value of isSorted (I agree that feels like a should-be-core thing, though)

mscuthbert: > Yeah, maybe ignoreSort needs to become exposed on coreInsert only?

yeah, I think that's a much better place.  And note that coreInsert still computes and returns whether the stream is guaranteed to be sorted.

But we shouldn't expose anything to the users where they could end up with a stream that needs manual sorting unless they set isSorted = False.  Feel free to remove.  Or just ignore the test and remove calls from this PR and I'll remove ignoreSort in insert later.

jacobtylerwalls: Reverted changes and just went with an explicit sort 👍 . I think I misdiagnosed exactly what was unsorted in the first place anyway.

mscuthbert:
Ah -- I never did a re-review.  Is this set for reviewing again?

jacobtylerwalls:
Yep!

music21/stream/base.py
mscuthbert: Generally I've used namedtuples because before it was a standard (anonymous) tuple, so people might still be using access by index or destructuring paradigms.

mscuthbert (approved):
Changes look good

music21/stream/base.py
mscuthbert: ah, yes that's fine.  You could sort with `key=lambda bqm: (bqm[-1], *bqm)` but yeah, too much trouble for backwards compatibility on an internal object.

I did though change it to use min -- there's no need to sort the whole thing if you're just getting the minimum

music21/stream/tests.py
```diff
@@ -4027,6 +4027,8 @@ def procCompare(srcOffset, srcDur, dstOffset, dstDur, divList):
                 n = note.Note()
                 n.quarterLength = srcDur[i]
                 s.insert(srcOffset[i], n)
+            # Must be sorted for quantizing to work optimally.
+            s.sort()
```
mscuthbert, line 4031: tiny bit confused by this addition -- Streams are autosorted by default.  This seems to be a problem of quantize if it somehow bypasses the normal "sort before iterate/access" paradigm.

music21/stream/base.py
mscuthbert: ah! That was old code back when the m21 team didn't know its own algorithmic complexity

music21/stream/tests.py
jacobtylerwalls: Might have been an artifact from an earlier commit. Does look unnecessary now. 😅

Issue #1507, opened 2023-01-01
Support for rendering (a.k.a. writing MusicXML) trills/mordents/turns with accidentals


gregchapman-dev:
**Motivation**
music21 supports half-step and whole-step trills, for example, but if the half or whole step up needs an accidental (due to key signature, or previous accidentals in the measure), they are missing.  

**Proposed implementation**
This requires new support in m21ToXml.py to add an `<accidental-mark>` as follows:
```
<ornaments>
    <trill-mark placement="above" />
    <accidental-mark placement="above">natural</accidental-mark>
</ornaments>
```

I believe this will also require support in makeAccidentals, and the addition of an Accidental field (or two) in Trill/Mordent/Turn objects that makeAccidentals will configure (and pay attention to), and parsers can also configure directly from their own analysis.  I would, for example, modify xmlToM21.py to support _reading_ such `<accidental-mark>` elements, which would end up setting (e.g.) `turn.upperAccid.displayStatus` and `turn.lowerAccid.displayStatus` appropriately.

I am looking for review of the above proposal, before I start blazing away.  I'm quite sure I'm not thinking of everything.

I _have_ started experimenting to see which MusicXML renderers support ornaments with an `<accidental-mark>`, and Finale is pretty good (except for double sharps and double flats, for some reason, but those will be fairly rare).  Musescore ignores `<accidental-mark>` entirely, but maybe I can help them fix that.  Dorico does OK, but is buggier than Finale.

If you want to play with other renderers (or see what I think m21ToXml.py should be generating), see the attached MusicXML file (zipped).  I have annotated the trills in the score with "half" and "whole" text fields so you can figure out what the accidentals should be as you look at any particular rendering of the score.
[trillaccid.musicxml.zip](https://github.com/cuthbertLab/music21/files/10328155/trillaccid.musicxml.zip)

**Intent**

[x] I plan on implementing this myself.
[ ] I am willing to pay to have this feature added.
[ ] I am starting a discussion with the hope that community members will volunteer their time to create this. I understand that individuals work on features of interest to them and that this feature may never be implemented.

mscuthbert:
Thank you.  I never thought about this, but of course it's another thing more complex than it seems.

Did not think that makeAccidentals would someday be looking at expressions!

mscuthbert:
(p.s. thanks for the huge pushes on other PRs last week -- I've seen them now and they're on my radar)

gregchapman-dev:
[...]

PR #1545, opened 2023-03-26
Support ornament accidentals

gregchapman-dev:
Fixes issue #1507


gregchapman-dev:
Implementation so far: Add accidental support to Turn, GeneralMordent, Trill.  Parse from/write to MusicXML.  

Last step (still to do): makeAccidentals needs to add support for ornament accidentals.

gregchapman-dev:
Yeah, I know, I have a bunch more tests to write.

gregchapman-dev:
@mscuthbert @jacobtylerwalls Can I get some review of this implementation (while I work on makeAccidentals)?

gregchapman-dev:
OK, initial implementation of everything is done.  I need to write a bunch of tests (especially for makeAccidentals stuff), but this should be reviewable now.

music21/expressions.py
```diff
@@ -517,21 +519,205 @@ def fillListOfRealizedNotes(
         fillObjects.append(firstNote)
         fillObjects.append(secondNote)
 
+    def resolveOtherPitches(self, srcPitch: pitch.Pitch):
```
mscuthbert, line 522: needs public docs -- perhaps give the demo here of what it does with those types.  Not a fan of the name "other pitches" -- doesn't tell me anything about what they are.

music21/expressions.py
```diff
@@ -517,21 +519,205 @@ def fillListOfRealizedNotes(
...
+        pitchPast: list[pitch.Pitch] | None = None,
+        pitchPastMeasure: list[pitch.Pitch] | None = None,
+        otherSimultaneousPitches: list[pitch.Pitch] | None = None,
+        alteredPitches: list[pitch.Pitch] | None = None,
+        cautionaryPitchClass: bool = True,
+        cautionaryAll: bool = False,
+        overrideStatus: bool = False,
+        cautionaryNotImmediateRepeat: bool = True,
+        lastNoteWasTied: bool = False,
+    ):
```
mscuthbert, line 539: Return value?

music21/expressions.py
```diff
@@ -517,21 +519,205 @@ def fillListOfRealizedNotes(
...
+        '''
+        Updates accidental display for Ornaments with accidental(s): turns, mordents, and trills.
+        '''
```
mscuthbert, line 542: needs doc tests -- I'd actually rather review doctests before reviewing implementation than other way around.  I don't care too much about how something is implemented, but rather what's there.

music21/expressions.py
```diff
@@ -517,21 +519,205 @@ def fillListOfRealizedNotes(
...
+        if isinstance(self, Turn):
```
mscuthbert, line 543: Why is this not implemented in Turn?  (Similarly for GeneralMordent/Trill?)  I don't like `if isinstance(self, SubClassOfSelf)` very much.

music21/expressions.py
```diff
@@ -517,21 +519,205 @@ def fillListOfRealizedNotes(
...
+            # upper first
+            p = self.upperPitch
+            if p is not None:
+                if self.upperAccid is not None:
```
mscuthbert, line 547: self.upperAccid is undefined for Ornament -- this is why it needs to go in Turn.

music21/expressions.py
```diff
@@ -517,21 +519,205 @@ def fillListOfRealizedNotes(
...
+
+        '''
+        theName: str = super().name
+        if self.accid is not None:
+            theName += ' (' + self.accid + ')'
+        return theName
+
+    @property
+    def accid(self) -> str | None:
+        return self._accid
```
mscuthbert, line 648: should be accidentalName -- don't abbreviate public attributes/properties.  thanks!

music21/expressions.py
```diff
@@ -517,21 +519,205 @@ def fillListOfRealizedNotes(
...
+            self._accid = pitch.makeAccidentalName(newAccid)
+
+    @property
+    def size(self) -> interval.IntervalBase | None:
+        if self._size is None:
+            raise ExpressionException(
+                'Cannot compute mordent size without srcPitch unless .size has been '
+                + 'explicitly set (use mordent.getSize(srcPitch) instead.)'
+            )
+        return self._size
```
mscuthbert, line 668: property getters are not allowed to raise exceptions as of v7.  The reason?  Many debuggers in IDEs treat properties as the same as attributes and call them all in the process of debugging.  What was wrong with GenericInterval(2) as before?

music21/expressions.py
```diff
@@ -517,21 +519,205 @@ def fillListOfRealizedNotes(
...
     '''
     Base class for all Mordent types.
     '''
-    def __init__(self, **keywords):
+    def __init__(self, *, accid: str | None = None, **keywords):
         super().__init__(**keywords)
-        self.direction = ''  # up or down
-        self.size = None  # interval.Interval (General, etc.) class
+        self._direction = ''  # up or down
+        self._accid: str | None = None
```
mscuthbert, line 616: (_accid to _accidentalName or _accidentalStr)  -- no abbreviations.

Why can't this use '' as a None value?  It's much easier for users in IDEs to have a single type as an answer.  Then just `if accidentalName...` latter.

music21/expressions.py
```diff
@@ -517,21 +519,205 @@ def fillListOfRealizedNotes(
...
+                'Cannot compute mordent size without srcPitch unless .size has been '
+                + 'explicitly set (use mordent.getSize(srcPitch) instead.)'
+            )
+        return self._size
+
+    @size.setter
+    def size(self, newSize: interval.IntervalBase | None):
+        if self._accid is not None:
+            raise ExpressionException('Cannot set mordent size if accid already set.')
+        self._size = newSize
```
mscuthbert, line 674: only raise exceptions in setters if absolutely needed.  Have the setter change the accidentalName to '' if needed.

music21/expressions.py
```diff
@@ -517,21 +519,205 @@ def fillListOfRealizedNotes(
...
+
+    @property
+    def direction(self) -> str:
+        return self._direction
```
mscuthbert, line 678: should direction have a setter?  Maybe not, but it's strange to expose a getter without a setter.

music21/expressions.py
```diff
@@ -517,21 +519,205 @@ def fillListOfRealizedNotes(
...
+        if self.direction == 'up':
+            otherPitch.transpose(interval.GenericInterval(2), inPlace=True)
+        else:
+            otherPitch.transpose(interval.GenericInterval(-2), inPlace=True)
+
+        if self.accid is not None:
+            otherPitch.accidental = pitch.Accidental(self.accid)
+        return interval.Interval(srcPitch, otherPitch)
+
+    def resolveOtherPitches(self, srcPitch: pitch.Pitch):
```
mscuthbert, line 702: what does this do?

music21/expressions.py
```diff
@@ -517,21 +519,205 @@ def fillListOfRealizedNotes(
...
+                otherPitch.accidental = pitch.Accidental(0)
+            otherPitch.accidental.displayStatus = True
+        self._otherPitches = (otherPitch,)
+
+    @property
+    def otherPitch(self) -> pitch.Pitch | None:
+        if self._otherPitches:
+            return self._otherPitches[0]
+
+    def realize(self, srcObj: note.Note, *, inPlace=False):
```
mscuthbert, line 720: Q -- can ornaments appear on Chords/Rests/Unpitched?  Should this be checked for?  (I think it's an overall music21 bug, so maybe not.

music21/expressions.py
```diff
@@ -517,21 +519,205 @@ def fillListOfRealizedNotes(
...
+        otherPitch: pitch.Pitch = copy.deepcopy(srcPitch)
+        otherPitch.accidental = None
+
+        if self.direction == 'up':
+            otherPitch.transpose(interval.GenericInterval(2), inPlace=True)
+        else:
+            otherPitch.transpose(interval.GenericInterval(-2), inPlace=True)
+
+        if self.accid is not None:
+            otherPitch.accidental = pitch.Accidental(self.accid)
```
mscuthbert, line 699: I'm confused from the comments why the ornament can't store accidental objects?  yes, this would then involve a copy process, but it seems cleaner -- turns have sharp-accidentals attached to them, not 'sharp' strings.

music21/expressions.py
```diff
@@ -572,19 +758,21 @@ def realize(self, srcObj: 'music21.note.Note', *, inPlace=False):
...
         currentKeySig = srcObj.getContextByClass(key.KeySignature)
         if currentKeySig is None:
             currentKeySig = key.KeySignature(0)
 
-        for n in mordNotes:
-            n.pitch.accidental = currentKeySig.accidentalByStep(n.step)
+        # second (middle) note might need an accidental from the keysig (but
+        # only if it doesn't already have an accidental from accid)
+        for noteIdx, n in enumerate(mordNotes):
+            noteNum: int = noteIdx + 1
```
mscuthbert, line 772: no need for typing here -- noteNum will be typed properly from `enumerate`

music21/expressions.py
```diff
@@ -610,18 +798,34 @@ class Mordent(GeneralMordent):
...
+    >>> m.getSize(pitch.Pitch('B3'))
+    <music21.interval.Interval M-2>
+
+    >>> mFlat = expressions.Mordent(accid='flat')
+    >>> mFlat.direction
+    'down'
+    >>> mFlat.getSize(pitch.Pitch('C4'))
+    <music21.interval.Interval M-2>
+    >>> mFlat.getSize(pitch.Pitch('B3'))
+    <music21.interval.Interval A-2>
```
mscuthbert, line 820: can this be demo'd in a key signature context?

music21/expressions.py
```diff
@@ -558,12 +744,12 @@ def realize(self, srcObj: 'music21.note.Note', *, inPlace=False):
         '''
```
mscuthbert, line 744: update docs to include demonstration of key context, accidental etc.

music21/expressions.py
```diff
@@ -517,21 +519,205 @@ def fillListOfRealizedNotes(
...
+    def size(self, newSize: interval.IntervalBase | None):
+        if self._accid is not None:
+            raise ExpressionException('Cannot set mordent size if accid already set.')
+        self._size = newSize
+
+    @property
+    def direction(self) -> str:
+        return self._direction
+
+    def getSize(
```
mscuthbert, line 680: how about naming `sizeOnPitch` etc. -- but won't this require a key context?  I think that this routine should be the place where the exact interval that will be played will be returned -- this way people can do analysis without needing to realize the ornament.  So better with a key context.

music21/expressions.py
```diff
@@ -647,11 +866,26 @@ class WholeStepMordent(Mordent):
...
     >>> m.size
-    <music21.interval.Interval M2>
+    <music21.interval.Interval M-2>
     '''
-    def __init__(self, **keywords):
+    def __init__(self, **keywords) -> None:
+        # no accidental supported here, just "WholeStep"
+        kw = keywords.copy()
+        if 'accid' in kw:
+            del kw['accid']
```
mscuthbert, line 875: this seems like a place to raise an exception rather than silently deleting it.  Better to fail fast rather than letting a bug slip through.

music21/expressions.py
```diff
@@ -739,17 +1019,78 @@ class Trill(Ornament):
...
+            self._accid = pitch.makeAccidentalName(newAccid)
+
+    @property
+    def size(self) -> interval.IntervalBase | None:
+        if self._size is None:
+            raise ExpressionException(
+                'Cannot compute trill size without srcPitch unless trill size has been '
+                + 'explicitly set (use trill.getSize(srcPitch) instead.)'
+            )
+        return self._size
```
mscuthbert, line 1086: again, no exceptions in getters.

music21/expressions.py
```diff
@@ -1032,15 +1442,53 @@ class Turn(Ornament):
     * Changed in v7: size is a Generic second.  removed unused nachschlag component.
     * Changed in v9: Added support for delayed vs non-delayed Turn.
```
mscuthbert, line 1443: update docs.

music21/expressions.py
```diff
@@ -1075,23 +1524,100 @@ def name(self) -> str:
...
+                theName += 'lower=' + self.lowerAccid
+            theName += ')'
+
+        return theName
 
-    def realize(self, srcObj: 'music21.note.Note', *, inPlace=False):
+    def getUpperSize(
+        self,
+        srcPitch: pitch.Pitch,
+    ) -> interval.IntervalBase:
```
mscuthbert, line 1561: I don't like the duplication of getUpperSize/getLowerSize -- why not a keyword to "getSize" or whatever we called it above.

music21/expressions.py
```diff
@@ -1190,12 +1716,16 @@ def realize(self, srcObj: 'music21.note.Note', *, inPlace=False):
         music21.expressions.ExpressionException: The note is not long enough to realize a turn
         '''
         from music21 import key
+
+        if not hasattr(srcObj, 'pitch'):
+            raise TypeError('Cannot realize turn with Unpitched srcObj')
```
mscuthbert, line 1721: we can change everything to `if not srcObj.pitches` and handles Rests, etc. -- but the typing is set to music21.note.Note so it's an error if any routine above is calling realize on anything that is not a Note object.

music21/musicxml/m21ToXml.py
```diff
@@ -1453,6 +1453,10 @@ def accidentalToMx(self, a):
         return mxAccidental
 
 
+    def accidentalToMxAccidentalMark(self, a: pitch.Accidental) -> Element:
+        return self.accidentalToMx(a, elName='accidental-mark')
```
mscuthbert, line 1457: all great -- thanks for noticing this.

music21/musicxml/testPrimitive.py
```diff
@@ -2724,7 +2724,7 @@
           <ornaments>
             <turn/>
             <accidental-mark placement="above">sharp</accidental-mark>
-            <accidental-mark placement="above">three-quarters-flat</accidental-mark>
+            <accidental-mark placement="below">flat</accidental-mark>
```
mscuthbert, line 2727: good catch, but can we change the mark itself to three-quarters-flat?  It's an important test of ability to do microtones.

music21/musicxml/test_xmlToM21.py
```diff
@@ -619,6 +619,18 @@ def testOrnamentC(self):
                     count += 1
         self.assertEqual(count, 1)
 
+        uppercount = 0
+        lowercount = 0
```
mscuthbert, line 623: camelCase or underscore_case.  (for accessibility)

music21/pitch.py
```diff
@@ -134,6 +163,28 @@ def _sortModifiers():
...
+        return True
+    # check against official modifiers (e.g. '--')
+    if name in accidentalNameToModifier.values():
+        return True
+    # check against alternate supported names (e.g. 'eses')
+    if name in alternateNameToAccidentalName:
+        return True
+    return False
+
+def makeAccidentalName(name: str) -> str | None:
```
mscuthbert, line 178: how about standardizeAccidentalName?

music21/pitch.py
```diff
@@ -134,6 +163,28 @@ def _sortModifiers():
...
+    if name in accidentalNameToModifier:
+        return name
+    if name in modifierToAccidentalName:
+        return modifierToAccidentalName[name]
+    if name in alternateNameToAccidentalName:
+        return alternateNameToAccidentalName[name]
+    return None
```
mscuthbert, line 185: could be better to raise an exception, then standardizeAccidentalName can always return a string and be used in string contexts.

music21/expressions.py
```diff
@@ -517,21 +519,205 @@ def fillListOfRealizedNotes(
...
+                        cautionaryPitchClass=cautionaryPitchClass,
+                        cautionaryAll=cautionaryAll,
+                        overrideStatus=overrideStatus,
+                        cautionaryNotImmediateRepeat=cautionaryNotImmediateRepeat,
+                        lastNoteWasTied=lastNoteWasTied)
+            return
+
+        if isinstance(self, (GeneralMordent, Trill)):
+            # just one otherPitch
+            p = self.otherPitches[0]
```
mscuthbert, line 585: what if otherPitches is empty?

music21/stream/base.py
```diff
@@ -6830,6 +6831,27 @@ def makeAccidentals(
...
+
+                    orn.updateAccidentalDisplay(
+                        pitchPast=pitchPast,
+                        pitchPastMeasure=pitchPastMeasure,
+                        alteredPitches=alteredPitches,
+                        cautionaryPitchClass=cautionaryPitchClass,
+                        cautionaryAll=cautionaryAll,
+                        overrideStatus=overrideStatus,
+                        cautionaryNotImmediateRepeat=cautionaryNotImmediateRepeat,
+                        lastNoteWasTied=False)
```
mscuthbert, line 6851: why is this always false?  There are notes tied into turns.

music21/stream/base.py
```diff
@@ -6858,11 +6880,53 @@ def makeAccidentals(
                     if n.tie is not None and n.tie.type != 'stop':
                         seenPitchNames.add(p.nameWithOctave)
 
+                    # handle this note-in-chord's ornaments' otherPitches
```
mscuthbert, line 6883: this is repeated three times in a routine that is already way too long.  Can't it be handled by a separate helper function?

music21/stream/base.py
```diff
@@ -10110,10 +10174,22 @@ def pitches(self) -> list[pitch.Pitch]:
...
             if isinstance(e, key.Key):
                 continue  # has .pitches but should not be added
-            # both GeneralNotes and Stream have a pitches properties; this just
-            # causes a recursive pitch gathering
-            elif isinstance(e, (note.GeneralNote, Stream)):
+            elif isinstance(e, Stream):
+                # Stream has a pitches properties; this just
+                # causes a recursive pitch gathering
                 post.extend(list(e.pitches))
+            elif isinstance(e, note.GeneralNote):
```
mscuthbert, line 10181: disagree, please revert -- `.pitches` on Stream should not include ornamental pitches -- that would be a major change in behavior.  If people want that, they can call `score.realizeOrnaments().pitches`.

music21/stream/base.py
```diff
@@ -6830,6 +6831,27 @@ def makeAccidentals(
...
+                        pitchPast=pitchPast,
+                        pitchPastMeasure=pitchPastMeasure,
+                        alteredPitches=alteredPitches,
+                        cautionaryPitchClass=cautionaryPitchClass,
+                        cautionaryAll=cautionaryAll,
+                        overrideStatus=overrideStatus,
+                        cautionaryNotImmediateRepeat=cautionaryNotImmediateRepeat,
+                        lastNoteWasTied=False)
+
+                    pitchPast += orn.otherPitches
```
mscuthbert, line 6853: even from this I still have no idea what .otherPitches is.  Can't otherPitches be None?

mscuthbert (changes requested):
Great work!  Love it!  Sorry that most individual comments are on the things I dislike, but it's a significant contribution.

Things requested to change are:
1) please document interface before requesting review next time (I'd prefer to review an empty interface w/o implementation rather than an implementation without docs).  It slows down the review immensely to have to guess what a routine does and how it handles corner cases.
2) otherPitches doesn't have any meaning to me -- what is this?
3) getSize -- issues with naming and the notion that the size returned will later be modified by key context etc.  -- one routine is preferred that takes in all the information it needs to spit out the exactly right interval.  Try to think of the use case where someone want to measure an ornament w/o realizing it.
4) don't change the behavior of Stream.pitches.
5) getters cannot raise exceptions.  See note for reason.
6) makeAccidentals is already way too long and complicated -- use helper functions to remove redundancy.

music21/stream/base.py
mscuthbert: btw -- `if n.pitches: p = n.pitches[-1]` will work for notes/chords/rests/unpitched etc.

gregchapman-dev:
Thank you for the quick review (and sorry for the lack of documentation).  I agree with nearly all of what you said, and the rest just raises questions, not disagreement. Hoping to find some time to work through most of this tomorrow.

music21/expressions.py
gregchapman-dev: Yeah, I think this needs to take the same params as realize().

music21/expressions.py
gregchapman-dev: Sounds good to me.

music21/expressions.py
gregchapman-dev: Ah yes, this is where I saw the Trill on an Unpitched (snare drum), which crashed until I added this code.  I'll fix the typing to be GeneralNote, and use `if not srcObj.pitches` to defend.

music21/musicxml/testPrimitive.py
gregchapman-dev: I changed this so that it would be something Finale was happy to render (so I could see it during debugging).  But I'm fine to change it back once I have things working again.

music21/pitch.py
gregchapman-dev: I like it!  Thanks.

music21/stream/base.py
gregchapman-dev: There can be notes tied into the turn's upper and lower "other" notes?  The turn's "main" note was already handled, this call is only about the "other" notes.

music21/stream/base.py
gregchapman-dev: I was wondering about that.  And in my defense I _did_ choose not to change GeneralNote.pitches to include ornamental pitches.  That seemed _way_ too much of a change.  (Hmmm... maybe otherPitches should be called ornamentalPitches.)

music21/expressions.py
gregchapman-dev: If I return GenericInterval(2), then I have a very common situation (`size` has not been explicitly set) where `size` will incorrectly return GenericInterval(2), and `getSize(src)` will return the correct interval.  That feels _really_ weird to me.  I would almost rather remove the `size` property entirely, and deal with the compatibility consequences (including in ornamentRecognizer).

music21/expressions.py
gregchapman-dev: I think not; it is implied by the class name (maybe it should even be a class variable? But still be gettable as a property.). 

I added this because I needed it myself at some point.  I forget where, but I can find it if you care to see it's use.

music21/expressions.py
gregchapman-dev: Huh.  I don't seem capable of anything between "no typing" and "always typing unless I really can't".  Unless you have a simple rule I can apply to find a comfortable spot in between those two places?

music21/stream/base.py
gregchapman-dev: I started out with `orn.pitches`, which seems reasonable (what pitches do you play when you play this ornament), but what I really want is "what pitches _besides the main note's pitch_ do you play when you play this ornament", so I came up with `orn.otherPitches`, which makes sense in context, but doesn't provide any context itself.  How about `orn.ornamentalPitches`?  They are the pitches that are added by the ornament.

music21/expressions.py
gregchapman-dev: Nope, can't do that exactly. `Ornament.updateAccidentalDisplay` doesn't have the necessary params in hand to pass to such a method.  It needs to be able to get and modify the ornamental pitches directly, assuming someone else has set them up.  So I will still need `Ornament.resolveOrnamentalPitches(src and maybe keysig)` as well as `Turn.upperOrnamentalPitch`,` Turn.lowerOrnamentalPitch`, and `Mordent/Trill.ornamentalPitch`, and maybe `Ornament.ornamentalPitches`, too, we'll see.

music21/expressions.py
gregchapman-dev: I'm going to return None for now.  We can discuss further.

music21/stream/base.py
mscuthbert: ah!  Yes, that makes sense!

music21/stream/base.py
mscuthbert: Ah, how about `ornamentalPitches`?  only a few more letters, but it makes clear that it doesn't include the main pitch

music21/expressions.py
mscuthbert: I'm fine with breaking the compatibility on this and removing .size -- since all mordents are generic interval of 2 of some sort anyhow.

music21/expressions.py
mscuthbert: okay

music21/expressions.py
mscuthbert: a trill on unpitched, yes.  It'd be the one trill to have a size of unison.  It's essentially a timpani roll.

music21/expressions.py
mscuthbert: I mean, it's fine to have it here, just redundant.  Generally you'll want typing on all inputs and outputs of functions, so the internal things that need typing are attributes (since they'll be queried/set and need to know what's allowed there) and especially variables that have containers, like `notes: list[note.Note] = []` 

Are you using an IDE that infers type when you mouse over a variable?  If the IDE can't infer the type, then it needs to be typed.  If a variable is a return value from a typed function (like enumerate) then it doesn't need a type (like nodeIdx), and `int + int` magically types an int to still be an int.

music21/stream/base.py
mscuthbert: Part of the reason is that most analysis ignores the ornamental notes.  So if you had three whole notes, C, C# and D, you could say that 2/3 of the notes fit in the key of C major,  but if you added `tr#` to the C#, then it'd be more like 10% do.  I wish more people would realize ornaments and expand repeats before doing analysis, but there needs to be a way that analytical methods that don't expand still work.

music21/expressions.py

music21/expressions.py
gregchapman-dev: There is none.

music21/expressions.py

music21/expressions.py

music21/expressions.py
gregchapman-dev: Done

music21/expressions.py

music21/expressions.py
gregchapman-dev: Oooh.  Done.  I tried to get ornamentRecognizer to work (it was setting/getting trill.size), but I failed.  Take a look at my attempt and see if you can help me.

music21/expressions.py

music21/expressions.py

music21/expressions.py
gregchapman-dev: Done (it's a private class variable now, with a property getter that returns it)

music21/expressions.py
gregchapman-dev: resolveOrnamentalPitches now.

music21/expressions.py
gregchapman-dev: I think I have to make another change to get an Unpitched Trill (e.g. snare drum roll) to work properly.  I don't know how a timpani roll would work in music21.  It's pitched, so a Trill would imply changing pitch.  Which ain't what a timpani roll is supposed to do.

music21/expressions.py
gregchapman-dev: Done.

gregchapman-dev: Oh wait, I did this on a different method.  Not on this one.  Still not done.

music21/expressions.py
gregchapman-dev: This is now documented in resolveOrnamentalAccidentals header doc.  Is that OK, or should I do it somewhere else?

music21/expressions.py
gregchapman-dev: Done.  It's still called getSize, but it takes the same params as realize() (and realize() calls getSize with those params).

music21/expressions.py

music21/expressions.py
gregchapman-dev: size is gone, so this is done.

music21/expressions.py
gregchapman-dev: Still need to do this in a few places, I think.

music21/expressions.py

music21/expressions.py
gregchapman-dev: Done.

music21/musicxml/testPrimitive.py
gregchapman-dev: Still need to do this.

music21/musicxml/test_xmlToM21.py
gregchapman-dev: Done.

music21/pitch.py
gregchapman-dev: Done.

music21/pitch.py
gregchapman-dev: Done.

music21/stream/base.py
gregchapman-dev: Done.  I added a new stream.ornamentalPitches property that gives me the rest of what I needed in makeAccidentalsForMeasureStream (or whatever that method is called).

music21/stream/base.py
gregchapman-dev: Done.  It's a really good name.

gregchapman-dev:
Almost all review has been incorporated.  A bunch more doc and test (and a couple more fixes) to go.

gregchapman-dev:
@mscuthbert OK, the last remaining test failure is in ornamentRecognizer.py (2nd of 3 test cases), where the test code recognizes a Trill that goes down a whole step.  The old implementation of Trill supported this by letting you set size to Interval('M-2'), even though Trill.direction is still 'up' in that case (ew).  But the new implementation knows full well that Trill goes up (and the non-existent InvertedTrill would go down if it existed), so the computed size is Interval('M2') for this Trill, and the test fails (M-2 != M2).

So, how do I fix this?  Should I implement InvertedTrill and make ornamentRecognizer generate an InvertedTrill for recognized trills that go down?  Or should I modify the test to no longer attempt to recognize an InvertedTrill?

music21/expressions.py
gregchapman-dev: Done (question about ornamentRecognizer is at the top level with more details now).

music21/expressions.py

music21/musicxml/testPrimitive.py
gregchapman-dev: Done.

music21/expressions.py
gregchapman-dev: Done.

music21/expressions.py
gregchapman-dev: Done.

music21/expressions.py
gregchapman-dev: Still working on this.

gregchapman-dev:
Still to do: 

1. Fix ornamentRecognizer.py.  There are actually several failing tests there, some because of unexpectedly positive size (inverted trills), but also some where the size is just wrong (A2 vs M2, for example).  I can't tell yet if the test is wrong/obsolete or my code is wrong.

2. I'm betting I'll need more test coverage.

music21/expressions.py
gregchapman-dev: Done.

gregchapman-dev:
I have implemented InvertedTrill, fixed TrillRecognizer to recognize InvertedTrills, and got the TrillRecognizer's tests passing again (including fixing a bug in Turn/GeneralMordent/Trill.getSize() having to do with transposing pitches with implicit octaves, which those tests were triggering).  Let's see how test coverage is looking.

gregchapman-dev:
I have `Trill`/`GeneralMordent`/`Turn` on `Unpitched` working and tested.  They are realized as lists of `Unpitched` with the usually expected duration (I didn't change the duration-computing code), and `.ornamentalPitches` returns an empty tuple.  `getSize(unp)` returns `interval.Interval('P1')`.  

Still working on improving test coverage.

gregchapman-dev:
Coverage is good, but I am not content: I need to add some tests for XML parse and write.

gregchapman-dev:
@mscuthbert I think I'm done here, ready for your re-review.  Please do validate that my changes to pitch.py's test expectations are correct (that's the very last commit above).

music21/test/test_expressions.py
```diff
@@ -161,11 +163,400 @@ def testTrillExtensionA(self):
...
+        with self.assertRaises(expressions.ExpressionException):
+            expressions.Turn().getSize(note.Note('C4'), which='bad: not upper or lower')
+
+    def testUnpitchedOrnaments(self):
         unp = note.Unpitched()
+
         mord = expressions.Mordent()
-        with self.assertRaises(TypeError):
-            mord.realize(unp)  # type: ignore
+        realized = mord.realize(unp)  # type: ignore
```
jacobtylerwalls, line 472: Do you think we should loosen the typing of the `srcNote` arg to `realize` and `fillListOfRealizedNotes` to `Note | Unpitched`?

music21/stream/base.py
```diff
@@ -10116,6 +10188,31 @@ def pitches(self) -> list[pitch.Pitch]:
...
+        '''
+        post = []
+        for e in self.elements:
+            if isinstance(e, Stream):
+                # recurse
+                post.extend(e.ornamentalPitches)
+            elif hasattr(e, 'expressions'):
+                for orn in e.expressions:
+                    if isinstance(orn, expressions.Ornament):
+                        post.extend(list(orn.ornamentalPitches))
```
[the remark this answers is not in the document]
gregchapman-dev: Done.

music21/test/test_expressions.py
```diff
@@ -161,11 +163,400 @@ def testTrillExtensionA(self):
...
+        self.assertEqual(len(realized), 3)
+        self.assertEqual(len(realized[0]), 2)
+        self.assertEqual(type(realized[0][0]), note.Unpitched)
```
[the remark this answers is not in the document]
gregchapman-dev: Done.

music21/test/test_expressions.py
gregchapman-dev: Yes.  Done.

gregchapman-dev:
I have removed the microtones-to-quarter-tones operation from pitch.accidental's float setter, and added it to the resolveOrnamentalPitches methods in Trill/Turn/GeneralMordent.  Note that I had to work around a different bug in transpose  where microtones get _completely_ lost if you transpose in place (that I will be writing an issue for shortly).  I also addressed Jacob's review comments.

gregchapman-dev:
I've decided I'm not done yet.  I've been playing with my Humdrum writer, making it write from these new Ornament objects, and I'm finding that sometimes the Ornament ends up with the wrong size, due to the fact that the ornamentalPitch was modified by an accidental on a note earlier in the measure (and no-one knows, so it ends up not modified).  This is actually a bug in my Humdrum importer, where I _know_ what the accidental should be, but I don't set accidentalName because I don't want to force it to be visible.  So I think @mscuthbert's idea earlier (that ornament.accidentalName should actually be an Accidental object instead of a name string) is correct, so my parser can set the accidental, and then set that accidental's displayStatus to True if it is supposed to be visible.  Then resolveOrnamentalPitches will need to copy that displayStatus to the resolved ornamentalPitch.accidental.  (And all of this is doubled for Turns, where there are two ornamentalPitches, two accidentalNames, etc).

gregchapman-dev:
OK, I've made that change.  I think I might be done here.

music21/expressions.py

mscuthbert: I wonder if eventually "Roll" should be a separate class from Trill -- but can have "displayAsTrill" set or something.  They use the same notation, but conceptually so different.

music21/expressions.py
mscuthbert: (usually that's doc'd as "-> None"

music21/expressions.py
```diff
@@ -835,27 +1093,31 @@ class HalfStepMordent(Mordent):
...
         # no accidental supported here, just "HalfStep"
-        kw = keywords.copy()
-        if 'accid' in kw:
-            del kw['accid']
-        super().__init__(**kw)
-        self._size: interval.IntervalBase = interval.Interval('m-2')
+        if 'accidental' in keywords:
+            raise ExpressionException('Cannot initialize HalfStepMordent with accidental')
+        super().__init__(**keywords)
+        self._minorSecondDown: interval.IntervalBase = interval.Interval('m-2')
```
mscuthbert, line 1104: I need to implement `FrozenInterval` to make it so we can cache these things and prevent them from being modified elsewhere.  If it existed then these intervals could be cached.

music21/expressions.py
```diff
@@ -1266,30 +1619,39 @@ def realize(
...
-                        n.pitch.accidental = currentKeySig.accidentalByStep(n.step)
+        if isTransposed:
+            setAccidentalFromKeySig = self._setAccidentalFromKeySig
+            if setAccidentalFromKeySig:
+                for n in trillNotes:
+                    if t.TYPE_CHECKING:
+                        # for some reason, this assertion doesn't convince mypy that n
+                        # has a pitch attribute, so we have to "type: ignore" below.
+                        assert isinstance(n, note.Note)
+                    if n.pitch.nameWithOctave != srcObj.pitch.nameWithOctave:  # type: ignore
```
mscuthbert, line 1648: mypy should be convinced -- the problem is that srcObj might be a Unpitched object

music21/expressions.py
```diff
@@ -1543,81 +1924,210 @@ def name(self) -> str:
...
+            else:
+                p.updateAccidentalDisplay(
+                    pitchPast=pitchPast,
+                    pitchPastMeasure=pitchPastMeasure,
+                    alteredPitches=alteredPitches,
+                    cautionaryPitchClass=cautionaryPitchClass,
+                    cautionaryAll=cautionaryAll,
+                    overrideStatus=overrideStatus,
+                    cautionaryNotImmediateRepeat=cautionaryNotImmediateRepeat,
+                    lastNoteWasTied=False)
```
mscuthbert, line 2101: better as a loop:

```
for p, acc in ((self.upperOrnamentalPitch, self.upperAccidental), (self.lowerOrnamentalPitch, self.lowerAccidental)):
    ...
```

then we're saving 15 lines of code.

music21/expressions.py
```diff
@@ -1715,16 +2225,16 @@ def realize(self, srcObj: note.Note, *, inPlace=False):
...
         if srcObj.duration.quarterLength == 0:
             raise ExpressionException('Cannot steal time from an object with no duration')
 
-        # here we compute transposeIntervals, and invert them if self.isInverted
+        # here we compute transposeIntervals, and invert them if self._isInverted
+        currentKeySig = keySig
+        if currentKeySig is None:
+            currentKeySig = srcObj.getContextByClass(key.KeySignature)
+            if currentKeySig is None:
+                currentKeySig = key.KeySignature(0)
```
mscuthbert, line 2237: as above

music21/expressions.py
```diff
@@ -1750,19 +2260,25 @@ def realize(self, srcObj: note.Note, *, inPlace=False):
...
+            secondTransposeInterval = self.getSize(srcObj, 'lower', keySig=currentKeySig)
         else:
-            firstTransposeInterval = self.getLowerSize(srcObj.pitch)
-            secondTransposeInterval = self.getUpperSize(srcObj.pitch)
+            firstTransposeInterval = self.getSize(srcObj, 'lower', keySig=currentKeySig)
+            secondTransposeInterval = self.getSize(srcObj, 'upper', keySig=currentKeySig)
 
-        turnNotes: list[note.Note] = []
+        # no need to check both intervals, they will both be perfectUnison, or neither will be.
+        isTransposed: bool = firstTransposeInterval != interval.Interval('P1')
```
mscuthbert, line 2271: ah! here's (and once above) is the other place to avoid creating an Interval just to check if it's a unison.

music21/pitch.py
```diff
@@ -4511,6 +4561,16 @@ def transpose(
         >>> pc6.transpose(10, inPlace=True)
         >>> pc6.spellingIsInferred
         True
+
+        Test an issue with inPlace not setting microtone.
+
+        >>> dFlatAndAHalf = pitch.Pitch('D2-`')
```
mscuthbert, line 4567: can you rewrite without the back-tick but instead explicitly set a 3qtr flat accidental object?  I'm planning on removing the backtick and ~ abbreviations -- probably a mistake to include them originally.

music21/stream/base.py
```diff
@@ -6682,6 +6682,39 @@ def makeBeams(self, *, inPlace=False, setStemDirections=True, failOnNoTimeSignat
...
+        pitchPast: list[pitch.Pitch] | None = None,
+        pitchPastMeasure: list[pitch.Pitch] | None = None,
+        otherSimultaneousPitches: list[pitch.Pitch] | None = None,
+        alteredPitches: list[pitch.Pitch] | None = None,
+        cautionaryPitchClass: bool = True,
+        cautionaryAll: bool = False,
+        overrideStatus: bool = False,
+        cautionaryNotImmediateRepeat: bool = True,
+    ):
+        for orn in noteOrChord.expressions:
```
mscuthbert, line 6698: -1:  undocumented, untested new addition to Stream.  Can this be documented and moved to stream.makeNotation....  -- I'm trying to trim down the complexity of the Stream class (and length of this file) and this seems highly unlikely for users to call on their own.  makeNotation.py is where the guts of `Stream.makeNotation()` needs to live.  I have as a todo to move makeBeams, makeAccidentals, etc. to makeNotation.py as well, but because they've appeared in demos, papers, etc. they're harder to move.

music21/stream/base.py
```diff
@@ -6831,26 +6864,16 @@ def makeAccidentals(
...
-                        alteredPitches=alteredPitches,
-                        cautionaryPitchClass=cautionaryPitchClass,
-                        cautionaryAll=cautionaryAll,
-                        overrideStatus=overrideStatus,
-                        cautionaryNotImmediateRepeat=cautionaryNotImmediateRepeat,
-                        lastNoteWasTied=False)
-
-                    pitchPast += orn.otherPitches
+                # handle this note's ornaments' ornamentalPitches
+                self.makeOrnamentalAccidentals(
```
mscuthbert, line 6868: just call `makeNotation.makeOrnamentalAccidentals(self, e, ...`

mscuthbert: There was also a note in the last review against writing this three times -- can it be written once with a helper call, like an embedded secondary function?

music21/stream/base.py
```diff
@@ -10174,22 +10182,35 @@ def pitches(self) -> list[pitch.Pitch]:
...
-                        if isinstance(orn, expressions.Ornament):
-                            # Figure out what the other pitches are, in case that hasn't
-                            # been done yet (this is a no-op if it has already been done).
-                            orn.resolveOtherPitches(e.pitches[-1])
-                            # Add them to the list
-                            post.extend(list(orn.otherPitches))
+        return post
+
+    @property
+    def ornamentalPitches(self) -> list[pitch.Pitch]:
```
mscuthbert, line 10192: remove from stream -- it can be in expressions.py, etc.  see above.  Unlikely to be used by most users and reducing music21 complexity is a goal.

mscuthbert: Or since it's only used in makeNotation it can I suppose live in makeNotation.py

music21/test/test_expressions.py
mscuthbert: Does this still need type: ignore now?

mscuthbert (changes requested):
Great work!  You'll be relieved to see that there are no conceptual changes requested, nothing about "wouldn't this be better if .accidental took a NeoRiemannianTransformation instead?" etc.  :-)

There are some places where I'd like refactors for clarity (and making Stream.base not so darn big) and some docs that are missing, etc.  

There are some significant typing improvements that can be done after this is merged.

Thanks!

mscuthbert:
The next review should be light enough that I'll get it on a mid-week review schedule when it comes in.  Needed to wait for a holiday for this one.

music21/expressions.py
```diff
@@ -62,8 +65,13 @@ def realizeOrnaments(srcObject):
     .. image:: images/expressionsMordentRealize.*
          :width: 218
 
-    :type srcObject: base.Music21Object
+    :type srcObj: note.GeneralNote
```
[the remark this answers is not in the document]
gregchapman-dev: ok.

music21/expressions.py
```diff
@@ -74,10 +82,15 @@ def realizeOrnaments(srcObject):
 
         loopBuster = 100
         while loopBuster:
+            if t.TYPE_CHECKING:
+                # if it was set to None, we break out of the loop, so we won't get here
+                assert srcObject is not None
```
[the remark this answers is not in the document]
gregchapman-dev: OK, I'll leave the better typing to you.

music21/expressions.py
```diff
@@ -463,17 +479,19 @@ def __init__(self, **keywords) -> None:
...
-                           list[note.Note]]:
+    def realize(
+        self,
+        srcObj: note.Note | note.Unpitched,
+        *,
+        keySig: key.KeySignature | None = None,
+        inPlace: bool = False
+    ) -> tuple[list[note.Note | note.Unpitched],
+                note.Note | note.Unpitched | None,
+                list[note.Note | note.Unpitched]]:
```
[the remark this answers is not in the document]
gregchapman-dev: Yeah, I'll leave that to you.  I'm guessing that NoteOrUnpitched might want to be defined someplace fairly central (in note.py maybe?).

music21/expressions.py
```diff
@@ -625,99 +773,177 @@ def name(self) -> str:
...
+            keySig = srcObj.getContextByClass(key.KeySignature)
+            if keySig is None:
+                keySig = key.KeySignature(0)
+
+        srcPitch: pitch.Pitch = srcObj.pitches[-1]
+
+        ornamentalPitch: pitch.Pitch = copy.deepcopy(srcPitch)
+        ornamentalPitch.accidental = None
+        if ornamentalPitch.octave is None:
+            ornamentalPitch.octave = ornamentalPitch.implicitOctave
```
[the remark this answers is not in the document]
gregchapman-dev: Ha!  I wish I had remembered it.  It caused a test failure where a half-step up from B (implicit octave = 4) was C (implicit octave = 4), so that half step up turned into a major 7th down.  Took me several minutes to figure out why.

music21/expressions.py
```diff
@@ -742,36 +968,37 @@ def realize(self, srcObj: note.Note, *, inPlace=False):
...
-        for noteIdx, n in enumerate(mordNotes):
-            noteNum: int = noteIdx + 1
-            if n.pitch.accidental is None and noteNum == 2:
-                n.pitch.accidental = currentKeySig.accidentalByStep(n.pitch.step)
+        if transposeInterval != interval.Interval('P1'):
+            # second (middle) note might need an accidental from the keysig (but
+            # only if it doesn't already have an accidental, from self.accidental)
+            for noteIdx, n in enumerate(mordNotes):
+                if t.TYPE_CHECKING:
+                    assert isinstance(n, note.Note)
```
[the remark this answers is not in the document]
gregchapman-dev: Ok, I'll leave that to you.

music21/expressions.py
gregchapman-dev: Ok, I'll leave that to you.

music21/expressions.py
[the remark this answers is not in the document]
gregchapman-dev: Done.

music21/expressions.py
```diff
@@ -535,88 +672,99 @@ def updateAccidentalDisplay(
...
-                        pitchPastMeasure=pitchPastMeasure,
-                        alteredPitches=alteredPitches,
-                        cautionaryPitchClass=cautionaryPitchClass,
-                        cautionaryAll=cautionaryAll,
-                        overrideStatus=overrideStatus,
-                        cautionaryNotImmediateRepeat=cautionaryNotImmediateRepeat,
-                        lastNoteWasTied=lastNoteWasTied)
+        Updates accidental display for an Ornament's ornamental pitch(es).
+        Defined exactly like Pitch.updateAccidentalDisplay with two changes:
+        1. We pay no attention to ties, since ornamental notes cannot be tied.
```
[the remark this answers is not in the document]
gregchapman-dev: Done (and found lots of other doc changes I wanted to make).

music21/expressions.py
gregchapman-dev: Done.

music21/expressions.py
gregchapman-dev: Nice!  Done.

music21/expressions.py
mscuthbert: It's not n it's srcObj I'm guessing.  You never asserted that IT was a note.

gregchapman-dev: Oh. Duh. 🤦🏻‍♂️

music21/test/test_expressions.py
gregchapman-dev: No. Fixed that.

music21/expressions.py
gregchapman-dev: Done.

music21/pitch.py
gregchapman-dev: Done.

music21/stream/base.py
gregchapman-dev: I tried, but the helper call would need to do different things between calling `e.pitch.updateAccidentalDisplay` and calling `makeNotation.makeOrnamentalAccidentals(e)` in each case.  I couldn't figure out a way to get that right.

But I have moved `makeOrnamentalAccidentals` to makeNotation.py, as requested.

gregchapman-dev: Done.

music21/stream/base.py
gregchapman-dev: Moved, and documented (but no doctests).  This routine is actually well covered by the test suite; do you also want doctests?

gregchapman-dev: Looks like both moved routines (`makeOrnamentalAccidentals` and `ornamentalPitches(s)` have 100% coverage.

gregchapman-dev:
I believe I have addressed all @mscuthbert's latest review comments.

music21/stream/base.py
mscuthbert: okay -- I'll try to get the function compacted myself later.

music21/stream/base.py
mscuthbert: Great -- no doctests needed if it's well tested and well documented.  Thanks for moving it.

music21/stream/base.py
```diff
@@ -6865,7 +6831,7 @@ def makeAccidentals(
                     tiePitchSet.add(e.pitch.nameWithOctave)
 
                 # handle this note's ornaments' ornamentalPitches
-                self.makeOrnamentalAccidentals(
+                makeNotation.makeOrnamentalAccidentals(
```
mscuthbert: Oh! it didn't even need `self` -- then definitely best not to live on Stream!

music21/stream/makeNotation.py
```diff
@@ -1708,6 +1709,68 @@ def makeAccidentalsInMeasureStream(
...
+    found in the ornaments on the notes and chords.
+
+    If you want a list of _all_ the pitches in a stream, including the ornamental
+    pitches, you can call s.pitches and makeNotation.ornamentalPitches(s),
+    and then combine the two resulting lists into one big list.
+    '''
+    from music21.stream import Stream
+    post = []
+    for e in s.elements:
+        if isinstance(e, Stream):
```
mscuthbert, line 1728: this can be done without the import with `if 'Stream' in e.classes` which is why .classes is still there.  NBD!

gregchapman-dev:
Thanks for taking it!

PR #1547, opened 2023-03-29
Implement useful `PercussionChord.pitches` property

jacobtylerwalls:
`PercussionChord.pitches` until now just fell back to the NotRest implementation that returns `()`. Now, it does something useful: setting, has parity with Chord; getting, has additional safety to filter out Unpitched instances in the `._notes` array.

TODO statement is just copied from Chord.pitches.


mscuthbert:
THANKS!

PR #1548, opened 2023-03-30
fix opfrac for small numbers

### LINKED ISSUE #1544: `opFrac()` could prefer 0.0 to Fraction(0, 1)
**music21 version**
dev

**Problem summary**
Low impact, but opFrac returns Fraction(0, 1) instead of 0.0 for small floats. Should it prefer 0.0 if the numerator is 0?

**Steps to reproduce**

```
>>> from music21.common import opFrac
>>> opFrac(0.000001)
Fraction(0, 1)
```

Expected: 0.0
Actual: Fraction(0, 1)

mscuthbert:
Ensure that opFrac never returns Fraction(0, 1)

Fixes #1544 

Thanks @jacobtylerwalls and @vanderstel

### DIFF
--- a/music21/common/numberTools.py
+++ b/music21/common/numberTools.py
@@ -286,13 +286,14 @@ def opFrac(num: OffsetQLIn | None) -> OffsetQL | None:
         #    (denominator & (denominator-1)) != 0
         # which is a nice test, but denominator here is always a power of two...
         # unused_numerator, denominator = num.as_integer_ratio()  # too slow
-        if num < 0.0000011:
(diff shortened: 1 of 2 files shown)


mscuthbert:
Noticed that these other Fractions were returned:

0.9999998 -> Fraction(1, 1)
1.4999998 -> Fraction(3, 2)

etc.  also fixed.

Slow down of 4% when a float is converted to a fraction; no change for common cases.

Also fixed a mypy warning about unchecked types in MusicXML which also found an error.

PR #1556, opened 2023-04-30
Fix stripTies when accidentals are neutral & none

TimFelixBeyer:
When a stream contains two notes/chords that contain the same pitch, but one has a "neutral" accidental and the other one doesn't have an accidental at all (None), the tied notes weren't matched even though they should. 
This happens e.g., for quite a lot musicxml files from the ASAP dataset.

It's debatable whether this change should really take place in the pitch.Pitch class, where the `__eq__` method could be adapted to interpret a "neutral" accidental to be equal to a None accidental.  
This seemed like the less intrusive option, but I'm happy to change the PR as well. 
Let me know what you think!

### DIFF
--- a/music21/features/jSymbolic.py
+++ b/music21/features/jSymbolic.py
@@ -1828,7 +1828,7 @@ class StrengthOfStrongestRhythmicPulseFeature(featuresModule.FeatureExtractor):
     ...     p.insert(0, tempo.MetronomeMark('Langsam', 70))
     >>> fe = features.jSymbolic.StrengthOfStrongestRhythmicPulseFeature(sch)
     >>> fe.extract().vector[0]
-    0.857...
+    0.853...
     '''
     id = 'R4'
 
@@ -1855,7 +1855,7 @@ class StrengthOfSecondStrongestRhythmicPulseFeature(
     ...     p.insert(0, tempo.MetronomeMark('Langsam', 70))
     >>> fe = features.jSymbolic.StrengthOfSecondStrongestRhythmicPulseFeature(sch)
     >>> fe.extract().vector[0]
-    0.119...
+    0.121...
     '''
     id = 'R5'
 
@@ -1891,7 +1891,7 @@ class StrengthRatioOfTwoStrongestRhythmicPulsesFeature(
     ...     p.insert(0, tempo.MetronomeMark('Langsam', 70))
     >>> fe = features.jSymbolic.StrengthRatioOfTwoStrongestRhythmicPulsesFeature(sch)
     >>> fe.extract().vector[0]
-    7.2
+    7.0
 
     '''
     id = 'R6'
@@ -1927,7 +1927,7 @@ class CombinedStrengthOfTwoStrongestRhythmicPulsesFeature(
     ...     p.insert(0, tempo.MetronomeMark('Langsam', 70))
     >>> fe = features.jSymbolic.CombinedStrengthOfTwoStrongestRhythmicPulsesFeature(sch)
     >>> fe.extract().vector[0]
-    0.976...
+    0.975...
     '''
     id = 'R7'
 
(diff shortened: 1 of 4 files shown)


mscuthbert:
Problem diagnosis probably correct.  Matching on midi is not the solution.  music21 matches based on spelling not midi.  I'll need to think about whether I want accidental == None to be equal to accidental == natural.  Probably not as it would be a major breaking change, so might need to special case this.

TimFelixBeyer:
Fair points, would you prefer this instead?
```
for pitchIndex in range(len(nLast.pitches)):
    pLast = nLast.pitches[pitchIndex]
    pInner = nInner.pitches[pitchIndex]
    if (pLast.octave != pInner.octave
            or pLast.step != pInner.step
            or pLast.microtone != pInner.microtone):
            return False
        # Special case: We have to treat None and "neutral" accidentals as equal
        if pLast.accidental != pInner.accidental:
            aLast = pLast.accidental
            aInner = pInner.accidental
            if not (aLast is None and aInner.name == "natural"
                         or aLast.name == "natural" and aInner is None):
                return False
```

TimFelixBeyer:
I just realised that there seems to be another issue with stripping chords of ties.
A failing example is IMSLP [#578434](https://imslp.org/wiki/File:PMLP931026-Dres-America_Schottisch.mid). 
```
from music21 import converter
s = converter.parse("IMSLP578434-PMLP931026-Dres-America_Schottisch.mid").flat.stripTies()
for n in s.notes:
    if n.tie is not None:
        print(f"Tie found {n}, {n.tie}, {n.offset}")
```
This occurs before and after implementing my suggestion - I will investigate.

mscuthbert:
> I just realised that there seems to be an additional issue with stripping chords of ties. 

This is intentional, I believe -- Stripties is designed to merge all notes that can be fully merged and have the same sound effect.  A chord with some notes tied and some notes not cannot be merged w/ the next chord w/o altering the sound.

TimFelixBeyer:
> > I just realised that there seems to be an additional issue with stripping chords of ties.
> 
> This is intentional, I believe -- Stripties is designed to merge all notes that can be fully merged and have the same sound effect. A chord with some notes tied and some notes not cannot be merged w/ the next chord w/o altering the sound.

If there was a chord A (1 quarter length) with notes XYZ and a chord B (same length) with notes TUVW, where note X is tied to T and Y is tied to U but Z is not tied to V and W is not tied to anything, I think it could be replaced like this:

```
X——T    X+T
Y——U —> Y+U
Z  V    Z    V
   W         W
```
We’d now have one chord XY+TU with length 2 quarter notes, one single note Z, and one chord VW.
I believe the result would sound the same and this would be the expected behavior for `stripTies` from my perspective as a user, maybe I am missing something though.

mscuthbert:
This would potentially take a single-voice measure and replace it with a measure with multiple voices.  This is exactly the sort of complexity that the stripTies routine is trying to help the user reduce.

TimFelixBeyer:
> Fair points, would you prefer this instead?
> 
> ```python
> for pitchIndex in range(len(nLast.pitches)):
>     pLast = nLast.pitches[pitchIndex]
>     pInner = nInner.pitches[pitchIndex]
>     if (pLast.octave != pInner.octave
>             or pLast.step != pInner.step
>             or pLast.microtone != pInner.microtone):
>         return False
>     # Special case: We have to treat None and "natural" accidentals as equal
>     if pLast.accidental != pInner.accidental:
>         aLast = pLast.accidental
>         aInner = pInner.accidental
>         if not (aLast is None and aInner.name == "natural"
>                 or aLast.name == "natural" and aInner is None):
>             return False
> ```

Would you like me to submit an updated version with this solution?

mscuthbert:
that looks great to me!

mscuthbert:
Thanks for the great PR!

Issue #1557, opened 2023-05-03
Saving a stream as musicxml alters the stream


TimFelixBeyer:
**music21 version**

`8.1.0`   

**Problem summary**
When saving a Stream `s` to musicxml, the object `s` is altered after saving to disk.
The documentation does not mention this side effect, in fact quite the opposite: "Some formats, including .musicxml, create a copy of the stream, pack it into a well-formed score if necessary, and run makeNotation()." which suggests that the original stream is left unaltered.

**Steps to reproduce**
Get e.g., this .mxl file: https://musescore.com/user/27439014/scores/5316979
```python3
from music21 import converter
path = "La_Campanella.mxl"
song = converter.parse(path, forceSource=True).expandRepeats().stripTies()
a = ([m.offset for m in song.flat.getElementsByClass('Note')])  # Note is just an example class, other's are affected too.
song.write("musicxml", "/tmp/_.musicxml")
b = ([m.offset for m in song.flat.getElementsByClass('Note')])
for a_, b_ in zip(a, b):
    if a_ != b_:
        print(a_, b_)
```

**Expected vs. actual behavior**
Expected: ```No output```
Actual: 
```
87.5 88.0
87.75 88.25
88.0 88.5
88.25 88.75
88.5 89.0
88.75 89.25
89.0 89.5
89.5 90.0
90.0 90.5
90.5 91.0
92.5 93.0
93.0 93.5
93.5 94.0
93.75 94.25
...
```

**More information**
I traced the changes and believe these lines are responsible: 
https://github.com/cuthbertLab/music21/blob/696545ae6f3a02d421ef29ca7aad5edcf23c1343/music21/stream/makeNotation.py#L852-L858

Please let me know if the behaviour is intended.
Thanks for your help!

mscuthbert:
Not intended!  Can you reduce the score to a minimum test set that displays the behavior?  Does it do this if you leave out the stripTies?  What happens around offset 87 that doesn't happen before in the piece?

I ran into another case where showing something changed the original the other day, but unrelated to this.  Show should not alter any non-private/"core" attributes

TimFelixBeyer:
I believe the problem occurs at notes that are tied across two measures (that's where La_Campanella.mxl goes awry as well). 
Here's a small example:
```
from music21 import stream, note, meter, tie

s = stream.Score(id='mainScore')
p0 = stream.Part(id='part0')

m01 = stream.Measure(number=1)
[...]

PR #1560, opened 2023-05-09
Avoid altering score on write

TimFelixBeyer:
We copy the score object early to avoid modifying it when calling `.write("musicxml", *)`
See #1557

@jacobtylerwalls


mscuthbert:
Looks good -- there was one additional deepcopy still there.  Just patched it up.  Will merge (or @jacobtylerwalls can) if I didn't introduce any lint errors/bugs.

jacobtylerwalls:
Super! In a minute I'll push up a regression test (given that this regressed once already).

jacobtylerwalls:
Thanks for the diagnosis and PR!

PR #1569, opened 2023-05-14
Remove byte order mark from corpus.virtual

jacobtylerwalls:
Was playing with [this tool](https://github.com/carljm/compfinder) and wondered why it couldn't parse the AST for `corpus.virtual`. Answer: it has a [byte order mark](https://www.freecodecamp.org/news/a-quick-tale-about-feff-the-invisible-character-cd25cd4630e7/).


mscuthbert:
Bom boom yuk. Thanks Jacob. I imagine the suite passed the tool otherwise?

jacobtylerwalls:
It did!

PR #1571, opened 2023-05-15
Fix #1561: Let makeTies iterate over parts

### LINKED ISSUE #1561: Let `makeTies()` iterate over parts
**music21 version**
`master`

**Problem summary**
makeTies throws `StreamException: cannot process a stream without measures` on score with measures.

**Steps to reproduce**

```python3
from music21 import stream, note, meter, tie

s = stream.Score(id='mainScore')
p0 = stream.Part(id='part0')

m01 = stream.Measure(number=1)
m01.append(meter.TimeSignature('4/4'))
d1 = note.Note('D', type="half", dots=1)
c1 = note.Note('C', type="quarter")
c1.tie = tie.Tie("start")
m01.append([d1, c1])
m02 = stream.Measure(number=2)
c2 = note.Note('C', type="quarter")
c2.tie = tie.Tie("stop")
c3 = note.Note("D", type="half")
m02.append([c2, c3])
p0.append([m01, m02])

s.insert(0, p0)
s.show("text")  # just to verify that there are measures in the score 
s = s.stripTies()
# s = s.makeMeasures()  # does not help
s = s.makeTies()
```

**More information**
Related to #1557

***
**3**
Your example is a score that has had `stripTies()` run on it, yielding an overfilled first measure and an underfilled second measure. I think I would expect that making rests on that would fill the second measure.

If you want to display the stripped ties score (are you sure you want to display that?) then you could makeTies() on it first, which restores the output I think you're expecting.

***

_Originally posted by @jacobtylerwalls in https://github.com/cuthbertLab/music21/issues/1557#issuecomment-1538586419_

TimFelixBeyer:
Fixes #1561

Allows makeTies to iterate over substreams and includes a regression test.
Let me know if there is anything you'd like to see addressed!


mscuthbert:
I think it would be better if makeTies were overloaded in Score and Opus -- I prefer superclasses not to need to think about subclass behavior (as much as possible -- there are already exceptions in the codebase, but I'm trying to remove them rather than add more).  It may be that Score, Part, Measure, etc. someday are moved to a different file from stream.Stream.

mscuthbert:

mscuthbert:
Turns out that it's not possible to overload this easily within Score/Opus, so it's fine as is for those parts.

However, the newly added `if returnObj.hasVoices()` part was untested and the code below already took into account streams with voices.  I'm removing it for now.  The Coveralls test results will show you which lines of code have been added or edited and are not yet tested.  I added tests for Score/Opus, but until Voices is tested cannot add that.

Screenshot 2023-07-01 at 14 42 25

I'm merging this once all tests pass and then will consider requests again after sabbatical.

PR #1572, opened 2023-05-15
Prevent doubly-flatted sevenths in chord symbols

### LINKED ISSUE #1570: ChordSymbols: addb7 gives doubly-flatted seventh
**music21 version**

`9.8.7`   

**Problem summary**

music21 is encoding several chords wrongly. 
1) Csus add b7, it seems no allowed this very common jazz chord, as it decoded it as ['C2', 'F2', 'G2', 'A3']
2) Slash chords with extensions need to be corrected.  For example, I'm parsing an XML file of a song with an F# b9 / B, and music21 decodes it as F#7/B add b9, which is a huge error. Is there a way to fix it? 

**Steps to reproduce**

```
element = 'Csus add b7'
tmp = m21.harmony.ChordSymbol(element)
theNotes = [str(p) for p in tmp.pitches]
print(theNotes)
```

**More information**

jacobtylerwalls:
Csusaddb7 now gives B-flat, not A.

Closes #1570

### DIFF
--- a/music21/harmony.py
+++ b/music21/harmony.py
@@ -1713,8 +1713,15 @@ class ChordSymbol(Harmony):
             pitchToAppend = sc.pitchFromDegree(hD.degree, rootPitch)
             if hD.interval and hD.interval.semitones != 0:
                 # added degrees are relative to dominant chords, which have all major degrees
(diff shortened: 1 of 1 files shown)


mscuthbert (approved):
LGTM, thanks!

PR #1575, opened 2023-05-16
Don't set status byte on Meta Message (Fix #1574)

### LINKED ISSUE #1574: MIDI MetaMessages shouldn't set status byte
**music21 version**

`master`  

**Problem summary**
When loading a MIDI file, messages can set a (running) status byte. However, meta messages don't set the status byte. 
This leads to problem on several MIDI files, for example 8584aaa73eb54bae39708da8be5e6282.mid from the lakh dataset.
Compare e.g. here: 
https://www.recordingblogs.com/wiki/midi-meta-messages
http://www.somascape.org/midi/tech/spec.html

See the linked PR for a solution.
Let me know if you'd like any changes!

TimFelixBeyer:
Fixes #1574
When loading a MIDI file, messages can set a (running) status byte. However, meta messages don't set the status byte. 

This leads to problem on several MIDI files, for example 8584aaa73eb54bae39708da8be5e6282.mid from the lakh dataset.

It could be argued that the `!= 0xff` check is too conservative as running status is only applicable for messages in the range `0x8n - 0xEn`, the proposal is this way to be aligned with other python midi libraries:
https://github.com/mido/mido/blob/bff6bf2a8711e33782de5661ed5ccdffb0bb1688/mido/midifiles/midifiles.py#L204-L206

### DIFF
--- a/music21/midi/__init__.py
+++ b/music21/midi/__init__.py
@@ -891,8 +891,9 @@ class MidiEvent(prebase.ProtoM21Object):
             midiBytes = rsb + midiBytes
             byte0 = midiBytes[0]
         else:
-            # store last status byte
-            self.lastStatusByte = midiBytes[0]
+            if midiBytes[0] != 0xff:
+                # store last status byte, unless it's a meta message
+                self.lastStatusByte = midiBytes[0]
 
         msgType: int = byte0 & 0xF0  # bitwise and to derive message type w/o channel
 
--- a/music21/midi/__init__.py
+++ b/music21/midi/__init__.py
@@ -890,10 +890,9 @@ class MidiEvent(prebase.ProtoM21Object):
             # and process as before
             midiBytes = rsb + midiBytes
             byte0 = midiBytes[0]
-        else:
-            if midiBytes[0] != 0xff:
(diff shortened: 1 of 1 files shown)


mscuthbert:
Only q: Should a meta 0xFF message set lastStatusByte to None instead or is the lastStatusByte still valid?

(btw -- preference for using `elif` in the patch rather than an `else` clause followed directly by `if` with no else)

TimFelixBeyer:
> Only q: Should a meta 0xFF message set lastStatusByte to None instead or is the lastStatusByte still valid?

I believe the lastStatusByte is still valid - the second source I mentioned in #1574 contains this quote about meta events “Also, they can be interleaved within other messages without affecting running status in any way” - that’s how Mido treats it as well.

mscuthbert:
> I believe the lastStatusByte is still valid - the second source I mentioned in #1574 contains this quote about meta events “Also, they can be interleaved within other messages without affecting running status in any way” - that’s how Mido treats it as well.

Great -- thank you for that clarification.  With it, I'm thrilled to merge.  (The elif is a nice bonus)

PR #1577, opened 2023-05-17
Increase default MIDI ticksPerQuarter

TimFelixBeyer:
According to comments in the code, the default MIDI `ticksPerQuarter` is set to 1024 to avoid a limitation in midi.py, which is no longer used. 
Therefore, I propose to set the `ticksPerQuarter` to `10080` which is the same value used for MusicXML divisions. This would allow exported MIDI to represent more durations exactly without quantization errors and without significant risk/downsides. 
I have also updated all relevant tests.
Let me know what you think!


mscuthbert:
Good idea.  Compatibility with buggy versions of midi.py is no longer a consideration, agreed.

Can you explain why the note was changed from 128th note to double the duration?  Is there a bug that emerged?

TimFelixBeyer:
Sure, 1024 = 2^10, so we can represent even a 1/4096th note using integer ticks.
10080 = 2^5 * 3^2 * 5 * 7, so the smallest standard (1/2^n) note it can represent exactly is a 1/128th note. 
The example effectively implemented a 3/256th note (dotted 128th), which would have required a non-integer tick. 

I‘d advise against increasing the _default_ ticksPerQuarter over 10080, as it is already on the high side for MIDI (typical range 120-960), and very high ticks may lead to compatibility problems with other libraries (https://github.com/craffel/pretty-midi/issues/175)
Users can always override the default if they want more precision and want to represent tiny notes (3/256th) exactly.

mscuthbert:
Thanks!  Got it!

PR #1580, opened 2023-05-23
FretBend supports alter (interval), prebend, and release

### LINKED ISSUE #1471: Improve management of bends
**Motivation**

As of now (version 8.1.0), it seems that bends (the technique of bending a string to make the note go higher, widely used in guitar for instance) can only be added as an articulation to notes with `music21.articulations.FretBend`.

In the doc, this articulation only has a `number` attribute which seems inherited from `articulations.FretIndication` while many more attributes are needed. Looking at the code, I see that some other attributes are mentioned though not documented nor actually used:
```python
# in music21/articulations.py
class FretBend(FretIndication):
    bendAlter: interval.IntervalBase | None = None
    preBend: t.Any = None
    release: t.Any = None
    withBar: t.Any = None
```

These are the attributes used in musicXML (see [the doc](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/bend/)) which is great so I guess what's *only* missing is to actually implement them in the musicXML reader/writer? 

edit: as a side note, I also wonder how Pitch should be managed when a note has a bend. For reading purposes, as it is done in musicXML, it probably makes more sense to keep the pitch of the note that has no bend. But for analysis, like in music21, a note with a bend should probably be considered as having the pitch of the destination of the bend, not the pitch of its starting point. This is an open discussion, I'd love to hear your thoughts on that.

**Feature summary**
- Complete the documentation on the `FretBend` articulation to mention all existing attributes (*I can do this*);
- Complete the class code to use the attributes, parse them from musicXML etc. (*I could probably do it with some help*);
- ~~Change the class inheritance because a bend does not need to be a FretIndication, I think?~~ Maybe a little complicated to change that and not that useful, I don't know;
- Make bends interact nicely between tablature staves and standard staves, see following example from GuitarPro for instance. I have no idea how to implement it, probably the hardest thing to do but it would be nice that a bend in a tablature staff automatically shows the corresponding pitch change on the related staff. This is also kinda related to the discussion on the pitch given to a note with a bend (see edit above).
image

**Proposed implementation**

To discuss

**Intent**

adhooge:
Fixes  #1471

Hi! First PR so I expect a few things to be added/modified but here's a first draft.
I've updated the `FretBend` articulation so that it includes the musicXML properties `bendAlter`, `preBend` and `release`.

It goes round from musicXML to m21 back to mXML.

I've not written any tests yet because I'm not sure what to add. Let me know of what could be good to add!

as a PR newbie, I also have some irrelevant commits that I manually reverted. If that's a problem I'll try to clean that up.


music21/articulations.py
```diff
@@ -84,6 +84,7 @@
 from music21.common.classTools import tempAttribute
 from music21 import environment
 from music21 import style
+from music21 import interval
```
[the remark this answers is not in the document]
mscuthbert: and if this is added it needs to be removed from they TYPE_CHECKING import below (but I don't think it will need to be, see below)

music21/articulations.py
```diff
@@ -582,10 +583,27 @@ class PullOff(spanner.Spanner, TechnicalIndication):
...
-    bendAlter: interval.IntervalBase | None = None
-    preBend: t.Any = None
-    release: t.Any = None
-    withBar: t.Any = None
+    bendAlter: interval.IntervalBase
+    preBend: bool = False
+    release: t.Optional[float]
+    withBar: t.Optional[bool]
+
+    def __init__(self, number=0, preBend=False, release=None, withBar=None, **keywords):
```
mscuthbert, line 591: number is too generic a term -- and we usually reserve the term for identifying which of multiple simultaneous items (like slur number=1 vs slur number=2).   Are fret-bends always chromatic intervals?  it seems like you can also have a diatonic interval of a fretbend.  hence why bendAlter was typed as IntervalBase not ChromaticInterval.  It seems better to have bendAlter as described above `interval.IntervalBase | None = None` where None indicates unspecified.  Having a default of 0, becoming ChromaticInterval(0) indicates that this fretBend does not alter the pitch at all, which would be a very strange fret bend.  This would also make it so that interval.py does not need to be imported here.

The other keywords should be typed in the `__init__` declaration in addition to the class.

We no longer use `t.Optional[X]` in the code base.  Use `X | None` instead.

music21/articulations.py
```diff
@@ -582,10 +583,27 @@ class PullOff(spanner.Spanner, TechnicalIndication):
...
+        '''
+        bend indication for fretted instruments
+
+        bend in musicxml
+
+        number is the interval of the bend in number of semitones, bend-alter in musicxml
+        preBend indicates wether the note is prebended or not
+        release is the offset value for releasing the bend, if Any
+        withBar indicates if the bend is done using a whammy bar movement
```
mscuthbert, line 600: Capitalization and use asterisks and line breaks for documentation -- run documentation/build.py and look at the difference between the docs produced here and elsewhere in the system.

`preBend indicates wether the note is prebended or not` -- I don't know what prebended means.

what is release measured in?  Is it an offset (which is relative to the start of the Measure/stream)?  That doesn't sound right.  Is it a quarterLength measured from the start of the note? or from the end of the note.

withBar doc is fine.

music21/musicxml/m21ToXml.py
```diff
@@ -5523,7 +5522,17 @@ def articulationToXmlTechnical(self, articulationMark):
             mxTechnicalMark.text = str(articulationMark.number)
         if musicXMLTechnicalName == 'fret':
             mxTechnicalMark.text = str(articulationMark.number)
-
+        if musicXMLTechnicalName == 'bend':
+            bendAlterSubElement = SubElement(mxTechnicalMark, 'bend-alter')
+            bendAlterSubElement.text = str(articulationMark.bendAlter.semitones)
```
mscuthbert, line 5527: this has enough ifs that it should be brought out as its own small method that allows for easier documentation and testing.  articulationMark will probably need to be cast as a FretBend.

We can solve the case of `GenericInterval` later, which doesn't have .semitones.  Perhaps it always needs to be a full Interval object or None.

music21/musicxml/m21ToXml.py
```diff
@@ -5523,7 +5522,17 @@ def articulationToXmlTechnical(self, articulationMark):
...
+            if articulationMark.preBend:
+                preBendSubElement = SubElement(mxTechnicalMark, 'pre-bend')
+            if articulationMark.release is not None:
+                releaseSubElement = SubElement(mxTechnicalMark, 'release')
+                releaseSubElement.set('offset', str(articulationMark.release))
```
mscuthbert, line 5532: musicxml's definition of offset is very different from music21's and (despite me now being editor of the musicxml spec) we shouldn't use musicxml offsets anywhere in music21 -- these need to be converted according to the current divisionsPerQuarter setting.

music21/musicxml/xmlToM21.py
```diff
@@ -3881,11 +3881,18 @@ def xmlTechnicalToArticulation(self, mxObj):
             if tag in ('heel', 'toe'):
                 if mxObj.get('substitution') is not None:
                     tech.substitution = xmlObjects.yesNoToBoolean(mxObj.get('substitution'))
+            if tag == 'bend':
+                tech.bendAlter = interval.Interval(int(mxObj.find('bend-alter').text))
```
mscuthbert, line 3885: this has too many potentials for crashing the processing.  What if there is no bend-alter or no bend-alter.text?  check that before typing to assign.

(Better to call out into a separate sub-method.  This method was designed just for one-to-two lines per technical

music21/musicxml/xmlToM21.py
```diff
@@ -3881,11 +3881,18 @@ def xmlTechnicalToArticulation(self, mxObj):
...
+                if mxObj.find('pre-bend') is not None:
+                    tech.preBend = True
+                if mxObj.find('release') is not None:
+                        tech.release = int(mxObj.find('release').get('offset'))
```
mscuthbert, line 3889: as above -- offset needs to be converted to music21 quarterLengths.

except clause without Try.

music21/musicxml/xmlToM21.py
```diff
@@ -3881,11 +3881,18 @@ def xmlTechnicalToArticulation(self, mxObj):
...
+                    except TypeError:
+                        # offset is not mandatory
+                        tech.release = 0
```
mscuthbert, line 3892: release is a float.  needs to be 0.0

music21/note.py
```diff
@@ -37,6 +37,7 @@
 from music21 import style
 from music21 import tie
 from music21 import volume
+from music21 import articulations
 
 if t.TYPE_CHECKING:
     from music21 import articulations
```
mscuthbert, line 43: as above -- alphabetical and remove from type-checking if imported at top.  (but won't be necessary).

music21/note.py
```diff
@@ -1784,6 +1785,49 @@ def pitchChanged(self):
         if self._chordAttached is not None:
             self._chordAttached.clearCache()
 
+    @property
+    def string(self):
```
mscuthbert, line 1789: this new property and all properties below are rejected.  Note is the most common object in music21 and users who are not working with guitar music will never use these routines but will need to figure out what they mean in the docs.  (we don't have "accent" or "staccato" etc. either) -- and parallel routines would need to be added to Chord, etc.

In general for any mature open-source project, don't make additions or substantive changes to core parts of the system without a prior discussion.

mscuthbert (changes requested):
Well on its way!  Ask around if you need help in writing tests, and for help on why music21's system of intervals is more complex than musicxml's.

(the changes to note.Note are rejected, so please remove them, thanks!)

mscuthbert:

adhooge:
Hello Michael, sorry for not being able to finish this PR sooner. Enjoy your sabbatical! Hopefully my PR will be ready for approval when you come back :crossed_fingers:

adhooge:
Hello @mscuthbert ! After one year of postponing, I'm finally back to this PR!
I've updated my repo with the updates in m21 so I have a lot of commits and I don't know how to remove them :sweat_smile: I'll answer to some of your comments here but I can also start a new PR to make this more readable !

music21/articulations.py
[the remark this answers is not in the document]
adhooge: Done with https://github.com/cuthbertLab/music21/pull/1580/commits/3cf71e0c154d80284eee1304b02be806f646cea8

music21/articulations.py
adhooge: Changed with https://github.com/cuthbertLab/music21/pull/1580/commits/d63c78ab54208bc0d8da8558190b821329900809

music21/articulations.py
adhooge: I've tried to improve it in https://github.com/cuthbertLab/music21/pull/1580/commits/03e6117e967d371851f8c80e73c3b2218a2ebf06 
Let me know if that seems good to you

music21/musicxml/m21ToXml.py
adhooge: New submethod added in https://github.com/cuthbertLab/music21/pull/1580/commits/85cc6a8cea3540b6397e309a5b377d45eb0c2d0c
Now I indeed have the problem that I need to retrieve the number of semitones. I could probably do something with `GenericInterval.value` but I don't know how to do the conversion properly

music21/musicxml/m21ToXml.py
adhooge: should work with https://github.com/cuthbertLab/music21/pull/1580/commits/d93af6d98c7b9bd1a7587c11d1c2d20f961a4460 though I need to write the tests

music21/musicxml/xmlToM21.py
adhooge: sub-method and bend-alter check in https://github.com/cuthbertLab/music21/pull/1580/commits/2e1ea432934e58cd877346be9b3b3d030c7a9a51

music21/musicxml/xmlToM21.py
adhooge: Both fixed with https://github.com/cuthbertLab/music21/pull/1580/commits/58eeb04646b04f5c0501795a30aa67e7083b23e6

music21/musicxml/xmlToM21.py
adhooge: also fixed with https://github.com/cuthbertLab/music21/pull/1580/commits/58eeb04646b04f5c0501795a30aa67e7083b23e6

music21/note.py
adhooge: oops sorry about that, they were just for my personal convenience. Removed with https://github.com/cuthbertLab/music21/pull/1580/commits/0b2074f084def3497f79d66da4a351c59c387263

music21/articulations.py
```diff
@@ -582,10 +580,47 @@ class PullOff(spanner.Spanner, TechnicalIndication):
...
+
+    PreBend indicates if the string is bent before
+    the onset of the note. Defaults to False.
+
+    Release is the quarterLength value from the start
+    of the note for releasing the bend, if Any. Defaults to None.
+
+    WithBar indicates if the bend is done using a whammy bar movement. Defaults to False.
+
+    >>> fb = articulations.FretBend(bendAlter=interval.ChromaticInterval(-2), release=0.5)
```
mscuthbert, line 601: add number to this to show that it appears in the fb representation below.

music21/articulations.py
```diff
@@ -582,10 +580,47 @@ class PullOff(spanner.Spanner, TechnicalIndication):
...
+    <music21.interval.ChromaticInterval -2>
+    >>> fb.release 
+    0.5
+    '''
+    bendAlter: interval.IntervalBase | None
+    preBend: bool
+    release: float | None
+    withBar: bool 
+
+    def __init__(self, number=0, bendAlter=None, preBend=False, release=None, withBar=False, **keywords):
```
mscuthbert, line 618: type needs to be repeated here.  Also make all but number keyword only so people don't have the remember the order.

```
def __init__(
    self, 
    number: int = 0, 
    *,
    bendAlter: interval.Interval | interval.ChromaticInterval | None = None, 
    preBend: bool = False, 
    release: OffsetQL | None = None, 
    withBar: bool = False, 
    **keywords
):```

File: music21/articulations.py [hunk 13]
```diff
@@ -582,10 +580,47 @@ class PullOff(spanner.Spanner, TechnicalIndication):
...
+        super().__init__(**keywords)
```
mscuthbert, line 619: bug.  needs to be `super().__init__(number=number, **keywords)`

music21/musicxml/m21ToXml.py
```diff
@@ -5477,6 +5485,46 @@ def articulationToXmlTechnical(self, articulationMark: articulations.Articulatio
...
+        <bend>
+          <bend-alter></bend-alter>
+        </bend>
+        '''
+        bendAlterSubElement = SubElement(mxh, 'bend-alter')
+        alter = bend.bendAlter
+        if alter is not None:
+            # musicxml expects a number of semitones but not sure how to get it
+            # from a GeneralInterval
+            pass
```
mscuthbert, line 5514: now with the change in typing above this can be possible.

music21/musicxml/m21ToXml.py
```diff
@@ -5477,6 +5485,46 @@ def articulationToXmlTechnical(self, articulationMark: articulations.Articulatio
         # mxArticulations.append(mxArticulationMark)
         return mxTechnicalMark
 
+    @staticmethod
+    def setBend(mxh: Element, bend: articulations.FretBend) -> None:
```
mscuthbert, line 5489: looks great!

music21/musicxml/xmlObjects.py
```diff
@@ -67,12 +67,12 @@
...
                                ('snap-pizzicato', articulations.SnapPizzicato),
                                ('string', articulations.StringIndication),
+                               ('bend', articulations.FretBend),
                                # hammer-on and pull-off not implemented because handled
                                # in method objectAttachedSpannersToTechnicals of m21ToXml.py
                                # ('hammer-on', articulations.HammerOn),
                                # ('pull-off', articulations.PullOff),
-                               # bend not implemented because it needs many subcomponents
-                               # ('bend', articulations.FretBend),
+                               ('bend', articulations.FretBend),
```
mscuthbert, line 75: this line appears to be added twice?  (see line 70)

music21/musicxml/xmlToM21.py
```diff
@@ -3861,17 +3861,32 @@ def xmlTechnicalToArticulation(self, mxObj):
...
             # TODO: musicxml 4: release sub-element as offset attribute
-
-
             self.setPlacement(mxObj, tech)
             return tech
         else:
             environLocal.printDebug(f'Cannot translate {tag} in {mxObj}.')
             return None
 
+    @staticmethod
```
mscuthbert, line 3875: Not going to be able to be a staticmethod -- see below.

music21/articulations.py
```diff
@@ -582,10 +580,47 @@ class PullOff(spanner.Spanner, TechnicalIndication):
...
+    >>> fb.withBar
+    False
+    >>> fb.bendAlter
+    <music21.interval.ChromaticInterval -2>
+    >>> fb.release 
+    0.5
+    '''
+    bendAlter: interval.IntervalBase | None
+    preBend: bool
+    release: float | None
```
mscuthbert, line 615: instead of float, type as OffsetQL  which is `float|Fraction`    Music21 represents 1/3 as Fraction(1, 3) not 0.3333333428 as an inexact float.

music21/musicxml/m21ToXml.py
```diff
@@ -5477,6 +5485,46 @@ def articulationToXmlTechnical(self, articulationMark: articulations.Articulatio
...
+            # from a GeneralInterval
+            pass
+        if bend.preBend:
+            SubElement(mxh, 'pre-bend')
+        if bend.release is not None:
+            # Specifies where the release starts in terms of 
+            # divisions relative to the current note.
+            releaseSubElement = SubElement(mxh, 'release')
+            quarterLengthValue = bend.release
+            divisionsValue = defaults.divisionsPerQuarter * quarterLengthValue
```
mscuthbert, line 5522:
```
divisionsValue = int(defaults.divisionsPerQuarter * quarterLengthValue)
```

Musicxml "prefers" that offset be an integer:  
https://www.w3.org/2021/06/musicxml40/musicxml-reference/data-types/divisions/

music21/musicxml/xmlToM21.py
```diff
@@ -3861,17 +3861,32 @@ def xmlTechnicalToArticulation(self, mxObj):
...
+        alter = mxh.find('bend-alter')
+        if alter is not None:
+            if alter.text is not None:
+                bend.bendAlter = interval.Interval(int(alter.text))
+        if mxh.find('pre-bend') is not None:
+            bend.preBend = True
+        if mxh.find('release') is not None:
+            try:
+                divisions = float(mxh.find('release').get('offset'))
+                bend.release = divisions / defaults.divisionsPerQuarter
```
mscuthbert, line 3886: In writing to musicxml, music21 uses defaults.divisionsPerQuarter.  But in reading from musicxml, we need to use the divisions defined earlier in the score, which is in self.divisions on the measure parser.  

Write

```
bend.release = opFrac(divisions / defaults.divisionsPerQuarter)
```

opFrac will convert it to Fraction(1, 3) if it's released within a triplet, etc. and leave it alone as 0.5 if it's a regular eighth note, etc.

mscuthbert (changes requested):
Hi!  It's again a Request Changes, but they're really minimal changes and I think this is very close to getting in.  :-)  CONGRATS on the amazing work so far and for putting up with a sabbatical that interrupted it!

music21/articulations.py
adhooge: done with 4c585ea510141ca6d0feca0c241eec83118ac311 and 099df36e2936e484b9daa124ac29001054f82d39

music21/articulations.py
adhooge: Done with f74771b08056943b035f3b929161232a37f6f085, thanks for the sample code!

music21/musicxml/m21ToXml.py
adhooge: indeed, should be fine with 95e00d733c4f63250acc66c801e1bbc84e92ade8

music21/musicxml/m21ToXml.py
adhooge: Fixed! d651fbd7ab6d2cb3098ccd573736bc3f8606cacc

music21/musicxml/xmlObjects.py
adhooge: Indeed, I might have had issues with merging new changes and the commented lines. Fixed with 91e9f8499390a678cbc5dc3622d63a95088af80b

music21/musicxml/xmlToM21.py
adhooge: Thanks, I changed it and added some tests (15eadd15586bfd88ac5343f08efbfa3e08907939, daa68483fc218604c0a8f70e575e193adaea40f9)

music21/articulations.py
```diff
@@ -596,27 +595,37 @@ class FretBend(FretIndication):
     Release is the quarterLength value from the start
     of the note for releasing the bend, if Any. Defaults to None.
 
-    WithBar indicates if the bend is done using a whammy bar movement. Defaults to False.
+    WithBar indicates what whammy bar movement is used, if any.
+    MusicXML supports 'scoop' or 'dip'. Defaults to None.
 
-    >>> fb = articulations.FretBend(bendAlter=interval.ChromaticInterval(-2), release=0.5)
+    >>> fb = articulations.FretBend(1,bendAlter=interval.ChromaticInterval(-2),release=0.5)
```
mscuthbert, line 601: spaces after commas

music21/articulations.py
```diff
@@ -596,27 +595,37 @@ class FretBend(FretIndication):
     Release is the quarterLength value from the start
     of the note for releasing the bend, if Any. Defaults to None.
 
-    WithBar indicates if the bend is done using a whammy bar movement. Defaults to False.
+    WithBar indicates what whammy bar movement is used, if any.
```
mscuthbert, line 598: If possible write camelCase variables w/ proper camelCase even at start of the line, using backticks `withBar` can make it look fine in the docs even at the start of the sentence.  Sorry I didn't get this last time.

music21/articulations.py
```diff
@@ -596,27 +595,37 @@ class FretBend(FretIndication):
...
-    >>> fb = articulations.FretBend(bendAlter=interval.ChromaticInterval(-2), release=0.5)
+    >>> fb = articulations.FretBend(1,bendAlter=interval.ChromaticInterval(-2),release=0.5)
     >>> fb
-    <music21.articulations.FretBend 0>
+    <music21.articulations.FretBend 1>
     >>> fb.preBend
     False
     >>> fb.withBar
-    False
+
```
mscuthbert, line 607: In the doctests it's best to show None as one of:

```
>>> fb.withBar is None
True
```

or

```
>>> print(fb.withBar)
None
```

Otherwise the empty space just looks a bit weird.

music21/articulations.py
```diff
@@ -596,27 +595,37 @@ class FretBend(FretIndication):
...
+    def __init__(
+        self,
+        number: int = 0,
+        *,
+        bendAlter: interval.Interval | interval.ChromaticInterval | None = None,
+        preBend: bool = False,
+        release: OffsetQL | None = None,
+        withBar: str | None = None,
+        **keywords
+        ):
```
mscuthbert, line 627: unindent this

music21/musicxml/m21ToXml.py
```diff
@@ -5557,31 +5557,48 @@ def setBend(mxh: Element, bend: articulations.FretBend) -> None:
...
         >>> mxh = Element('bend')
 
-        >>> MEXClass.setBend(mxh, a)
-        >>> MEXClass.dump(mxh)
+        >>> MEXclass.setbend(mxh, a)
+        >>> MEXclass.dump(mxh)
+        <bend>
+          <bend-alter>2</bend-alter>
+        </bend>
+        >>> a.pre-bend = True
```
mscuthbert, line 5570: this test is not being run since `a.pre-bend = True` is  `a.pre  -  bend = True` which is an incorrect statement in Python.

mscuthbert:
```
In [5]: a.pre-bend = True
  Cell In[5], line 1
    a.pre-bend = True
    ^
SyntaxError: cannot assign to expression here. Maybe you meant '==' instead of '='?

```

mscuthbert (changes requested):
Hi `adhooge` I'm realizing that for some reason you don't have permissions to run the Github Actions here automatically (I think you get them the first time you open a PR that passes)

Make sure that you're running `python music21/test/multiprocessTest.py` before pushing here -- you have tests that aren't passing -- as soon as I noticed that I stopped reviewing for now.  

If you want to open up a dummy PR that changes one letter (like add a period somewhere it's missing) point to this comment and I'll merge it so you have full actions rights.  (Or point me to some web resource that tells me how to exempt you from the hold back until you have it).

adhooge:
Hi `mscuthbert` apologies for the missed test, I updated them and also made a dummy PR so that I can run github actions to check linting etc.:
https://github.com/cuthbertLab/music21/pull/1776

adhooge:
It seems I got all the coveted green ticks! There might still be a few changes required. I'm thinking in particular about the fact that I had to import `interval` and `OffsetQL` to get the typing tests to pass. They should probably only be imported for type checking.

Thank you very much for your reactivity and detailed answers `mscuthbert`, it was very helpful!

mscuthbert (approved):
Congrats!  Looks great!

mscuthbert:
This is a really significant contribution for a first PR.  Congrats!  Hope it was a good enough experience that there will be many more.

I'm planning on doing a release by July 1 or when we get the Piano Pedal PRs in whichever comes first.

adhooge:
Thank you very much for your kind help throughout the process (that spanned over several years!!)
It was a great experience and I'm very happy and proud to contribute to this great library :)

I won't hesitate to contribute more in the future!

PR #1581, opened 2023-05-27
Add preserveVoices option to stripTies to fix MIDI export

TimFelixBeyer:
Follow-up from previous discussion in #1556 
The current `stripTies` function sometimes does not strip all ties from a score to preserve voicing.
However, in certain cases (like MIDI export) we need to strip *all* ties to prevent double-sounding notes to be exported. 
This proposal would add a `preserveVoices` flag that defaults to `True`, keeping backwards compatibility. 
In addition, the new `stripTies` would have more explicit and informative error handling and warnings built-in. It is currently quite an ugly draft with a lot of redundancy between the current logic and the new branch, but I'd be more than happy to spend some time simplifying/combining the changes with existing code, adding a breaking test, and changing the `preserveVoices` flag in the ConverterMIDI export flow.
Let me know what you think!

### DIFF
--- a/music21/stream/base.py
+++ b/music21/stream/base.py
@@ -7110,7 +7110,8 @@ class Stream(core.StreamCore, t.Generic[M21ObjType]):
     def stripTies(
         self,
         inPlace=False,
-        matchByPitch=True
+        matchByPitch=True,
+        preserveVoices=True
     ):
         # noinspection PyShadowingNames
         '''
@@ -7260,7 +7261,7 @@ class Stream(core.StreamCore, t.Generic[M21ObjType]):
             # returnObj.parts for this...
             for p in returnObj.getElementsByClass(Stream):
                 # already copied if necessary; edit in place
-                p.stripTies(inPlace=True, matchByPitch=matchByPitch)
+                p.stripTies(inPlace=True, matchByPitch=matchByPitch, preserveVoices=preseveVoices)
             if not inPlace:
                 return returnObj
             else:
@@ -7269,7 +7270,7 @@ class Stream(core.StreamCore, t.Generic[M21ObjType]):
         if returnObj.hasVoices():
             for v in returnObj.voices:
                 # already copied if necessary; edit in place
-                v.stripTies(inPlace=True, matchByPitch=matchByPitch)
+                v.stripTies(inPlace=True, matchByPitch=matchByPitch, preserveVoices=preserveVoices)
             if not inPlace:
                 return returnObj
             else:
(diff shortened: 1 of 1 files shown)


music21/stream/base.py
```diff
@@ -7110,7 +7110,8 @@ def extendDuration(self, objClass, *, inPlace=False):
     def stripTies(
         self,
         inPlace=False,
-        matchByPitch=True
+        matchByPitch=True,
+        preserveVoices=True,
```
mscuthbert, line 7114: add a note about what this does to docs.

music21/stream/base.py
```diff
@@ -7287,6 +7288,21 @@ def stripTies(
         posConnected = []  # temporary storage for index of tied notes
         posDelete = []  # store deletions to be processed later
 
+        def pitchesEqual(p1, p2):
```
mscuthbert, line 7291: since this is redundant code from the previous PR, we should factor it out -- I'm okay with having it as a method on Pitch that is used in eq... wait -- I'm going to do that right now...

music21/stream/base.py
```diff
@@ -7110,7 +7110,8 @@ def extendDuration(self, objClass, *, inPlace=False):
     def stripTies(
```
mscuthbert, line 7110: this is getting too big -- I'm trying to reduce the size of stream/base.py (it's starting to choke certain checkers) -- so my suggestion is to finish this review until it's basically done, then we'll move it to stream/tools.py and just leave it as a stump:

```
def stripTies(
    self,
    *,
    inPlace: bool = False,
    matchByPtch: bool = True,
    preserveVoices: bool = True,
):
    '''docs'''
    return tools.stripTies(self, inPlace=inPlace, matchByPitch...etc.)
```

this is what we've already been doing with makeTies(), etc. in stream/makeNotation.py

Thanks!

mscuthbert (commented):
Review in progress, but I think I have an idea to make this better...

mscuthbert:
See #1582 for a way to do the kind of pitch comparisons you want:

`if p1.isEnharmonic(p2) and p1.step == p2.step`

TimFelixBeyer:
@mscuthbert I'd like to seek input on how to handle certain edge cases/incorrect tie formatting.

There are the obviously correct cases:
1) `start -> stop`
2) `start -> continue -> stop` (n x continue)
However, according to the [music21 documentation](https://web.mit.edu/music21/doc/moduleReference/moduleTie.html), "you can tie n1 to n2 just with a tie start on n1", which adds the following case as valid:
3) `start -> None`

I'd like to know your position on what to do (warn/error/nothing) in the following cases:
4) `None -> stop`
5) `None -> continue -> stop` (n x continue)
6) `start -> continue -> None` (n x continue)
7) `None -> continue -> None` (n x continue)
8) `start -> start -> stop`

In my testing, especially 8) comes up quite often, and in many cases has led to me discovering incorrectly formatted MusicXML. 

Presently, if the argument `matchByPitch=False`, only 1) + 2) are accepted. 
If `matchByPitch=True`,
4) is not accepted/ignored
5) is accepted (continue is treated as start)
6) is accepted
7) is accepted (continue is treated as start)
8) is accepted as `start -> continue -> stop`
There are no warnings of any kind.

Here are my thoughts:
if `matchByPitch=False`:
Handle only 1) + 2)
Warn if the other cases come up.
 
If `matchByPitch=True`:
4) don't ignore, accept
5) + 6) + 7) + 8) treat like now, but warn. 
For all cases: if no matching pitch is ever found until the start/end of the stream, raise an error.

Another topic: tied grace notes are common and are currently ignored/mishandled due to the `el.quarterLength > 0` filtering, but work completely fine if the filtering is removed - is there a good reason to keep the filtering?

(By the way, in my opinion, something like `strict` or `onlyStripPairedTies` might be a better keyword than `matchByPitch`.)

mscuthbert:
> @mscuthbert I'd like to seek input on how to handle certain edge cases/incorrect tie formatting.

Thanks for asking. 

> There are the obviously correct cases:
> 
> 1. `start -> stop`
> 2. `start -> continue -> stop` (n x continue)
>    However, according to the [music21 documentation](https://web.mit.edu/music21/doc/moduleReference/moduleTie.html), "you can tie n1 to n2 just with a tie start on n1", which adds the following case as valid:
> 3. `start -> None`

Yes!

> I'd like to know your position on what to do (warn/error/nothing) in the following cases:

For each of these I'll give acceptable first followed by // two slashes and then ideal handling:
(I wrote this before reading your next part -- so it's largely redundant)

> 4. `None -> stop`

 just ignore  //  Ignore and drop stop tie.

> 5. `None -> continue -> stop` (n x continue)

just ignore.  // treat as None -> Start -> Stop and merge final two. 

> 6. `start -> continue -> None` (n x continue)

merge all three as if it's proper start -> continue -> stop  // same as acceptable here.

> 7. `None -> continue -> None` (n x continue)

ignore  // treat as None -> Start -> Stop and merge final two.

> 8. `start -> start -> stop`

merge any two // merge all three (treating second start immediately following a start as a continue)


> In my testing, especially 8) comes up quite often, and in many cases has led to me discovering incorrectly formatted MusicXML.

The important point is that start or continue is enough to indicate a desire for that note to be tied to the next note and that overrules anything that the next note says about its tie status.

 
> Presently, if the argument `matchByPitch=False`, only 1) + 2) are accepted. If `matchByPitch=True`, 3) is accepted 4) is not accepted/ignored 5) is accepted (continue is treated as start) 6) is accepted 7) is accepted (continue is treated as start) 8) is accepted as `start -> continue -> stop` There are no warnings of any kind.

Wow, it's good that I did this independently because I came to the exact same conclusion as the current usage.

> Here are my thoughts: if `matchByPitch=False`: Handle only 1) + 2) Warn if the other cases come up.

Agreed except that I don't think warning is necessary or appropriate here, especially since matchByPitch=False is a setting that people need to opt into so they should know what they're doing.  (in general, warnings should happen when importing/exporting data, not when manipulating in music21)

> If `matchByPitch=True`: 4) don't ignore, accept 3) + 5) + 6) + 7) + 8) treat like now, but warn. For all cases: if no matching pitch is ever found until the start/end of the stream, raise an error.

Disagree, especially on 3) -- we've been teaching people that an appropriate way to set a tie is with n.tie = tie.Tie('start') and that's enough.  I'd need to know a compelling case for changing a design decision that goes back to before v1.0.  

It sounds like you're wanting another routine called `sanitizeTies()` that has various functions for fixing broken ties or returning (not warning -- return something that can be worked with) references to notes etc. that could not be sanitized for various reasons.  That'd be a good routine, but it's not what stripTies is about.

(I have for another project a "sanitize_ties()" routine, but it's not at music21 quality to be worth getting permission to open source it.  But I do find it very handy)

> 
> (By the way, in my opinion, something like `strict` or `onlyStripPairedTies` might be a better keyword than `matchByPitch`.)

<s>One thing that does come to mind that I'd be happy to take as a separate PR (this is already getting too big) is to rename matchByPitch, which is very confusing, to allowIncompleteTies which describes better what is happening here.  (The way to do this though is with a one-to-two version deprecation cycle where matchByPitch is also allowed and, if changed from a default of None, overrides allowIncompleteTies and -- for this, a deprecation warning is acceptable).  I always have to look up what matchByPitch does...</s>

OH! you wrote the same thing I was thinking, yes.  But we should not reverse the boolean logic, so whatever it is, the default should remain True and work like matchByPitch=True, so it needs to be an affirmative `allowUnpairedTies` or something.  (But again, a separate PR).


> Another topic: tied grace notes are common and are currently ignored/mishandled due to the `el.quarterLength > 0` filtering, but work completely fine if the filtering is removed - is there a good reason to keep the filtering?

There's good reason to not allow non-grace notes of zero-duration to be considered (these come up in certain processing, especially with ChordSymbols) but there's nothing in the docs that talk about grace notes so that's an acceptable proposal, but it's a separate PR.  (I don't think I've ever seen a tied grace note except in a cadenza/duration-taking context).  PR scope-creep -- I'm forgetting what the original problem this PR wanted to solve was.

(Really this whole routine should be completely rewritten to use the Tree structure to run orders of magnitude faster, but that's scope-creep on MY part, and I'd only accept a PR for that which didn't alter behavior at all)

Thanks!

mscuthbert:
Yes -- I actually forgot that I stopped reviewing the PR once I saw that the same accidental(None == natural) code was being reimplemented and wrote a PR to show how to do it more simply instead.  I'm not sure that this is fully in scope for stripTies.   I'm not going to approve a PR that adds voices where they did not exist before (that was already rejected in the last PR), and something that removes notes from partially tied chords doesn't seem like part of what stripTies should do.  It all sounds like a separate method that one would call before calling stripTies, and I'm not sure the use case is sufficient for including in music21.  If there's a problem with MIDI output, then that's something that can be patched in the midi modules.

TimFelixBeyer:
The reason I proposed this functionality as part of `stripTies` is that MIDI export already calls the existing `stripTies`, so it seemed like the least intrusive part to change (with end-user functionality as a nice bonus). I thought correct behaviour (e.g. export tied MusicXML notes as sounding once) would be important to music21.

I believe that the use case *is* important enough to be exposed to users as part of stripTies (probably biased here because I'd like the functionality for my own work ;-)).
However, there are other cases too: many of the feature extractors in `music21.features.jSymbolic` rely on the existing `stripTies` and as they were originally designed for MIDI-like data, expect it to work as if `preserveVoices=False`. Other cases of this can be found in `analysis.discrete.MelodicIntervalDiversity` or e.g. if one wanted to implement Krumhansl-Schmuckler with note attack weighting.
Either way, I find it highly unlikely that anyone reading the documentation of `stripTies` would ever expect it *not* to merge partially tied chords since the docs state that it would merge "all tied notes".
Note that that is not what I am proposing, it would simply be an option for the user.

Finally, after reading your stance on the different cases, I am confused as to why a `continue`-tie is even implemented in `music21` at all, since it is always treated exactly like a `start` tie would be (except for some inconsistent chord-related edge cases, see below).

Maybe the following example can help you understand my frustrations.
Consider this setup:
```python3
from music21 import chord, tie, stream
c0 = chord.Chord('C4 E4 G4')
c0.tie = tie.Tie('start')
c1 = chord.Chord('C4 E4 G4')
c2 = chord.Chord('C4 E4 G4')
c2.tie = tie.Tie('stop')
m = stream.Measure()
m.append([c0, c1, c2])
```
Now try to predict the output of this piece of code:
```python3
tie_combinations = [
    ('start', 'start', 'start'),
    ('continue', 'start', 'start'),
    ('start', 'continue', 'start'),
    ('continue', 'continue', 'start'),     # in docs: doesn't merge
    ('start', 'start', 'continue'),
    ('continue', 'start', 'continue'),
    ('start', 'continue', 'continue'),
    ('continue', 'continue', 'continue'),  # in docs: merges
    ('start', 'continue', 'stop'),
    ('stop', 'continue', 'start'),
]
for comb in tie_combinations:
    c1.notes[0].tie = tie.Tie(comb[0])
    c1.notes[1].tie = tie.Tie(comb[1])
    c1.notes[2].tie = tie.Tie(comb[2])
    stripped = m.stripTies(matchByPitch=False)
    print(comb, stripped.elements)
```

I will of course accept your opinion, if you decide that preserving voices takes priority over everything, I will not bring this up again and close the draft.

jacobtylerwalls:
@TimFelixBeyer the fact that partially-tied chords aren't handled by stripTies() is discussed in #620. There it's discussed that some sort of explode() method needs to be implemented. Was that the main motivation for this PR?

TimFelixBeyer:
The main motivation for this PR was indeed correctly handling partially tied chords, however during the process I found that the current implementation `stripTies` is quite hard to read and buggy (see e.g. examples above), so I thought it'd be nice to cleanly reimplement the existing functionality as well (which is why i asked for feedback regarding all the cases mentioned above). In fact, after reading #620, I believe that my implementation would be something along the lines of what a option like `ChordNotesToParts.FLAT` would do. A question that just popped into my mind: Voices are basically "just" Streams, so wouldn't it be possible to split the chords with partial ties into its constituents but keep all resulting notes in the original voice, keeping all voicing information?

mscuthbert:
Hi @TimFelixBeyer -- I appreciate the work that you'd like to do here, but as #620 suggests, I think that this should be implemented elsewhere as an independent script or project and if it ends up being useful to a lot of cases after it's been well tested and in the wild, I can consider a PR later that replaces stripTies with this.  For now, I'm not willing to take on the maintenance burden of replacing a core part of the system that much 3rd-party code depends on functioning a certain way so that it works differently in a particular case.  That sounds exactly like what a different function should do.  Sorry, but I'm going to need to close this for now and we can think again on it for v10.

TimFelixBeyer:
> Hi @TimFelixBeyer -- I appreciate the work that you'd like to do here, but as #620 suggests, I think that this should be implemented elsewhere as an independent script or project and if it ends up being useful to a lot of cases after it's been well tested and in the wild, I can consider a PR later that replaces stripTies with this. For now, I'm not willing to take on the maintenance burden of replacing a core part of the system that much 3rd-party code depends on functioning a certain way so that it works differently in a particular case. That sounds exactly like what a different function should do. Sorry, but I'm going to need to close this for now and we can think again on it for v10.

No worries - I'd nonetheless (highly!) encourage you to run the Chord example from I posted last week as it shows that the core part currently does not work as the documentation suggests anyways.

jacobtylerwalls:
I did run the example last week but didn't see where it contradicted the docs. Can you point to a line number in the source?

This seems like it describes the difference between your two cases: one of them has a start; one doesn't.
https://github.com/cuthbertLab/music21/blob/d9676fb9268a387f815f28326b7e3042405d2409/music21/stream/base.py#L7195-L7197

TimFelixBeyer:
On my machine I get this output (line numbers added):

```
1, ('start', 'start', 'start') (<music21.chord.Chord C4 E4 G4>,)
2, ('continue', 'start', 'start') (<music21.chord.Chord C4 E4 G4>, <music21.chord.Chord C4 E4 G4>, <music21.chord.Chord C4 E4 G4>)
3, ('start', 'continue', 'start') (<music21.chord.Chord C4 E4 G4>,)
4, ('continue', 'continue', 'start') (<music21.chord.Chord C4 E4 G4>, <music21.chord.Chord C4 E4 G4>, <music21.chord.Chord C4 E4 G4>)
5, ('start', 'start', 'continue') (<music21.chord.Chord C4 E4 G4>,)
6, ('continue', 'start', 'continue') (<music21.chord.Chord C4 E4 G4>, <music21.chord.Chord C4 E4 G4>, <music21.chord.Chord C4 E4 G4>)
7, ('start', 'continue', 'continue') (<music21.chord.Chord C4 E4 G4>,)
8, ('continue', 'continue', 'continue') (<music21.chord.Chord C4 E4 G4>,)
9, ('start', 'continue', 'stop') (<music21.chord.Chord C4 E4 G4>,)
10, ('stop', 'continue', 'start') (<music21.chord.Chord C4 E4 G4>, <music21.chord.Chord C4 E4 G4>)
```
I think people would expect cases 2, 3, 4, 5, 6 or 7 to be handled the same (don't merge since not all ties are continue, which is what the strict option requires according to the docs), yet 3,5 and 7 are merged while the rest aren't. While the docs don't say anything specifically about options 9 and 10, they are handled differently even though the ties are effectively the same (just shuffled the pitches). 
(Just to be clear, the 3-tie combination is assigned to the members of the middle chord, while the first chord is always (start, start, start) and the third is always (stop, stop, stop)).

PR #1594, opened 2023-06-05
Speed up quantize()

TimFelixBeyer:
During quantization, rests that end up with duration 0 are deleted.
Previously, the list of rests to be deleted was not reset after finishing with a substream, yielding extremely slow performance on big scores as the code attempted to delete all previously found (and already deleted) rests again in each following substream.


mscuthbert:
Thanks Tim!

PR #1599, opened 2023-06-10
Remove xmlSoundParser

jacobtylerwalls:
Not used in the implementation from #1579 of the `<sound>` tag.


mscuthbert:
I'm going to hold off on this until after v9 is released because I actually want to go the opposite direction and move that PR to the soundParser module instead.  The xmlToMusic21 file is already too big.

mscuthbert:
This is one of the pre-July 1 PRs that I'm committing to continue to work on during the sabbatical, since it was in good shape before July 1.

jacobtylerwalls:
Sounds good. Unfortunately I don't have any current plans to work on it, so I'll just close.

mscuthbert:
Thanks @jacobtylerwalls -- it's in as #1627

Issue #1606, opened 2023-06-14
v9 incompatibility: Converter.subconvertersList vs Converter.subConvertersList


gregchapman-dev:
Can we get a deprecated old guy here, too?  I use this in converter21's `__main__.py` if you screw up the command line, so I can list available formats you can use.

mscuthbert:
Sure!  I'll look at 55207a2 and put back things that were changed.

These will not be exported by "from music21.converter import *" -- hope that's okay.

gregchapman-dev:
That's fine.

gregchapman-dev:
My tests have all run, and I have found no other issues.  Thanks!

mscuthbert:
Fixed in #1607

PR #1607, opened 2023-06-14
substitute deprecation for removal: converter.py

mscuthbert:
Converter.subconvertersList, Converter.defaultSubconverters , Converter.getSubconverterFormats, and setSubconverterFromFormat are all restored.

Note that most of these have been converted in their newer capital-C form to also be callable from the Class (staticmethods) -- this has not been added to the old forms which behave like before.

also keep taking every opportunity to see what readthedocs.io wants...


mscuthbert:
Fixes #1606

PR #1614, opened 2023-06-19
Feature/figuredbass object

mxordn:
I'm opening this as a smaller step towards representing and importing figures bass information. PR #1613 is the big PR that is split here into three parts as suggested there.

One of the first things is to add the FiguredBass Object, that stores the information in the score/stream.

I'm comenting on the remarks by [mscuthbert](https://github.com/mscuthbert):

> Leave the repr for fb.notation.Figure alone for the common case where hasExtension is False. Add "+extension" or "--" or something for that case. in general, touch as little as needed.

Done so now.

> Make isExtension a property that computes automatically or at least is updated if number or figure is changed.

The Figure __init__ does this now.

> When adding an additional less-used feature to an existing method, generally make it keyword only. (such as extenders)

> need typing on extenders -- no idea what it is expecting. (all new additions to m21 must be typed). Try to avoid X | None types where falsey X (like an empty tuple) can suffice.

I changes the handling of extenders now slightly and decided that they only can be parsed as underscores. Extenders are a rather big problem. For now also the musicxml and mei parser can work with the underscore char.

> only camelCase not underscore_case for exposed properties, etc.

changed the snake case things

> what is "originalString" doing in Modifier that is different than modifierString?

it stores an "original" version with the original values. E. g. a figure number 3 would not be added to a pure '#'

> in FiguedBassIndication:
> - rename to FiguredBass

Done

> - figs -- neds typing of list of what? remove None option and make default be () or ''

Done

> - watch indentation there. See format elsewhere -- all arguments on one line or each argument on its own line,.

Alright didn't know that. Done.

> - Python's not TypeScript: isFigure, part, _figs should be defined on the instance not the Class.

Done

> - do not use str for "part" -- people will expect that "part" refers to a stream.Part object

renamed to "correspondingPart"

> - fig_notation --> notation, since it returns a Notation object (and no snake case). needs typing and docs. Currently violates these property: f = harmony.FiguredBassIndication(...);  f.fig_notation = f.fig_notation should not change f.fig_notation

Should be solved. Although I'm not sure, if I got your point.

Hope this is going better in this way.


music21/figuredBass/notation.py
```diff
@@ -30,6 +30,33 @@
                      (2,): (6, 4, 2),
                      }
 
+prefixes = ['+', '#', '++', '##']
+suffixes = ['\\']
+
+modifiersDictXmlToM21 = {'sharp': '#',
```
mscuthbert, line 36: indent as 

```
modifiersDictXmlToM21 = {
    'sharp' :'#',
    'flat': 'b',
    ...etc.
}
```

so that all keys begin on the same column and have four spaces showing they're inside a top-level dictionary. (same below)

music21/figuredBass/notation.py
```diff
@@ -189,12 +234,15 @@ def __init__(self, notationColumn=None):
         self.origModStrings = None
         self.numbers = None
         self.modifierStrings = None
+        self.extenders:list[bool] = []
```
mscuthbert, line 237: space before `list[bool]`

music21/figuredBass/notation.py
```diff
@@ -189,12 +234,15 @@ def __init__(self, notationColumn=None):
...
+        self.hasExtenders: bool = False
         self._parseNotationColumn()
         self._translateToLonghand()
 
         # Convert to convenient notation
         self.modifiers = None
         self.figures = None
+        self.figuresFromNotationColumn: list(Figure) = []
```
mscuthbert, line 245: list[Figure]  -- square brackets not parentheses.

music21/figuredBass/notation.py
```diff
@@ -247,17 +303,40 @@ def _parseNotationColumn(self):
...
                 modifierString = m2[0].strip()
 
             numbers.append(number)
             modifierStrings.append(modifierString)
+            self.extenders.append(extender)
 
         numbers = tuple(numbers)
         modifierStrings = tuple(modifierStrings)
 
+        # extenders come from the optional argument when instantionting the object.
```
mscuthbert, line 331: typo.  instantiating.

music21/figuredBass/notation.py
```diff
@@ -247,17 +303,40 @@ def _parseNotationColumn(self):
...
+        # If nothing is provided, no extenders will be set.
+        # Otherwise we have to look if amount of extenders and figure numbers match
+        #
```
mscuthbert, line 334: remove blank `#`

music21/figuredBass/notation.py
```diff
@@ -433,6 +560,13 @@ def _reprInternal(self):
                     '++': '##',
                     '+++': '###',
                     '++++': '####',
+                    '\u266f': '#',
```
mscuthbert, line 563: ...a place where existing code doesn't match the modern style mentioned above.  :-O  ah well, can't be perfect.  We didn't have good linters in 2008

music21/figuredBass/notation.py
```diff
@@ -410,15 +518,34 @@ class Figure(prebase.ProtoM21Object):
...
+            ''',
+        'isExtender': '''
+            A bool value that returns true if an extender is part of the figure but no
+            number is given. Pure extender if you will.
+            It is set by evaluating the number and extender arguments.
+            '''
     }
 
-    def __init__(self, number=1, modifierString=None):
+    isExtender: bool
```
mscuthbert, line 532: remove -- not a class property.

music21/figuredBass/notation.py
```diff
@@ -410,15 +518,34 @@ class Figure(prebase.ProtoM21Object):
...
+
+    def __init__(self, number=1, modifierString=None, extender=False):
         self.number = number
         self.modifierString = modifierString
         self.modifier = Modifier(modifierString)
+        # look for extender's underscore
+        self.hasExtender: bool = extender
+        self._updateIsExtenderProperty()
```
mscuthbert, line 540: remove -- see below.

music21/figuredBass/notation.py
```diff
@@ -410,15 +518,34 @@ class Figure(prebase.ProtoM21Object):
...
+            A bool value that returns true if an extender is part of the figure but no
+            number is given. Pure extender if you will.
+            It is set by evaluating the number and extender arguments.
+            '''
     }
 
-    def __init__(self, number=1, modifierString=None):
+    isExtender: bool
+
+    def __init__(self, number=1, modifierString=None, extender=False):
```
mscuthbert, line 534: change to:

```
def __init__(self, number=1, modifierString=None, *, extender: bool = False):
```
so that extender has to be given as a keyword.  Otherwise people will do this in code:

```
n = notation.Figure(3, '#', True)
```

and I'll be scratching my head about what the True means.

music21/figuredBass/notation.py
```diff
@@ -410,15 +518,34 @@ class Figure(prebase.ProtoM21Object):
...
         self.number = number
         self.modifierString = modifierString
         self.modifier = Modifier(modifierString)
+        # look for extender's underscore
+        self.hasExtender: bool = extender
+        self._updateIsExtenderProperty()
+        
+    def _updateIsExtenderProperty(self):
+        self.isExtender = (self.number == 1 and self.hasExtender)
```
mscuthbert, line 543: what I meant is if I create a figure object like:

```
>>> n = notation.Figure(1, '#', extender=True)
```

then the `.isExtender` property can get out of date:

```
>>> n.isExtender
True
>>> n.hasExtender = False
>>> n.isExtender
True
```

instead don't store `isExtender` anywhere and recalculate it on the fly.  Let's rename it
to `pureExtender` or `isPureExtender` since that makes it clear what the difference with
hasExtender is:

```
@property
def pureExtender(self):
    '''
    Read-only boolean property that returns True if an extender is part of the 
    figure but no number is given (a number of 1 means no-number). It is a pure extender.

    >>> n = notation.Figure(1, '#', extender=True)
    >>> n.isExtender
    True
    >>> n.number = 2
    >>> n.isExtender
    False
    '''
    return self.number == 1 and self.hasExtender
```
and then it will always be up-to-date no matter what.

note -- should a pureExtender also require notationString to be '' or None?

music21/figuredBass/notation.py
```diff
@@ -410,15 +518,34 @@ class Figure(prebase.ProtoM21Object):
...
 
     def _reprInternal(self):
         mod = repr(self.modifier).replace('music21.figuredBass.notation.', '')
+        if self.isExtender or self.hasExtender:
+            return f'{self.number} {mod} extender: {True}'
```
mscuthbert, line 548: isExtender and hasExtender is redundant, right?  since there should never be a case where something hasExtender but isExtender is False.

Maybe:

```
def _reprInternal(self):
    if self.isPureExtender:
        return 'pure extender'
    mod = repr(self.modifier).replace('music21.figuredBass.notation.', '')
    extender = '___' if self.hasExtender else ''
    return f'{self.number} {mod}{extender}'
```

music21/harmony.py
```diff
@@ -2499,6 +2500,57 @@ def transpose(self: NCT, _value, *, inPlace=False) -> NCT | None:
...
+
+    The single figures are stored as figuredBass.notation.Figure objects:
+    >>> fb.figNotation.figures[0]
+    <music21.figuredBass.notation.Figure 3 <Modifier # sharp>>
+    '''
+
+
+    def __init__(self,
+                 figureString: str | list[str] | None = None,
+                 correspondingPart: str | None=None,
```
mscuthbert, line 2521: sorry, I wasn't clear about this.  Anything ending in Part (part, correspondingPart, activePart, etc.) looks like it should be a Part object.  partId would be clear that it is a string or None.

But I don't know why this would need this information at all -- the Part that an object is part of can always be retrieved by:

```
fb.getContextByClass(stream.Part)
```

then if the Figure is moved to a new Part or there are two parts with the same id, the FiguredBass will always be with the right part.  This is why Note, Clef, etc. don't need a correspondingPart object.

music21/harmony.py
```diff
@@ -2499,6 +2500,57 @@ def transpose(self: NCT, _value, *, inPlace=False) -> NCT | None:
...
+                 **keywords):
+        super().__init__(**keywords)
+
+        self.isFigure: bool = True
+        self.corresPart: str | None = None
```
mscuthbert, line 2526: it's moot if we go the route above, but no abbreviations.

music21/harmony.py
```diff
@@ -2499,6 +2500,57 @@ def transpose(self: NCT, _value, *, inPlace=False) -> NCT | None:
...
+    '''
+
+
+    def __init__(self,
+                 figureString: str | list[str] | None = None,
+                 correspondingPart: str | None=None,
+                 **keywords):
+        super().__init__(**keywords)
+
+        self.isFigure: bool = True
```
mscuthbert, line 2525: Why is this needed?

music21/harmony.py
```diff
@@ -2499,6 +2500,57 @@ def transpose(self: NCT, _value, *, inPlace=False) -> NCT | None:
...
+            if isinstance(figureString, list):
+                self._figs = ','.join(figureString)
+            elif isinstance(figureString, str):
+                if ',' in figureString:
+                    self._figs = figureString
+                else:
+                    self._figs = ','.join(figureString)
+        else:
+            self._figs = ''
+        self._figNotation = notation.Notation(self._figs)
```
mscuthbert, line 2539: needs typing.

music21/harmony.py
```diff
@@ -2499,6 +2500,57 @@ def transpose(self: NCT, _value, *, inPlace=False) -> NCT | None:
...
+    <FiguredBass figures: #,6# correspondingPart: P1>
+
+    The single figures are stored as figuredBass.notation.Figure objects:
+    >>> fb.figNotation.figures[0]
+    <music21.figuredBass.notation.Figure 3 <Modifier # sharp>>
+    '''
+
+
+    def __init__(self,
+                 figureString: str | list[str] | None = None,
```
mscuthbert, line 2520: After music21 moved to the world of typing extensions (around v6) figureString should just take a string.  If someone wants to pass in a list of figures that can be a different keyword-only argument like figureStrings.

There are lots of cases in the current code that haven't moved to this paradigm because they're grandfathered in, but we're trying to make it so that new code mostly takes one type of object.

music21/harmony.py
```diff
@@ -2499,6 +2500,57 @@ def transpose(self: NCT, _value, *, inPlace=False) -> NCT | None:
...
+        self._figNotation = notation.Notation(self._figs)
+        self.corresPart = correspondingPart
+
+    @property
+    def figNotation(self) -> notation.Notation:
+        return self._figNotation
+
+    @figNotation.setter
+    def figNotation(self, figs: str | list[str] | None):
+        self._figNotation = notation.Notation(figs)
```
mscuthbert, line 2548: as noted above, this doesn't pass reflexive setters.  (also no abbreviations.  This should just be `notation`)

Needs docs and tests.

it seems like there should be two sets of properties.  One for getting/setting the Notation object, and one for getting/setting the string of figures -- changing one should change the other.

music21/harmony.py
```diff
@@ -2499,6 +2500,57 @@ def transpose(self: NCT, _value, *, inPlace=False) -> NCT | None:
 
 # ------------------------------------------------------------------------------
 
+class FiguredBass(Harmony):
+    '''
+    The FiguredBassIndication objects store information about thorough bass figures.
+    It is created as a representation for <fb> tags in MEI and <figured-bass> tags in MusicXML.
+    The FiguredBassIndication object derives from the Harmony object and can be used
+    in the following way:
+    >>> fb = harmony.FiguredBass('#,6#', correspondingPart='P1')
```
mscuthbert, line 2509: space before.

music21/harmony.py
```diff
@@ -2499,6 +2500,57 @@ def transpose(self: NCT, _value, *, inPlace=False) -> NCT | None:
...
+    @property
+    def figNotation(self) -> notation.Notation:
+        return self._figNotation
+
+    @figNotation.setter
+    def figNotation(self, figs: str | list[str] | None):
+        self._figNotation = notation.Notation(figs)
+
+    def __repr__(self):
+        return f'<{self.__class__.__name__} figures: {self.figNotation.notationColumn} correspondingPart: {self.corresPart}>'
```
mscuthbert, line 2551: Music21Objects don't define `__repr__` just `_reprInternal`, which would be just the part from `figures: {...}` to the end before `>` -- let the Music21Object `__repr__` take care of the class name, etc.

Generally the representation just provides the most relevant information, so I think it could just be `return self.notation.notationColumn`.  in .show('text') it'll be obvious what Part the FiguredBass object belongs to.

mscuthbert (changes requested):
On the right track -- some changes requested (and more tests and documentation).

In general, objects do not keep track of what Part they are in; so I'd need a very compelling reason why FiguredBass needs a correspondingPart when ChordSymbols etc. don't.

Thanks!

music21/figuredBass/notation.py
mxordn: I see. Not storing this seperatly it makes sense. Thanks for adding this. I will include it.

music21/harmony.py
mxordn: I would then add the two keywords figureString and figureStrings. If nothing at all is handed over, it will result in an empty FiguredBass.

mscuthbert:
Keeping track of the associated bass Note or Pitch object for the figure and the Key object seems like it could be important--since unlike RomanNumerals or ChordSymbols, a FB doesn't make sense without a bass (and unlike ChordSymbols but like RN it doesn't make sense without a Key object). 

One thing that there are not yet tests on is .pitches -- we need to be able to get reasonable pitches out of the FB object. That's where breaking this PR up into its component parts has helped with the reasoning. 

There may be parts of the RN code that might be factored out to make more reusable there. Do you mind if I push code to this PR? I think I have an idea of how things may work...

mxordn:
Sure you can contribute to the PR. It’d be a pleasure.

In general, it would be nice to have a pitch associated with the fb notation. Although I was careful with that, because there might be e.g. a rest with a figure in a score. In these cases the figure might just indicate the sonority, that sounds at a specific point (with the related bass note somewhere else in the score).

I’d like to be able to preserve the possibility to either to have FiguredBass without a bass note explicitly specified or to be able to compute the related note out of a score. Therefore an offset is might be sufficient.

mscuthbert:
yes -- there are definitely cases where the pitch being attached to is other than the one sounding (definitely in the case of 4-3 suspensions, etc.) so that's where a reference to a previous pitch would be helpful.

music21/harmony.py
mxordn: Good idea. I'm pushing a proposal.

mscuthbert:
I'm a little bit behind w/ other work, so I'll extend work on this into my sabbatical for a little bit (until July 10) -- if we can get this merged by then, I'll continue working on it.

music21/figuredBass/notation.py
```diff
@@ -366,11 +447,26 @@ def _getFigures(self):
         for i in range(len(self.numbers)):
             number = self.numbers[i]
             modifierString = self.modifierStrings[i]
+            if self.extenders:
+                if i < len(self.extenders):
+                    extender = self.extenders[i]
+            else:
+                extender = False
+            figure = Figure(number, modifierString, extender)
             figure = Figure(number, modifierString)
```
mscuthbert, line 456: this is immediately wiped out -- mistake?  Not tested enough.

mscuthbert (approved):

PR #1615, opened 2023-06-23
Remove useless Pylint disables

jacobtylerwalls:
Not the most substantial PR, but it's nice to remove useless pylint disables every once in a while.

Found this with the `--enable=useless-suppression` argument.


mscuthbert:
Thank you!  :-)  These were all former bug-fixes that now just work!

mypy failure is unrelated; from a recent upgrade.  #1616 got hit by it too and fixes the problem.  Will re-run tests and merge this once #1616 is in.

PR #1618, opened 2023-06-26
Mini-optimization to getElementAtOrBefore

TimFelixBeyer:
Replaces an unnecessary sort with max and simplifies the remaining logic of the function.

### DIFF
--- a/music21/stream/base.py
+++ b/music21/stream/base.py
@@ -4144,22 +4144,18 @@ class Stream(core.StreamCore, t.Generic[M21ObjType]):
             #   'e.offset', e.offset, 'self.elementOffset(e)', self.elementOffset(e), 'e', e])
             if span < 0:
                 continue
-            elif span == 0:
+            elif span == nearestTrailSpan:
                 candidates.append((span, e))
+            elif span < nearestTrailSpan:
+                candidates = [(span, e)]
                 nearestTrailSpan = span
-            else:
-                # do this comparison because may be out of order
-                if span <= nearestTrailSpan:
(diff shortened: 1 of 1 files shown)


mscuthbert:
Hi: the max is approved but some of the other logic changed--before if span was zero it was also added to candidates regardless of nearestTailSpan. Can you justify the change? Thanks!

(I think that the TODO caution is likely obsolete after the inclusion of sortTuple but will need to go back in the code to be sure)

TimFelixBeyer:
Sure:
First, since `nearestTrailSpan` is always >= 0, “ we will still always add candidates that have a span of 0 (either creating a new candidates list or appending to the existing one). 
Secondly, only candidates with the lowest span can end up as the final candidate, making it ok to discard all previous candidates if we find one with a lower span.
Therefore the logic actually remains unchanged.

mscuthbert:
I did the initial review on my phone where I couldn't reread all the logic.  Now that I do, I understand why these are equivalent, we could actually cut out one more `if` to simply the logic just to one `<=` because setting `nearestTrailSpan = span` in the case of equal is faster than doing two equality comparisons, but none of this really matters, since we really should be switching this to either trees or using a bisect search on `_elements` (if sorted) -- a lot of this code is still from the time that autosort was opt-in rather than opt-out (and turning it off was supposed to not have hugely negative consequences -- now it is allowed to).

The other possible improvement is to use .iter(restoreActiveSites=False) and only set the activeSite on the element that is returned, and if the stream is sorted to break once it's found the best candidate.  I'll merge this and make a second PR for that.

PR #1619, opened 2023-06-26
Mini-optimization to updateEndTimes

TimFelixBeyer:
Very slightly improves the performance of `updateEndTimes` which is called millions of times in some tree-related tasks.
Does not affect functionality. Same for `recurseGetTreeByClass`.


mscuthbert:
Thanks!

PR #1620, opened 2023-06-26
Speed up Duration deepcopy

TimFelixBeyer:
Instead of updating the components during deepcopy (which is expensive), we can defer this update until it is actually needed and just copy the object for now. This speeds up tasks that use derivations (e.g. `quantize`) by up to 30% on my machine.


mscuthbert:
Ah yes. That makes sense that the update would be slower than just doing a standard copy.

PR #1621, opened 2023-06-26
Speed up hasElement by 2x

TimFelixBeyer:
Instead of comparing the result of the `id` function between objects, we can use the built-in `is` operator to achieve the same in half the time. `coreHasElementByMemoryLocation` was only used this one time and can therefore be removed as well.


TimFelixBeyer:
Maybe it would make sense to combine the offsetDict check into `__contains__` which implements the same functionality as `hasElement` I believe?

mscuthbert:
happy to merge this now.  I'll defer the `__contains__` check until I'm off the dev sabbatical.

PR #1624, opened 2023-06-27
speed up sortTuple for streams

TimFelixBeyer:
This avoids the computation of the duration attribute when creating a stream's `sortTuple`, which is expensive as it requires iterating over all elements to compute `highestTime`.
Yields significant speedups on tasks that require sorting (12% overall for a simple load + quantize benchmark on my machine).

### DIFF
--- a/music21/base.py
+++ b/music21/base.py
@@ -2725,10 +2725,8 @@ class Music21Object(prebase.ProtoM21Object):
             offset = t.cast(OffsetQL, foundOffset)
             atEnd = 0
 
-        if self.duration.isGrace:
-            isNotGrace = 0
-        else:
-            isNotGrace = 1
+        # avoids expensive duration computation for streams, which can never be grace notes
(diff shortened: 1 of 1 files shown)


mscuthbert:
Yes! This is significant!

I had the idea at some point of a GraceStream which was a stream entirely of grace notes for things like complex ornaments that have beams and their own accent patterns etc which take up zero conceptual time. But I don't think I ever implemented that and if I did then we can add a special case. 

Streams are one of the few subclasses of Music21Object that the base class is allowed to be aware of (that and Spanner) so putting this code in the main class is okay. 

I'll merge but please note that i'm less of a fan of "x = 1 if ...compound-tests-that-are-long... else 0" type of expressions rather than the simple if x: ... else: ... type. I try to avoid them except in generators/comprehensions/lambdas. And even then only if the normal case is left of the if.

TimFelixBeyer:
I’ll keep that in mind for future PRs!

PR #1627, opened 2023-07-01
SoundParser to a separate file

mscuthbert:
Going in the opposite direction as #1599 -- move MusicXML `<sound>` tag parsing to a separate file along with several helpers to helpers.py -- both m21ToXML and xmlToM21 are getting huge! so anything to try to make these files more manageable is the right direction.

While fixing mypy errors, noticed a bunch of warnings that things that were typed were not being checked in noteworthy/binaryTranslate.py -- checked these and fixed.  (TODO: make the contents of untyped functions are not checked become an error in CI)

### DIFF
--- a/music21/musicxml/helpers.py
+++ b/music21/musicxml/helpers.py
@@ -12,9 +12,20 @@
 from __future__ import annotations
 
 import copy
+import typing as t
 from xml.etree.ElementTree import tostring as et_tostring
 
+from music21 import common
 from music21 import meter
+from music21.musicxml import xmlObjects
+from music21 import prebase
+
+if t.TYPE_CHECKING:
+    from collections.abc import Callable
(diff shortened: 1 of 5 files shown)


jacobtylerwalls:
Maybe [`--check-untyped-defs`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-check-untyped-defs)?

mscuthbert:
> Maybe [`--check-untyped-defs`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-check-untyped-defs)?

It's something to do with changing status of `annotation-unchecked` from note to error.  [This question](https://stackoverflow.com/questions/74578185/suppress-mypy-notes) shows how to make it go away completely, but not how to make it go from note to error.

PR #1629, opened 2023-07-08
Speed up __contains__ by 2x and deprecate hasElement

TimFelixBeyer:
Follow up to #1621.
I am aware of the dev sabbatical, just opening the PR now so I don't forget about it later...

Proposal:
`hasElement` has identical functionality to `__contains__`. Since the Zen of Python teaches us that "There should be one-- and preferably only one --obvious way to do it.", I propose deprecating `hasElement` in favor of the more concise and pythonic `__contains__`.
In addition, I have  optimized the speed of the function by adding the fast `offsetDict` check to `__contains__`, and after stumbling on [this discussion](https://stackoverflow.com/questions/44802682/python-any-unexpected-performance), and changing the calls to `any`, the changes resulted in about a 2x speed-up on my machine, which is significant since the contains check is used a lot throughout the package.
For `self._elements` with fewer than 200 elements, this would be slightly faster (10%) still:

```python3
if len(self._elements) > 200:
    return any(True for x in self._elements if x is obj)
for el in self._elements:
    if el is obj:
        return True
return False
```

Since the absolute time spent in `__contains__` is dominated by checks with large `self._elements`, I think the trade-off favors the more readable variant over this slightly faster, optimized version. Let me know what you think!
Here's a comparison graph vs the speed of `hasElement` before #1621:
Figure_1

I have also removed all uses of `hasElement` in the music21 codebase since the method would be deprecated.

### DIFF
--- a/music21/analysis/reduction.py
+++ b/music21/analysis/reduction.py
@@ -286,12 +286,12 @@ class ScoreReduction:
             return
 
         removalIndices = []
-        if m.hasElement(n):
+        if n in m:
             offset = n.getOffsetBySite(m)
         else:  # it is in a Voice
             offset = 0.0
             for v in m.voices:
-                if v.hasElement(n):
+                if n in v:
                     offset = n.getOffsetBySite(v)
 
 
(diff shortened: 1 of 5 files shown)


mscuthbert:
drove by on the way to something else and looks good -- thanks for not doing the more complex > 200 version -- added complexity not worth the small speedup there, but this is.  When I started music21 I didn't know about dunder methods for `in` (I'm not sure Perl had them) so that's why hasElement was added.  -- back to sabbatical

PR #1632, opened 2023-07-25
Fix ugly MusicXML durations & offsets

TimFelixBeyer:
We all know the ugly offsets (3875/480, 161/180, etc.) that happen after importing some musicXML files.
This usually occurs because the file contains tuplets with strange durations that cannot be represented exactly using integer `<duration>` tags in musicXML. For this reason, rounding errors can accumulate and lead to over/underfull measures.
Currently, if a measure is overfull, music21 just accepts that as a fact and increases the duration of the measure, shifting all following objects and resulting in them having very odd offsets. 

This proposal deals with this by automatically detecting such cases, warning the user, and fixing them. 
The change only affects behavior if the measure duration resulting from the contained notes differs from the duration dictated by the time signature by less than 0.5 and is a very ugly duration. 
In my testing, this fixes the vast majority of imports without breaking anything else.

Let me know what you think!

### DIFF
--- a/music21/_version.py
+++ b/music21/_version.py
@@ -47,7 +47,7 @@ Changing this number invalidates old pickles -- do it if the old pickles create 
 '''
 from __future__ import annotations
 
-__version__ = '9.2.0b1'
+__version__ = '9.2.0b2'
 
 def get_version_tuple(vv):
     v = vv.split('.')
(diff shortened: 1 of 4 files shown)


mscuthbert:
this looks good -- can you rename "PP" in the tests to "pp" -- capital letters should only be used to represent classes, not instances of classes.  Then anyone w/ merge rights should feel free to merge.

TimFelixBeyer:
I actually fully agree, just did this to be in-line with all the other tests. 
I've now lowercased all instances of PartParser that i could find.

mscuthbert (approved):
Thanks! Sorry for letting the inconsistencies in in the first place

PR #1635, opened 2023-08-19
Roman Numerals: Don't substitute `d` for `0` in `add10`

malcolmsailor:
`add10` annotations (which occur when converting some DCML files) were causing a bug because the 0 was being converted to `o`, leading to parsing exceptions.

This PR updates the replace logic for `0` so as not to match this case, and adds a test case to the unit test.


jacobtylerwalls:
Thanks for the fix and test!

jacobtylerwalls:
For future work with regexes, esp. in such a common code path like RN instantiation, it would be good to do timing tests when considering alternatives. I think without the lookbehind you can get an improvement:
```python
>>> from timeit import timeit
>>> timeit("re.sub(r'(?<!\d)0', 'o', 'vii07')", setup='import re')
0.4586234579910524
>>> timeit("re.sub(r'(\\D)0', r'\1o', 'viio7')", setup='import re')
0.324576957966201
```

PR #1636, opened 2023-08-20
Fix #1633

### LINKED ISSUE #1633: Musicxml import wrong for voices that don't span whole bar
**music21 version**

`9.8.7`   

**Problem summary**
When dealing with voices that don't span a whole bar, music21 wrongly interprets notes and voices. It adds rests that don't exist and causes simultaneous notes to be not on the same beat.

**Steps to reproduce**
consider the voices in the treble. I attached the xml file (one bar) to reproduce this.

```
from music21 import *

# load the musicxml file
file_path = 'test.musicxml'
score = converter.parse(file_path)
score.show()

for el in score.recurse():
    print(el)
```

**corresponding file**
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE score-partwise PUBLIC "-//Recordare//DTD MusicXML 3.1 Partwise//EN" "http://www.musicxml.org/dtds/partwise.dtd">
<score-partwise version="3.1">
  <part-list>
    <score-part id="P1">
      <part-name>Piano</part-name>
      <score-instrument id="P1-I1">
        <instrument-name>Piano</instrument-name>
        </score-instrument>
      <midi-device id="P1-I1" port="1"></midi-device>
      <midi-instrument id="P1-I1">
        <midi-channel>1</midi-channel>
        <midi-program>1</midi-program>
        <volume>78.7402</volume>
        <pan>0</pan>
        </midi-instrument>
      </score-part>
    </part-list>
  <part id="P1">
    <measure number="1" width="295.92">
      <print>
        <system-layout>
          <system-margins>
            <left-margin>85.60</left-margin>
            <right-margin>0.00</right-margin>
            </system-margins>
          <top-system-distance>182.91</top-system-distance>
          </system-layout>
        <staff-layout number="2">
          <staff-distance>73.55</staff-distance>
          </staff-layout>
        </print>
      <attributes>
        <divisions>4</divisions>
        <key>
          <fifths>-3</fifths>
          </key>
        <time symbol="common">
          <beats>4</beats>
          <beat-type>4</beat-type>
          </time>
        <staves>2</staves>
        <clef number="1">
          <sign>F</sign>
          <line>4</line>
          </clef>
        <clef number="2">
          <sign>F</sign>
          <line>4</line>
          </clef>
        </attributes>
      <direction placement="below">
        <direction-type>
          <dynamics default-y="-40.00" relative-y="-40.00">
            <ff/>
            </dynamics>
          </direction-type>
        <staff>1</staff>
        <sound dynamics="124.44"/>
        </direction>
      <direction placement="above">
        <direction-type>
          <words default-x="-40.91" default-y="46.91" relative-y="20.00" font-weight="bold" font-family="Times New Roman" font-size="12">Largo</words>
          </direction-type>
        <staff>1</staff>
        <sound tempo="30"/>
        </direction>
      <note default-x="123.38" default-y="-5.00">
        <pitch>
          <step>G</step>
          <octave>3</octave>
          </pitch>
        <duration>4</duration>
        <voice>1</voice>
        <type>quarter</type>
        <stem>down</stem>
        <staff>1</staff>
        <notations>
          <slur type="start" placement="above" number="1"/>
          </notations>
        </note>
      <note default-x="123.38" default-y="10.00">
        <chord/>
        <pitch>
          <step>C</step>
          <octave>4</octave>
          </pitch>
        <duration>4</duration>
        <voice>1</voice>
        <type>quarter</type>
        <stem>down</stem>
        <staff>1</staff>
        </note>
      <note default-x="123.38" default-y="20.00">
        <chord/>
        <pitch>
          <step>E</step>
          <alter>-1</alter>
          <octave>4</octave>
          </pitch>
        <duration>4</duration>
        <voice>1</voice>
        <type>quarter</type>
        <stem>down</stem>
        <staff>1</staff>
        </note>
      <note default-x="123.38" default-y="30.00">
        <chord/>
        <pitch>
          <step>G</step>
          <octave>4</octave>
          </pitch>
        <duration>4</duration>
        <voice>1</voice>
        <type>quarter</type>
        <stem>down</stem>
        <staff>1</staff>
        </note>
      <note default-x="161.09" default-y="0.00">
        <pitch>
          <step>A</step>
          <alter>-1</alter>
          <octave>3</octave>
          </pitch>
        <duration>4</duration>
        <voice>1</voice>
        <type>quarter</type>
        <stem>down</stem>
        <staff>1</staff>
        </note>
      <note default-x="161.09" default-y="10.00">
        <chord/>
        <pitch>
          <step>C</step>
          <octave>4</octave>
          </pitch>
        <duration>4</duration>
        <voice>1</voice>
        <type>quarter</type>
        <stem>down</stem>
        <staff>1</staff>
        </note>
      <note default-x="161.09" default-y="20.00">
        <chord/>
        <pitch>
          <step>E</step>
          <alter>-1</alter>
          <octave>4</octave>
          </pitch>
        <duration>4</duration>
        <voice>1</voice>
        <type>quarter</type>
        <stem>down</stem>
        <staff>1</staff>
        </note>
      <note default-x="161.09" default-y="35.00">
        <chord/>
        <pitch>
          <step>A</step>
          <alter>-1</alter>
          <octave>4</octave>
          </pitch>
        <duration>4</duration>
        <voice>1</voice>
        <type>quarter</type>
        <stem>down</stem>
        <staff>1</staff>
        </note>
      <note default-x="198.80" default-y="20.00">
        <pitch>
          <step>E</step>
          <alter>-1</alter>
          <octave>4</octave>
          </pitch>
        <duration>3</duration>
        <voice>1</voice>
        <type>eighth</type>
        <dot/>
        <stem>up</stem>
        <staff>1</staff>
        <beam number="1">begin</beam>
        </note>
      <note default-x="198.80" default-y="30.00">
        <chord/>
        <pitch>
          <step>G</step>
          <octave>4</octave>
          </pitch>
        <duration>3</duration>
        <voice>1</voice>
        <type>eighth</type>
        <dot/>
        <stem>up</stem>
        <staff>1</staff>
        </note>
      <note default-x="232.25" default-y="15.00">
        <pitch>
          <step>D</step>
          <octave>4</octave>
          </pitch>
        <duration>1</duration>
        <voice>1</voice>
        <type>16th</type>
        <stem>up</stem>
        <staff>1</staff>
        <beam number="1">end</beam>
        <beam number="2">backward hook</beam>
        </note>
      <note default-x="232.25" default-y="25.00">
        <chord/>
        <pitch>
          <step>F</step>
          <octave>4</octave>
          </pitch>
        <duration>1</duration>
        <voice>1</voice>
        <type>16th</type>
        <stem>up</stem>
        <staff>1</staff>
        </note>
      <note default-x="256.61" default-y="-15.00">
        <pitch>
          <step>E</step>
          <alter>-1</alter>
          <octave>3</octave>
          </pitch>
        <duration>4</duration>
        <voice>1</voice>
        <type>quarter</type>
        <stem>down</stem>
        <staff>1</staff>
        <notations>
          <slur type="stop" number="1"/>
          </notations>
        </note>
      <note default-x="256.61" default-y="-5.00">
        <chord/>
        <pitch>
          <step>G</step>
          <octave>3</octave>
          </pitch>
        <duration>4</duration>
        <voice>1</voice>
        <type>quarter</type>
        <stem>down</stem>
        <staff>1</staff>
        </note>
      <note default-x="256.61" default-y="10.00">
        <chord/>
        <pitch>
          <step>C</step>
          <octave>4</octave>
          </pitch>
        <duration>4</duration>
        <voice>1</voice>
        <type>quarter</type>
        <stem>down</stem>
        <staff>1</staff>
        </note>
      <note default-x="256.61" default-y="20.00">
        <chord/>
        <pitch>
          <step>E</step>
          <alter>-1</alter>
          <octave>4</octave>
          </pitch>
        <duration>4</duration>
        <voice>1</voice>
        <type>quarter</type>
        <stem>down</stem>
        <staff>1</staff>
        </note>
      <backup>
        <duration>16</duration>
        </backup>
      <forward>
        <duration>8</duration>
        </forward>
      <note default-x="198.80" default-y="-5.00">
        <pitch>
          <step>G</step>
          <octave>3</octave>
          </pitch>
        <duration>4</duration>
        <voice>2</voice>
        <type>quarter</type>
        <stem>down</stem>
        <staff>1</staff>
        </note>
      <note default-x="198.80" default-y="5.00">
        <chord/>
        <pitch>
          <step>B</step>
          <octave>3</octave>
          </pitch>
        <duration>4</duration>
        <voice>2</voice>
        <type>quarter</type>
        <accidental>natural</accidental>
        <stem>down</stem>
        <staff>1</staff>
        </note>
      <backup>
        <duration>12</duration>
        </backup>
      <note default-x="123.38" default-y="-173.55">
        <pitch>
          <step>C</step>
          <octave>2</octave>
          </pitch>
        <duration>4</duration>
        <voice>5</voice>
        <type>quarter</type>
        <stem>up</stem>
        <staff>2</staff>
        </note>
      <note default-x="123.38" default-y="-138.55">
        <chord/>
        <pitch>
          <step>C</step>
          <octave>3</octave>
          </pitch>
        <duration>4</duration>
        <voice>5</voice>
        <type>quarter</type>
        <stem>up</stem>
        <staff>2</staff>
        </note>
      <note default-x="161.09" default-y="-193.55">
        <pitch>
          <step>F</step>
          <octave>1</octave>
          </pitch>
        <duration>4</duration>
        <voice>5</voice>
        <type>quarter</type>
        <stem>up</stem>
        <staff>2</staff>
        </note>
      <note default-x="161.09" default-y="-158.55">
        <chord/>
        <pitch>
          <step>F</step>
          <octave>2</octave>
          </pitch>
        <duration>4</duration>
        <voice>5</voice>
        <type>quarter</type>
        <stem>up</stem>
        <staff>2</staff>
        </note>
      <note default-x="198.80" default-y="-188.55">
        <pitch>
          <step>G</step>
          <octave>1</octave>
          </pitch>
        <duration>4</duration>
        <voice>5</voice>
        <type>quarter</type>
        <stem>up</stem>
        <staff>2</staff>
        </note>
      <note default-x="198.80" default-y="-153.55">
        <chord/>
        <pitch>
          <step>G</step>
          <octave>2</octave>
          </pitch>
        <duration>4</duration>
        <voice>5</voice>
        <type>quarter</type>
        <stem>up</stem>
        <staff>2</staff>
        </note>
      <note default-x="256.61" default-y="-173.55">
        <pitch>
          <step>C</step>
          <octave>2</octave>
          </pitch>
        <duration>4</duration>
        <voice>5</voice>
        <type>quarter</type>
        <stem>up</stem>
        <staff>2</staff>
        </note>
      <note default-x="256.61" default-y="-153.55">
        <chord/>
        <pitch>
          <step>G</step>
          <octave>2</octave>
          </pitch>
        <duration>4</duration>
        <voice>5</voice>
        <type>quarter</type>
        <stem>up</stem>
        <staff>2</staff>
        </note>
      <note default-x="256.61" default-y="-138.55">
        <chord/>
        <pitch>
          <step>C</step>
          <octave>3</octave>
          </pitch>
        <duration>4</duration>
        <voice>5</voice>
        <type>quarter</type>
        <stem>up</stem>
        <staff>2</staff>
        </note>
      </measure>   
    </part>
  </score-partwise>
´´´

### PR DESCRIPTION
**Pull request by TimFelixBeyer**
No longer create hidden rests while parsing forward tags in musicXML files, as this led to problems. If there are unfilled gaps, they will be filled later by makeRests anyways if required.
This also allows us to remove some code to deal with trailing rests.

### DIFF
--- a/music21/musicxml/xmlToM21.py
+++ b/music21/musicxml/xmlToM21.py
@@ -31,6 +31,7 @@ from music21 import defaults
 from music21 import duration
 from music21 import dynamics
 from music21.common.enums import OrnamentDelay
+from music21.common.numberTools import opFrac
 from music21 import editorial
 from music21 import environment
 from music21 import exceptions21
@@ -1760,31 +1761,7 @@ class PartParser(XMLParserBase):
         for mxMeasure in self.mxPart.iterfind('measure'):
             self.xmlMeasureToMeasure(mxMeasure)
 
-        self.removeEndForwardRest()
         part.coreElementsChanged()
-
-    def removeEndForwardRest(self):
-        '''
-        If the last measure ended with a forward tag, as happens
-        in some pieces that end with incomplete measures,
-        and voices are not involved,
-        remove the rest there (for backwards compatibility, esp.
-        since bwv66.6 uses it)
-
-        * New in v7.
-        '''
-        if self.lastMeasureParser is None:  # pragma: no cover
-            return  # should not happen
-        lmp = self.lastMeasureParser
-        self.lastMeasureParser = None  # clean memory
-
-        if lmp.endedWithForwardTag is None:
-            return
-        if lmp.useVoices is True:
-            return
-        endedForwardRest = lmp.endedWithForwardTag
-        if lmp.stream.recurse().notesAndRests.last() is endedForwardRest:
-            lmp.stream.remove(endedForwardRest, recurse=True)
 
     def separateOutPartStaves(self) -> list[stream.PartStaff]:
         '''
@@ -2233,7 +2210,7 @@ class PartParser(XMLParserBase):
                     else:
                         self.lastMeasureWasShort = False
 
-        self.lastMeasureOffset += mOffsetShift
+        self.lastMeasureOffset = opFrac(self.lastMeasureOffset + mOffsetShift)
 
     def applyMultiMeasureRest(self, r: note.Rest):
         '''
@@ -2389,13 +2366,6 @@ class MeasureParser(SoundTagMixin, XMLParserBase):
 
         # what is the offset in the measure of the current note position?
         self.offsetMeasureNote: OffsetQL = 0.0
-
-        # keep track of the last rest that was added with a forward tag.
-        # there are many pieces that end with incomplete measures that
-        # older versions of Finale put a forward tag at the end, but this
-        # disguises the incomplete last measure.  The PartParser will
-        # pick this up from the last measure.
-        self.endedWithForwardTag: note.Rest | None = None
 
     @staticmethod
     def getStaffNumber(mxObjectOrNumber) -> int:
@@ -2596,10 +2566,8 @@ class MeasureParser(SoundTagMixin, XMLParserBase):
         '''
         mxDuration = mxObj.find('duration')
         if durationText := strippedText(mxDuration):
-            change = common.numberTools.opFrac(
-                float(durationText) / self.divisions
-            )
-            self.offsetMeasureNote -= change
+            change = opFrac(float(durationText) / self.divisions)
+            self.offsetMeasureNote = opFrac(self.offsetMeasureNote - change)
             # check for negative offsets produced by
             # musicxml durations with float rounding issues
             # https://github.com/cuthbertLab/music21/issues/971
@@ -2611,22 +2579,9 @@ class MeasureParser(SoundTagMixin, XMLParserBase):
         '''
         mxDuration = mxObj.find('duration')
         if durationText := strippedText(mxDuration):
-            change = common.numberTools.opFrac(
-                float(durationText) / self.divisions
-            )
-
-            # Create hidden rest (in other words, a spacer)
-            # old Finale documents close incomplete final measures with <forward>
-            # this will be removed afterward by removeEndForwardRest()
-            r = note.Rest(quarterLength=change)
-            r.style.hideObjectOnPrint = True
-            self.addToStaffReference(mxObj, r)
-            self.insertInMeasureOrVoice(mxObj, r)
-
+            change = opFrac(float(durationText) / self.divisions)
             # Allow overfilled measures for now -- TODO(someday): warn?
-            self.offsetMeasureNote += change
-            # xmlToNote() sets None
-            self.endedWithForwardTag = r
+            self.offsetMeasureNote = opFrac(self.offsetMeasureNote + change)
 
     def xmlPrint(self, mxPrint: ET.Element):
         '''
@@ -2785,8 +2740,7 @@ class MeasureParser(SoundTagMixin, XMLParserBase):
             self.nLast = c  # update
 
         # only increment Chords after completion
-        self.offsetMeasureNote += offsetIncrement
-        self.endedWithForwardTag = None
+        self.offsetMeasureNote = opFrac(self.offsetMeasureNote + offsetIncrement)
 
     def xmlToChord(self, mxNoteList: list[ET.Element]) -> chord.ChordBase:
         # noinspection PyShadowingNames
@@ -3578,7 +3532,7 @@ class MeasureParser(SoundTagMixin, XMLParserBase):
         mxDuration = mxNote.find('duration')
         if mxDuration is not None:
             noteDivisions = float(mxDuration.text.strip())
-            qLen = common.numberTools.opFrac(noteDivisions / divisions)
+            qLen = opFrac(noteDivisions / divisions)
         else:
             qLen = 0.0
 
@@ -5510,7 +5464,7 @@ class MeasureParser(SoundTagMixin, XMLParserBase):
                 meth(mxSub)
             # NOT to be done: directive -- deprecated since v2.
             elif tag == 'divisions':
-                self.divisions = common.opFrac(float(mxSub.text))
+                self.divisions = opFrac(float(mxSub.text))
             # TODO: musicxml4: for-part including part-clef
             # TODO: instruments -- int if more than one instrument plays most of the time
             # TODO: part-symbol

### CODE REVIEW

File: music21/musicxml/xmlToM21.py [hunk 1]
```diff
@@ -866,6 +867,19 @@ def xmlRootToScore(self, mxScore, inputM21=None):
...
+                if v:  # do not bother with empty voices
+                    # the musicDataMethods use insertCore, thus the voices need to run
+                    # coreElementsChanged
+                    v.coreElementsChanged()
+                    # Fill mid-measure gaps, and find end of measure gaps by ref to measure stream
+                    # https://github.com/cuthbertlab/music21/issues/444
+                    v.makeRests(refStreamOrTimeRange=m,
+                                fillGaps=True,
+                                inPlace=True,
+                                hideRests=True)
```
jacobtylerwalls, line 882: I think this is too presumptive on the part of the parser. We should be interpolating/inventing as few elements as possible that are not in the document.

music21/musicxml/xmlToM21.py
```diff
@@ -2588,18 +2560,16 @@ def xmlBackup(self, mxObj: ET.Element):
...
         0.0
         '''
         mxDuration = mxObj.find('duration')
         if durationText := strippedText(mxDuration):
-            change = common.numberTools.opFrac(
-                float(durationText) / self.divisions
-            )
-            self.offsetMeasureNote -= change
+            change = opFrac(float(durationText) / self.divisions)
+            self.offsetMeasureNote = opFrac(self.offsetMeasureNote - change)
```
jacobtylerwalls, line 2572: This is a useful change that I'd be glad to accept in isolation.

jacobtylerwalls (changes requested):
Hi @TimFelixBeyer thanks for submitting a patch. I want to suggest a different solution, which is not that we get stop interpreting `<forward>` tags as hidden rests but that we just infer the correct voice number after a `<backup>` tag, i.e. that we make the solution from #1030 more robust. WDYT?

music21/musicxml/xmlToM21.py
TimFelixBeyer: I agree! 
I just put the makeRests there because previously the parser also ran `makeRests`, but I'd be more than happy to remove it.

TimFelixBeyer:
> Hi @TimFelixBeyer thanks for submitting a patch. I want to suggest a different solution, which is not that we stop interpreting `<forward>` tags as hidden rests but that we just infer the correct voice number after a `<backup>` tag, i.e. that we make the solution from #1030 more robust. WDYT?

I completely agree with your logic about not 'inventing' elements where possible. In my mind it would therefore follow that we should also not interpret <forward> elements as rests (since they aren't rests).
Feel free to let me know your thoughts!
The ideal solution for me personally would be to not create hidden rests for forward tags & to also not run makeRests, since they both alter the score as it was saved to XML, as users could always run those functions later as well.

jacobtylerwalls:
I guess without getting into the ontology of nothingness 😰, I think the reason for the status quo is that creating hidden rests in notation programs (at least Finale) will be exported as forward tags. And there is a slight difference between an explicit nothing marked as "do not show" versus an implicit nothing "document silent/incomplete".

TimFelixBeyer:
I'm not super familiar with notation programs like finale, if forward tags are usually used to represent hidden rests, it totally makes sense to me to keep them. On the other hand if we decide that `<forward>` tags are always to be interpreted as hidden rests, I don't think there is a way for 'nothingness' to be represented in musicXML (from music21's perspective) at all, gaps will always have hidden rests (which is also what the makeRests call currently ensures). I don't really know whether this matters much/is intended - feel free to lmk what you think.

jacobtylerwalls:
But makeRests() is only called on export, so gaps/silence can still be represented on import or on export with makeNotation=False.

TimFelixBeyer:
Maybe I’m missing something, but I don’t think this is accurate: https://github.com/cuthbertLab/music21/blob/fdb126c2c22c53a919a42cf5c0354f3b2aed4c99/music21/musicxml/xmlToM21.py#L2565

`makeRests` is called every time on import already, irrespective of the makeNotation keyword

jacobtylerwalls:
Ah, but only in voices, not measures. I suppose the reason could be that incomplete voices are more likely to lead to display problems than incomplete measures but I suppose we could test out removing that call.

TimFelixBeyer:
The call to `makeRests` was introduced in #444 so removing it would likely lead to the same problems described in that issue. 
The only CI error I get is here, which is expected: https://github.com/cuthbertLab/music21/blob/f457a2ba52ea3f978d805cd52fa101dfb0dd8577/music21/musicxml/test_xmlToM21.py#L1245

I think it just depends on what the preference of the `music21` maintainers is: how should `<forward>` tags be interpreted (as hidden rests, or as empty space)? My vote would be that even if programs like finale abuse the `forward` tag to represent hidden rests, that its not a faithful representation of the information in the musicXML format if we change forward tags to hidden rests.
Another alternative (most complex, but probably most accurate) would be to parse the tags differently depending on the `<software>` tag in the XML file.

jacobtylerwalls:
I think a survey of other major musicxml writers would be helpful. What do MuseScore and Sibelius do? Dorico?

Daniel63656:
In musicxml this is just "empty space". In fact, it is even less than that. <Forward> and <Backward> are just used to reposition a "cursor" so voices can also start mid-measure. If this is not handled, correct xml import is guaranteed to be wrong in a lot of cases as I illustrated in #1633

jacobtylerwalls:
Sorry for the long delay. And thanks for the thoughtful responses. Agree we need to do something here, so I'll try to help suggest the least breaking change to get this in.

> The call to makeRests was introduced in https://github.com/cuthbertLab/music21/issues/444

Actually it was there before, but #444 just extended the "gap finding" to apply two additional arguments in order to find mid-measure and trailing rests. (makeRests defaults to only finding leading rests for some reason).

> Another alternative (most complex, but probably most accurate) would be to parse the tags differently depending on the <software> tag in the XML file.

Actually, that doesn't sound so bad. I bet we could just guard the fix for #444 under a check for the software tag, which has already been parsed (although somewhat inconveniently inserted with coreInsert(), so at this juncture has to be checked against _elements):

```py
(Pdb) self.parent.parent.stream._elements[0].software
('Finale v26.3 for Mac')
```

Then #444 becomes something like:
```py
if finale:
    v = makeRests(all four arguments)
else:
    v = makeRests(the two arguments before #444)
```

Then, @TimFelixBeyer you are right that removing the code in xmlForward to create leading hidden rests solves #1633, so I'm okay with that change and the related cleanups.

(Finally, the leading hidden rest that remains once we put #444 under a Finale guard doesn't cause problems with parsing and showing #1633, so I don't think we need to deal with that here to get this change in. If we want to remove it eventually, we can discuss it separately.)

Does that sound like a reasonable way to get #1633 fixed? LMK!

TimFelixBeyer:
Thanks for your reply!

What do you think about this alternative:
```
if finale:
    v = makeRests(all four arguments)
else:
    pass
```
Or alternatively we just have two implementations for the forward tag, one for finale, which inserts hidden rests and one which doesn't do anything. Then we could skip all calls to `makeRests` during import.

jacobtylerwalls:
Alright, I'm on board. Getting rid of makeRests altogether on import for non-Finale sounds good.

TimFelixBeyer:
@jacobtylerwalls It should now all be implemented as discussed and should fix #1633 and #1715.

I also added a test for finale and non-finale hidden-rest-on-<forward>-tag behavior. 
Sorry for the messy commit history, I was using Github's Editor - feel free to squash them before merging. 

Also had to include a minor mypy fix to `common/enums.py` to get the linting test to pass.

music21/musicxml/test_xmlToM21.py
```diff
@@ -1290,7 +1301,7 @@ def testHiddenRestImpliedVoice(self):
 
         self.assertEqual(len(MP.stream.voices), 2)
         self.assertEqual(len(MP.stream.voices[0].elements), 1)
-        self.assertEqual(len(MP.stream.voices[1].elements), 2)
+        self.assertEqual(len(MP.stream.voices[1].elements), 1)
```
jacobtylerwalls, line 1304: Would you mind updating the unit test name to avoid mentioning "hidden rests" now that we no longer create them? Maybe: gap?

music21/musicxml/xmlToM21.py
```diff
@@ -867,6 +867,21 @@ def xmlRootToScore(self, mxScore, inputM21=None):
             self.spannerBundle.remove(sp)
 
         s.coreElementsChanged()
+        # Fill gaps with rests where needed
+        for m in s[stream.Measure]:
```
jacobtylerwalls, line 871: Why move this out of the MeasureParser? If it's just to have access to the `md` instance, we could just query for it in the measure parser (with getContextByClass, which is fast). To do that, we likely need to avoid `coreInsert`'ing it when inserting for the reason [mentioned](https://github.com/cuthbertLab/music21/pull/1636#issuecomment-2143924582) earlier.

jacobtylerwalls (commented):
Great! Two little cleanups, and then in my opinion this will be g2g.

music21/musicxml/xmlToM21.py
```diff
@@ -867,6 +867,21 @@ def xmlRootToScore(self, mxScore, inputM21=None):
...
+            for v in m.voices:
+                if v:  # do not bother with empty voices
+                    # the musicDataMethods use insertCore, thus the voices need to run
+                    # coreElementsChanged
+                    v.coreElementsChanged()
```
jacobtylerwalls, line 876: Note to self: if we decide to leave this here after all, this is probably not needed (but still think MeasureParser is better if I'm not missing something)

music21/musicxml/xmlToM21.py
```diff
@@ -1761,32 +1776,8 @@ def parseMeasures(self):
...
-        and voices are not involved,
-        remove the rest there (for backwards compatibility, esp.
-        since bwv66.6 uses it)
-
-        * New in v7.
-        '''
-        if self.lastMeasureParser is None:  # pragma: no cover
-            return  # should not happen
-        lmp = self.lastMeasureParser
-        self.lastMeasureParser = None  # clean memory
```
jacobtylerwalls, line 1780: I think we can remove this instance attribute now.

music21/musicxml/xmlToM21.py
mscuthbert: It's been a while since I've thought through this PR, but I think that this is correct.

mscuthbert:
Nice -- I suppose -- to know that there's not a future version of Finale on its way which might break this.

music21/musicxml/xmlToM21.py
TimFelixBeyer: I had a look - I moved it there since it didn't play nice with partwise musicxml files. I think this is because the rests inserted by `makeRests` don't have an entry in `staffReference`, and therefore can't be placed in the correct part until after the parts have been pulled apart. 

One example is schoenberg op19.2. - `testHiddenRests` fails when keeping this inside the measure parser because a hole in voice 2 is filled with a half rest. this voice is supposed to be the top staff, but the rest ends up in the bottom staff, where we end up with two voices instead of no voices (one with the newly inserted rest, and one with the actual data for the bottom staff).

TimFelixBeyer: I wasn't able to quickly resolve this in another way - lmk if you have any thoughts.

TimFelixBeyer: I think this test wasn't failing before because we got luckky and this particular case happened in the last measure of the piece, where the incorrect rest was removed by `removeEndForwardRest`.

jacobtylerwalls: Thanks for the ping. I wanted to see what you were talking about, so here's what I stubbed out:
```diff
diff --git a/music21/musicxml/test_xmlToM21.py b/music21/musicxml/test_xmlToM21.py
index ab1e99879..c8fe4421c 100644
--- a/music21/musicxml/test_xmlToM21.py
+++ b/music21/musicxml/test_xmlToM21.py
@@ -1547,4 +1547,4 @@ class Test(unittest.TestCase):
 
 if __name__ == '__main__':
     import music21
-    music21.mainTest(Test)
+    music21.mainTest('noDocTest', Test, runTest="testHiddenRests")
diff --git a/music21/musicxml/xmlToM21.py b/music21/musicxml/xmlToM21.py
index 0caa04f55..b385aeaab 100644
--- a/music21/musicxml/xmlToM21.py
+++ b/music21/musicxml/xmlToM21.py
@@ -821,7 +821,7 @@ class MusicXMLImporter(XMLParserBase):
             self.musicXmlVersion = mxVersion
 
         md = self.xmlMetadata(mxScore)
-        s.coreInsert(0, md)
+        s.insert(0, md)
 
         mxDefaults = mxScore.find('defaults')
         if mxDefaults is not None:
@@ -867,21 +867,7 @@ class MusicXMLImporter(XMLParserBase):
             self.spannerBundle.remove(sp)
 
         s.coreElementsChanged()
-        # Fill gaps with rests where needed
-        for m in s[stream.Measure]:
-            for v in m.voices:
-                if v:  # do not bother with empty voices
-                    # the musicDataMethods use insertCore, thus the voices need to run
-                    # coreElementsChanged
-                    v.coreElementsChanged()
-                    # Fill mid-measure gaps, and find end of measure gaps by ref to measure stream
-                    # https://github.com/cuthbertlab/music21/issues/444
-                    # but only when the score comes from Finale
-                    if any('Finale' in software for software in md.software):
-                        v.makeRests(refStreamOrTimeRange=m,
-                                    fillGaps=True,
-                                    inPlace=True,
-                                    hideRests=True)
+
         s.definesExplicitSystemBreaks = self.definesExplicitSystemBreaks
         s.definesExplicitPageBreaks = self.definesExplicitPageBreaks
         for p in s.parts:
@@ -2558,8 +2544,25 @@ class MeasureParser(SoundTagMixin, XMLParserBase):
                 if methName is not None:
                     meth = getattr(self, methName)
                     meth(mxObj)
+        md = self.parent.parent.stream[metadata.Metadata].first()
         for v in self.stream[stream.Voice]:
-            v.coreElementsChanged()
+            if v:  # do not bother with empty voices
+                # the musicDataMethods use insertCore, thus the voices need to run
+                # coreElementsChanged
+                v.coreElementsChanged()
+                # Fill mid-measure gaps, and find end of measure gaps by ref to measure stream
+                # https://github.com/cuthbertlab/music21/issues/444
+                # but only when the score comes from Finale
+                if md and any('Finale' in software for software in md.software):
+                    old_ids = {id(el) for el in v}
+                    v.makeRests(refStreamOrTimeRange=self.stream,
+                                fillGaps=True,
+                                inPlace=True,
+                                hideRests=True)
+                    for el in v:
+                        if id(el) not in old_ids:
+                            self.addToStaffReference(self.mxMeasure, el)
+
         self.stream.coreElementsChanged()
 
         if (self.restAndNoteCount['rest'] == 1
@@ -6238,4 +6241,4 @@ class MeasureParser(SoundTagMixin, XMLParserBase):
 
 if __name__ == '__main__':
     import music21
-    music21.mainTest()  # doctests only
+    music21.mainTest('noDocTest', runTest='testHiddenRests')  # doctests only
```

Now I get:
```py
  File "/Users/jwalls/music21/music21/musicxml/test_xmlToM21.py", line 1282, in testHiddenRests
    self.assertIsInstance(hiddenRest, note.Rest)
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/unittest/case.py", line 1294, in assertIsInstance
    self.fail(self._formatMessage(msg, standardMsg))
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/unittest/case.py", line 715, in fail
    raise self.failureException(msg)
AssertionError: <music21.chord.Chord E-5 F#5 B-5 D6> is not an instance of <class 'music21.note.Rest'>
```

But when I look at this piece on IMSLP, that seems correct. No rests at the end of the file. So my next thought was to look at how the file is encoded in the version in the music21 corpus.

I have to run, but my intent is to get you an answer in the next few days, unless you already know the answer and can enlighten me :-)

Thanks for your persistence.

mscuthbert: > I had a look - I moved it there since it didn't play nice with partwise musicxml files. 

Do you mean MusicXML files where some parts have more than one staff?  Almost all MusicXML files are Partwise, and I don't think music21 even imports Timewise since I've never seen one in the wild.

TimFelixBeyer: Yeah I meant multi-staff parts, sorry. Basically anything with a `<staff>` tag that forces m21 to later use `separateOutPartStaves` to move something to a different staff.

TimFelixBeyer: @jacobtylerwalls i'm not 100% how `addToStaffReference` is supposed to work, but I'm pretty sure measure elements don't have a `<staff>` element assigned to them, only notes or rests do. Therefore the added rests will probably always be added to the "NO_STAFF_ASSIGNED" bucket. I think the right way to do this would be to find either the following or preceding note in the same voice and put the added rests in the same staff as that note.
Maybe `separateOutPartStaves` is smart enough to do that already?

mscuthbert:
btw -- I am mostly staying out of this for now, because Jacob is really the expert on the `<forward>` `<backward>` and `<staff>` parsing -- Chris Ariza wrote the original code, the refactoring I did didn't touch this code, and Jacob did a ton of work on it later, so it's one of the (not many) parts of music21 that I'm not an expert on.

music21/musicxml/xmlToM21.py
jacobtylerwalls: > No rests at the end of the file. So my next thought was to look at how the file is encoded in the version in the music21 corpus.

Forget that, I must have not used `forceSource=True` or something -- I get a much more understandable failure now.

@TimFelixBeyer to your question, you're absolutely right that we need the staff reference. The bulletproof way to do that would be to get it from the `<forward>` tag and rely on that for hidden rest creation, which is what we were doing before. Instead off adding this makeRests() call under a Finale guard, what do you think about putting the old `xmlForward` tag code that created hidden rests and assigned staff references under a Finale guard? You should get the same overall amount of net simplicity, no?

gregchapman-dev:
Any chance we can pick this up ASAP?  I really need it for PR #1766 (sometimes turning a forward tag into an invisible rest is actually incorrect).

gregchapman-dev:
FWIW, my opinion is that (unconditionally) removing the hidden rest from xmlForward is the absolutely critical part of this fix (I have MusicXML files I can share where that hidden rest will overlap with other notes; it's really wrong).

jacobtylerwalls:
@gregchapman-dev Did my reply at https://github.com/cuthbertLab/music21/pull/1636/files#r1921565041 make sense? If so, would you like to pair up on this?

gregchapman-dev:
@jacobtylerwalls I understood what you said as "how about creating a hidden rest in xmlForward, but only for Finale docs?", but I disagree.  That hidden rest from xmlForward can be completely wrong, overlapping with existing notes in the voice (or measure, if no voices).  Imagine a MusicXML file that lays out all the notes in a measure/voice, then does a `<backup>` and lays in the directions for that measure/voice, skipping ahead to the next direction each time with `<forward>`.  xmlForward must _not_ create hidden rests in this case, or they will overlap with the notes.  Calling makeRests after the fact is great (and honestly I'm happy to do it for everyone, not just for Finale docs; I hate gaps) because it will only create hidden rests to fill existing gaps (i.e. never overlapping).  The only problem remaining is how to get the right staff number (if the hidden rest is in a multi-staff part).  But couldn't we just call makeRests _after_ splitting out the PartStaffs?  Then the hidden rests would always end up in the correct staff, without us having to know the right staff number.

I'm happy to discuss/pair up on this PR; but I already have two PRs in progress, that overlap with each other, and both of them overlap with this one.  That's too much overlapping for my brain (and also my development environment).  So you would have to do the actual driving of this PR from _your_ development environment, and I would merge those changes into my two PRs, to test.

jacobtylerwalls:
For the moment let's grant for the sake of argument there are simply no such Finale files that back up and write spanners as you suggest. (I'll check some files this weekend.) Also grant me that the forward tags are the only place to parse the info because making rests afterward would be lossy (will catch encoding omissions and treat them like actually "encoded" hidden rests).

With those two points granted, would you still be unwilling to make an exception for a specific software?

And yes happy to do what I can to unblock this!

gregchapman-dev:
When there are SpannerAnchors in a music21 Score, music21 itself writes files like this.  And unfortunately, if the source MusicXML file for that Score was written by Finale, the music21-written MusicXML will also say it was written by Finale (and music21; we add to the "software" metadata, never delete), so bingo, there's a Finale doc with lots of `<forward>` tags that would generate invalid hidden rests.  I will, however, grant you that filling gaps with hidden rests after import is not exactly correct, so we need a better solution than the one I proposed.

I don't think we can remove Finale from the "software" metadata on write; we never have deleted a "software" metadata value before.  Could we notice that Finale is not "most recent" in the "software" metadata list, and decide the file wasn't really written by Finale, for purposes of this xmlForward decision?  I'm not sure of the definition of "most recent".  I know where we add "music21", but other software might add at the other end of the list.

gregchapman-dev:
@jacobtylerwalls See PR #1777 where I propose a simple(ish) solution.

jacobtylerwalls:
Still planning to get to this soon. I found your reply convincing re: we can't special-case Finale because m21 adds to the software tags.

Honestly my interest in parsing hidden rests as hidden rests and not as gaps stemmed from a project dealing with loads of them in choral scores (where you can imagine constant stems up / stems down nonsense expressed graphically with hidden rests)

gregchapman-dev:
I think PR #1777 does a pretty good job of adding hidden rests for documents that were definitively written by Finale.

PR #1651, opened 2023-10-02
wrap siteDict access in try/catch

jhansen317:
Not sure if this has unintended side effects, but seems to work. The issue affected <10% of the midi files I am working with (i hesitate to put a number on it because I haven't really counted, but a minority.) I didn't see any open issues that looked germane, but I did not undertake an exhaustive search.

### DIFF
--- a/music21/base.py
+++ b/music21/base.py
@@ -2728,11 +2728,14 @@ class Music21Object(prebase.ProtoM21Object):
         # avoids expensive duration computation for streams, which can never be grace notes
         isNotGrace = 1 if self.isStream or not self.duration.isGrace else 0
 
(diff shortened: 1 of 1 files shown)


jacobtylerwalls:
Hi @jhansen317 :wave:

Thanks for raising the issue. What version of music21 were you using? If you can reproduce on the latest, can you share the smallest MIDI file you have that reproduces the problem? I think we'd prefer to understand and fix the root cause rather than add a try. Thanks.

mscuthbert:
closing for lack of discussion / tests / etc.  Please feel free to reopen if this becomes active again.

PR #1652, opened 2023-10-02
wrap pitch access in try/catch

jhansen317:
This was failing with 
```
'PercussionChord/Unpitched' object has no attribute 'pitch'
```
when calling chordify() on, say, orchestral scores with timpani, etc.


mscuthbert:
conditionalAdd is typed as requiring note.Note -- could you shift the guard to be at the point where conditionalAdd is being called?  Thanks

mscuthbert:
Closing in favor of #1678 which implements the same safety but in a more type-aware manner.  Thanks for the great idea.

PR #1656, opened 2023-10-19
Improve commonName for enharmonic equivalent to minor seventh chords

vanderstel:
This PR extends the approach from https://github.com/cuthbertLab/music21/pull/1245 by accounting for minor 7th chords that are spelled enharmonically as C-E-G-Bbb.

Pinging @jacobtylerwalls since he wrote the original code, but would welcome a review from any contributor.


mscuthbert:
thanks!

PR #1657, opened 2023-10-23
check if HOME directory is writeable in getSettingsPath

Leo1998:
I came across this problem when running a script on a compute cluster where the home dir exists, but is mounted read-only when running on the actual compute node.

In this case we should store the .music21rc file in /tmp.


mscuthbert:
Thanks Leo1998 -- great first commit!

PR #1665, opened 2023-11-18
Fix argument annotation in `converter.parse()` signature

jacobtylerwalls:
Fixes error reported in https://github.com/cuthbertLab/music21/pull/1664#issuecomment-1817500436.

Adds missing coverage.

PR #1666, opened 2023-11-19
Avoid testing truth value of ElementTree.Element 

jacobtylerwalls:
Resolves deprecation warnings in python 3.12
Refs https://github.com/python/cpython/pull/31149

(mypy failure fixed in #1665)


mscuthbert (approved):
Thank you!

PR #1677, opened 2024-01-02
add some casts for PyCharm

mscuthbert:
Not a problem in mypy but PyCharm is complaining that "all" is still possible.


mscuthbert:
not worth it

PR #1683, opened 2024-01-10
Avoid calling hyphenToCamelCase twice per note

alexandermorgan:
`hyphenToCamelCase` from common.stringTools.py currently gets called twice for every note in a piece when importing an xml file. This can be avoided by passing an `attributeName` argument when using the `_setAttributeFromTagText` method to set the step and octave properties of notes.

With this change, when importing a file like:
`score = corpus.parse('joplin/maple_leaf_rag.mxl', forceSource=True)`
`_setAttributeFromTagText` goes from being called over 30k times to about 120 times. Since all the string manipulation in hyphenToCamelCase has no impact on one-word strings, all these avoided calls were unnecessary.

### DIFF
--- a/music21/musicxml/xmlToM21.py
+++ b/music21/musicxml/xmlToM21.py
@@ -3160,8 +3160,8 @@ class MeasureParser(SoundTagMixin, XMLParserBase):
             if mxPitch is None:  # whoops!!!!
                 return p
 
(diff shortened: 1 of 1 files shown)


jacobtylerwalls (approved):
Thanks! Same spirit as 7a9a8f56e3e3435eb456f90cb2990eccb1fe0928.

PR #1684, opened 2024-01-19
In kern allow non-eighth type & non-slashed grace notes, improve default duration of non-first chord notes, decrease import time by ~15%

alexandermorgan:
This PR has some very minor substantive changes with accompanying tests and several small optimizations for kern import

Substantive changes:
1. Grace notes without slashes are now detected when two lowercase q's are present
2. Grace notes apply their stated duration rather than always having eighth note representations. About these two grace note points, you can see in [this example](https://verovio.humdrum.org/?t=KiprZXJuCjE2YS1xCjJnCjhmI3FxCjJnCj09CiotCg==) that humdrum is sensitive to the stated duration of a grace note, and also that the second q will omit the slash.
3. In chords where only the first note has a stated duration, the other notes of the chord will take on the duration of the first note. The same exact duration object is used, similar to how chords already currently reuse the duration of their first note. Although standard humdrum syntax calls for a duration on each note of a chord, you can see [here](https://verovio.humdrum.org?t=KiprZXJuCjRnIDRiCjRmIGEKMmUgZwoxZCBmCj09CiotCg==) that when only the first note has a stated duration, other chords will use the same rather than defaulting to a quarter note.

Kern import optimizations (mostly in hdStringToNote)
1. `if str.count('x'):` becomes `if 'x' in str:`
2. Move duration parsing of notes to the beginning of hdStringToNote so that notes and rests can be instantiated with their correct duration from the beginning rather than being created and then having their duration replaced. This prevents an unused default duration from being created on each note and rest. Even better for chords, this pr makes them inherit the duration of their first note when the chord is created rather than right after creation, which means that instead of creating 1 unused duration per chord, now no new ones are created. I didn't check, but this super simple change for chords could be a quick and easy win for other import types if they construct chords in the same way.
3. Combine the normal duration and rational duration regexes into one. Likewise for note name and accidental searches.
4. Use walrus operator where possible to avoid duplicating operations
5. Add short-circuit to stop processing a note/rest if all the characters have been accounted for after the duration and note name/accidental have been processed
6. Comment out parsing that only results in pass statements
7. Memoize the duration and note name regex searches inside of hdStringToNote. New durations, notes, and rests are still instantiated for each event.

### DIFF
--- a/music21/pitch.py
+++ b/music21/pitch.py
@@ -553,7 +553,10 @@ def _convertHarmonicToCents(value: int|float) -> int:
 # -----------------------------------------------------------------------------
 
 
-def _dissonanceScore(pitches, smallPythagoreanRatio=True, accidentalPenalty=True, triadAward=True):
+def _dissonanceScore(pitches: list[Pitch],
+                     smallPythagoreanRatio: bool = True,
+                     accidentalPenalty: bool = True,
+                     triadAward: bool = True):
     r'''
     Calculates the 'dissonance' of a list of pitches based on three criteria:
     it is considered more consonant if 1. the numerator and denominator of the
@@ -577,8 +580,12 @@ def _dissonanceScore(pitches, smallPythagoreanRatio=True, accidentalPenalty=True
 
     if smallPythagoreanRatio or triadAward:
         try:
-            intervals = [interval.Interval(noteStart=p1, noteEnd=p2)
-                        for p1, p2 in itertools.combinations(pitches, 2)]
+            intervals = []
+            for p1, p2 in itertools.combinations(pitches, 2):
+                p2 = copy.deepcopy(p2)
+                p2.octave = None
+                this_interval = interval.Interval(noteStart=p1, noteEnd=p2)
+                intervals.append(this_interval)
         except interval.IntervalException:
             return math.inf
     if smallPythagoreanRatio:
@@ -586,8 +593,7 @@ def _dissonanceScore(pitches, smallPythagoreanRatio=True, accidentalPenalty=True
         for this_interval in intervals:
             # does not accept weird intervals, e.g. with semitones
             ratio = interval.intervalToPythagoreanRatio(this_interval)
-            # d2 is 1.0
-            penalty = math.log(ratio.numerator * ratio.denominator / ratio) * 0.03792663444
+            penalty = math.log(ratio.denominator) * 0.07585326888
             score_ratio += penalty
 
         score_ratio /= len(pitches)
@@ -607,7 +613,7 @@ def _dissonanceScore(pitches, smallPythagoreanRatio=True, accidentalPenalty=True
                                                                  + accidentalPenalty + triadAward)
 
 
-def _bruteForceEnharmonicsSearch(oldPitches, scoreFunc=_dissonanceScore):
+def _bruteForceEnharmonicsSearch(oldPitches: list[Pitch], scoreFunc=_dissonanceScore):
     '''
     A brute-force way of simplifying -- useful if there are fewer than 5 pitches
     '''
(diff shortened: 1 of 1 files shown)


alexandermorgan:
Is there any input on this pr? It fixes two bugs for me when reading kern files so I would like to see it integrated if possible. I can also file those two bugs as issues if that would help. Thanks.

mscuthbert:
Hi Alex -- as I've mentioned on the list, I'm taking a hiatus from music21 except for paid support, unless I happen to have a bit of extra time.  I'll try to get this when I can.  Note that there's some conflict w/ the PR of Tim #1693 -- whichever gets merged first will need to have some edits on the other.

music21/humdrum/spineParser.py
```diff
@@ -2119,7 +2122,27 @@ def getAllOccurring(self):
         return retEvents
 
 
-def hdStringToNote(contents):
+@lru_cache(maxsize=256)
+def _noteMemos(contents):
```
mscuthbert, line 2126: new music21 routines require typing.  Can you type the input and output of this routine?

music21/humdrum/spineParser.py
```diff
@@ -2119,7 +2122,27 @@ def getAllOccurring(self):
...
+    else:
+        matchedNote = re.search('([a-gA-G]+)([n#-]*)', contents)
+        if not matchedNote:
+            raise HumdrumException(f'Could not parse {contents} for note information')
+        accountedFor += len(matchedNote.group(0))
+        step = matchedNote.group(1)[0].lower()
+    return (dur_regex, dotCount, matchedNote, step, accountedFor)
+
+
+def hdStringToNote(contents, defaultDuration=None):
```
mscuthbert, line 2145: add typing at least to the new attribute (but it'd be kind to add typing to all)

music21/humdrum/spineParser.py
mscuthbert: Give some docs about what this does.

music21/humdrum/spineParser.py
```diff
@@ -2119,7 +2122,27 @@ def getAllOccurring(self):
...
+        accountedFor += 1
+        matchedNote = None
+        step = None
+    else:
+        matchedNote = re.search('([a-gA-G]+)([n#-]*)', contents)
+        if not matchedNote:
+            raise HumdrumException(f'Could not parse {contents} for note information')
+        accountedFor += len(matchedNote.group(0))
+        step = matchedNote.group(1)[0].lower()
+    return (dur_regex, dotCount, matchedNote, step, accountedFor)
```
mscuthbert, line 2142: don't mix underscore and camelCase in internal variable names.  Choose one or the other.

music21/humdrum/spineParser.py
```diff
@@ -2199,197 +2222,186 @@ def hdStringToNote(contents):
...
+
+                dur.appendTuplet(newTup)
+                if JRP is True and dotCount:
+                    dur.dots = dotCount
+                # call Duration.TupletFixer after to correct this.
+    elif defaultDuration is not None:
+        dur = defaultDuration
+    else:
+        if 'q' in contents:
+            dur = duration.Duration(.5, dots=dotCount)
```
mscuthbert, line 2281: style: 0.5 not .5

music21/humdrum/spineParser.py
```diff
@@ -2199,197 +2222,186 @@ def hdStringToNote(contents):
     >>> humdrum.spineParser.flavors['JRP'] = storedFlavors  #_DOCS_HIDE
 
     '''
-
-    # http://www.lib.virginia.edu/artsandmedia/dmmc/Music/Humdrum/kern_hlp.html#kern
+    dur_regex, dotCount, matchedNote, step, accountedFor = _noteMemos(contents)
+
+    if dur_regex:
+        foundNumber, foundRational = dur_regex.groups()
```
mscuthbert, line 2228: all of this takes an already very complex set of routines and makes them even more complex -- now with a set of nested if statements.  Together with the rewriting of the count routines, I can't make heads or tails of what has changed.  Please refactor into a routine that gets the duration from contents and one that gets the pitch from contents.  I have no idea what "accountedFor" means or what a dur_regex is supposed to represent.

mscuthbert (changes requested):
Hi  -- sorry but I cannot accept this in the state it is in because I cannot follow the logic that it is supposed to use to speed up.  The refactor adds much more complexity to an already overly complex routine.

PR #1686, opened 2024-02-07
Store "numbered ending" within barline

OmreeG:
When parsing musicxml files which contain numbered endings, it would be nice to be able to access them within the "Measure" object.
I believe one reasonable way to do this would be to store the numbered ending in the Barline
object, be it a "Repeat" or not. Then it is easily accessed by the "numberedEnding" field.

### DIFF
--- a/music21/bar.py
+++ b/music21/bar.py
@@ -129,6 +129,8 @@ class Barline(base.Music21Object):
     classSortOrder = -5
 
     equalityAttributes = ('type', 'pause', 'location')
+    
+    numberedEnding = None
 
     def __init__(self,
                  type=None,  # pylint: disable=redefined-builtin
@@ -156,6 +158,9 @@ class Barline(base.Music21Object):
 
     def _setType(self, value):
         self._type = standardizeBarType(value)
+
+    def setNumberedEnding(self, value):
+        self.numberedEnding = value
 
     type = property(_getType, _setType,
         doc='''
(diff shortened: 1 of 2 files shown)


music21/bar.py
```diff
@@ -157,6 +159,9 @@ def _getType(self):
     def _setType(self, value):
         self._type = standardizeBarType(value)
 
+    def setNumberedEnding(self, value):
+        self.numberedEnding = value
```
mscuthbert, line 163: In Python we don't generally use setters for things that can be set directly.

music21/bar.py
```diff
@@ -129,6 +129,8 @@ class Barline(base.Music21Object):
     classSortOrder = -5
 
     equalityAttributes = ('type', 'pause', 'location')
+    
+    numberedEnding = None
```
mscuthbert, line 133: All new attributes need to be typed.  `numberedEnding: int | None = None`

music21/musicxml/xmlToM21.py
```diff
@@ -4825,6 +4825,7 @@ def xmlBarline(self, mxBarline):
         # barline objects also store ending objects, that mark begin
         # and end of repeat bracket designations
         mxEndingObj = mxBarline.find('ending')
+        barline.setNumberedEnding(mxEndingObj)
```
mscuthbert, line 4828: What type of variable is this storing?  Is it a str? int?

mscuthbert (changes requested):
Hi OmreeG -- thanks for the first contribution to the project.  I have some concerns about the code, but more about the general concept here.  What does this solve that isn't better solved by calling `.getContextByClass(spanner.RepeatBracket).number` on the measure object?  This way if the RepeatBracket's number is changed then the information propagates to all other sources.

Since it's a niche use case that hasn't come up for discussion before and there is an easy way to solve it, I'm inclined to reject the PR  (But so grateful for the contribution!)

mscuthbert:
Closing -- nearly a year w/o responses -- repeatBracket offers similar functionality.  Feel free to reopen if disagreement.

PR #1690, opened 2024-02-27
Fix #1689: Make `_dissonanceScore` independent of octaves

TimFelixBeyer:
Fixes #1689 where the spelling of a pitch was dependent on its octave.
This was a result of the `_dissonanceScore` computing Pythagorean ratios between the keyContext (which contributes a Pitch without Octave), and the given pitches. Now, the ratios are always computed between Pitches without Octaves to ensure correct scoring.

Adds a regression test and code simplifications as well.

Also incorporates new pylint recommendations about using `min`/`max` and `yield from`.

PR #1691, opened 2024-02-27
Fix #1382: Simplify ChordBase duration creation

### LINKED ISSUE #1382: Check Chord duration creation
**music21 version**

`8alpha`   

**Problem summary**

Check whether the duration code in ChordBase can be simplified:

```
        super().__init__(**keywords)

        # inherit Duration object from GeneralNote
        # keep it here in case we have no notes
        durationKeyword = None
        if 'duration' in keywords:
            durationKeyword = keywords['duration']

        durationKeyword = self._add_core_or_init(notes, useDuration=durationKeyword)

        if durationKeyword is not None:
            self.duration = durationKeyword
        elif 'type' in keywords or 'quarterLength' in keywords:  # dots dont cut it
            self.duration = Duration(**keywords)
```

it seems like all this information already is handled in Music21Object

TimFelixBeyer:
Fixes #1382. Had a look at the mentioned thread and found that the duration code can indeed be simplified a bit.
Chords basically have two ways of setting their duration.
Either the duration is specified explicitly in the keywords of the constructor (via `duration`, `type`, or `quarterLength`), or it must be inferred from the notes contained in the chord (via `_add_core_or_init`).

### DIFF
--- a/music21/chord/__init__.py
+++ b/music21/chord/__init__.py
@@ -138,25 +138,17 @@ class ChordBase(note.NotRest):
         # if provided.
 
         super().__init__(**keywords)
-
-        # inherit Duration object from GeneralNote
-        # keep it here in case we have no notes
-        durationKeyword = None
-        if 'duration' in keywords:
-            durationKeyword = keywords['duration']
-
-        durationKeyword = self._add_core_or_init(notes, useDuration=durationKeyword)
-
-        if durationKeyword is not None:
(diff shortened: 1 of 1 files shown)


TimFelixBeyer:
Linting issues are addressed in #1690

mscuthbert:
Closing and reopening to get a clean lint.  Approved!  THANKS!

PR #1692, opened 2024-02-29
Fix #1653: Correct scale caching mechanism.

### LINKED ISSUE #1653: Repeated calls of ConcreteScale() and romanNumeral return wrong romanNumeral object
**music21 version**

`9.1.0`   

**Problem summary**

Repeated call of `ConcreteScale()` and `romanNumeral` with different pitches results in wrong RomanNumeral object. I.e. subsequent calls return the same romanNumeral object as the first call.

Single call returns the right object. See below.

**Steps to reproduce**

```python
>>> pitches_lists = [
>>>     ('C', 'D', 'E', 'F', 'G', 'A', 'B'),
>>>     ('C', 'D', 'E-', 'F', 'G', 'A', 'B'),
>>> ]
>>> for pitches in pitches_lists:
>>>     mcs = scale.ConcreteScale('c', pitches=pitches)
>>>     rn = mcs.romanNumeral('IV7')
>>>     print([p.unicodeName for p in rn.pitches])
['F', 'A', 'C', 'E']
['F', 'A', 'C', 'E'] # <- here I would expect ['F', 'A', 'C', 'E♭'] (see below)
```

```python
>>> pitches_lists = [
>>>     #('C', 'D', 'E', 'F', 'G', 'A', 'B'),
>>>     ('C', 'D', 'E-', 'F', 'G', 'A', 'B'),
>>> ]
>>> for pitches in pitches_lists:
>>>     mcs = scale.ConcreteScale('c', pitches=pitches)
>>>     rn = mcs.romanNumeral('IV7')
>>>     print([p.unicodeName for p in rn.pitches])
['F', 'A', 'C', 'E♭']
```

TimFelixBeyer:
Fixes #1653: We cannot use the name of a `ConcreteScale` to cache the scales as there are sometimes different scales with the same name but different pitches. Using the pitch list and names now to be more precise and avoid collisions.
Also adds a regression test. (and a small unrelated simplification).

### DIFF
--- a/music21/roman.py
+++ b/music21/roman.py
@@ -3489,10 +3489,11 @@ class RomanNumeral(harmony.Harmony):
             elif 'Scale' in keyClasses:
                 # Need unique 'hash' for each scale. Names is not enough as there are
                 # multiple scales with the same name but different pitches.
+                scaleHash = keyOrScale.name
                 if getattr(keyOrScale, 'isConcrete', False):
-                    scaleHash = "_".join([p.nameWithOctave for p in keyOrScale.pitches])
-                else:
-                    scaleHash = keyOrScale.name
+                    scaleHash += "_".join(
+                        [p.nameWithOctave for p in keyOrScale.pitches]
+                    )
                 if scaleHash in _scaleCache:
                     keyOrScale = _scaleCache[scaleHash]
                 else:
--- a/music21/roman.py
+++ b/music21/roman.py
@@ -3491,7 +3491,7 @@ class RomanNumeral(harmony.Harmony):
                 # multiple scales with the same name but different pitches.
                 scaleHash = keyOrScale.name
                 if getattr(keyOrScale, 'isConcrete', False):
-                    scaleHash += "_".join(
+                    scaleHash += '_'.join(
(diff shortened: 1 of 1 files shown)


music21/roman.py
```diff
@@ -2488,10 +2488,10 @@ def _parseFigure(self):
         if not isinstance(self._figure, str):  # pragma: no cover
             raise RomanNumeralException(f'got a non-string figure: {self._figure!r}')
 
-        if not self.useImpliedScale:
-            useScale = self._scale
-        else:
+        if self.useImpliedScale:
```
mscuthbert, line 2491: perhaps a bit more elegant, but this change is not really in the scope of the PR.  :-)

mscuthbert (commented):
Thanks.  Good catch.  I ran tests here myself, but try to make sure when making changes to a caching system (which is generally there for necessity) that the use of the cache is still tested (generally with timing tests to back it up).  This slows down the creation of roman numerals by 8%, which is acceptable given the bug it fixes.  If it slowed it down by 200% then we would need to make some other fixes (such as id hashing, etc.)

PR #1694, opened 2024-03-03
Fix pylint recommendations

TimFelixBeyer:
Pylint now recommends using `yield from` and the use of `min/max` to simplify if statements.

### DIFF
--- a/music21/analysis/reduceChordsOld.py
+++ b/music21/analysis/reduceChordsOld.py
@@ -79,15 +79,14 @@ class ChordReducer:
         4.0
         '''
         from music21 import note
-        if measureObj.isFlat is False:
+        if not measureObj.isFlat:
             mObj = measureObj.flatten().notes.stream()
         else:
             mObj = measureObj.notes.stream()
 
         chordWeights = self.computeMeasureChordWeights(mObj, weightAlgorithm)
 
-        if numChords > len(chordWeights):
-            numChords = len(chordWeights)
+        numChords = min(numChords, len(chordWeights))
 
         maxNChords = sorted(chordWeights, key=chordWeights.get, reverse=True)[:numChords]
         if not maxNChords:
--- a/music21/analysis/reduceChordsOld.py
+++ b/music21/analysis/reduceChordsOld.py
@@ -79,7 +79,7 @@ class ChordReducer:
         4.0
         '''
         from music21 import note
-        if not measureObj.isFlat:
+        if measureObj.isFlat is False:
             mObj = measureObj.flatten().notes.stream()
         else:
             mObj = measureObj.notes.stream()
--- a/music21/braille/text.py
+++ b/music21/braille/text.py
@@ -599,7 +599,10 @@ class BrailleTextLine(prebase.ProtoM21Object):
         '''
         searchLocation = max(self.highestUsedLocation, self.textLocation)
         addSpaceAmount = 1 if addSpace else 0
-        return (searchLocation + len(text) + addSpaceAmount) <= self.lineLength
+        if (searchLocation + len(text) + addSpaceAmount) > self.lineLength:
+            return False
+        else:
+            return True
 
     def canInsert(self, textLocation, text):
         '''
--- a/music21/braille/text.py
+++ b/music21/braille/text.py
@@ -336,8 +336,7 @@ class BrailleText(prebase.ProtoM21Object):
                 if self.allLines[i].isHeading:
                     break
                 lineLength = self.allLines[i].textLocation
-                if lineLength > maxLineLength:
-                    maxLineLength = lineLength
+                maxLineLength = max(maxLineLength, lineLength)
             for j in range(indexStart, indexFinal):
                 brailleTextLine = self.allLines[j]
                 lineStrToCenter = str(brailleTextLine)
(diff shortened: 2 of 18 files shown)


music21/analysis/reduceChordsOld.py
```diff
@@ -79,15 +79,14 @@ def reduceMeasureToNChords(self,
         4.0
         '''
         from music21 import note
-        if measureObj.isFlat is False:
+        if not measureObj.isFlat:
```
jacobtylerwalls, line 82: Please revert -- this is probably fine, but to be sure would need to brush up on whether measureObj is ever None and if so what is appropriate behavior then.

music21/braille/text.py
```diff
@@ -600,15 +597,9 @@ def canAppend(self, text, addSpace=True):
...
-            searchLocation = self.highestUsedLocation
-        else:
-            searchLocation = self.textLocation
+        searchLocation = max(self.highestUsedLocation, self.textLocation)
         addSpaceAmount = 1 if addSpace else 0
-        if (searchLocation + len(text) + addSpaceAmount) > self.lineLength:
-            return False
-        else:
-            return True
+        return (searchLocation + len(text) + addSpaceAmount) <= self.lineLength
```
jacobtylerwalls, line 602: Please revert this unrelated change.

music21/figuredBass/realizer.py
```diff
@@ -186,10 +186,7 @@ def addLyricsToBassNote(bassNote, notationString=None):
     n = notation.Notation(notationString)
     if not n.figureStrings:
         return
-    maxLength = 0
-    for fs in n.figureStrings:
-        if len(fs) > maxLength:
-            maxLength = len(fs)
+    maxLength = max([len(fs) for fs in n.figureStrings])
```
jacobtylerwalls, line 189: While here, may as well
```suggestion
    maxLength = max(len(fs) for fs in n.figureStrings)
```

music21/stream/base.py
```diff
@@ -11183,10 +11181,7 @@ def makeVoices(self, *, inPlace=False, fillGaps=True):
         # environLocal.printDebug(['makeVoices(): olDict', olDict])
         # find the max necessary voices by finding the max number
         # of elements in each group; these may not all be necessary
-        maxVoiceCount = 1
-        for group in olDict.values():
-            if len(group) > maxVoiceCount:
-                maxVoiceCount = len(group)
+        maxVoiceCount = max([len(group) for group in olDict.values()] + [1])
```
jacobtylerwalls, line 11184: This isn't necessarily better than the prior version, so although it's not worth reverting, it might have been worth adding a disable. To keep in mind for future lint additions.

music21/voiceLeading.py
```diff
@@ -1554,11 +1554,7 @@ def getShortestDuration(self):
         >>> vs1.getShortestDuration()
         1.0
         '''
-        leastQuarterLength = self.objects[0].quarterLength
-        for obj in self.objects:
-            if obj.quarterLength < leastQuarterLength:
-                leastQuarterLength = obj.quarterLength
-        return leastQuarterLength
+        return min([obj.quarterLength for obj in self.objects])
```
jacobtylerwalls, line 1557:
```suggestion
        return min(obj.quarterLength for obj in self.objects)
```

music21/voiceLeading.py
```diff
@@ -1574,11 +1570,7 @@ def getLongestDuration(self):
         >>> vs1.getLongestDuration()
         4.0
         '''
-        longestQuarterLength = self.objects[0].quarterLength
-        for obj in self.objects:
-            if obj.quarterLength > longestQuarterLength:
-                longestQuarterLength = obj.quarterLength
-        return longestQuarterLength
+        return max([obj.quarterLength for obj in self.objects])
```
jacobtylerwalls, line 1573:
```suggestion
        return max(obj.quarterLength for obj in self.objects)
```

music21/figuredBass/realizer.py
TimFelixBeyer: Using a generator is slower than a list comprehension and really has no particular benefits in this case: https://stackoverflow.com/questions/47789/generator-expressions-vs-list-comprehensions.

music21/analysis/reduceChordsOld.py
TimFelixBeyer: It is never `None`, you can check by searching the codebase for any assignment to `isFlat`.
Will revert anyways

music21/figuredBass/realizer.py
jacobtylerwalls: Ah, good point! Honestly I just wanted to defend against the possibility of a cosmetic lint message for it in the future, but I guess it's unlikely there will be one, given the reasoning you link out to. Thanks.

music21/analysis/reduceChordsOld.py
jacobtylerwalls: Oh, for sure, I know I can check, I'm just saying I'd prefer not to in this PR. :-)

music21/stream/base.py
TimFelixBeyer: I agree, I'm pretty sure there should be a simpler option (maybe `max([..], default=1)`, which would be ok if we know at least one group is not empty), but I wasn't able to confirm it quickly.

PR #1699, opened 2024-03-26
Add 'fa up' notehead type

TimFelixBeyer:
This PR adds the 'fa up' notehead type, which is included in the official [MusicXML spec](https://www.w3.org/2021/06/musicxml40/musicxml-reference/data-types/notehead-value/).
While it is rarely used, it should still be included as it currently leads to import errors for MusicXML files that use it.

### DIFF
--- a/music21/chord/__init__.py
+++ b/music21/chord/__init__.py
@@ -4247,6 +4247,7 @@ class Chord(ChordBase):
         'diamond'
         'do'
         'fa'
+        'fa up'
         'inverted triangle'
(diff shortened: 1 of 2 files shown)


mscuthbert:
Thanks Tim.  I have no idea what a "fa up" is.  But easier to support than to make this a case where music21 is less expressive than musicxml for no good reason.

(Not importing Goblins as an instrument from MIDI is one of the cases where we don't do something for good reason...)

PR #1702, opened 2024-04-03
A few ChordSymbol fixes

gregchapman-dev:
(was 'Bbpedal', should be 'B-pedal')

### DIFF
--- a/music21/harmony.py
+++ b/music21/harmony.py
@@ -1516,7 +1516,7 @@ class ChordSymbol(Harmony):
 
     You can also create a Chord Symbol by writing out each degree,
     and any alterations to that degree:
-    
+
     You must explicitly indicate EACH degree (a triad is NOT necessarily implied)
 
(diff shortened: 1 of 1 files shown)


gregchapman-dev:
I did modify one test to actually test that the new code did the right thing.  But it's on the same line as before, which was already tested, so no coverage change.

gregchapman-dev:
Added another fix: CHORD_TYPES['Neapolitan'] had '1,2-,3,5-' but should be '1,-2,3,-5'.

gregchapman-dev:
Third fix: when applying ChordSymbol modifications, if a modification is an 'alter' for a degree not in the ChordSymbol, turn it into an 'add' instead (and apply it).

mscuthbert:
Ah, this was leftover from an early version of ChordSymbol where b was used instead of -.  Just fixed a doc that also referenced the old way.  Will merge when tests pass.

PR #1693, opened 2024-03-01
Small simplifications and speed-ups

TimFelixBeyer:
Just some small tweaks in the spirit of #1631 and #1683 that don't change the functionality.

PR #1706, opened 2024-05-02
Enhance kern grace note importing, fix durations of partially notated chords

alexandermorgan:
This pr is a rework of a [previous pr](https://github.com/cuthbertLab/music21/pull/1684) that I accidentally closed when resyncing my fork, sorry if that's confusing. I've addressed the comments that were made on that previous pr. This enhances kern grace note imports to allow "qq" to mean an unslashed grace note, and for the unlinked duration to take that of the written note if one is provided. This makes the unlinked duration of imported kern grace notes behave the same everywhere else in music21. Also, for kern chords with durations only specified on the first note in the chord ("8C E G"), non-first notes will now reuse that duration instead of defaulting to a quarter note duration. I also added 4 tests to the spineParser.py file to cover all of this new import behavior.

This pr also adds types to a couple functions in spineParser.py. I had a bit of a hard time with the mypy types because my local mypy setup doesn't give the same exact output as the CI test does. Please be aware that I had to use the comment `# type: ignore` 4 times in this file so I hope that's not an issue.

Additionally, this pr separates duration processing into its own function, and decreases the overall import time for kern scores by ~12%. The speed up is primarily due to the following:

1. Move duration parsing of notes to the beginning of hdStringToNote so that notes and rests can be instantiated with their correct duration from the beginning rather than being created and then having their duration replaced. This prevents an unused default duration from being created on each note and rest.
2. Combine the normal duration and rational duration regexes into one. Likewise for note name and accidental searches.
3. Place variables such tat duplicate operations are avoided.
4. Comment out parsing that only leads to pass statements.

The previous version of this pr also included memoization and safe short-circuiting, but those were discarded in favor of simplicity. That earlier pr also replaced `str.count('x')` with `if x in str` where possible but that has already been addressed by [this pr](https://github.com/cuthbertLab/music21/pull/1693).

Looking quickly, this same basic idea of figuring out the duration part before creating a note could apply in tinyNotation, abc, and m21ToXml.py->GeneralObjectExporter.fromDuration/fromScale/fromDiatonicScale
And a similar idea, but figuring out pitch first, might apply in lots of places.


music21/humdrum/spineParser.py
```diff
@@ -1372,12 +1373,14 @@ def processChordEvent(self, eventC):
...
+        defaultDuration = None  # used for non-first notes without a duration
+        for i, noteToProcess in enumerate(notesToProcess):
+            thisNote = hdStringToNote(noteToProcess, defaultDuration)
+            if i == 0:
+                defaultDuration = thisNote.duration
             chordNotes.append(thisNote)
-        eventChord = chord.Chord(chordNotes, beams=chordNotes[-1].beams)
-        eventChord.duration = chordNotes[0].duration
-
+        eventChord = chord.Chord(chordNotes, beams=chordNotes[-1].beams,  # type: ignore
```
mscuthbert, line 1382: Can you explain the `type: ignore` here?

music21/humdrum/spineParser.py
```diff
@@ -2118,7 +2121,85 @@ def getAllOccurring(self):
         return retEvents
 
 
-def hdStringToNote(contents):
+def _hdStringToDuration(contents: str,
+            defaultDuration: duration.Duration|None = None) -> duration.Duration:
```
mscuthbert, line 2125: lint -- inconsistent indentation.

I still don't understand why a routine called XToDuration would take in a Duration object?  This does not look right or make sense to me.

music21/humdrum/spineParser.py
```diff
@@ -2306,93 +2400,18 @@ def hdStringToNote(contents):
...
-            # call Duration.TupletFixer after to correct this.
-
     # 3.2.9 Grace Notes and Groupettos
-    if 'q' in contents:
-        thisObject = thisObject.getGrace()
-        thisObject.duration.type = 'eighth'
+    if (qCount := contents.count('q')):
+        thisObject.getGrace(inPlace=True)
+        if qCount == 2:
+            thisObject.duration.slash = False  # type: ignore
```
mscuthbert, line 2407: why type ignore -- this generally means that there's an error in the code logic.  Please remove these.

mscuthbert: My guess is that a GraceDuration should be created separately and then typed in order to change its slash value.  (It seems like an error on music21's logic in some way -- the slash is really on the GeneralNote not on the duration itself... but that's a larger PR)

music21/humdrum/spineParser.py
```diff
@@ -2201,49 +2282,62 @@ def hdStringToNote(contents):
 
     # http://www.lib.virginia.edu/artsandmedia/dmmc/Music/Humdrum/kern_hlp.html#kern
 
-    # 3.2.1 Pitches and 3.3 Rests
-
-    matchedNote = re.search('([a-gA-G]+)', contents)
-    thisObject = None
+    # Determine duration part first to avoid making an unused duration
+    thisDuration = _hdStringToDuration(contents, defaultDuration)
```
mscuthbert, line 2286: Is this line supposed to be something like:

```
if defaultDuration:
    thisDuration = defaultDuration
else:
    thisDuration = _hdStringToDuration(contents)
```

??? this is the part I'm most confused by.

mscuthbert (commented):
This is a lot closer than the last PR.  Congrats.  But I still have some questions esp. about defaultDuration (and make sure that you're working with an IDE that does indentation that follows the house style).

music21/humdrum/spineParser.py
alexandermorgan: I had to use type ignore where I couldn't reproduce the music21 type-checking locally and kept failing when pushing even though the types were right. It might have been an issue with things being set to None initially and then replaced with a different type according to some logic. I'll give this typing another try.

music21/humdrum/spineParser.py
alexandermorgan: Looking back at my pr description, it looks like I didn't copy over my previous example. The issue is that you occasionally get a humdrum chord in one voice that only notates the duration of the first note, as in:
8C E G
Humdrum at least allows for this and the Humdrum Verovio Viewer would parse these three notes all as eighth notes. Here's a [minimal example](https://verovio.humdrum.org/?t=KiprZXJuCjRnIDRiCjRmIGEKMmUgZwoxZCBmCj09CiotCg==) showing this behavior. music21 currently parses the C as an eighth note but the E and the G are seen as not having a duration so they get the default duration of a duration object, which is a quarter note. A quarter note is a reasonable default duration to have in music21 but not in this circumstance. In this case it is helpful to be able to say "in humdrum chords, use the first note's duration as the default duration for the other notes in the chord if they don't specify one themselves". This is just like what music21 does for the chord object itself. The chord object not only has the same duration as the first note, it uses the same literal duration object. So this pr extends that same behavior to the other notes in the chord if they don't specify their own duration. So 8C E G would be a chord of three eighth notes but 8C 2E 2G would be a chord of one eighth note and two half notes. I encountered this issue in a Joplin rag kern file, but I don't know how common it is in general. Tests are also added to check this new behavior.

music21/humdrum/spineParser.py
alexandermorgan: I think I tried to avoid creating a new grace note separately because that may have created an extra or even two extra duration objects. This seemed like a great application of the `inPlace` parameter. I'll double-check this typing issue as well.

music21/humdrum/spineParser.py
alexandermorgan: The idea is that the `defaultDuration` argument you pass will be `None` most of the time, but in the case of chords, you pass the duration object of the first note in the chord. Then, you check the second note's duration part of its string. If the second note has a duration part to its string, then you use that no matter what it is. It's only if the second note doesn't specify its own duration that we want to use the first note's duration. But using your if/else above, that would use the first note's duration (the `defaultDuration` argument) for all other notes in the chord no matter what. Maybe the problem is just a naming one, does `fallbackDuration` make more sense?

alexandermorgan: Hi @mscuthbert does my explanation above clarify this? Would you be ok with calling it `fallbackDuration`? Passing the first note's duration to all other notes in that chord is key to fixing the bug explained in my "minimal example" given in a comment above. I'm on board with all your other comments, so if we can agree on this then I can update the pr.

mscuthbert:
Hi Alex -- sorry for being AWOL during my other work/life commitments for a while.  I recently had a chance to re-read the whole spineParser code and am back up-to-speed on what is happening here a bit.  If you're willing to try to make a go at this over the summer, I'm game to try it too.

alexandermorgan:
Hi and thanks for digging this pr up! Yeah it would be great to clean this up a bit and get it merged. I saw that you did an overhaul of spine parsing (congrats on v10!), so why don't I resolve those and try to make the changes a bit clearer. I'm a bit busy at the moment, but I can definitely have that by the end of June.

mscuthbert:
That'd be wonderful -- Thanks!

mscuthbert:
Hi Alex -- thanks for the latest pushes -- if you don't mind, I'm going to use AI to pull apart the various PR components, hopefully we can get them in with the smallest number of changes and get this in fast.  Thanks!

music21/humdrum/spineParser.py
```diff
@@ -2947,9 +2964,383 @@ def testCopyAndDeepcopy(self) -> None:
         testCopyAll(self, globals())
 
 
+        hf1 = HumdrumDataCollection(testFiles.mazurka6)
```
mscuthbert: These tests were all moved to humdrum/tests.py

alexandermorgan:
Hey Myke, I'm just seeing now that you already merged a slightly earlier commit. That's fine if that's easier to work with, but for what it's worth the latest commit here fixes the couple of remaining issues there were, namely:
- uses `typing.cast` instead of `# type: ignore`
- reinstates the original meaning of 'Q' in kern syntax
- uses `KernSpine.parse()` to address some other typing issues.
One major benefit here which is independent of the speed-up and added kern syntax support is that the massive function `hdStringToNote` gets a big chunk of it factored out to `_hdStringToDuration`.
But whichever way is easier for you is fine by me, so no worries if you'd rather decline these last couple of commits.

mscuthbert:
Hi @alexandermorgan -- I'm working to break this PR into smaller components that can be individually tested and which do not create large diffs.   #1941 and #1943 take on the first two major parts of this (including leaving the duration component alone except to pass in a defaultDurationOrNone).   I'll put in a third PR with the regex speedups and commenting out TODO parts.  And then if there's anything left I'd be glad to look at the remaining things I missed.  Thanks!

mscuthbert:
Hi @alexandermorgan -- sorry that there are now conflicts -- but could you see if the main original aims of the PR have now been fulfilled via the broken down step PRs (1941, 1943, #1944).  I don't know about `KernSpine.parse()` change but maybe that can be separate?  I think I'd only want to have `hdStringToDuration(contents)` as a separate function if it really was separate and only called if it returned a Duration and didn't take in the defaultDuration object (like have the system already figure out whether it needs to be called or not before calling the function).  Thanks!

alexandermorgan:
I made some comments on the #1944 pr though I see now that it has already been merged.

mscuthbert:
Hi Alex -- I think that everything from this PR has now landed (see #1946).  THANK YOU!   it helped me a ton to break it down step by step, focusing on the ideas rather than individual code and helped me to find a way that any `createDuration(contents)` function did not take a Duration object as an argument, which is what held me back.  I'm going to close this for now, but please feel free to reopen if you think there are code elements missing.

PR #1708, opened 2024-05-06
A few leadsheet-related fixes

gregchapman-dev:
MusicXML import: forward tags with voice should trigger the same behavior as note tags with voice.

I see this quite a bit in leadsheets where these are used to position chordsym changes within a single melody note.  The result should be a second voice containing only hidden rests, but instead, they land in the measure with the notes, and they overlap.

### DIFF
--- a/music21/musicxml/xmlToM21.py
+++ b/music21/musicxml/xmlToM21.py
@@ -6202,7 +6202,7 @@ class MeasureParser(SoundTagMixin, XMLParserBase):
     def updateVoiceInformation(self):
         # noinspection PyShadowingNames
         '''
-        Finds all the "voice" information in <note> tags and updates the set of
+        Finds all the "voice" information in <note> and <forward> tags and updates the set of
         `.voiceIndices` to be a set of all the voice texts, and if there is
         more than one voice in the measure, sets `.useVoices` to True
         and creates a voice for each.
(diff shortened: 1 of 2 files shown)


gregchapman-dev:
Huh.  That test failure has nothing to do with this PR.

gregchapman-dev:
A second fix: a missing opFrac call in OffsetHierarchyFilter that I ran into.

music21/musicxml/xmlToM21.py
```diff
@@ -6231,14 +6231,27 @@ def updateVoiceInformation(self):
...
+        [<music21.stream.Voice 1>, <music21.stream.Voice 2>]
         '''
         mxm = self.mxMeasure
-        for mxn in mxm.findall('note'):
-            voice = mxn.find('voice')
-            if vIndex := strippedText(voice):
-                self.voiceIndices.add(vIndex)
-                # it is a set, so no need to check if already there
-                # additional time < 1 sec per ten million ops.
+        for tagSearch in ('note', 'forward'):
```
mscuthbert, line 6248: Should we also be searching in `<direction>` tags?  They're the only other thing that can contain a voice tag.  But maybe unnecessary.

music21/stream/filters.py
```diff
@@ -490,7 +490,7 @@ def __call__(self, e, iterator=None):
         if not hasattr(iterator, 'iteratorStartOffsetInHierarchy'):
             raise FilterException('Can only run OffsetHierarchyFilter on a RecursiveIterator')
 
-        offset = s.elementOffset(e) + iterator.iteratorStartOffsetInHierarchy
+        offset = opFrac(s.elementOffset(e) + iterator.iteratorStartOffsetInHierarchy)
```
mscuthbert, line 493: Thanks!

mscuthbert (approved):
LGTM   Thanks!

PR #1717, opened 2024-06-14
Update FiguredBass.Examples numbers

mscuthbert:
this module is not run with every test run.  Presumably some bugfix along the way slightly lowered the number of solutions when maximum separation is removed.


mscuthbert:
going to do it all in #1716 instead

PR #1741, opened 2024-10-30
Pyproject updates

jacobtylerwalls:
Congrats on the release! Spotted these potential additions while approving a PR from a conda bot to package 9.3 for conda.

Thanks for shout out in the docs! Spotted an "o" in my middle name, but decided to just skip the whole word, since I don't really go by that anywhere but gmail and my url, although I may need to in the future to disambiguate me from the other Jacob Walls on spotify...

### DIFF
--- a/documentation/source/about/about.rst
+++ b/documentation/source/about/about.rst
@@ -91,7 +91,7 @@ to this project.  Their contributions and generosity are greatly appreciated.
    support and for their contributions to the open source music notation projects,
    including the Bach Goldberg Variations and the Handel Arias included.
 
-* `Jacob Tylor Walls`_, contributed greatly to the type-safety, speed, and test coverage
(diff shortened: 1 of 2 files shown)


mscuthbert:
OOOPS!  So sorry -- Something looked wrong about the middle name to me too -- I was like "there's no "a" between t and y" I even looked up in two places that there was no Tayl... -- and somehow my spider sense didn't tell me "look at the second syllable you dolt!"  Good to know that we're not using it anyhow!.

jacobtylerwalls:
no worries!

> Will Conda need a new 9.3.1 release to get the correct packaging?

nah, I added the <2.0.0 for numpy in the [conda PR](https://github.com/conda-forge/music21-feedstock/pull/12), and the rest is only helpful for people browsing/querying pypi

PR #1745, opened 2025-01-23
Astroid 1015 + misc improvements jan

mscuthbert:
upgrade astroid and get better information about overloads -- this was a long-awaited improvement.  Thanks to @jacobtylerwalls for the mypy contributions!

some attempts at fixes of graph for microtonal but no work there.

Information about deprecation of StrEnum in the distant future

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -1,10 +1,11 @@
 -r requirements.txt
 wheel
+astroid>=3.2.0
 pylint>=2.16.0
 flake8<7.0.0
 flake8-quotes>=3.3.2
 hatchling
-mypy>=1.4.0
+mypy>=1.13.0
 coveralls
 scipy
 sphinx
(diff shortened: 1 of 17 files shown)


requirements_dev.txt
```diff
@@ -1,10 +1,11 @@
 -r requirements.txt
 wheel
+astroid>=3.2.0
 pylint>=2.16.0
```
jacobtylerwalls, line 4: Thanks for the shout out. I'm contributing less often to pylint these days, but what I remember from when I did more of it was advising folks not to manage their own astroid dependency. There's a very narrow band of compatibility between pylint and astroid. They're independent projects in name only, a speculative decision made many eons ago.

Long story short, I'd just pin pylint >= 3.2.0 (which requires astroid 3.2) and turn off whatever new messages you find annoying :D

PR #1746, opened 2025-01-31
calrify type of pitch constructor argument

float3:
Line 1926 says that the argument can be a Pitch too so the constructor should reflect that

### DIFF
--- a/music21/analysis/reduceChords.py
+++ b/music21/analysis/reduceChords.py
@@ -15,6 +15,7 @@ Automatically reduce a MeasureStack to a single chord or group of chords.
 from __future__ import annotations
 
 import collections
+from collections.abc import Sequence
 import itertools
 import unittest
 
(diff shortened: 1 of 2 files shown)


mscuthbert:
Hi @float3 (hill) thanks for the PR -- it's nearly perfect for what we should have here given the code

I'm leaning instead though to removing the line -- it comes form a type of coding paradigm in 2014 that I don't think we would do today (from before Typed python and automatic code generation).  I think that we're better off getting rid of it once I see if it's used anywhere in music21 -- the whole test suite passes so I think not. 

So I want to thank you for the PR but I think we'll go in the opposite direction and trust the Docs and change the code to match)

There's also a bug in the code for documenting Pitch as acceptable to `__init__`.  Here it will fail:

```
class Pitch2(pitch.Pitch): pass

p = pitch.Pitch('C4')
Pitch2(p)   # will raise error
````

we did not check `isinstance(name, pitchPitch)` instead matched by type(self), so the docs would still be incorrect!

I think for that reason better to get rid of it.  Will make creating pitches a few ms faster which is always appreciated.

Will leave open for a day or two to see if there are any objections to this approach.

float3:
I see, I've adjusted the PR, thanks for your response

mscuthbert (approved):
Approved!

mscuthbert:
Some mypy errors were from updating mypy and unrelated.  Fixed in #1747

mscuthbert:
Thank you -- congrats on your first commit to music21!  :-)  The community will long be grateful!

float3:
> Thank you -- congrats on your first commit to music21!  :-)  The community will long be grateful!

thank you and thank you for music21! hope to contribute again if I find anything else

PR #1748, opened 2025-02-07
Code of Conduct Clarify

mscuthbert:
Using correct pronouns, not choice of pronouns (previous wording wasn't good no matter what viewpoint a person has! :-) )


mscuthbert:
Some mypy errors were from updating mypy and unrelated. Fixed in #1747

Issue #1744, opened 2025-01-22
Transpose(P8) turns E-5 to D#6, e.g. in toSounding(piccoloPart))


gregchapman-dev:
**music21 version**

`9.2.0b2`

**Problem summary**
When I do toSounding(piccoloPart), a transpose(P8) happens, and everything looks good _except_ that E-5 gets transposed to D#6 instead of to E-6.

**Steps to reproduce**
I can reproduce this easily, and I will be working on it myself.  I'm just checking with you (@mscuthbert and @jacobtylerwalls) to see if this is something you already know about, and can point me in the right direction.

gregchapman-dev:
Edited because I had the to/from reversed.

mscuthbert:
So weird!  What is the `.transposition` on the Piccolo Instrument object (is it an interval.Interval P8 or something like interval.ChromaticInterval 12)?  How was the Piccolo instrument created?

The only two file formats whose music21 converters seem to have the possibility of creating incorrect transpositions that I can see are MEI and Noteworthy Composer -- is your file from one of them?

gregchapman-dev:
The transposition that screwed up is definitely P8 (not just 12 semitones). I don't think it's related to an importer (I parsed a MusicXML file and then called toSounding on the resulting score).

Bummer. I was hoping it was some edge case you knew about. Ok, I will just debug it through.

mscuthbert:
I think that the MusicXML file is likely corrupted in the instrument definition (or it's a problem in our XML parsing), because this works fine:

```
>>> s = stream.Stream()
>>> picc = instrument.Piccolo()
>>> picc.transposition
<music21.interval.Interval P8>
>>> s.atSoundingPitch = False
>>> s.insert(0, picc)
>>> n = note.Note('D#5')
>>> s.append(n)
>>> s.toSoundingPitch()[-1].nameWithOctave
'D#6'
```

gregchapman-dev:
Thanks! That helps me know where to look.

mscuthbert:
Any progress on this?

gregchapman-dev:
No, sorry, I've been distracted by other work.

mscuthbert:
no problem -- just wanted to make sure there wasn't something I did holding it up -- I'll be here when it's needed!

gregchapman-dev:
[...]

PR #1752, opened 2025-02-18
Fix for issue #1744 (mis-transpose up by an octave)

gregchapman-dev:


mscuthbert:
Fixes #1744

mscuthbert:
Thanks for the diagnosis and fix!  Yes, this is it -- and I had also forgotten how the MusicXML diatonic tag is written so that 0 is a totally acceptable number.

PR #1757, opened 2025-03-07
Move MIDI init to base.py

mscuthbert:
This PR moves MIDI's `__init__.py` functions into `midi.base` continuing a series of improvements and updates that have happened to stream, meter, etc. where `__init__` does not have core functionality.  (still needed: abcFormat and noteworthy).

The previous functionality was fine in a world without strong typing and when isinstance checking was very slow.  Now it is crucial that sub modules like midi.translate can import things like MidiEvent, which wasn't possible (without lots of hacks) when everything was in `__init__` -- the prior organization was fine for its time but now we must move on.

Slightly backwards incompatible but only if people are doing deep checking of inheritance.  Should still be able to keep it for v9.6


mscuthbert:
I'm adding typing to try to figure out where our O(n^2) lookup for channels is taking place, and I can't do that until I add more typing and I couldn't do that until I had this PR in.

PR #1760, opened 2025-03-19
PedalMark support

gregchapman-dev:


gregchapman-dev:
Oops, this one ain't ready yet.

music21/expressions.py
```diff
@@ -325,6 +325,114 @@ def nextMark(self):
         return RehearsalMark(self.nextContent(), numbering=self.numbering)
 
 
+# ------------------------------------------------------------------------------
+class PedalMark(spanner.Spanner):
+    '''
+        A pedal mark spanner contains a pedaled set of notes.
+        The spanner starts at the moment of down-pedal, and
```
mscuthbert, line 332: Dedent 4 spaces please

music21/expressions.py
```diff
@@ -325,6 +325,114 @@ def nextMark(self):
...
+        ends at the moment of up-pedal.  There can be any
+        number of pedal "bounces" in the middle of the spanner.
+
+        The visual form of the pedal mark can vary.  The following
+        examples use a pedal mark with one "bounce" in the middle.
+
+        Pedal marks can be lines:            |_______^________|
+        Pedal marks can be normal symbolic:  Ped.    * Ped.   *
+        Pedal marks can be altered symbolic: Ped.    Ped.     *
+        Pedal marks can be symbol and line:  Ped.____^________|
```
mscuthbert, line 342: there's a way of marking this in RST so it will appear in monospaced font in the docs, otherwise this will be gibberish on the web.

music21/expressions.py
```diff
@@ -325,6 +325,114 @@ def nextMark(self):
...
+
+        Pedal marks that are symbolic can be abbreviated: e.g. P. instead
+        of Ped., S. instead of Sost.
+
+        Pedal marks that are lines can have non-printed portions
+        (gaps) in them; these are usually started with "simile",
+        but not necessarily.
+    '''
+    validPedalTypes: tuple[str, ...] = ('sustain', 'sostenuto', 'soft', 'silent')
+    validPedalForms: tuple[str, ...] = ('line', 'symbol', 'altsymbol', 'symline')
```
mscuthbert, line 357: Can we make these be Enums? -- I'm slowing trying to migrate all special strings to be enums.  I'm also thinking maybe enums don't need to be all caps -- it's the one thing I hate about them, how many SCREAMING CAPITALS we're getting in the code.  But in any case, can be do SYMBOL_ALT or SymbolAlt instead of altsymbol and SymbolLine instead of symline -- sym seems like a misspelling of simile line or something.

Great!

music21/expressions.py
```diff
@@ -325,6 +325,114 @@ def nextMark(self):
...
+        from music21 import note
+        self.fillElementTypes = [note.GeneralNote]
+
+        # startOffset is completely optional.  Can be set by parsers (such
+        # as MusicXML) that might need to get this offset back from the
+        # spanner before the start note has been added to the spanner.
+        self.startOffset: OffsetQL|None = startOffset
+
+        self.pedalType: str = 'sustain'
+        self.pedalForm: str = 'symbol'
```
mscuthbert, line 375: So these would also change to Enums

music21/expressions.py
```diff
@@ -325,6 +325,114 @@ def nextMark(self):
...
+        **keywords
+    ):
+        super().__init__(*spannedElements, **keywords)
+        from music21 import note
+        self.fillElementTypes = [note.GeneralNote]
+
+        # startOffset is completely optional.  Can be set by parsers (such
+        # as MusicXML) that might need to get this offset back from the
+        # spanner before the start note has been added to the spanner.
+        self.startOffset: OffsetQL|None = startOffset
```
mscuthbert, line 372: I don't think this is necessary -- Spanners are Music21Objects so you can just use the spanner's offset right?

music21/expressions.py
```diff
@@ -325,6 +325,114 @@ def nextMark(self):
...
+
+        self.pedalType: str = 'sustain'
+        self.pedalForm: str = 'symbol'
+        self.abbreviated: bool = False
+        self.placement: str|None = None
+
+    @property
+    def pedalType(self) -> str:
```
mscuthbert, line 380: docs for properties (even just short ones) are nice to have.  But if Enums are used for these then there is no longer a need to check type on setter (users' IDEs should do that for them) so these can just be regular attributes.

music21/expressions.py
```diff
@@ -325,6 +325,114 @@ def nextMark(self):
...
+        self._pedalType = newType
+
+    @property
+    def pedalForm(self) -> str:
+        return self._pedalForm
+
+    @pedalForm.setter
+    def pedalForm(self, newForm: str):
+        if newForm not in PedalMark.validPedalForms:
+            raise ValueError(f'invalid pedalForm "{newForm}"')
```
mscuthbert, line 396: If we keep it like this, use `f'invalid pedalForm {newForm!r}'`  instead of enclosing with quotes, so it's easy to see the difference between None and "None" etc.

music21/expressions.py
```diff
@@ -325,6 +325,114 @@ def nextMark(self):
...
+
+class PedalGapStart(PedalObject):
+    '''
+        This object, when seen in a PedalMark spanner, represents a disappearance of
+        the pedal line, usually with a TextExpression('simile').  The pedaling should
+        continue (similarly to before), but the line is invisible during this gap.
+    '''
+
+
+class PedalGapEnd(PedalObject):
```
mscuthbert, line 429: I like all these! (except the indentation)

mscuthbert (commented):
Great work!  Even though the PR is closed for now, I wanted to endorse the general direction of the PR and give a few suggestions for when you want to reopen it.

PR #1761, opened 2025-03-19
Support enclosure == Enclosure.NO_ENCLOSURE which is different from enclosure == None

gregchapman-dev:


music21/style.py
```diff
@@ -105,7 +107,8 @@ def __init__(self) -> None:
     @property
     def enclosure(self) -> Enclosure|None:
         '''
-        Get or set the enclosure as a style.Enclosure enum or None.
+        Get or set the enclosure as a style.Enclosure enum (or
+        None, which means the enclosure is left unspecified).
 
         Valid names are:
```
mscuthbert, line 113: include literal None in the list below at the end.

At end of docs, mark as a "Changed in v9.7 -- ..."  as in other places in the docs.

Thanks!

music21/musicxml/m21ToXml.py
```diff
@@ -6167,8 +6165,7 @@ def codaToXml(self, coda: repeat.Coda) -> Element:
         >>> MEX.dump(mxCodaText)
         <direction>
           <direction-type>
-            <words default-y="20" enclosure="none"
-                justify="center">Coda</words>
+            <words default-y="20" justify="center">Coda</words>
```
mscuthbert, line 6168: Maybe, as a test that XML `enclosure="none"` output is still working, you could set this object (and just this Coda object) to give an explicit no enclosure?

```
>>> c.enclosure = Enclosure.NO_ENCLOSURE
```

and then see that the old-output is still possible?  I'm not seeing any tests of the musicxml output of NO_ENCLOSURE.

Thanks!

mscuthbert (changes requested):
One small docs change requested and one little test change -- tests look like they're passing so I'll merge today if you can get these in.

music21/musicxml/m21ToXml.py
gregchapman-dev: Done.

music21/style.py
gregchapman-dev: Done. (I also allowed the enclosure setter to take str, since the code actually handles that.)

gregchapman-dev:
Looks like the tests are passing.  Thanks for the quick review!

mscuthbert (approved):
New changes are great!  Merging once we get the Coveralls report (merging whatever the report says, but just want to get the baseline number)

PR #1762, opened 2025-03-21
PedalMark support

gregchapman-dev:
Ready for a (second) review, although I'm not going to call it done until I have finished implementing Humdrum read/write and MEI read/write in converter21.

### DIFF
--- a/music21/expressions.py
+++ b/music21/expressions.py
@@ -326,6 +326,114 @@ class RehearsalMark(Expression):
 
 
 # ------------------------------------------------------------------------------
+class PedalMark(spanner.Spanner):
+    '''
+        A pedal mark spanner contains a pedaled set of notes.
+        The spanner starts at the moment of down-pedal, and
+        ends at the moment of up-pedal.  There can be any
+        number of pedal "bounces" in the middle of the spanner.
+
+        The visual form of the pedal mark can vary.  The following
+        examples use a pedal mark with one "bounce" in the middle.
+
+        Pedal marks can be lines:            |_______^________|
+        Pedal marks can be normal symbolic:  Ped.    * Ped.   *
+        Pedal marks can be altered symbolic: Ped.    Ped.     *
+        Pedal marks can be symbol and line:  Ped.____^________|
+
+        Pedal marks, whether lines, symbols, or a combination, can
+        represent the sustain pedal (Ped.), the sostenuto pedal
+        (Sost.), the soft (una corda) pedal, or (more rarely) the
+        silent (muting) pedal.
+
+        Pedal marks that are symbolic can be abbreviated: e.g. P. instead
+        of Ped., S. instead of Sost.
+
+        Pedal marks that are lines can have non-printed portions
+        (gaps) in them; these are usually started with "simile",
+        but not necessarily.
+    '''
+    validPedalTypes: tuple[str, ...] = ('sustain', 'sostenuto', 'soft', 'silent')
+    validPedalForms: tuple[str, ...] = ('line', 'symbol', 'altsymbol', 'symline')
+
+    def __init__(
+        self,
+        *spannedElements,
+        startOffset: OffsetQL|None = None,
+        **keywords
+    ):
+        super().__init__(*spannedElements, **keywords)
+        from music21 import note
+        self.fillElementTypes = [note.GeneralNote]
+
+        # startOffset is completely optional.  Can be set by parsers (such
+        # as MusicXML) that might need to get this offset from the spanner
+        # before the start note has been added to the spanner.
+        self.startOffset: OffsetQL|None = startOffset
+
+        self.pedalType = 'sustain'
+        self.pedalForm = 'symbol'
+
+        self.abbreviated = False
+
+    @property
+    def pedalType(self) -> str:
+        return self._pedalType
+
+    @pedalType.setter
+    def pedalType(self, newType: str):
+        if newType not in PedalMark.validPedalTypes:
+            raise ValueError(f'invalid pedalType "{newType}"')
+        self._pedalType = newType
+
+    @property
+    def pedalForm(self) -> str:
+        return self._pedalForm
+
+    @pedalForm.setter
+    def pedalForm(self, newForm: str):
+        if newForm not in PedalMark.validPedalForms:
+            raise ValueError(f'invalid pedalForm "{newForm}"')
+        self._pedalForm = newForm
+
+
+class PedalObject(base.Music21Object):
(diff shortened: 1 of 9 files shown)


gregchapman-dev:
Weird that the Amy Beach piece has just one pedal mark, but it's true (I checked).  So I had to increment the expected spanner count by 1 in that test.

gregchapman-dev:
@mscuthbert do you want me to add support for 'half down' pedal markings (I don't have any examples, but MEI docs say they support them).

music21/expressions.py
```diff
@@ -325,6 +325,119 @@ def nextMark(self):
...
+
+class PedalForm(common.StrEnum):
+    Line = 'line'
+    Symbol = 'symbol'
+    SymbolAlt = 'symbolalt'
+    SymbolLine = 'symbolline'
+
+class PedalMark(spanner.Spanner):
+    '''
+        A pedal mark spanner contains a pedaled set of notes.
```
mscuthbert, line 343: Make sure to get the comments from the #1760 PR -- Music21 docs don't indent another 4 spaces.

music21/expressions.py
```diff
@@ -325,6 +325,119 @@ def nextMark(self):
...
+
+    @property
+    def pedalType(self) -> PedalType|None:
+        return self._pedalType
+
+    @pedalType.setter
+    def pedalType(self, newType: PedalType|None):
+        if newType is not None and newType not in PedalType:
+            raise ValueError(f'invalid pedalType {newType!r}')
+        self._pedalType = newType
```
mscuthbert, line 391: Properties are fine here, but just wanted to note that with modern Python typing, we can leave type error detecting to the static checker and not use getters and setters for mere type validation.

music21/expressions.py
```diff
@@ -325,6 +325,119 @@ def nextMark(self):
...
+    ) -> None:
+        super().__init__(*spannedElements, **keywords)
+        from music21 import note
+        self.fillElementTypes = [note.GeneralNote]
+
+        self._pedalType: PedalType|None = None
+        self._pedalForm: PedalForm|None = None
+        self.abbreviated: bool = False
+        self.placement: str|None = None
+        self.startOffset: OffsetQL|None = None  # only needed temporarily by xmlToM21.py
```
mscuthbert, line 381: If this is only needed temporarily, there are two better options:

1. Set an entry in PedalMark.editorial['startOffset'] = ...  (easy)
2. (better) In xmlToM21 create a weakref.WeakKeyDictionary()  (like a Javascript WeakMap) like:  `pedalToStartOffset = weakref.WeakKeyDictionary() ;  pm1 = PedalMark();  pedalToStartOffset[pm1] = 3.0`

There are still some places in the code where a converter is leaving a permanent key on an object, but we're trying to wipe them out.

music21/musicxml/m21ToXml.py
```diff
@@ -3508,10 +3511,20 @@ def getProc(su, innerTarget):
...
+                            mxPedalLine = (
+                                self.makePedalResumeLineXml(thisSpanner)
+                            )
+                            preList.append(mxPedalLine)
+
         return preList, postList
 
     @staticmethod
-    def _spannerStartParameters(spannerClass, sp):
+    def _spannerStartParameters(spannerClass: str, sp: spanner.Spanner) -> dict[str, t.Any]:
```
mscuthbert, line 3527: Thank you!

music21/musicxml/xmlToM21.py
```diff
@@ -4060,7 +4060,13 @@ def xmlOrnamentToExpression(
         self.setPlacement(mxObj, orn)
         return orn
 
-    def xmlDirectionTypeToSpanners(self, mxObj):
+    def xmlDirectionTypeToSpanners(
+        self,
+        mxObj: ET.Element,
+        mxDirection: ET.Element|None = None,
+        staffKey: int|None = None,
+        totalOffset: OffsetQL|None = None
```
mscuthbert, line 4068: You've added three (optional) arguments to the method but they don't appear in the docs.  They should be explained with a demo.

mscuthbert: Btw -- seeing the parallel to the arguments in `setDynamicsDirection` the addition of these arguments is approved just so long as they're documented here (or if they're properly documented in setDynamicsDirection, you can just say that they have the same role as they have there...)

mscuthbert (changes requested):
Looking good -- m21ToXml for pedal marks seems pretty well tested, but xmlToM21 still needs demonstrations that it parses properly.  A few other small changes.

Probably better to make the Humdrum import/export a separate PR after this so it doesn't get too big and out of scope.  An object class and one importer/exporter is usually enough to get the first PR in.

As far as half-pedaling -- I think better to have a future use "pedalProportion: float" from 0 to 1, because I've seen 1/4 pedaling etc. but we can make 0.5 (which is a nice round number in binary) also be a special value for half pedal.  But best to get this in before adding more complexity.

music21/musicxml/xmlToM21.py
gregchapman-dev: I'll document them.

music21/expressions.py

music21/expressions.py
gregchapman-dev: Removed them.

music21/expressions.py

music21/musicxml/xmlToM21.py

gregchapman-dev:
I agree with coveralls; there's whole sections of untested code here.  Back to the testing room.

mscuthbert:
Yeah -0.05% is a big amount for such a large codebase. We aim for positive. At most 1/10th as much untested. Thanks!

mscuthbert:
Yeah -0.05% is a big amount for such a large codebase. We aim for positive. At most 1/10th as much untested. Thanks!

gregchapman-dev:
Better, but not good enough.  More tests on their way.

gregchapman-dev:
-0.01% now.  That's way better than -0.05%, but there's still a section of untested code.  It's in an area (export to MusicXML from PedalBounce with symbols) that wouldn't be exercised when exporting to MusicXML from MusicXML; only MEI describes an up/down as a bounce when there are no lines, just symbols.  The untested code is the export to MusicXML code that would take such a symbolic bounce, and turn it into two <pedal> elements: a stop and a start.  I'll write tests for that tomorrow.  Thanks for your patience!

mscuthbert:
Yes!  We have pedals!  Thanks for the extensive tests also.

PR #1766, opened 2025-04-04
New PedalMarks implementation

gregchapman-dev:
New implementation. Also fixed a bug where positions were being imported but not exported.  Note that I apparently lied when I said that <pedal "*"><pedal "Ped"> was being carefully imported as a bounce.  I have added TODO: comments to mark where that needs to be done.

PR #1768, opened 2025-04-16
More accurate spanner stop/start on musicxml import

gregchapman-dev:
Import of `<direction>` spanners now always generate SpannerAnchors, and take direction@offset into account.  This also fixes some bugs where self.nLast and "whatever the next GeneralNote is" weren't the right things to use.

Export of SpannerAnchor-based spanners produces "extra jumpy" MusicXML files (lots more `<backup>` and `<forward>` elements), which are valid, but unusual, and import of those "extra jumpy" files needed some fixing here as well.


gregchapman-dev:
Time to work on some test coverage.

gregchapman-dev:
@mscuthbert This is ready for review.

mscuthbert (changes requested):
Hi Greg -- sorry to take some time on this.  I'm noticing some changes on Crescendo attachment that I want to discuss a bit more -- returning to conversation.

mscuthbert:
As it should be I'm seeing a number of additional SpannerAnchors in places where they weren't before.  However, some of them seem excessive or incorrect:

Take this from `corpus.parse('verdi')`, which is MusicXML 2.0 originally encoded by MuseScore 1.1:

```
      <note default-x="105.25" default-y="-10.00">
        <pitch>
          <step>D</step>
          <octave>5</octave>
          </pitch>
        <duration>12</duration>
        <voice>1</voice>
        <type>eighth</type>
        <stem>down</stem>
        <staff>1</staff>
        <beam number="1">continue</beam>
        <notations>
          <slur type="stop" number="1"/>
          </notations>
        <lyric number="1">
          <syllabic>begin</syllabic>
          <text>piu</text>
          </lyric>
        </note>
      <direction placement="below">
        <direction-type>
          <wedge type="crescendo"/>
          </direction-type>
        <staff>1</staff>
        </direction>
      <note default-x="157.33" default-y="-10.00">
        <pitch>
          <step>D</step>
```

with corresponding stop:

```
    <measure number="21" width="183.77">
      <note default-x="12.00" default-y="5.00">
        <grace slash="yes"/>
        <pitch>
          <step>G</step>
          <alter>-1</alter>
          <octave>5</octave>
          </pitch>
        <voice>1</voice>
        <type>eighth</type>
        <accidental>flat</accidental>
        <stem>up</stem>
        <staff>1</staff>
        </note>
      <direction placement="below">
        <direction-type>
          <wedge type="stop"/>
          </direction-type>
        <staff>1</staff>
        </direction>
      <note default-x="48.84" default-y="-20.00">
        <pitch>
          <step>B</step>
          <alter>-1</alter>
```

[laDonnaEMobile.musicxml.zip](https://github.com/user-attachments/files/19900192/laDonnaEMobile.musicxml.zip)

Despite two possible staves, this seems like a clear case of the wedge being encoded to go from D to Gb (it probably should've been encoded to go after the next chord, but it wasn't...).   Here's how Music21 currently encodes it and some nearby spanners:

```
<music21.dynamics.Crescendo <music21.note.Note D><music21.note.Note G->>
<music21.spanner.Slur <music21.note.Note F><music21.note.Note F>>
<music21.spanner.Slur <music21.chord.Chord B-4 G5><music21.note.Note F>>
<music21.dynamics.Diminuendo <music21.note.Rest 16th><music21.note.Rest 16th>>
```

Here's how it's encoded after this PR:

```
<music21.dynamics.Crescendo <music21.spanner.SpannerAnchor at 1.0><music21.spanner.SpannerAnchor at 0.0>>
<music21.spanner.Slur <music21.note.Note F><music21.note.Note F>>
<music21.spanner.Slur <music21.chord.Chord B-4 G5><music21.note.Note F>>
<music21.dynamics.Diminuendo <music21.spanner.SpannerAnchor at 0.75><music21.spanner.SpannerAnchor at 1.5>>
```

and that notes are not part of the spanned elements at all, even after calling .fill()

```
>>> cr.getSpannedElements()
[<music21.spanner.SpannerAnchor at 1.0>, <music21.spanner.SpannerAnchor at 0.0>]
>>> cr.fill()
>>> cr.getSpannedElements()
[<music21.spanner.SpannerAnchor at 1.0>, <music21.spanner.SpannerAnchor at 0.0>]
```

Though fill doesn't seem to be working properly even on master, drat...but it never really worked.

My thought was that SpannerAnchors would only be created for places that there wasn't a logical note to attach directions to.  They were supposed to be an object for unusual cases not a standard part of music21 that users would encounter anytime they were working with standard objects like crescendos or diminuendos unless the crescendo did something like end on beat 3 of a whole note.  Was I mistaken?

mscuthbert:
Thanks!

re fill(): Yeah, I don't see an example in the docs of filling between multiple measures so it's possible that hadn't been properly implemented.  sigh.  :-D

gregchapman-dev:
I think that fill() should _require_ a stream argument (and perhaps allow a list of streams, for the case where pedal marks should be filled from both left and right hand staves of a piano part within an orchestral score).  I think I was being too helpful when I allowed None, implying sp.getFirst().activeSite.  That's almost never useful.

gregchapman-dev:
But perhaps fill() changes should be in a different PR.  I'm making good progress on "only use SpannerAnchors if you have to" here.

mscuthbert:
definitely let's do fill() separately, but I'm not going to change it to make giving the stream mandatory -- there are good default options (activeSite, firstElement.activeSite, closestCommonContainer(firstElement, lastElement), etc.) that should be explored first.

music21/musicxml/test_m21ToXml.py
```diff
@@ -788,9 +786,15 @@ def check(s1, s2, classType):
                 # check that the spanners start and stop at exactly the same score offset
                 s1StartOffset = s1sp.getFirst().getOffsetInHierarchy(s1)
                 s2StartOffset = s2sp.getFirst().getOffsetInHierarchy(s2)
+                if s1StartOffset != s2StartOffset:
+                    print('hey')
```
mscuthbert, line 790: Remove test code. :-)

music21/musicxml/xmlToM21.py
```diff
@@ -4067,20 +4102,22 @@ def xmlDirectionTypeToSpanners(
...
+        >>> MP.stream.insert(1.0, n1)
         >>> MP.nLast = n1
 
         >>> len(MP.spannerBundle)
         0
         >>> mxDirectionType = EL('<wedge type="crescendo" number="2"/>')
         >>> retList = MP.xmlDirectionTypeToSpanners(mxDirectionType, 1, 0.0)
         >>> retList
-        [<music21.dynamics.Crescendo <music21.spanner.SpannerAnchor at 0.0>>]
+        [<music21.dynamics.Crescendo>]
```
mscuthbert, line 4114: Thank you -- this is much closer to prior usage.  Approve!

music21/musicxml/xmlToM21.py
```diff
@@ -4221,9 +4271,12 @@ def xmlDirectionTypeToSpanners(
                         sp.endHeight = float(height)
                     sp.lineType = mxObj.get('line-type')
 
-                stop = spanner.SpannerAnchor()
-                self.insertCoreAndRef(totalOffset, staffKey, stop)
-                sp.addSpannedElements(stop)
+                if targetLast is not None and offsetAfterLast == totalOffset:
+                    sp.addSpannedElements(targetLast)
```
mscuthbert, line 4275: thank you!

music21/musicxml/xmlToM21.py
```diff
@@ -4221,9 +4271,12 @@ def xmlDirectionTypeToSpanners(
...
+                else:
+                    stop = spanner.SpannerAnchor()
```
mscuthbert, line 4277: Not for this PR, but should SpannerAnchor have an (optional) `type='start|stop'` so when iterating we can filter on them?

music21/musicxml/xmlToM21.py
```diff
@@ -4194,9 +4241,12 @@ def xmlDirectionTypeToSpanners(
                     sp.startTick = mxObj.get('line-end')
                     sp.lineType = mxObj.get('line-type')  # redundant with setLineStyle()
 
-                start = spanner.SpannerAnchor()
-                self.insertCoreAndRef(totalOffset, staffKey, start)
-                sp.addSpannedElements(start)
+                self.spannerBundle.setPendingFirstSpannedElementAssignment(
+                    sp,
+                    'GeneralNote',
```
mscuthbert, line 4246: we're going to get to this routine in a bit, but if it's possible to pass `note.GeneralNote` rather than the string `'GeneralNote'` it's the direction the library is going.  Sometimes it's not possible because of circular imports though.

music21/spanner.py
```diff
@@ -471,6 +471,46 @@ def addSpannedElements(
...
+        >>> n5 = note.Note('c')
+
+        >>> sl = spanner.Spanner()
+        >>> sl.addSpannedElements(n2, n3)
+        >>> sl.addSpannedElements([n4, n5])
+        >>> sl.insertFirstSpannedElement(n1)
+        >>> sl.getSpannedElementIds() == [id(n) for n in [n1, n2, n3, n4, n5]]
+        True
+        '''
+        from music21 import stream
```
mscuthbert, line 491: stream is only used for typing, so please put at the top of the module:

```
if t.TYPE_CHECKING:
    from music21 import stream
```

There is an overhead to these in-function imports.

music21/spanner.py
```diff
@@ -471,6 +471,46 @@ def addSpannedElements(
...
+        # add firstEl first
+        self.addSpannedElements(firstEl)
+
+        # add all the rest in order
+        self.addSpannedElements(elements)
+
+        # restore all elements' activeSite and offset
+        for el, elOffset, elActiveSite in zip(elements, elOffsets, elActiveSites):
+            el.activeSite = elActiveSite
+            el.offset = elOffset
```
mscuthbert, line 512: please explain why all other elements need to be removed and then re-added?  It seems very inefficient, so I don't think we want to use this very much.

music21/musicxml/xmlToM21.py
mscuthbert: Ah, that's the way it always was, got it.

music21/spanner.py
```diff
@@ -1275,10 +1317,12 @@ def getByClassIdLocalComplete(self, className, idLocal, completeStatus):
         return self.getByClass(className).getByIdLocal(
             idLocal).getByCompleteStatus(completeStatus)
 
-    def setPendingSpannedElementAssignment(
+    def setPendingFirstSpannedElementAssignment(
```
mscuthbert, line 1320: -1 on rename -- it's not the first spanned element that is pending, it is the next one -- generally the last one.

Plus there are docs (including the doc below) that weren't changed to the new name.   It's backwards incompatibility for a very small gain.

music21/spanner.py
```diff
@@ -1275,10 +1317,12 @@ def getByClassIdLocalComplete(self, className, idLocal, completeStatus):
...
         self,
         sp: Spanner,
         className: str,
+        offsetInScore: OffsetQL,
+        staffKey: int
```
mscuthbert, line 1325: these should be optional with sensible defaults to keep backwards compatible.  Trust me, there is at least one library that will break immediately if this is released without a backwards incompatible tag.

music21/spanner.py
```diff
@@ -771,10 +811,12 @@ def getLast(self):
 
 
 # ------------------------------------------------------------------------------
-class _SpannerRef(t.TypedDict):
+class PendingAssignmentRef(t.TypedDict):
```
mscuthbert, line 814: Since this has been moved from a private to a public class it needs documentation.

mscuthbert: And see below on why offsetInScore should be optional, and staffKey might be not needed.

music21/spanner.py
mscuthbert: The two new attributes are not discussed in the docs -- as a reader I would have no idea what setting them to anything but 0. and 0 are.  Give defaults of 0. and 0 and then add documentation of what it means to have another number besides them.  Also noticing that there are no tests on the new attributes.

What happens if I don't care about the next offset in score for the pending attribute?  Note that spanners exist in other formats than musicxml, so this change potentially breaks those parsers (most of which aren't tested nearly as well as musicxml).  So this is a big -1 for me unless it is optional.

Where is staffKey used?  It appears to be set and then never read.

mscuthbert (changes requested):
Hi Greg -- lots of good changes, but it's also frustrating for me to see backwards incompatible changes introduced late into the PR review -- these routines are used by others (outside music21) and also in places other than musicxml parsing; old code can't break for this improvement, and new optional tags need to be documented and tested.  Thanks!

music21/spanner.py
gregchapman-dev: It was the only way I could think of to insert at the first element.  Do you have a better way?

music21/spanner.py
gregchapman-dev: In xmlToM21.py's usage it is always the first element of a spanner that is pending.  I was confused about this for a bit, but I finally figured out that in MusicXML, direction start comes before the first note (pending until you parse it), and direction stop comes after the last note (self.nLast).  And within music21, xmlToM21.py is the only client.  I'm happy to restore the old API for other clients, but then I will need this new API for xmlToM21.py, since I have an example in the test suite (directions31a) where the last element lands in the spanner before the (pending) first element does, so you end up with an out-of-order first/last element.  It's a rare-ish case (caused by lots of offsets), but perfectly valid, and I needed to handle it.  It's the crescendo in measure 14 of directions31a.  You can watch it happen in freePendingFirstSpannedElementAssignment: the last element (a quarter Rest) is already there when freePending fills in the first element.

music21/spanner.py
gregchapman-dev: Does the MNX parser you're working on use these "pending" spanner APIs?  No other parser in music21 does.  

The non-use of staffKey is a bug (I just ran into it).  I _should_ be using staffKey (so that PartStaff splits put the spanner in the correct staff), and I am not.

I'm happy to default offsetInScore and staffKey to 0. and 0, respectively.

music21/spanner.py
gregchapman-dev: ok

music21/musicxml/xmlToM21.py
gregchapman-dev: Maybe?  Seems useful, but I worry about it triggering a "person with two watches doesn't know what time it is" problem, if a SpannerAnchor is marked as 'start', but it's the last element in the spanner.

music21/musicxml/test_m21ToXml.py
gregchapman-dev: <snort> That's embarrassing.

music21/spanner.py
gregchapman-dev: Oh, actually, maybe they should both default to None, so the caller can say "I don't care", and freePending...() can ignore offsetInScore when looking for a match.

gregchapman-dev:
@mscuthbert Well, I was going to work on implementing your review comments today, but I ran into a gnarly MusicXML import bug that this PR introduced where (apparently) the ending SpannerAnchor sometimes ends up at the correct offset in the wrong measure.  So I need to track that down first.  Yikes.

gregchapman-dev:
Well, that was fun.  I _think_ this might even be a pre-existing bug.  ActorPreludeSample.musicxml is an example of a score that has a wedge start with no matching wedge stop in the part (in the bass drum part "P16", in measure 35, there is a wedge start, and that is the only wedge element in that part).  This causes almost all the wedges in the string section (actually in all the parts below the bass drum) to be way too long, and overlap each other.  All better now.  

I have a bug or two more to fix before I get to the structural changes you've requested.

gregchapman-dev:
Ugh.  The test failures are obviously due to the fact that I forgot about ArpeggioMarkSpanners, which span across parts.  Sigh.

mscuthbert:
Staff connectors are also spanners that range across parts. Slurs etc can also in music21 (also in MusicXML if you can spare the "number" -- but only 6 in the whole piece)

mscuthbert:
But don't bother with the slurs.

gregchapman-dev:
StaffGroups?  They are created, filled, and marked complete in MusicXMLImporter.partGroups, so we'll never see them uncompleted.  Should I add a comment about this in the "uncompleted" removal loop?  Or should I actually check for them, just in case partGroups() changes in the future?

gregchapman-dev:
Found/fixed a longstanding bug when parsing MusicXML files that music21 wrote ("jumpy" files because of exporting SpannerAnchors).

gregchapman-dev:
Hmmm... some interesting regressions (I didn't catch them because I ran the tests on the wrong PR; sigh...)

music21/spanner.py
gregchapman-dev: I just tried simply doing self.spannerStorage.coreInsert(0, firstEl) followed by a coreElementsChanged call, but it still puts it at the end (which is correct, it is at offset 0, but I wanted it to be first).  @mscuthbert Is there some even more internal call I can make to insert at the start of spannerStorage.elements?

music21/spanner.py
gregchapman-dev: Since I have now put the original setPendingSpannedElementAssignment/freePendingSpannedElementAssignment APIs back in place _without_ offsetInScore and staffKey, my new APIs (setPendingFirstSpannedElementAssignment and freePendingFirstSpannedElementAssignment) still require both offsetInScore and staffKey, and are always called with both from xmlToM21.py.

music21/musicxml/test_m21ToXml.py
gregchapman-dev: Done.

music21/spanner.py
gregchapman-dev: Done.

music21/spanner.py
gregchapman-dev: This is still pending.

music21/spanner.py
gregchapman-dev: I have put the original APIs back (for external clients), and left my new ones in place (for xmlToM21.py).

music21/spanner.py
gregchapman-dev: Still need to document this.

gregchapman-dev:
I think it's at least a bit faster; assignment of elements does a subset of reinsertion, but it's still O(n).

gregchapman-dev:
@mscuthbert OK, I think I've addressed everything.  Please review. (and thanks for your patience)

mscuthbert (changes requested):
See last comment.

mscuthbert:
I think that this is heading in the wrong direction now.  The complexity of having two similar but slightly different code paths is not something worth maintaining for only a small gain.  

I suggest going back to the original spanner code and figuring out how to add these features in a backwards compatible way that does not add a ton of complexity.  I cannot maintain the code going forward with this level of complexity for what I am sure has a much simpler solution.

gregchapman-dev:
I merged the new behavior into the old API and removed the new API.  I made the assumption that inserting the pending element as first in the spanner was just a bug fix that the old clients would want as well.  If that's not right, I can add a new optional parameter, addAsFirst: bool = False, if you like.

gregchapman-dev:
Ugh. I had completely forgotten about the partial fix from PR #1636 that I had taken, and that Jacob and I were to work on a better fix together.  I now have _three_ nested PRs in flight.  Sigh...

gregchapman-dev:
Wow, that's a lot of test failures in completely unrelated areas (after a fairly simple change to the MusicXML reader).  What is going on?

gregchapman-dev:
Got it.  Another merge failure; I lost the "remove that last hidden rest from Finale documents" code.  Damn.

gregchapman-dev:
A description of the three related PRs (this is one of them) has been added to PR #1766.

gregchapman-dev:
I'm going to mark this as a draft PR until we get #1777 (xmlForward should only make hidden rests for Finale-written docs) approved and merged into this one.

gregchapman-dev:
@mscuthbert Tests look like they will pass, so this PR is ready to go.  The most controversial change here is, I believe, that I modified makeRests to insert split rests instead of a complex duration rest.  If you're not up for that, I can undo it; it was briefly necessary, but I believe it isn't necessary anymore.

gregchapman-dev:
Sigh... I merged 9.7.1 (master) back into this PR, and tests are failing.  I'll investigate.

mscuthbert:
> Sigh... I merged 9.7.1 (master) back into this PR, and tests are failing. I'll investigate.

I'm so sorry -- I don't remember making changes that were supposed to affect this.

gregchapman-dev:
Oh I'm definitely not blaming you! My current theory is a bad merge (user error, or git error). I'll figure it out.

mscuthbert:
Hi -- is this about fixing behavior if `makeNotation=False`?   I am not interested in having any changes that only improve this setting.  Thanks.

gregchapman-dev:
For the record, run_tests (3.12) failure was a "coveralls is not responding" error.  Hopefully that will be fixed well before I turn this back into a non-draft PR.

mscuthbert:
Greg -- I'm so sorry that I've let all this stall for a year -- after PyCon I want to look again.  Thanks!

gregchapman-dev:
I'm actually no longer sure that supporting offset fully for MusicXML elements is a good idea. Sometimes it's really important (e.g. I've seen pedal marks in MusicXML files that will miss a spanner element without it), and sometimes it's not so smart (e.g. I've seen directions in MusicXML files that are just being shifted left for typographical reasons, including to a negative offset within the measure).

What do you think?

mscuthbert:
<s>I made the changes needed for the merge to #1807 -- is this downstream from pedal or upstream?</s>  -- Upstream.

mscuthbert:
> I'm actually no longer sure that supporting offset fully for MusicXML elements is a good idea. Sometimes it's really important (e.g. I've seen pedal marks in MusicXML files that will miss a spanner element without it), and sometimes it's not so smart (e.g. I've seen directions in MusicXML files that are just being shifted left for typographical reasons, including to a negative offset within the measure).
> 
> What do you think?

I think that since Finale 2010ish and definitely in modern software like Dorico (and I think MuseScore), there is now an easy way of distinguishing between moving an element for typographical reasons and doing it for offset/playback reasons, so I think going forward more and more musicxml files will have correct offset elements, so we should respect it.

Nice to have feature later would be to be able to pass converter.parse('x.xml',  ignoreDirectionOffsets=True) so those old scores work out better -- but that's definitely a different PR/downstream.

Let me know if you're on board with the new staffKey handling or if we need to think of something else.

Oh, and xmlToM21.py line 8 addition in my recent push is long-long overdue.

Issue #356, opened 2019-01-07
Load lyrics from MIDI


earboxer:
When parsing a midi file, the lyrics aren't loaded.

Example midi file with lyrics: https://hymnary.org/media/fetch/99027/hymnary/audio/PsH/PsH_328.mid

```python
>>> from music21 import *
>>> m = converter.parse("~/Downloads/PsH_328.mid")
>>> text.assembleLyrics(m)
''
```

(MuseScore, however, is able to load the lyrics from this midi file, and similar ones)

mscuthbert:
Lyrics are not currently supported from MIDI files -- it's not on the project roadmap, but happy to take a pull request and review.

PR #1769, opened 2025-04-17
Support loading lyrics from midi

oxygen-dioxide:
Midi files can contain lyrics. Many music creation softwares that deal with lyrics, especially sheet music editor and singing voice synthesizers, such as Musescore and OpenUtau, can create midi file with lyrics. After this PR, we can load lyrics from midi files.

#356 will be solved.


mscuthbert:
hi @oxygen-dioxide -- can you add tests including doctests for `getLyricsFromEvents`.  Thanks.

oxygen-dioxide:
Added lyrics import tests. The song for testing is a small part of Haydn's String Quartet with Chinese and Korean lyrics.
Different softwares use different encoding when writing midi file, and some of them use your system encoding. Music21's parser uses UTF-8 encoding by default. To specify encoding, use `converter.parse(fp, encoding_type = encoding)`. Tests for UTF-8, GBK and EUC-KR encodings are added.

mscuthbert:
Almost there -- can you fix the errors in flake8 run:

```
flake8 music21
```

fixing those should fix the pylint also.  

Because `note` is a module name we don't use it as a variable name (same w/ pitch, interval, etc.) so use `n` or `n1` etc.

oxygen-dioxide:
I also want to implement "`-` to Slur". In singing synthesis softwares, we use `-` to make the current note and the previous note a slur, which means the syllable sings across two notes. So I think we should convert midi lyric `-` to Slurs in music21.
Example:
image
image

mscuthbert:
Closing and Reopening to get a clean slate.

mscuthbert:
Hi @oxygen-dioxide -- Great PR -- there are still a few changes I wanted to make but they were easier for me to do because they mostly touched on parts of the system that were pre-existing.

As far as "-" getting translated to slurs automatically, I'm not going to put that into the music21 MIDI parsing routines unless it becomes a standard (or de-facto standard) across all of MIDI.  I'd be glad to put example code of how to do this after the fact somewhere in the documentation system.  Glad to know that automatic singing systems are also using music21 --I'd very much enjoy talking to you about what you're doing sometime!

oxygen-dioxide:
For the "'-' to slurs", I found that music21 doesn't allow "-" as note lyric. I understand that we don't make it builtin because it isn't a standard, but if a developer wants to convert "-" to slurs in their own python code, they can't, because all the infos about "-" lyrics are cleared when loading the midi file into music21.

PR #1777, opened 2025-05-15
Interpret empty <forward> tags as rests on Finale documents only (derived from PR #1636, thanks @TimFelixBeyer!)

gregchapman-dev:
Note that I had to check md['software'] as it was being constructed, instead of afterward, because Metadata() actually adds a software item for the music21 version that read the file, and I don't want to be confused by that.


gregchapman-dev:
@jacobtylerwalls What do you think?

TimFelixBeyer:
FWIW I don't think this is enough to fully fix the issue. A lot of additional changes in my PR are centered around avoiding unnecessary calls to  `makeRests`. With this version we still add rests that don't exist.

gregchapman-dev:
Yeah, I just realized that.  Trying again.

gregchapman-dev:
(and sorry about jumping all over this, but I just remembered that my 2 nested PRs depend on this, and I need to get it done ASAP; remind me NEVER to have PRs that depend on each other again, this sucks a lot.)

TimFelixBeyer:
While I personally would love this change, when I proposed this version in my PR (https://github.com/cuthbertLab/music21/pull/1636#issuecomment-1694764970) it wasn't accepted (explanation here: https://github.com/cuthbertLab/music21/pull/1636#issuecomment-2143924582) 

I know it's a lot, but I highly encourage you to read through the full thread in #1636 to understand the maintainers' perspective. It's quite a nuanced issue I suppose.

gregchapman-dev:
Yeah, this is my attempt to get what you and I both need, and also make Jacob happy. (Hi @jacobtylerwalls!)

gregchapman-dev:
Hoping we can ignore the slight decrease in coverage.  I promise my next two PRs will increase coverage in MusicXML read/write a lot.

gregchapman-dev:
OK, @jacobtylerwalls, I think this meets everyone's needs.  This is the first of three PRs that need to go in sequentially, the next one is PR #1768 (fixes many Spanner and SpannerAnchor-related bugs) and the last one is PR #1766, a new (better) implementation of PedalMarks that replaces the one that was merged onto master recently. My goal is to get that merged onto master before @mscuthbert makes the next music21 release (otherwise I think we might have to remove the PedalMark support that's on master).  Can we expedite getting this one in, since it is the first of the blockers for that PedalMark work?

mscuthbert:
| (and sorry about jumping all over this, but I just remembered that my 2 nested PRs depend on this, and I need to get it done ASAP; remind me NEVER to have PRs that depend on each other again, this sucks a lot.)

Hi -- this is some very complex and subtle code changes whose discussions etc. are now spread over 4 PRs, and it's getting more than a bit frustrating to me.  I'm not going to review any more code until I have a clear prose explanation for how the proposed solution to the problem given here differs from and is superior to the previous proposed PRs/solutions and, importantly, why it is significantly superior to the status quo.

A superior solution, btw., does not add epicycles within epicycles of added complexity for all parts of the system to improve one thing.  I haven't pushed enough against the complexity in terms of future dev's time and energy of some proposed solutions.  Thanks!

gregchapman-dev:
@jacobtylerwalls I think you should speak up here. I _think_ this is what you were asking for, but perhaps I have missed something.

gregchapman-dev:
And for the record, Myke, what's happening is that as I work on pedal marks, I am finding bugs in other parts of the system that block me (granted, some of them are my bugs, but I'm not adding layers of complexity, I'm fixing long-standing bugs).

gregchapman-dev:
I will work on some prose about these three related PRs, though. You make a good point that I haven't described this for you very well at all.

gregchapman-dev:
A description of the three related PRs (this is one of them) has been added to PR #1766.

jacobtylerwalls:
Taking a look

music21/musicxml/xmlToM21.py
```diff
@@ -769,6 +769,7 @@ def __init__(self):
         self.parts = []
 
         self.musicXmlVersion = defaults.musicxmlVersion
+        self.wasWrittenByFinale = False
```
jacobtylerwalls, line 772: I think it would be better to localize this in xmlForward() along the lines I suggested [here](https://github.com/cuthbertLab/music21/pull/1636#discussion_r1912235012), checking `metadataObj.software`. A helper function could be extracted to do this checking.

I see from your commit message you tried this and found that `setEncoding()` is lossy -- we lose information about whether music21 was _already_ in the software tag on the document -- but I think that shouldn't stop us from doing the right thing; we should just have at most one parser-level variable about _that_ rather than one parser-level variable (and lower-order component variables) about Finale-ness.

Checking in xmlForward should be fast, since we can depend on the metadata object always existing at the the top of the stream (no long O(n) searches finding no metadatas anywhere):

https://github.com/cuthbertLab/music21/blob/52fc784d1f18398a73cd6f1d7a7c969f159af03e/music21/musicxml/xmlToM21.py#L822-L823

You'd have to change that coreInsert (as modeled in that diff I linked).

@mscuthbert do you agree that would achieve better "locality of behavior" than adding more parser attributes for edge cases?

music21/musicxml/xmlToM21.py
```diff
@@ -2575,16 +2584,7 @@ def parse(self):
 
         if self.useVoices is True:
             for v in self.stream.iter().voices:
-                if v:  # do not bother with empty voices
-                    # the musicDataMethods use insertCore, thus the voices need to run
-                    # coreElementsChanged
-                    v.coreElementsChanged()
-                    # Fill mid-measure gaps, and find end of measure gaps by ref to measure stream
-                    # https://github.com/cuthbertlab/music21/issues/444
-                    v.makeRests(refStreamOrTimeRange=self.stream,
```
jacobtylerwalls, line 2584: This is the crux of this PR -- I agree this is a useful change, all discussion on #1636 was in agreement that we want to make this change and only account for Finale's abuse of the spec, it just stalled out on code review comments. Nice job keeping this PR small.

music21/musicxml/xmlToM21.py
```diff
@@ -2630,18 +2630,19 @@ def xmlForward(self, mxObj: ET.Element):
...
             change = opFrac(float(durationText) / self.divisions)
 
-            # Create hidden rest (in other words, a spacer)
-            # old Finale documents close incomplete final measures with <forward>
-            # this will be removed afterward by removeEndForwardRest()
-            r = note.Rest(quarterLength=change)
-            r.style.hideObjectOnPrint = True
-            self.addToStaffReference(mxObj, r)
-            self.insertInMeasureOrVoice(mxObj, r)
+            if self.parent.parent.wasWrittenByFinale:
```
jacobtylerwalls, line 2633: We need a test that the Finale behavior was preserved. There was one on #1636 you could include and give @TimFelixBeyer co-author credit (which we should do in any case!)

music21/musicxml/test_xmlToM21.py
```diff
@@ -1328,12 +1328,13 @@ def testHiddenRests(self):
...
+        self.assertEqual(v2.duration.quarterLength, 3.0)
 
-        restV1 = v1.getElementsByClass(note.Rest)[0]
-        self.assertTrue(restV1.style.hideObjectOnPrint)
-        restsV2 = v2.getElementsByClass(note.Rest)
-        self.assertEqual([r.style.hideObjectOnPrint for r in restsV2], [True, True])
+        restsV1 = list(v1.getElementsByClass(note.Rest))
+        self.assertEqual(restsV1, [])
+        restsV2 = list(v2.getElementsByClass(note.Rest))
+        self.assertEqual(restsV2, [])
```
jacobtylerwalls, line 1337: note for @mscuthbert: this change implicitly tests that hidden rests are no longer created in documents with both Finale and something else (music21) in the software tags.

This is ~okay -- would be better to have a more explicit test for that.

jacobtylerwalls (changes requested):
Thanks, as agreed in #1636 this is an improvement in correctness (limit workaround to Finale only) and net simplicity (given fewer makeRests() calls mixed in with parsing).

jacobtylerwalls:
Also, please update the PR title to something that will be more helpful years from now.

music21/musicxml/xmlToM21.py
```diff
@@ -2575,16 +2584,7 @@ def parse(self):
...
-                    # the musicDataMethods use insertCore, thus the voices need to run
-                    # coreElementsChanged
-                    v.coreElementsChanged()
-                    # Fill mid-measure gaps, and find end of measure gaps by ref to measure stream
-                    # https://github.com/cuthbertlab/music21/issues/444
-                    v.makeRests(refStreamOrTimeRange=self.stream,
-                                fillGaps=True,
-                                inPlace=True,
-                                hideRests=True)
+                v.coreElementsChanged()
```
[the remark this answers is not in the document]
gregchapman-dev: ok.

music21/musicxml/xmlToM21.py
gregchapman-dev: Assuming I'm understanding this correctly, I'm not sure much simplicity is gained by replacing self.wasWrittenByFinale with self.music21WasActuallyInTheDocsSoftwareMetadata (or some other better name).  Maybe a rename with the existing semantics?  self.applyFinaleWorkarounds, or something like that, that feels more like it is a legit parser-level variable.

music21/musicxml/xmlToM21.py
gregchapman-dev: Argh.  I actually picked up that test in the next PR up (before I split this PR out from that one).  Sigh.  I'll bring it here.

music21/musicxml/xmlToM21.py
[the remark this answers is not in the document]

music21/musicxml/test_xmlToM21.py
gregchapman-dev: Done.

music21/musicxml/xmlToM21.py
gregchapman-dev: Done.

music21/musicxml/xmlToM21.py
gregchapman-dev: I'm also going to argue that performing the "loop over the software metadata" for every `<forward>` element is going to be significantly more expensive than doing it once per document.

gregchapman-dev:
All done except for the continuing discussion of "locality of behavior" regarding self.wasWrittenByFinale.

jacobtylerwalls (approved):
Thanks! I'll wait a few days to see if @mscuthbert wants to weigh in.

music21/musicxml/xmlToM21.py
mscuthbert: Yep, do it once.  I believe that the order of software tags written out by `music21` is the order we found was most common at the time (17 years ago) for how to layer them.  So we really only need to look if the top layer is Finale.

mscuthbert: `self.applyFinaleWorkarounds` sounds very good.  Or if we want to make it very generic `self.softwareWorkarounds: set[str] = {'finale'}`

mscuthbert:
Huge thanks to @gregchapman-dev and to @jacobtylerwalls and to @TimFelixBeyer  -- I finally get what through all the weeds and difficulties what we're essentially all in agreement we're trying to do.  And based on finally understanding everything, my normal worry-wart objection "what if the next version of Finale fixes this?" is now moot by experience.  Let me do a last readthrough and see what to do.

music21/musicxml/xmlToM21.py
```diff
@@ -2630,18 +2633,19 @@ def xmlForward(self, mxObj: ET.Element):
...
             change = opFrac(float(durationText) / self.divisions)
 
-            # Create hidden rest (in other words, a spacer)
-            # old Finale documents close incomplete final measures with <forward>
-            # this will be removed afterward by removeEndForwardRest()
-            r = note.Rest(quarterLength=change)
-            r.style.hideObjectOnPrint = True
-            self.addToStaffReference(mxObj, r)
-            self.insertInMeasureOrVoice(mxObj, r)
+            if self.parent.parent.wasWrittenByFinale:
```
mscuthbert, line 2636: One of the few times I wish Python were Javascript -- need to check for grandparent existence...

```
if (self.parent
    and self.parent.parent
    and self.parent.parent.wasWrittenbyFinale
):
```

Vote YES on [PEP 505](https://peps.python.org/pep-0505/)

mscuthbert:
Thanks everyone -- I realize now that the details prevented me from seeing the big picture.

In my last push I renamed some things to help me understand what things (like storing the final rest made from `<forward>`) existed just to get through Finale workarounds.

I added a test of the encoding parser (long overdue before this even) and now I think I fully understand how this PR improves the system without making it too much more complex.

Approved and merging.

Please take a minute after this to close issues/PRs that are now fixed -- or to reformulate them (renaming helps a LOT!) so that I can understand their relevance, and let's get some more Piano marks etc. in.

Greg -- seeing on the list that m21 isn't currently working on Python 3.13 for Windows users, I might make a quick 9.7 release ASAP to help them even if PianoMarks aren't in fast.  But I'll do 9.9 quickly afterwards to get that in, promise.

(In case eyebrows go up -- since m21 9 worked with the latest Python w/ numpy 1.26 and now less than a year later no longer works with the latest Python w/ recent numpy, I'm considering it more of a bug-fix release than a non-backwards-compatible release, since our dependencies didn't think about 18-24 month support on their end).

mscuthbert:
Closing #1636 ([Harvard PR](https://www.harvard.edu/about/history/#:~:text=On%20October%2028%2C%201636%2C%20Harvard,400%20books%20to%20the%20School))

PR #1778, opened 2025-05-16
midi.translate improvements

mscuthbert:
Followup from #1769 gave me an opportunity to find places to improve midi translate.  Pinging @oxygen-dioxide since some changes took place since then

* Change `getNotesFromEvents()` from potentially O(n^2) to O(n) while retaining sort order.

* Pass encoding to parsing of instrument names in addition to lyrics (renamed "encoding_type" to "encoding" -- missed in prior review)

* create TimingNoteEvent which is a typed NamedTuple that keeps track of when a MidiEvent NoteOn started, stopped, and what the NoteOn event was.  Replaced the old tuple of tuples.

* Faster gathering of Chords (remove potential O(n^2) search).

* Remove unused special case for single-note streams.

* opFrac many offsets and durations in MIDI for absolutely correct tuplets, etc.

* Misc: Update braille.translate SegmentKey to typing.NamedTuple

* Misc: Search code for places where we were still casting to float for Python 2 division!

* Misc: Add typing to text/Trigram and some layout things

This is part of the non-backwards compatible MIDI.translate guts rewrite discussed on the music21 list.

Coverage % is expected to drop because the amount of code has been reduced a lot.

### DIFF
--- a/music21/_version.py
+++ b/music21/_version.py
@@ -50,7 +50,7 @@ so change it if a bug or new feature creates a problem with using old pickles.
 '''
 from __future__ import annotations
 
-__version__ = '9.6.0b4'
+__version__ = '9.6.0b5'
 
 def get_version_tuple(vv):
     v = vv.split('.')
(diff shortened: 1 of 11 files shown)


mscuthbert:
Coverage didn't drop -- I guess some of the code was not well tested!

PR #1779, opened 2025-05-18
Support writing lyrics to midi

oxygen-dioxide:
Midi files can contain lyrics. After this PR, the midi file exported from music21 will contain lyrics, which will show if you open it with musescore.

I've implemented reading lyrics from midi in #1769

Users can select the encoding which encoding to use when exporting by using `stream.write('midi', fp=filename, encoding=encoding)`. The default encoding is UTF-8.

### DIFF
--- a/music21/midi/translate.py
+++ b/music21/midi/translate.py
@@ -294,6 +294,7 @@ def getEndEvents(
 def music21ObjectToMidiFile(
     music21Object: base.Music21Object,
     *,
+    encoding: str = 'utf-8',
     addStartDelay=False,
 ) -> MidiFile:
     '''
@@ -307,12 +308,14 @@ def music21ObjectToMidiFile(
             music21Object = music21Object.toSoundingPitch()
 
         return streamToMidiFile(t.cast(stream.Stream, music21Object),
-                                addStartDelay=addStartDelay)
+                                addStartDelay=addStartDelay,
+                                encoding=encoding)
     else:
         m21ObjectCopy = copy.deepcopy(music21Object)
         s: stream.Stream = stream.Stream()
         s.insert(0, m21ObjectCopy)
-        return streamToMidiFile(s, addStartDelay=addStartDelay)
+        return streamToMidiFile(s, addStartDelay=addStartDelay,
+                                encoding=encoding)
 
 
 # ------------------------------------------------------------------------------
@@ -463,6 +466,7 @@ def midiEventsToNote(
 def noteToMidiEvents(
     inputM21: note.Note|note.Unpitched,
     *,
+    encoding: str = 'utf-8',
     includeDeltaTime: bool = True,
     channel: int = 1,
 ) -> list[DeltaTime|MidiEvent]:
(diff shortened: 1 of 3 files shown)


mscuthbert:
Now that encoding is mentioned there, can you check if there are other written out encodings in Midi export (such as instrument name) that should also respect this encoding instead of defaulting to utf-8 (I can't remember if we actually write out the instrument name).

Could you also put encoding last on the attributes (before **keywords if any) -- when we add new features we usually add them last, especially if they're less fundamental than say "channel: int = 1"

Needs tests.  You can probably just test by round tripping.  This time can you test that lyrics work also in English.  Thanks!

oxygen-dioxide:
In `ConverterMidi`, `encoding` is implemented as a property because there are `self.subConverter.keywords = keywords` which bypass the `__init__` of `ConverterMidi`. for example, in music21\converter\__init__.py line 565.

music21/midi/translate.py
```diff
@@ -2590,6 +2616,7 @@ def streamHierarchyToMidiTracks(
     *,
     acceptableChannelList=None,
     addStartDelay=False,
+    encoding: str = 'utf-8',
```
mscuthbert, line 2619: If you're not able to fix the mypy errors that adding `encoding: str = 'utf-8'` caused, just change this to `encoding='utf-8'` and the mypy errors will go away.  I'll be able to add typing to that anyhow.

Once the mypy errors are gone, I'm happy to merge -- the tiny drop in coverage is very small.

mscuthbert:
Thanks OO₂ !  Very happy to merge!  _btw -- since you don't have a "real name" recorded here, is there a way you'd like to be acknowledged in the m21 release notes except "oxygen-dioxide"? some people have used m21 contributions for promotions or tenure in the past_

oxygen-dioxide:
> Thanks OO₂ ! Very happy to merge! _btw -- since you don't have a "real name" recorded here, is there a way you'd like to be acknowledged in the m21 release notes except "oxygen-dioxide"? some people have used m21 contributions for promotions or tenure in the past_

You don't have to acknowledge me. I want to keep myself anonymous, at least currently.

oxygen-dioxide:
Also, I'm working on some other stuffs related to music21. They're currently not finished. I contribute to open source projects in my spare time, so they don't get updates frequently.
[music21-zh-docs](https://github.com/oxygen-dioxide/music21-zh-docs): A Chinese translation of music21's documentation (I started making this because machine translations performs badly on musical terms)
[music21-svs-formats](https://github.com/oxygen-dioxide/music21-svs-formats): Parser and dumper for singing voice synthesis formats. This would help us develop musicology tools integrated with singing voice synthesizers.

mscuthbert:
Thanks "Oxy" (may I use that abbreviation or 氧 ? )  I no longer have funding obligations but I have in the past and I may in the future and they require some name attached to every contribution, even if it is a pseudonym, so I'll go with oxygen-dioxide (anon.) and Oxy in short, unless you object

> Also, I'm working on some other stuffs related to music21. They're currently not finished. I contribute to open source projects in my spare time, so they don't get updates frequently. [music21-zh-docs](https://github.com/oxygen-dioxide/music21-zh-docs): A Chinese translation of music21's documentation (I started making this because machine translations performs badly on musical terms) [music21-svs-formats](https://github.com/oxygen-dioxide/music21-svs-formats): Parser and dumper for singing voice synthesis formats. This would help us develop musicology tools integrated with singing voice synthesizers.

Both of these projects make me so very happy to see!  Gives me a good excuse to continue writing the docs.  And (as a basic reader of Chinese characters mostly from Japanese, nice to see 小 next to discussions of TinyNotation!)

PR #1794, opened 2025-06-19
Fix mypy 1.16 issues

mscuthbert:
Improve typing for music21 based on issues uncovered in mypy 1.16.

Fix error in `Interval.noteEnd = None` -- setting noteStart, noteEnd, pitchStart, or pitchEnd to None raised exceptions

### DIFF
--- a/music21/analysis/discrete.py
+++ b/music21/analysis/discrete.py
@@ -153,7 +153,7 @@ class DiscreteAnalysis:
                 post.append(solution)
         return post
 
-    def solutionLegend(self, compress=False):
+    def solutionLegend(self, compress: bool = False) -> list[list[str|list[tuple[int|str, str|None]]]]:
         '''
         A list of pairs showing all discrete results and the assigned color.
         Data should be organized to be passed to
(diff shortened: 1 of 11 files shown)


music21/analysis/discrete.py
```diff
@@ -153,7 +153,9 @@ def getSolutionsUsed(self):
                 post.append(solution)
         return post
 
-    def solutionLegend(self, compress=False):
+    def solutionLegend(self, compress: bool = False) -> list[
+        list[str|list[tuple[int|str, str|None]]]
```
ahankinson, line 157: For complex types like this that you re-use, you can define a type alias. It might make it a bit clearer to understand when the same type is returned from different functions, or annotated on different variables.

https://docs.python.org/3/library/typing.html#typing.TypeAlias

mscuthbert: Thanks Andrew -- we use TypeAliases all over the place -- I'm leaving this in for now so that I can see that this is something to change/fix/standardize in the next not-backwards compatible version.  There's no way if we had had Python typing when we wrote this code that anyone involved with music21 would have ever have let this come in.  Next version will definitely be something cleaner like `list[tuple[str, list[tuple[str, str]]]]` and THAT will be worthy of a typealias.

PR #1796, opened 2025-06-19
Also merge spanners when calling stream.Score.partsToVoices

guang-yng:
## What does this PR do?

When calling `stream.Score.partsToVoices`, only `notesAndRests` are merged into the new part. We can also merge the spanners from those parts into the new one.

### DIFF
--- a/music21/stream/base.py
+++ b/music21/stream/base.py
@@ -14159,6 +14159,8 @@ class Score(Stream):
         165
 
         '''
+        from music21 import spanner
+
         sub: list[Part] = []
         bundle = []
         if isinstance(voiceAllocation, int):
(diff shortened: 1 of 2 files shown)


mscuthbert:
Hi [guang-yng](https://github.com/guang-yng) -- thanks for the contribution -- I'm looking forward to being able to merge this.  Make sure that `spanner` is not used as a variable since it overshadows the name of the module (`sp` is what I usually use); use an editor that follows .editorconfig formats so you don't break music21 format; and most importantly please add tests to show that the spanners are collected.

What about spanners within the part's measures?  Should we be collecting them?  In which case

```
for sp in p[spanner.Spanner]:
```

might be the better loop variable.

mscuthbert:
For now, one PR = one feature. We can do dynamics after you have a successful merge etc.

music21/stream/base.py
```diff
@@ -14238,7 +14238,7 @@ def partsToVoices(self,
                     v = Voice()
                     v.id = pIndex
                     # for now, just take notes, including rests
-                    for e in m.notesAndRests:  # m.getElementsByClass():
+                    for e in m.getElementsByClass(['Note', 'Chord', 'Rest', 'Spanner']):
```
mscuthbert, line 14241: the replacement for `notesAndRests` is "note.GeneralNote" so this should be:  `for e in m.getElementsByClass([note.GeneralNote, spanner.Spanner])`

this was pretty close, but would lose note.Unpitched, percussion.PercussionChord, etc.

otherwise ready to merge!

guang-yng: Bug fixed! Also added the spanner import at the start of this function.

guang-yng:
Hi @mscuthbert, just checking in—would it be okay to merge this PR now?

mscuthbert:
Thanks for the bump!

Issue #1791, opened 2025-06-05
translate.py uses old GUILE syntax


jeremy9959:
`9.7.0`   

**Operating System(s) checked**
Mac OS 15.5

**Problem summary**
The lilypond translation uses old GUILE syntax.  So for example
```
\override VerticalAxisGroup #'remove-first = ##t
```
should be
```
\override VerticalAxisGroup.remove-first=##t
```
There are many places in `translate.py` where this happens. 

**Steps to reproduce**
```
import music21 as m21

music=[m21.note.Note('G4'),m21.note.Note('A4')]
s = m21.stream.Score(music)

s.show('lily')
```

Note also that lilypond `\RemoveEmptyStaffContext` is now `\RemoveAllEmptyStaves`.

**Expected vs. actual behavior**

**More information**

jeremy9959:
A few comments not related to the issue here but worth investigating.
- the lilypond conversion to png doesn't work well on things that are longer than one line, because the png file gets created in pieces.  
- because of this, it's better to use pdf (as in s.show('lily.pdf')) which handles bigger pieces fine.  But for this little example I stick with png.

The colors seem to work.   Here's a sample.

```
from music21 import *

a = note.Note('A4')
a.style.color = 'blue'
b = note.Note('G4')
b.style.color = 'green'
s = stream.Score()
s.append([a,b])
s.show('lily')
```
yields:

![Image](https://github.com/user-attachments/assets/1b2aa89b-1a1d-4e8e-9237-6c520416ab69)

The lilypond errors are gone.
```
$ python demo.py
GNU LilyPond 2.24.4 (running Guile 2.2)
Changing working directory to: `/tmp/music21'
Processing `/tmp/music21/tmp9h6wcb6z.ly'
Parsing...
Interpreting music...
Preprocessing graphical objects...
Calculating line breaks... 
Drawing systems... 
Converting to PNG...
Success: compilation successfully completed
```

What's odd is that the code inserts a scheme function called color into the lilypond, but doesn't actually use it. I wonder if this is a relic of past implementations?  Or it has some other purpose?

PR #1800, opened 2025-06-23
Update of property assignment syntax for Lilypond conversion (issue #1791)

jeremy9959:
The property assignment syntax for lilypond has been updated, so that instead of

```
\override Property #'Attribute = value
```
we have
```
\override Property.Attribute = value
```

This responds to issue #1791

### DIFF
--- a/music21/lily/lilyObjects.py
+++ b/music21/lily/lilyObjects.py
@@ -1482,11 +1482,11 @@ class LyPropertyOperation(LyObject):
 
     >>> lpo = lily.lilyObjects.LyPropertyOperation('override', 'simple', 'x', 'y')
     >>> str(lpo)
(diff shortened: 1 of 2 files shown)


jeremy9959:
The flake errors refer to "" vs '' and most have nothing to do with changes I made, so I hesitate to get involved in that.

mscuthbert:
closing and reopening to get a clear flake sign.

mscuthbert:
Hi -- those flake things were "caused" by you -- I'll do a quick fix.

When you modified the `Rest #'Property` to `Rest.Property` you removed the single-quote within double-quotes.  Therefore there was no longer a reason why we should use double quotes.  Easy fix for me  as penance for letting this slip.

PR #1807, opened 2025-07-26
spanner.py backward compatible changes

gregchapman-dev:
New features in spanner.py: (1) Let pending spanner element assignments have an associated required offset and clientInfo (if not specified, behaves as before). (2) new API insertFirstSpannedElement (3) new API popPendingSpannedElements.  Also a fix that saves/restores spanner element offset/activeSite around operations that clear them.


gregchapman-dev:
This is the second split-out PR from PR #1768

gregchapman-dev:
Argh.  Gotta test those new APIs that no-one is calling yet.  Hang on.

music21/spanner.py
```diff
@@ -1275,44 +1346,60 @@ def setPendingSpannedElementAssignment(
 
         Now set up su1 to get the next note assigned to it.
 
-        >>> sb.setPendingSpannedElementAssignment(su1, 'Note')
+        >>> sb.setPendingSpannedElementAssignment(su1, 'Note', 0.)
```
mscuthbert, line 1349: little favor -- can you make these all 0.0, 1.0, etc. -- just a music21 style thing to always have a digit on both sides of a decimal point -- I think it makes it much clearer that we're working with floats not ints.

music21/_version.py
```diff
@@ -50,7 +50,7 @@
 '''
 from __future__ import annotations
 
-__version__ = '9.7.2a4'
+__version__ = '9.7.2a6'
```
mscuthbert, line 53: little thing since you're changing this a lot.  Call it version '9.8.0' and then '9.8.1.' etc. -- we use even minor digits to signify betas/not-released so we don't need 'alpha' or 'beta' on it.

(Plus this isn't what I call alpha -- which means we're pretty sure this code doesn't work -- this should be beta at lowest, since we're merging to master).

music21/spanner.py
```diff
@@ -752,10 +803,17 @@ def getLast(self):
...
+    '''
+    An object containing information about a pending first spanned element
+    assignment. See setPendingFirstSpannedElementAssignment for documentation
+    and tests.
+    '''
     # noinspection PyTypedDict
     spanner: 'Spanner'
     className: str
+    offsetInScore: OffsetQL|None
+    clientInfo: t.Any|None
```
mscuthbert, line 816: I still have no idea what `clientInfo` is for or why it should be here.  `t.Any` seems like a bad typing.

music21/spanner.py
```diff
@@ -752,10 +803,17 @@ def getLast(self):
...
 # ------------------------------------------------------------------------------
-class _SpannerRef(t.TypedDict):
+class PendingAssignmentRef(t.TypedDict):
+    '''
+    An object containing information about a pending first spanned element
+    assignment. See setPendingFirstSpannedElementAssignment for documentation
+    and tests.
+    '''
     # noinspection PyTypedDict
     spanner: 'Spanner'
```
mscuthbert, line 813: Remnant from before `from __future__ import annotations` can be changed to `Spanner` without quotes

music21/spanner.py
```diff
@@ -752,10 +803,17 @@ def getLast(self):
 
 
 # ------------------------------------------------------------------------------
-class _SpannerRef(t.TypedDict):
+class PendingAssignmentRef(t.TypedDict):
```
mscuthbert, line 806: This is a good rename.

music21/spanner.py
```diff
@@ -1253,16 +1318,22 @@ def setPendingSpannedElementAssignment(
...
-        is looking for an element of class (className) to complete it. Any future
-        element that matches the className which is passed to the SpannerBundle
-        via freePendingSpannedElementAssignment() will get it.
+        A SpannerBundle can be set up so that a particular spanner (sp) is looking
+        for an element of class (className) to be set as first element. Any future
+        future element that matches the className (and offsetInScore, if specified)
+        which is passed to the SpannerBundle via freePendingFirstSpannedElementAssignment()
+        will get it.  clientInfo is not used in the match, but can be used by the client
+        when cleaning up any leftover pending assignments, by creating SpannerAnchors
+        at the appropriate offset.
```
mscuthbert, line 1331: please demo why we would want this.

music21/spanner.py
mscuthbert: Actually, I still need to understand why we had something that worked fine without 0.0 before and now needs a 0.0 to be passed to it.  I must be incredibly dense that I'm still not seeing the contribution here...

mscuthbert: I think you need to set up a scenario where the wrong spanned-element would be fried with the prior style of calling `freePendingSpannedElementAssignment` with a default `offset` and no `clientInfo`.   Right now all I see is extra work that people using spanner.py would have to do with no benefit.

mscuthbert (changes requested):
Hi Greg -- can you talk through your proposed contribution with another music21 user to help get down on paper what exactly the benefit to the library of this contribution is?  I feel like I've tried many times to understand what is happening and I still don't see it, and I need to get to other projects until I do.  Thanks.

music21/_version.py
gregchapman-dev: Ah, I was trying real hard not to burn numbers.  I will do as you say.

music21/spanner.py
gregchapman-dev: In the next PR, xmlToM21.py uses it to stash off the staffKey for each pending assignment (staffKey is used when splitting a Part into multiple PartStaffs).  When it finds pending assignments that were not ever resolved (by calling popPendingSpannedElementAssignments), it creates a SpannerAnchor for each one, using that stashed off staffKey to put the anchor in the correct split-out staff (of a piano score, for example).  I could name/type it as staffKey: int (and in fact I did for a while), but that seems very specific to one client.  "clientInfo: t.Any|None" makes it so that any client can put anything they need to remember about this particular pending assignment there, and then read it back out after calling popPendingSpannedElementAssignments.

music21/spanner.py
gregchapman-dev: ok.

music21/spanner.py
gregchapman-dev: Will do.

music21/spanner.py
gregchapman-dev: This is a classic (a.k.a. real old, like me) software engineering pattern.  We used to call clientInfo "refcon".

music21/spanner.py
gregchapman-dev: Also, there is no extra work for existing clients: these are optional params.  Existing clients will continue to work as before, with no changes.  I need to demonstrate that in the doctests.

gregchapman-dev:
Maybe I should discuss this with @jacobtylerwalls, since he has been in the MusicXML reader a lot.

gregchapman-dev:
I've tried to explain carefully in the doctests exactly what is intended here, in terms of how these APIs have been used in the past (which still works) and how they can be used in the future (and why).

mscuthbert:
make it `staffKey: int|None` -- now it makes sense why we might only want to free up spanned elements on one staff.  Making something too hypothetical doesn't help me get why we want to add this complexity to the system.  

Just please show that the way things were working before will still work -- I don't normally need to know the offset of the pending assignment in order to get it.  But if there are multiple pending I can free just one by offset?  Just show that -- then it's a new feature.

gregchapman-dev:
Ooops, I was editing my comment while you were commenting on my previous comment:  Here's my new one again:

I've tried to explain carefully in the doctests exactly what is intended here, in terms of how these APIs have been used in the past (which still works) and how they can be used in the future (and why).

I restored the old doctests to show that they still pass (although I tweaked the order of calls a bit to match the order of the "old-style" calls that come from xmlToM21.py). Then I added the new doctests, and enhanced them (and their comments) to show how a parser can use the new parameters to deal with the fact that sometimes there is no "next note" that is at the correct offset to start the spanner, so a SpannerAnchor should be used instead.

The next PR will be the one that changes xmlToM21.py to take advantage of the new-style calls, to handle <direction> with <offset> by inserting a SpannerAnchor at the correct offset.

music21/spanner.py
gregchapman-dev: Done.  With much more commentary as well.

music21/spanner.py
gregchapman-dev: Switched this to staffKey: int|None as requested.

music21/_version.py
gregchapman-dev: Done.  Switched to 9.7.3 for this PR.

music21/spanner.py
gregchapman-dev: Floating point numbers all have a digit after the decimal point now.  And I have added scenarios and descriptions in the doctests.

gregchapman-dev:
Oops, forgot to change the version number(s).  All ready now.

gregchapman-dev:
Hopefully the new doctests and description will make the benefit of this new feature clear.  I am adding a feature that xmlToM21.py will use soon to import accurately `<direction>`s that have `<offset>`s.

gregchapman-dev:
Oh wow, changing version from 9.7.2a7 to 9.7.3 lowered coverage by 3 lines, because "a7" was not parsed.  Before I made that change, coverage was actually increased (a very little) by this PR.

mscuthbert:
Closing and reopening for new tests, etc. (after resolving conflicts)

music21/spanner.py
```diff
@@ -1282,36 +1372,104 @@ def setPendingSpannedElementAssignment(
...
+        >>> su1.getSpannedElements()
+        [<music21.note.Note C>]
+
+        >>> n1.getSpannerSites()
+        [<music21.spanner.Slur <music21.note.Note C>>]
+
+        Now set up su1 to get the next note assigned to it.  Stash off the staffKey (1) that
+        should be used for the SpannerAnchor, should a SpannerAnchor be needed.
+
+        >>> sb.setPendingSpannedElementAssignment(su1, 'Note', 0.0, staffKey=1)
```
mscuthbert: eventually all of these should take in an actual Class/type object instead of str.

mscuthbert:
Hi Greg -- I worked my way through the PR again and it is a significant improvement that will greatly enhance getting spanners from MusicXML parsing correct.

I've accepted the `offsetInScore` path, but in the last passes, I've removed `staffKey` completely from `spanner.py`. That offset is used to filter pending assignments but staffKey is not - ended up confusing me and the spanner.py code too much.  But PendingAssignementRef is now a frozen dataclass and is returned from setPendingSpannedElementAssignment which offers an easy path to using staff keys.  For the MusicXML parsing branch I suggest doing this:

```
self.pendingAssignmentRefIdToStaff = {}
...
ref = bundle.setPendingAssignmentRef(spanner, 'Note', offsetInScore)
self.pendingAssignmentRefIdToStaff[id(ref)] = staffKey
```

Let me know if that works -- merging this now, and we can get other changes in a fresh PR.

mscuthbert (approved):
Approving!

PR #1836, opened 2025-10-28
Update pygame code

### LINKED ISSUE #1835: StreamPlayer not working anymore
**music21 version**

`9.9.1`

macOS Sequoia 15.7.1 (Apple Silicon, M2)

---

**Problem summary**  
`music21.midi.realtime.StreamPlayer` fails on macOS ARM (Apple Silicon) with Python 3.14, even when `pygame` is correctly installed.  
It raises `StreamPlayerException('StreamPlayer requires pygame. Install first')` although `import pygame` works fine.

---

**Steps to reproduce**

```python
from music21 import *

b = corpus.parse('bwv66.6')
sp = midi.realtime.StreamPlayer(b)
sp.play()
```

Full traceback

```
  File "test.py", line 5, in <module>
    sp = midi.realtime.StreamPlayer(b)
  File "/music21/midi/realtime.py", line 94, in __init__
    raise StreamPlayerException('StreamPlayer requires pygame.  Install first')
music21.midi.realtime.StreamPlayerException: StreamPlayer requires pygame.  Install first
```

Expected: StreamPlayer should detect pygame and allow real-time playback.
Actual: Raises an exception claiming pygame is missing even though it’s installed. 

**More information**

- Python version: 3.14.0 (via Homebrew but it's the same with normal installer)
- music21 version: 9.9.1
- pygame version: 2.6.1 (pip show pygame confirms installation)
- Architecture: ARM64 (Apple Silicon)

mscuthbert:
midi.realtime did not work with Pygame 2.0 or newer.  Update

Fixes #1835

Breaks compatibility with Pygame 1.8 and earlier, but only in an error situation.

### DIFF
--- a/documentation/source/installing/installAdditional.rst
+++ b/documentation/source/installing/installAdditional.rst
@@ -126,5 +126,4 @@ On the Mac, run::
 
     pip3 install pygame
 
-
-
+Pygame 1.9 or later is required for correct error handling
--- a/music21/midi/realtime.py
+++ b/music21/midi/realtime.py
@@ -6,7 +6,7 @@
 # Authors:      Michael Scott Asato Cuthbert
 #               (from an idea by Joe "Codeswell")
 #
-# Copyright:    Copyright © 2012 Michael Scott Asato Cuthbert
+# Copyright:    Copyright © 2012-25 Michael Scott Asato Cuthbert
 # License:      BSD, see license.txt
 # ------------------------------------------------------------------------------
 '''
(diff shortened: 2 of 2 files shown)


trawn3333:
Was just about to file an issue and fix for this. Would be great to see in a version 9.9.2. Thanks to all involved.

mscuthbert:
I'll try to do a back port. But there should also be a v10 before PyCon next month. I just need to get some more modern integration with MIDI in Jupyter set up before a talk there.

PR #1877, opened 2026-04-27
Fix pyproject.toml

mscuthbert:
I didn't realize that dependency-groups and optional-dependencies were different in `uv` land.  Only `dependency-groups[dev]` gets pulled in by default in uv sync.

Update developerGuidelines.

AI-assisted (but Myke wrote this commit and stand behind it)


mscuthbert:
Apparently, for testing purposes, even libraries should commit a uv.lock and run `uv sync --locked` (equivalent of `npm ci`).

PR #1895, opened 2026-05-16
Stream.measures(): collect MetronomeMark by default

oxygen-dioxide:
Summary
Add 'MetronomeMark' to the default collect tuple in Stream.measures().

Motivation
When extracting a region of measures via measures(), the method collects context elements (Clef, TimeSignature, Instrument, KeySignature, etc.) from the surrounding stream so the excerpt is self-contained and playable. However, MetronomeMark was missing from the default collection list.

This means if a tempo indication (e.g. MetronomeMark(number=80)) appears before the extracted region but not inside it, the excerpt will have no tempo marking, causing playback to fall back to a default tempo (often 120 quarter notes per minute in music21's default, but missing the actual intended tempo). Adding 'MetronomeMark' ensures the most recent tempo marking is preserved in the extracted excerpt, matching the behavior of the other context elements already collected.

### DIFF
--- a/music21/stream/base.py
+++ b/music21/stream/base.py
@@ -13929,7 +13929,13 @@ class Score(Stream):
     # this is Score.measure
     def measure(self,
                 measureNumber,
-                collect=(clef.Clef, meter.TimeSignature, instrument.Instrument, key.KeySignature, tempo.MetronomeMark),
+                collect=(
+                    clef.Clef,
+                    meter.TimeSignature,
+                    instrument.Instrument,
+                    key.KeySignature,
+                    tempo.MetronomeMark
+                ),
                 gatherSpanners=GatherSpanners.ALL,
                 indicesNotNumbers=False):
         '''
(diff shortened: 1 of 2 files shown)


music21/stream/base.py
```diff
@@ -13929,7 +13929,13 @@ def measures(self,
     # this is Score.measure
     def measure(self,
                 measureNumber,
-                collect=(clef.Clef, meter.TimeSignature, instrument.Instrument, key.KeySignature),
+                collect=(
+                    clef.Clef,
+                    meter.TimeSignature,
+                    instrument.Instrument,
+                    key.KeySignature,
+                    tempo.MetronomeMark
```
mscuthbert: comma at end.

mscuthbert: when each has its own line, we do comma dangle.  when in line (as above) we don't

mscuthbert:
i'm +1 on this.  

the PR demos that collect should eventually be typed and listed (as classes not strings) in one place (as defaultCollect) but do that later.  Get the tests fixed in ASAP and it'll make the next release (today)

mscuthbert:
Thanks!

mscuthbert:
ah -- i did complete the release before bed last night -- i'll make a 10.1.1 tomorrow or the next day.  this is good to get in.

PR #1900, opened 2026-05-24
Add missing opFrac() to extendDuration()

jacobtylerwalls:
I had this lying around (and unfortunately no longer remember the context where I needed it), but it seems like we should add `opFrac()` here?


mscuthbert:
Thank you!!

PR #1905, opened 2026-05-31
In MusicXML, a clef with no number is for staff 1, not for all staves in a part

gregchapman-dev:
Looked it up, and this applies also to staff-layout and staff-details.

### DIFF
--- a/music21/musicxml/xmlToM21.py
+++ b/music21/musicxml/xmlToM21.py
@@ -6017,7 +6017,7 @@ class MeasureParser(SoundTagMixin, XMLParserBase):
         >>> MP = musicxml.xmlToM21.MeasureParser()
         >>> MP.handleClef(mxClef)
         >>> MP.lastClefs
-        {0: <music21.clef.TrebleClef>}
+        {0: <music21.clef.TrebleClef>, 1: <music21.clef.TrebleClef>}
 
         >>> mxClefBC = ET.fromstring('<clef number="2"><sign>F</sign><line>4</line></clef>')
         >>> MP.handleClef(mxClefBC)
--- a/music21/musicxml/xmlToM21.py
+++ b/music21/musicxml/xmlToM21.py
@@ -2556,9 +2556,9 @@ class MeasureParser(SoundTagMixin, XMLParserBase):
             try:
                 staffObject = mxObject.find('staff')
                 if staffObject is not None:
-                    t = staffObject.text
-                    if t is not None:
-                        k = t.strip()
+                    staffText = staffObject.text
+                    if staffText is not None:
+                        k = staffText.strip()
                         return int(k)
             except (AttributeError, TypeError, ValueError):
                 pass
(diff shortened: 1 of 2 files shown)


gregchapman-dev:
I thought I had a good example MusicXML file, but I was wrong.  I made this fix locally several months ago, and am having trouble remembering exactly what files prompted it; I do remember realizing that there were quite a few files with this issue (non-numbered clefs in multi-staff parts) out in the wild, and I tested against Finale and Musescore, who both parsed them correctly (putting the non-numbered clef in staff 1 of a multi-staff part), where music21 was incorrectly putting the clef in both staves of a multi-staff part.  I am scanning all my test files to find examples of non-numbered clefs in multi-staff parts.

gregchapman-dev:
Ah, I see.  It is music21's MusicXML writer that is writing non-numbered clefs in multi-staff parts.  I attach two MusicXML files, one (with clefs numbered) that was read by music21, and the other (some clefs not numbered) that was subsequently written by music21.  The written MusicXML file, when read again by music21 (without this PR), ends up with the clef change incorrectly happening in both piano staves, instead of just in the right hand staff.  I attach both MusicXML files, as well as PDFs of both as rendered by music21, with the extra inserted clefs called out (in red, with text notation) by musicdiff.  Note that Finale and Musescore both render the written file correctly, as does music21 if I apply this PR.

[BrahWiMeSample.musicxml.txt](https://github.com/user-attachments/files/28485454/BrahWiMeSample.musicxml.txt)
[BrahWiMeSample_Written.musicxml.txt](https://github.com/user-attachments/files/28485463/BrahWiMeSample_Written.musicxml.txt)
[BrahWiMeSample.pdf](https://github.com/user-attachments/files/28485487/BrahWiMeSample.pdf)
[BrahWiMeSample_Written.pdf](https://github.com/user-attachments/files/28485505/BrahWiMeSample_Written.pdf)

mscuthbert:
Ah!  This is a real misunderstanding by me  Thank you!

music21/musicxml/xmlToM21.py
```diff
@@ -6011,13 +6016,17 @@ def handleClef(self, mxClef):
 
         >>> MP = musicxml.xmlToM21.MeasureParser()
         >>> MP.handleClef(mxClef)
-        >>> MP.lastClefs
-        {0: <music21.clef.TrebleClef>}
+        >>> MP.lastClefs[1]
+        <music21.clef.TrebleClef>
+        >>> MP.lastClefs[0]
+        <music21.clef.TrebleClef>
```
mscuthbert: I think this test was insufficient before and it's insufficient now too -- since we have no tests of what this did before.  I'll make a push onto this.

mscuthbert:
Good idea -- but I think that this will not be sufficient for dealing with rests w/ display-step & display-octave w/o staff numbers for mono-staff instruments -- I worked on this for about 20 minutes and realized it was a bigger refactor -- right now we counted on the fact that clefs w/o staff numbers were assigned to 0 (NO_STAFF_ASSIGNED) and then rests w/o staff numbers could look up the last clef for NO_STAFF_ASSIGNED.  With this change, rests will go anywhere and we don't have sufficient test coverage in our current system to make sure that things still work, so I'll need to add that coverage first. 

Thanks!

mscuthbert:
got it...

PR #1906, opened 2026-06-02
Fix roman.RomanNumeral Key cache colliding across modes

mscuthbert:
Creating a `RomanNumeral` in a non-major/minor mode (like Lydian or Dorian) poisons the cache for the major key (Lydian) or minor (Dorian, Phyrgian, etc.) on the same tonic. 

After creating `RomanNumeral('I', key.Key('C', 'lydian'))`, a subsequent `RomanNumeral('I', 'C')` returns a **C lydian** key instead of C major.

## Cause

`roman.py`'s `_keyCache` was keyed by `Key.tonicPitchNameWithCase`, which only distinguishes major (upper case) from minor (lower case). Every other mode falls back to the upper-case tonic, so `C lydian` was stored under `'C'` and collided with `C major`.

## Fix

Add a `_keyCacheKey()` helper used at both cache-store sites:

- **major/minor** keep the case-only form (`'C'`, `'c'`) so existing fast string lookups still hit.
- **any other mode** is qualified with its mode name (`'C lydian'`) so it can never be returned for a `C major` request.

AI-assisted (Claude)

### DIFF
--- a/music21/roman.py
+++ b/music21/roman.py
@@ -70,6 +70,22 @@ FIGURES_IMPLYING_ROOT: tuple[tuple[int, ...], ...] = (
 )
 
 
+def _keyCacheKey(keyObj: key.Key) -> str:
+    '''
+    Return the string used to cache a Key in `_keyCache`.
+
+    `tonicPitchNameWithCase` only distinguishes major (upper case) from minor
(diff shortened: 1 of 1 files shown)


mscuthbert:
Bug since at least 2015.  Maybe older

PR #1913, opened 2026-06-16
MISC Cleanups

mscuthbert:
Saw some cleanups while reviewing PR #1905 but I decided to make this separate.

"Let Python be Python" -- move string concat in tests to use implicit.

Fix a bug in ABC Parsing of ABCNote vs subclass ABCChord (order of attributes in parse were reversed -- was okay because they were called by keyword).

Standardize ElementTree.fromstring as `EL` not `ET.fromstring` or `El`


mscuthbert:
Misc...not music cleanups.  hehe

PR #1944, opened 2026-06-22
humdrum: remove unneeded Regexes, fix partial-beam direction

mscuthbert:
Third and last PR broken out from #1706 

Reduce the number of regex searches needed to parse a humdrum scope -- Chopin moves from 6000 to 3400.  Simplify the code and make it more elegant.

But barely moves on performance.  hdStringToNote speeds up by 5% 8.59 µs/call to 8.14.   Full parse speeds up by about 1%

Performance/cleanup, with identical output:

- accidentals: count '#'/'-' directly (the count is the alteration) and build a pitch.Accidental, instead of two `re.search` calls;
- rational durations: only run the `\d+%\d+` search when '%' is present;
- pitch: only search when the event is not a rest (walrus in the elif).

Also, from ideas in PR #1706 (Alexander Morgan): comment out the phrase-mark, slur, and arpeggiation blocks that only ran `pass`, and rename the unused beaming loop variable from `i` to `_`.

AI-assisted (Claude).  Read closely by Myke -- co-author Alexander Morgan

### DIFF
--- a/music21/humdrum/spineParser.py
+++ b/music21/humdrum/spineParser.py
@@ -83,6 +83,7 @@ from music21 import key
 from music21 import note
 from music21 import meter
 from music21 import metadata
+from music21 import pitch
 from music21 import roman
 from music21 import prebase
 from music21 import stream
@@ -2353,17 +2354,18 @@ def hdStringToNote(
 
     # 3.2.1 Pitches and 3.3 Rests
 
-    matchedNote = re.search('([a-gA-G]+)', contents)
-    # Detect rests first, because rests can contain manual positioning information,
-    # which is also detected by the `matchedNote` variable above.
-    thisObject: note.GeneralNote
+    # Detect rests first, because rests can contain manual positioning information
+    # (a pitch letter) that the note search would otherwise pick up; only notes
+    # need that search at all.
+    #
     # Build with defaultDurationOrNone (None gives the usual default).  A note
     # that omits its own duration then shares this exact Duration object rather
     # than allocating a copy; one that has its own duration replaces it below.
+    thisObject: note.GeneralNote
     if 'r' in contents:
         thisObject = note.Rest(duration=defaultDurationOrNone)
 
-    elif matchedNote:
+    elif matchedNote := re.search('([a-gA-G]+)', contents):
         kernNoteName = matchedNote.group(1)
         step = kernNoteName[0].lower()
         if step == kernNoteName[0]:  # middle C or higher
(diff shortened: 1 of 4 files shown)


music21/humdrum/spineParser.py
```diff
@@ -2353,17 +2354,18 @@ def hdStringToNote(
...
+    # Detect rests first, because rests can contain manual positioning information
+    # (a pitch letter) that the note search would otherwise pick up; only notes
+    # need that search at all.
+    #
     # Build with defaultDurationOrNone (None gives the usual default).  A note
     # that omits its own duration then shares this exact Duration object rather
     # than allocating a copy; one that has its own duration replaces it below.
+    thisObject: note.GeneralNote
     if 'r' in contents:
         thisObject = note.Rest(duration=defaultDurationOrNone)
```
alexandermorgan, line 2366: Even if you don't want to factor out duration evaluation to its own function, I would still recommend doing it before anything else in this hdStringToNote function. The reason is that, e.g. for the case of rests here on 2366, you're still creating an initial duration that will just get replaced with the actual duration when you calculate it at the end of the function. So that's still 2 duration objects per rest which might account for the speedup being a little less than what I measured (10-12%). But that will increase the diff size quite a bit.

music21/humdrum/tests.py
```diff
@@ -82,6 +82,15 @@ def testSingleNote(self):
         self.assertEqual(b.duration.dots, 0)
         self.assertEqual(b.duration.tuplets[0].durationNormal.dots, 2)
```
alexandermorgan, line 84: I had added these 4 tests and I would recommend keeping them to cover the new functionality. I had to add `KernSpine.parse()` because the new version of the `KernSpine` class got rid of the `__init__` constructor, so some of the class properties were only available if I called `.parse()` first.
If you don't like calling `.parse()` there's probably a different way to run these tests, but I would definitely keep the tests in some form.

```
    def testGraceNote(self):
        ks = KernSpine()
        ks.parse()
        # noinspection SpellCheckingInspection
        a = ks.processNoteEvent('4Cq')
        self.assertEqual(a.duration.type, 'quarter')
        self.assertEqual(a.duration.slash, True)

    def testGraceNote2(self):
        ks = KernSpine()
        ks.parse()
        # noinspection SpellCheckingInspection
        a = ks.processNoteEvent('16Cqq')
        self.assertEqual(a.duration.type, '16th')
        self.assertEqual(a.duration.slash, False)

    def testChord(self):
        ks = KernSpine()
        ks.parse()
        # noinspection SpellCheckingInspection
        c = ks.processChordEvent('8C 8E')
        self.assertEqual(len(c.notes), 2)
        self.assertEqual(c.notes[0].duration, c.duration)

    def testChord2(self):
        # noinspection SpellCheckingInspection
        ks = KernSpine()
        ks.parse()
        c = ks.processChordEvent('8C E')
        self.assertEqual(c.notes[0].duration, c.notes[1].duration)
```

PR #1956, opened 2026-06-25
Add Typing to Features

mscuthbert:
* Features are typed
* Take many X|None attributes and make them X = '' or other falsey values.
* Bring `__init__: self.name = ....` to be class variables.
* Put defaults for dimensions, discrete, etc. and remove inits that mirror the default.
* Fix grammar, capitalization, and punctuation.
* Bug fix: add `.stream()` to values that were returning StreamIterator but which expectedSteams.

### DIFF
--- a/music21/features/base.py
+++ b/music21/features/base.py
@@ -11,7 +11,7 @@
 from __future__ import annotations
 
 from collections import Counter
-from collections.abc import KeysView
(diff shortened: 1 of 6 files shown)


mscuthbert:
b299663, 8e45d1c, and 2954f69 were all from an adversarial review with another agent.  Looking good.

PR #1962, opened 2026-06-30
(AI) Preserve microtones through Interval.transpose (fix accidental corruption / AccidentalException)

binggao1230:
### The bug

Transposing a `Pitch` (or `Note`) that carries a microtone corrupts its spelling, and for larger microtones raises an exception — even when the transposition is a no-op:

```python
from music21 import pitch, interval

p = pitch.Pitch('C4'); p.microtone = 30
p.transpose(interval.Interval('P1'))     # C~4(-20c)   -- expected C4(+30c)

p = pitch.Pitch('G#5'); p.microtone = 50
p.transpose(interval.Interval('A1'))     # AccidentalException: 2.5 is not a supported accidental type
```

Transposing by a **perfect unison** must be the identity, so `C4(+30c)` → `P1` → `C4(+30c)` is unambiguous. The corruption kicks in above 25¢ (25¢ survives, 26¢ flips to a half-sharp) and affects ordinary intervals too — `Note('C4', microtone=30).transpose('P5')` gives `G~4(-20c)` instead of `G4(+30c)`. It's also internally inconsistent: a 25¢ microtone *is* preserved by `P5`, and a 30¢ microtone *is* preserved by intervals that resolve to a sharp/flat — so identical inputs behave differently depending on the interval.

### Root cause

`music21/interval.py`, `Interval._diatonicTransposePitch`:

```python
pitch2.microtone = None
interval2 = Interval(pitch1, pitch2)                       # pitch1 still carries its microtone
halfStepsToFix = self.chromatic.semitones - interval2.chromatic.semitones
...
pitch2.accidental = halfStepsToFix                         # fractional alter -> corrupt accidental
```

`interval2` is measured from the *microtonal* `pitch1`, so the source cents (0.30 semitone) leak into `halfStepsToFix`. That fractional value is then assigned to the accidental, which snaps to the nearest quarter-tone (half-sharp = 0.5) and dumps the −0.20 remainder into a spurious microtone — or, when the alter lands on an unsupported value like 2.5, raises `AccidentalException`.

The method even contains a dead comment, `# centsOrigin = pitch1.microtone.cents  # unused!!`, marking the intended-but-never-wired-up fix.

### The fix

Resurrect `centsOrigin`, peel it off `halfStepsToFix` before the accidental is set, and restore it as a microtone afterward. The whole thing is guarded by `if centsOrigin:` so the common, non-microtonal (performance-critical) path is byte-identical:

```python
centsOrigin = pitch1.microtone.cents
...
halfStepsToFix = self.chromatic.semitones - interval2.chromatic.semitones
if centsOrigin:
    halfStepsToFix = round(halfStepsToFix - centsOrigin / 100.0, 9)
...
if centsOrigin:
    pitch2.microtone = pitch2.microtone.cents + centsOrigin
```

Now `C4(+30c)` → `P1` → `C4(+30c)`, `G#5(+50c)` → `A1` → `G##5(+50c)` (no crash), and a combined quarter-tone-plus-microtone pitch like `C~4(+30c)` round-trips through a unison unchanged.

### Tests

`testTransposeMicrotonePreserved` covers the unison identity, the formerly-crashing `A1` case, a combined accidental+microtone pitch, and a sweep over eight intervals × four microtone magnitudes asserting the cents are preserved, the sounding pitch (`ps`) moves by exactly the interval, and no quarter-tone accidental is introduced on an integer-semitone interval. It fails on `master` and passes with the fix. The existing `interval`, `pitch`, and `note` test/doctest suites stay green; `ruff`, `pylint -j0`, and `mypy` are clean.

---

*Disclosure (per CONTRIBUTING.md): prepared with AI assistance under my direction. I reproduced the bug, wrote and reviewed the fix and tests, and verified it locally — full `interval`/`pitch`/`note` suites, ruff/pylint/mypy, and an ~1,200-case oracle (intervals × microtones × accidentals) with zero microtone/ps failures and zero crashes.*

### DIFF
--- a/music21/interval.py
+++ b/music21/interval.py
@@ -3460,7 +3460,7 @@ class Interval(IntervalBase):
         pitch1 = p
         pitch2 = copy.deepcopy(pitch1)
         oldDiatonicNum = pitch1.diatonicNoteNum
-        # centsOrigin = pitch1.microtone.cents  # unused!!
+        centsOrigin = pitch1.microtone.cents
         distanceToMove = self.diatonic.generic.staffDistance
 
         newDiatonicNumber = oldDiatonicNum + distanceToMove
@@ -3475,8 +3475,14 @@ class Interval(IntervalBase):
 
         # We have the right note name but not the right accidental
         interval2 = Interval(pitch1, pitch2)
-        # halfStepsToFix already has any microtones
+        # interval2 is measured from the microtonal pitch1, so halfStepsToFix
+        # carries the source microtone. Peel it off so it adjusts only the
+        # accidental; it is restored as a microtone once the accidental is set.
+        # Without this, the leaked cents corrupt the spelling (e.g. a +30c pitch
+        # transposed by P1 becomes a half-sharp) or raise AccidentalException.
         halfStepsToFix = self.chromatic.semitones - interval2.chromatic.semitones
+        if centsOrigin:
+            halfStepsToFix = round(halfStepsToFix - centsOrigin / 100.0, 9)
 
         # environLocal.printDebug(['self', self, 'halfStepsToFix', halfStepsToFix,
         #    'centsOrigin', centsOrigin, 'interval2', interval2])
@@ -3526,6 +3532,10 @@ class Interval(IntervalBase):
                 if (oldPitch2Accidental is not None
                         and oldPitch2Accidental.name == 'natural'):
                     pitch2.accidental = oldPitch2Accidental
+
+        # restore the source microtone that was peeled off halfStepsToFix above
+        if centsOrigin:
+            pitch2.microtone = pitch2.microtone.cents + centsOrigin
 
         if useImplicitOctave:
             pitch2.octave = None
--- a/music21/interval.py
+++ b/music21/interval.py
@@ -3460,7 +3460,10 @@ class Interval(IntervalBase):
         pitch1 = p
         pitch2 = copy.deepcopy(pitch1)
         oldDiatonicNum = pitch1.diatonicNoteNum
-        centsOrigin = pitch1.microtone.cents
+        if not pitch1.isTwelveTone():
+            centsOrigin = pitch1.microtone.cents
+        else:
+            centsOrigin = 0.0
         distanceToMove = self.diatonic.generic.staffDistance
 
         newDiatonicNumber = oldDiatonicNum + distanceToMove
@@ -3475,13 +3478,9 @@ class Interval(IntervalBase):
 
         # We have the right note name but not the right accidental
         interval2 = Interval(pitch1, pitch2)
-        # interval2 is measured from the microtonal pitch1, so halfStepsToFix
-        # carries the source microtone. Peel it off so it adjusts only the
-        # accidental; it is restored as a microtone once the accidental is set.
-        # Without this, the leaked cents corrupt the spelling (e.g. a +30c pitch
-        # transposed by P1 becomes a half-sharp) or raise AccidentalException.
         halfStepsToFix = self.chromatic.semitones - interval2.chromatic.semitones
         if centsOrigin:
+            # peel the source microtone off so it adjusts only the accidental
             halfStepsToFix = round(halfStepsToFix - centsOrigin / 100.0, 9)
 
         # environLocal.printDebug(['self', self, 'halfStepsToFix', halfStepsToFix,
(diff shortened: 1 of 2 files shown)


music21/interval.py
```diff
@@ -3460,7 +3460,7 @@ def _diatonicTransposePitch(self,
         pitch1 = p
         pitch2 = copy.deepcopy(pitch1)
         oldDiatonicNum = pitch1.diatonicNoteNum
-        # centsOrigin = pitch1.microtone.cents  # unused!!
+        centsOrigin = pitch1.microtone.cents
```
mscuthbert: this should gate on `if not pitch1.isTwelveTone():` ... `else: centsOrigin = 0.0`

music21/interval.py
```diff
@@ -3475,8 +3475,14 @@ def _diatonicTransposePitch(self,
 
         # We have the right note name but not the right accidental
         interval2 = Interval(pitch1, pitch2)
-        # halfStepsToFix already has any microtones
+        # interval2 is measured from the microtonal pitch1, so halfStepsToFix
```
mscuthbert: pitch1 isn't necessarily microtonal.  This comments should be trimmed to essentials and moved into the `if centsOrigin:` clause -- in general agents write too much about their particular task and leave too much baggage for people interested in other tasks.  Think about 10-15 words

music21/interval.py
```diff
@@ -3527,6 +3533,10 @@ def _diatonicTransposePitch(self,
                         and oldPitch2Accidental.name == 'natural'):
                     pitch2.accidental = oldPitch2Accidental
 
+        # restore the source microtone that was peeled off halfStepsToFix above
```
mscuthbert: this does not restore the source microtone, it restores the microtone's `.cents`.  And again comment inside the if.

music21/interval.py
```diff
@@ -3527,6 +3533,10 @@ def _diatonicTransposePitch(self,
...
+        if centsOrigin:
+            pitch2.microtone = pitch2.microtone.cents + centsOrigin
```
mscuthbert, line 3545: pitch2.microtone.cents = ...

music21/test/test_interval.py
```diff
@@ -350,6 +350,51 @@ def testIntervalMicrotonesB(self):
         i = interval.Interval(note.Note('c4'), note.Note('c~4'))
         self.assertEqual(str(i), '<music21.interval.Interval A1 (-50c)>')
 
+    def testTransposeMicrotonePreserved(self):
```
mscuthbert, line 353: good test

mscuthbert (changes requested):
Great PR -- a few smaller changes requested, with guidance to you and agent for future PRs.

Btw -- version 11 or a future version might deprecate the ~ and \`  representations of half-flats and sharps, so if you want to make future microtonal contributions please use `Accidental('half-flat')` instead of `\``  etc.  Thanks!

music21/interval.py
mscuthbert: I have a patch that should make `isTwelveTone` run faster -- will push in next PR.  `isTwelveTone` guards against creating Microtone objects unnecessarily.

mscuthbert: See #1963

binggao1230:
Thanks, all addressed in a50d769:

- Gated `centsOrigin` on `if not pitch1.isTwelveTone(): ... else: 0.0` so no Microtone is created for the common twelve-tone case.
- Trimmed the `halfStepsToFix` comment to one line and moved it inside `if centsOrigin:`.
- Fixed the restore comment (it restores the cents, not the source microtone) and moved it inside the `if`.

On `pitch2.microtone.cents = ...`: `Microtone.cents` is documented as not settable ("To set the value in cents, simply use that value as a creation argument"), so I kept the `Pitch.microtone` setter, which takes the cents value directly — `pitch2.microtone = pitch2.microtone.cents + centsOrigin`. Happy to change it if you'd prefer a different idiom. Noted #1963 for the faster `isTwelveTone`.

binggao1230:
Full CI surfaced a regression I'd missed (present since the first commit): `testSieveScaleA`/`testScalaScaleA` broke. Scale realization chains `transposePitch` through microtonal intervals, and the microtonal interval already supplies the result's cents — so unconditionally carrying pitch1's microtone double-counted, and feeding a microtonal pitch1 into `Interval(pitch1, pitch2)` produced a fractional chromatic with no specifier (`IntervalException: ... 4.5`).

Fixed in 8c82424:
- Only carry the source microtone through when the applied interval is **not** itself microtonal (`self.chromatic.semitones` is integral). Scale intervals are microtonal, so they're untouched and keep their old behaviour.
- Measure the accidental from a microtone-free copy of pitch1, so the internal `Interval()` never sees a fractional chromatic.

Now green locally: `test_interval` (13), full `scale.test_scale_main` (26), `interval` doctests (111), plus the new microtone test — all pass; ruff + mypy clean. Let me know if you'd prefer the microtonal-interval case to also compose the source microtone (source + interval cents) rather than defer to the interval — that's a larger semantic call and I kept this PR to the integer-interval case that was the original bug.

mscuthbert (approved):
Thanks for the changes! looks good

mscuthbert:
Thanks for the contribution Gaoflow!

PR #1973, opened 2026-07-04
Fix invalid MusicXML for an unpitched grace note in a voice

### LINKED ISSUE #1732: Unpitched grace note in a voice exports a malformed musicxml file
**music21 version**

`9.1.0`

**Problem summary**
If there is an unpitched grace note in a measure with multiple voices, the exported musicxml is malformed, with the error being that the 'voice' element is not defined.

**Steps to reproduce**

```
from music21 import note, duration, stream
grace = note.Unpitched(duration=duration.GraceDuration())
v1 = stream.Voice(grace)
v2 = stream.Voice(note.Note())
stream.Measure([v1, v2]).show("musicxml")
```

The `<voice>` element should be after the `<unpitched>` element, but it comes before.

**More information**
I changed the line
`helpers.insertBeforeElements(mxNote, mxUnpitched, tagList=['duration', 'type'])`
in the `unpitchedToXml` method in the `m21ToXml.py` file to have the element `'voice'` in the variable `tagList` and the output was well-formed. Not sure if it can cause a problem elsewhere.

pablopupo:
Fixes #1732. Following up on the September 2024 comment that you'd be glad to take a fix for this one.

An unpitched grace note in a voice exported `<voice>` ahead of `<unpitched>`, which fails schema validation, since `<unpitched>` has to come right after `<grace>` in the note content model. `unpitchedToXml` retrofits `<unpitched>` into the already-built `<note>` by inserting before the earliest of `<duration>` or `<type>`, but a grace note has no `<duration>`, so anything `noteToXml` had appended in between slipped ahead. The tagList now lists every element that can follow `<unpitched>` within `<note>`, which also fixes the same misplacement for a tied grace note (`<tie>` landed ahead of `<unpitched>` too).

Before, from the issue's repro:

```xml
<note>
  <grace slash="yes" />
  <voice>1</voice>
  <unpitched>
    <display-step>B</display-step>
    <display-octave>4</display-octave>
  </unpitched>
  <type>eighth</type>
</note>
```

After:

```xml
<note>
  <grace slash="yes" />
  <unpitched>
    <display-step>B</display-step>
    <display-octave>4</display-octave>
  </unpitched>
  <voice>1</voice>
  <type>eighth</type>
</note>
```

Added a doctest for the voice case, a regression test in test_m21ToXml.py, and bumped the version to 11.0.0b7 for the writer change. pylint (10.00/10), ruff, mypy, and the musicxml module tests all pass locally.

### DIFF
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -4645,12 +4645,13 @@ class MeasureExporter(XMLExporterBase):
         _setTagTextFromAttribute(up, mxUnpitched, 'display-step')
         _setTagTextFromAttribute(up, mxUnpitched, 'display-octave')
 
-        # anything that can follow <unpitched> within <note> must be listed
-        # here, since a grace note has no <duration> to insert before (#1732)
-        helpers.insertBeforeElements(
+        # within <note>, only <grace>, <cue>, and <chord> may precede
+        # <unpitched>, so insert after those rather than trying to list
+        # everything that can follow it (#1732)
+        helpers.insertAfterElements(
             mxNote,
             mxUnpitched,
-            tagList=['duration', 'tie', 'instrument', 'footnote', 'level', 'voice', 'type']
+            tagList=['grace', 'cue', 'chord']
         )
 
         return mxNote
--- a/music21/musicxml/m21ToXml.py
+++ b/music21/musicxml/m21ToXml.py
@@ -4622,6 +4622,22 @@ class MeasureExporter(XMLExporterBase):
           </unpitched>
           <type>quarter</type>
         </note>
+
+        <unpitched> must come before <voice>, even when there is
+        no <duration>, as on a grace note in a voice:
+
+        >>> MEX.currentVoiceId = 1
+        >>> mxUnpitched = MEX.unpitchedToXml(graceUp)
+        >>> MEX.dump(mxUnpitched)
+        <note>
+          <grace slash="yes" />
+          <unpitched>
+            <display-step>D</display-step>
+            <display-octave>5</display-octave>
+          </unpitched>
+          <voice>1</voice>
+          <type>quarter</type>
+        </note>
         '''
         mxNote = self.noteToXml(up, noteIndexInChord=noteIndexInChord, chordParent=chordParent)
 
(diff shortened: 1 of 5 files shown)


music21/musicxml/m21ToXml.py
```diff
@@ -4622,14 +4622,36 @@ def unpitchedToXml(self,
...
         _setTagTextFromAttribute(up, mxUnpitched, 'display-step')
         _setTagTextFromAttribute(up, mxUnpitched, 'display-octave')
 
-        helpers.insertBeforeElements(mxNote, mxUnpitched, tagList=['duration', 'type'])
+        # anything that can follow <unpitched> within <note> must be listed
+        # here, since a grace note has no <duration> to insert before (#1732)
+        helpers.insertBeforeElements(
+            mxNote,
+            mxUnpitched,
+            tagList=['duration', 'tie', 'instrument', 'footnote', 'level', 'voice', 'type']
```
jacobtylerwalls: Looking at the [spec](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/note/), it seems there are a few more (dot, accidental, lyric, etc.). In practice, this seems not an issue, since grace notes always have `<type>` in music21's export workflow, as far as I can tell, despite being optional in the spec. Still, it seems like maybe insertBeforeElements with a quasi-exhaustive list is not our only option. Did you consider writing an insertAfter helper that inserts directly after the` <grace>`?

jacobtylerwalls (commented):
Thanks for the PR -- One question for your consideration.

music21/musicxml/m21ToXml.py
pablopupo: Good call, that is cleaner than trying to enumerate the successors. Pushed 800cab13, which adds an insertAfterElements helper mirroring insertBeforeElements, and unpitchedToXml now inserts after grace, cue, or chord, the only elements the spec allows before <unpitched> within <note>. Output is unchanged for the existing doctests and the full musicxml suite is still green.

jacobtylerwalls (approved):
Thanks 👍

mscuthbert:
Thanks all. This is great!

PR #1974, opened 2026-07-05
Remove Path -> str -> Path roundtrip in uncompressMXL()

jacobtylerwalls:
Since Python 3.6.2, `ZipFile.extract` takes path-like objects ([docs](https://docs.python.org/3.15/library/zipfile.html#zipfile.ZipFile.extract)).


mscuthbert (approved):
Thanks! Looks good!

PR #1988, opened 2026-07-23
Support sus4 chords in root-finding algorithm

### LINKED ISSUE #1650: Support sus4 chords in root-finding algorithm
**Motivation**
sus4 chords are common enough that we could support them in the root-finding algorithm.

**Feature summary**
```py
>>> from music21 import chord
>>> chord.Chord('F B- C E-').root()  # expected F
<music21.pitch.Pitch C>

**Proposed implementation**
`rootnessFunction()` might possibly prefer adjacent values (e.g. having a fifth and seventh trumps having a third and a seventh), but this would need to be tested.

**Intent**
<!-- Please check one or more boxes -->

[?] I plan on implementing this myself.

Reported on [m21 list](https://groups.google.com/g/music21list/c/Dqp4HyRLJfk/m/Dz73fh_TIgAJ) Sept. 2023

p2rez:
Fixes #1650.

**Note: This contribution was developed with AI assistance (Claude/Anthropic), as declared per the contributing guidelines.**

**Problem**

chord.Chord('F B- C E-').root() incorrectly returned C instead of F. The _findRoot method's rootnessFunction only scored notes based on thirds, fifths, sevenths, etc. above them, but gave no credit for a fourth, so sus4 chord roots were never correctly identified.

**Solution**

Added sus4 chord detection to _findRoot. Before scoring, we check whether any note in the chord has both a third and fifth above it. If none do (indicating a sus4 chord), we apply a bonus to notes that have both a fourth and fifth above them, giving sus4 roots the same base score as notes with a third above them in regular chords.

This avoids incorrectly boosting notes in chords that happen to have an eleventh (like C11 = C E G B- D F), where C has both a third and fifth above it, so the sus4 path is not triggered.

**Tests**

Added testSus4RootFinding in test/test_chord.py covering:

Basic sus4 seventh chord from the issue (F B- C E- → F)
Simple sus4 triad (C F G → C)
sus4 with seventh (C F G B- → C)
Regression: regular major, minor, dominant seventh still correct
Regression: C11 chord still returns C, not F

All 27 tests pass.

### DIFF
--- a/music21/chord/__init__.py
+++ b/music21/chord/__init__.py
@@ -1668,7 +1668,7 @@ class Chord(ChordBase):
         Generally use root() instead, since if a chord doesn't know its root,
         root() will run ._findRoot() automatically.
         '''
-        def rootnessFunction(rootThirdList):
+        def rootnessFunction(rootThirdList, is_sus4_chord=False):
             '''
             Returns a value for how likely this pitch is to be a root given the
             number of thirds and fifths above it.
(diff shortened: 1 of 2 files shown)


mscuthbert:
tests do not cover more difficult cases.  unwieldy 2-part sol (thanks for AI attempt)

PR #2000, opened 2026-08-19
Fix MEI importer silently dropping <bTrem> and shifting subsequent onsets (AI)

### LINKED ISSUE #1994: MEI importer silently drops <bTrem> (bowed tremolo) and its child note/chord, shifting all subsequent onsets
**music21 version:** 9.1.0 (also verified against the current MEI dispatch table)

**Problem summary:** `music21.mei.base` has no handler for `<bTrem>` (absent from the layer dispatch table), and an unhandled element's children are not imported. The result is not a cosmetic loss: the tremolo's note/chord contributes **zero duration**, so **every onset after it in that voice shifts earlier** by the tremolo's duration.

In the piece where I hit this (a Humdrum score rendered to MEI by verovio, which correctly emits `<bTrem unitdur="8">` for a beamed tremolo), one dropped chord shifted the entire remainder of the piece by a quarter note — 169 spurious / 171 missing onsets out of ~1,900 when compared event-wise against the source.

**Minimal repro** (self-contained, 4/4 bar with notes on beats 1, 2, 3):

```python
import music21

mei = '''<?xml version="1.0" encoding="UTF-8"?>
<mei xmlns="http://www.music-encoding.org/ns/mei" meiversion="4.0.1">
 <music><body><mdiv><score>
  <scoreDef meter.count="4" meter.unit="4">
   <staffGrp><staffDef n="1" lines="5" clef.shape="G" clef.line="2"/></staffGrp>
  </scoreDef>
  <section>
   <measure n="1">
    <staff n="1">
     <layer n="1">
      <note xml:id="n1" dur="4" oct="4" pname="c"/>
      <bTrem xml:id="bt1" form="meas" unitdur="8">
       <chord xml:id="ch1" dur="4">
        <note xml:id="n2" oct="4" pname="e"/>
        <note xml:id="n3" oct="5" pname="e"/>
       </chord>
      </bTrem>
      <note xml:id="n4" dur="2" oct="4" pname="g"/>
     </layer>
    </staff>
   </measure>
  </section>
 </score></mdiv></body></music>
</mei>'''

s = music21.converter.parse(mei, format="mei")
for n in s.flatten().notes:
    print(n.offset, [p.nameWithOctave for p in n.pitches], n.duration.quarterLength)
```

**Actual output:**

```
0.0 ['C4'] 1.0
1.0 ['G4'] 2.0
```

The `<bTrem>` chord (E4+E5 on beat 2) is gone, and G4 lands on beat **2** instead of beat **3**.

**Expected:** the chord imported with its notated duration (offset 1.0, quarterLength 1.0) and G4 at offset 2.0 — ideally with a `Tremolo` expression attached, but timing correctness is the important part.

**Suggested minimal fix short of full tremolo support:** treat `<bTrem>` (and `<fTrem>`) transparently in the layer dispatch — import child `<note>`/`<chord>` elements as ordinary notes so durations survive, optionally attaching a `Tremolo`/`TremoloSpanner` expression when `unitdur`/`@measperf` is present. Silent deletion is the one behaviour that corrupts everything downstream of the tremolo.

youdie006:
Fixes #1994

## Problem
`music21.mei.base._processEmbeddedElements` dispatches each child element through a per-container tag->function mapping. A child whose tag is not in the mapping is silently dropped and never recursed into (only a debug message is emitted). `<bTrem>` (measured / bowed tremolo, which wraps the `<note>`/`<chord>` carrying the notated duration) was registered in none of the dispatch dicts, so the tremolo note/chord and its duration disappeared and every later onset in the layer shifted earlier -- data corruption, not a cosmetic loss. As @weselyj reported, one dropped tremolo chord shifted the remainder of a real piece by a quarter note.

## Fix
- Add `bTremFromElement`, mirroring `beamFromElement`/`tupletFromElement`: it imports the wrapped `<note>`/`<chord>` with its notated duration and returns it, so timing is preserved. When `@unitdur` is present it attaches an `expressions.Tremolo` with `numberOfMarks` derived from the unit duration (eighth->1 mark, sixteenth->2, ...).
- Register `<bTrem>` in the layer, beam, and tuplet dispatch tables.
- Scope limited to `<bTrem>`; the two-note `<fTrem>` is intentionally out of scope.
- Docstring "Implemented"/"not Implemented" element lists updated to match.

## Tests
Added focused tests in `music21/mei/test_base.py` using the reporter's snippet: the `<bTrem>` chord now imports on beat 2 with its correct duration, subsequent onsets no longer shift (3 notes at offsets 0.0/1.0/2.0, ql 1/1/2), and the Tremolo expression is attached with the right number of marks. Red-green verified (without the fix the layer imports only 2 notes); the full `music21/mei/test_base.py` module and the module doctests pass with no regressions; ruff, pylint, and mypy are clean.

Thanks to @weselyj for the precise repro and to @mscuthbert for inviting the PR.

## AI disclosure
This contribution was prepared with AI assistance; I take responsibility for its correctness and have verified it against the project's test suite and style checks.

### DIFF
--- a/music21/mei/base.py
+++ b/music21/mei/base.py
@@ -188,6 +188,7 @@ from music21 import clef
 from music21 import duration
 from music21 import environment
 from music21 import exceptions21
+from music21 import expressions
 from music21 import instrument
 from music21 import interval
 from music21 import key
@@ -270,6 +271,21 @@ _UNIMPLEMENTED_IMPORT = 'Importing {} without {} is not yet supported.'
 _UNPROCESSED_SUBELEMENT = 'Found an unprocessed <{}> element in a <{}>.'
 _MISSED_DATE = 'Unable to decipher the composition date "{}"'
 _BAD_VERSE_NUMBER = 'Verse number must be an int (got "{}")'
+
+
+# Mapping from an MEI <bTrem> @unitdur (the note value of one tremolo repetition) to the number of
+# tremolo marks used by :class:`~music21.expressions.Tremolo`: an eighth-note unit is one mark, a
+# sixteenth two, and so on (log2(unitdur) - 2).  Durations outside this table produce no marks.
+_MEI_UNITDUR_TO_TREMOLO_MARKS = {
+    '8': 1,
+    '16': 2,
+    '32': 3,
+    '64': 4,
+    '128': 5,
+    '256': 6,
+    '512': 7,
+    '1024': 8,
+}
 
 
 # Module-level Functions
@@ -2632,7 +2648,7 @@ def beamFromElement(elem, slurBundle=None):
 
     **Attributes/Elements Implemented:**
 
-    - <clef>, <chord>, <note>, <rest>, <space>, <tuplet>, <beam>, <barLine>
+    - <clef>, <chord>, <note>, <rest>, <space>, <tuplet>, <beam>, <bTrem>, <barLine>
 
     **Attributes/Elements Ignored:**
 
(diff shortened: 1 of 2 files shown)


mscuthbert (approved):
Great! Thank you

mscuthbert:
Congrats on being issue/PR 2000!

PR #2004, opened 2026-08-21
remove type: ignore

float3:
https://github.com/python/mypy/issues/3004 

Issue has been closed


mscuthbert:
Thanks! Fixed in mypy 18510

PR #2021, opened 2026-09-01
Give the flat-side accidentals their own unicode spellings (AI)

dylanpulver:
`unicodeFromModifier` (`music21/pitch.py:115`) gives three flat-side accidentals a glyph that belongs to something else, so `Accidental.unicode` cannot tell them apart:

| accidental | shipped | is also | should be |
|---|---|---|---|
| `triple-flat` | `U+266D` | `flat` | `U+266D U+1D12B` |
| `half-flat` | `U+1D132` QUARTER TONE SHARP | `half-sharp` | `U+1D133` QUARTER TONE FLAT |
| `one-and-a-half-flat` | `U+266D U+1D132` | — | `U+266D U+1D133` |

Thirteen accidental names produce eleven distinct spellings. The sharp side is already right (`###` is sharp + double sharp, `~` is `U+1D132`), and `U+1D133 MUSICAL SYMBOL QUARTER TONE FLAT` is in the Unicode Musical Symbols block but unused here.

Nothing internal reads these entries: `graph.utilities.accidentalLabelToUnicode` iterates the table but skips every modifier except `-` and `#`, so the flat-side values reach users only through `Accidental.unicode`, `Pitch.unicodeName` and `unicodeNameWithOctave`.

Measurement, same venv both legs: `uv run pytest music21/pitch.py music21/graph music21/roman.py` 213 -> 214. Mutants, each verified in the file before running: (A) restore the three shipped values -> the new test fails with `11 != 13`; (B) make all three distinct but by the lazy route (`♭♭♭`, and `U+1D12C` FLAT UP for the quarter tones) -> uniqueness passes, the value assertion fails, which is what pins the test to the codepoints rather than to mere distinctness; (C) break `###` -> the sharp-side assertion fails, so it is not vacuous. `uv run ruff check music21`, `uv run mypy music21` and `uv run pylint -j0 music21/pitch.py` are clean (10.00/10).

Not tested: `testSingleCoreAll` will not start here without lilypond installed, so I ran `uv run pytest music21` instead — 3747 passed on master and on this branch, with the same 29 lilypond-dependent failures in both legs. No rendering back end was checked for glyph coverage; `U+1D133` is astral-plane like the double sharp and double flat already in the table.

AI-assisted (Claude). (Entirely AI written)

### DIFF
--- a/music21/pitch.py
+++ b/music21/pitch.py
@@ -5,7 +5,7 @@
 # Authors:      Michael Scott Asato Cuthbert
 #               Christopher Ariza
 #
-# Copyright:    Copyright © 2008-2019 Michael Scott Asato Cuthbert
+# Copyright:    Copyright © 2008-2026 Michael Scott Asato Cuthbert
 # License:      BSD, see license.txt
 # ------------------------------------------------------------------------------
 '''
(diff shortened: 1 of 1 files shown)


mscuthbert:
Great -- in the future we might also allow for SMuFL output (since there's actually more support for that), but this is a good fix. Thanks!

PR #2026, opened 2026-09-07
Ignore trailing text after a Scala pitch value

float3:
ScalaData.parse skips a line only when it starts with '!', so a degree line carrying a comment after the value is handed to ScalaPitch, whose _setSrc keeps every digit, '.' and '/' from anywhere in the string. The comment's digits are spliced onto the end of the number rather than dropped, so `10/9  ! A\  ;  8 |mi|la|re|` is read as '10/98' -- -3951.34 cents where 10/9 is 182.40, and no exception is raised.

The Scala format ignores any characters following the pitch value, which is why the bundled files write comments there. Take the first whitespace-delimited token on the pitch-count line and on each degree line.

23 of the 3932 files in scale/scala/scl/ were misread. 21 raised ValueError, including every file whose comment contributes a second '/'; 2 more parsed without error and gave wrong cents, newton_15_out_of_53 and sparschuh-jsbloops440. Trailing text without a '!' took the same path: rvf1, rvf2 and rvf3 write a degree as `454.75 (427)`. All of them read correctly now, and no other file's parsed values change.

Two tests. testCommentAfterAPitchValue covers a '!' comment, bare trailing text, and the two bundled files that used to be read silently wrong, pinning their cents. testEveryBundledScalaFileParses reads the whole archive and asserts every file declares a pitch count, yields that many values, and gives a finite cents for each -- so a file added or edited upstream that this parser cannot read fails there rather than wherever some scale is later asked for. It takes about 0.8s. The description line is not checked, since the format allows it to be empty.

### DIFF
--- a/music21/scale/scala/__init__.py
+++ b/music21/scale/scala/__init__.py
@@ -288,10 +288,12 @@ class ScalaData:
                     self.description = line
             elif count == 2:
                 if line != '':
-                    self.pitchCount = int(line)
+                    # only the first token is data; the format ignores any
+                    # trailing characters, so a comment can follow the value
+                    self.pitchCount = int(line.split()[0])
             else:  # remaining counts are pitches
                 if line != '':
-                    sp = ScalaPitch(line)
+                    sp = ScalaPitch(line.split()[0])
                     sp.parse()
                     self.pitchValues.append(sp)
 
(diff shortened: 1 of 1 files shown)


mscuthbert:
Thank you!  Wonderful PR and big help.  My only concern originally was whether testEveryBundledScalaFileParses() was going to take too long to make it run every time we do the test suite, but 100ms is definitely fast enough to be worth avoiding this kind of bug in the future.

```python
[14]: %timeit t.testEveryBundledScalaFileParses()
105 ms ± 775 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

float3:
Happy to help (:

PR #2027, opened 2026-09-07
Tell a negative octave from a flat in removeRedundantPitches

float3:
`Pitch.nameWithOctave` spells both B-flat in octave 1 and B-natural in octave -1 as `'B-1'`, because the `'-'` is the flat sign and the minus sign both. `removeRedundantPitches` compared pitches by that string, so the second read as a repeat of the first and was thrown away.

Revised after review: `_removePitchByRedundantAttribute` now has the special knowledge of `'nameWithOctave'` itself and compares by name and octave in its place. Its signature is unchanged, no tuple form is exposed, and both call sites and the pitch-class and pitch-name reductions are exactly as they were.

While rewriting it I noticed the earlier version had a second problem, so this one also pins it: the octave is compared together with `octaveIsImplicit`, so a pitch given no octave stays distinct from one placed in the default octave. `nameWithOctave` already made that distinction (`'C'` against `'C4'`), and since v11 `Pitch.octave` returns 4 for both, comparing the octave alone would have quietly merged them. There is a doctest and a test case for it.

The docstring documented the old behaviour as a known bug ("doesn't seem a bug worth squashing at this moment"); it now documents what the method does, and `test_chord` carries a regression test.

### DIFF
--- a/music21/chord/__init__.py
+++ b/music21/chord/__init__.py
@@ -3625,24 +3625,28 @@ class Chord(ChordBase):
         <music21.chord.Chord C2 E3 G4 C5>
 
         Pitches with extreme octaves (whose `.nameWithOctave`s
-        look identical) are still distinguished.
-        
-        >>> p1 = pitch.Pitch('B')
-        >>> p1.octave = -1
+        look identical) are still distinguished.  B-flat in octave 1 and
+        B-natural in octave -1 both spell 'B-1', the '-' being the flat sign
+        in the one and the minus sign in the other:
+
+        >>> p1 = pitch.Pitch('B-')
+        >>> p1.octave = 1
         >>> p2 = pitch.Pitch('B')
-        >>> p2.octave = 1
+        >>> p2.octave = -1
+        >>> p1.nameWithOctave == p2.nameWithOctave
+        True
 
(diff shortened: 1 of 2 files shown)


mscuthbert:
I would rather have removeByRedunantPitches have special knowledge of 'nameWithOctave' and substitute 'nameWithOctave' with a special routine (either as a tuple or not) -- make the equivalence not something users have to think about.  

Based on that, do not demonstrate ('name', 'octave') as the usage of a tuple in the docs -- I was trying to figure out some other usage of the tuple form -- but i can't think of one -- everything i tried ('pitchClass', 'octave') had an equivalent single attribute like `ps` -- so unless there's another good use for tuple form, do not expose that.  Complexity reduction is a direction we're trying to go with -- this is the opposite.  :-)  Thanks!

float3:
is this what you had in mind?

mscuthbert:
> is this what you had in mind?

Nope!  My bug was floating space -- i wanted the rest of my changes kept.  We generally create chords from lowest to highest so p1 should be octave -1 and p2 be octave 1.  And I didn't want extra information about octave -1 in the docs -- human attention is valuable so we try not to draw attention where it doesn't need to be.

PR #2037, opened 2026-09-20
Fix #1190: a pitch repeated from a chord is an immediate repeat (AI)

### LINKED ISSUE #1190: cautionaryNotImmediateRepeat should account for chords
**music21 version**
dev

**Problem summary**
`cautionaryNotImmediateRepeat` doesn't treat a C# following a C# in a chord as an immediate repeat.

**Steps to reproduce**

```
>>> s = stream.Stream([chord.Chord("C#4 G#5"), note.Note('C#4')])
>>> s.show()  # default for cautionaryNotImmediateRepeat is True
```

Expected no accidental on the second C#, as it's an immediate repeat.

Open the file in Finale (MuseScore does its own magic). Or inspect the xml tags.

 
Screen Shot 2022-01-08 at 4 22 51 PM

ruiling-smartbear:
Fixes #1190.

`cautionaryNotImmediateRepeat` counted the other pitches of the immediately preceding chord
as intervening, so a pitch that had just sounded in that chord was given a cautionary
accidental. Those pitches sound at the same time as it, not in between, so the repeat is
still immediate.

`Stream.makeAccidentals()` now passes the immediately preceding chord's pitches down to
`Pitch.updateAccidentalDisplay()` as `previousSimultaneousPitches`, and those pitches no
longer break a continuous repeat. An intervening note, rest, or different chord still does,
so the cautionary accidental is still shown when it is actually warranted.

```
>>> from music21 import chord, note, stream
>>> s = stream.Stream([chord.Chord('C#4 G#5'), note.Note('C#4')])
>>> s.makeAccidentals(inPlace=True)
>>> s[1].pitch.accidental.displayStatus
False
```

## Review follow-up

Review asked for tests of the other accidental strategies and of measure boundaries. Both are
added.

- **Measure boundaries** — `testMakeAccidentalsAfterChordAcrossBarline`: a pitch repeated from
  the last chord of the previous measure still gets its courtesy accidental, because
  accidentals do not carry across a barline; the same two pitches within one measure do not.
- **Other strategies** — `testMakeAccidentalsAfterChordStrategies`: `cautionaryAll=True` still
  shows every accidental for an immediate repeat too, `cautionaryNotImmediateRepeat=False` still
  suppresses the cautionary for a non-immediate repeat, and a key signature,
  `cautionaryPitchClass=False` and `overrideStatus=True` each keep their existing behaviour.

I also ran every one of those combinations against the unfixed code and diffed the results. The
only behaviour that moves is the repeated pitch inside a measure; the barline path,
`cautionaryAll` and `cautionaryNotImmediateRepeat=False` are all untouched.

AI-assisted (DeepSeek). The submitting human reviewed the change and takes responsibility for it.

---

Supersedes #2036, which was closed asking for those tests and for the AI involvement to be
disclosed. GitHub will not reopen a PR whose branch was force-pushed after it closed, so this is
a fresh PR from the same two commits rather than a reopen.


mscuthbert:
Closing until evidence of other engagement within the system.  This is not the place for drive-by AI PRs.

ruiling-smartbear:
> Closing until evidence of other engagement within the system. This is not the place for drive-by AI PRs.

I have run the test myself but I will not contribute to this project using AI in the future. Thank you.

PR #2038, opened 2026-09-21
Do not append to the caller's list in AbstractOctaveRepeatingScale (AI Assisted)

float3:
buildNetwork completed the octave with mode.append(iComplement),

    mode = ['M3', 'M3']
    scale.OctaveRepeatingScale('c4', mode)
    mode  # ['M3', 'M3', <music21.interval.Interval d4>]

Reusing that list for a second scale then built a different scale.


float3:
I saw that agents.md mentions changing the copyright years, which I didn't do for previous PRs

mscuthbert:
> I saw that agents.md mentions changing the copyright years, which I didn't do for previous PRs

no worries -- it's not a big deal.  :-)

mscuthbert:
this is fine -- we might tighten the typing of mode to something like list[interval.Interval] and rename it from mode.  But a lot of these Abstract scales probably need to be rethought anyhow...

mscuthbert (approved):
Thanks!

